PR #260, opened 2014-02-28
Switched to ModeShapeEngine for deploy/undeploy/shutdown of repository

gregjan:
Edits to repo.xml files reflect the new ModeShapeEngine and destroy hook on ModeShapeRepositoryFactoryBean
Retooled several ITs to create just a container context and get at beans via singleton pattern
Some ITs were starting two repositories (in two different application contexts)


awoods:
This PR was the core of the eventual resolution: https://github.com/fcrepo4/fcrepo4/pull/476

PR #356, opened 2014-05-14
Updating range retrieval test to have configurable file size

escowles:
* Allows testing large files by setting a system property (fcrepo.rangetest.skip)

Related to https://www.pivotaltracker.com/story/show/71236292


osmandin:
https://jira.duraspace.org/browse/FCREPO-1693

PR #405, opened 2014-06-03
More unit tests and minor updates based on code review

awoods:


osmandin:
It seems this branch is preventing the deletion of the stale branch fcrepo4:jms-msg-format. See FCREPO-1693 for context. Original branch awoods:jms-msg-format can remain.

PR #504, opened 2014-10-07
Remove Lock Button from UI

osmandin:
https://www.pivotaltracker.com/story/show/79188218


bseeger:
Hi @osmandin and @awoods, 

It looks like this pull request was replaced by this one: https://github.com/fcrepo4/fcrepo4/pull/521

Is that true?  If so, could one of you close this one? 

Thanks!

PR #522, opened 2014-10-13
Removing POST from node type endpoint and slight factoring on HTTP ITs

ajs6f:
https://www.pivotaltracker.com/story/show/80574160


osmandin:
https://jira.duraspace.org/browse/FCREPO-1693

PR #668, opened 2014-11-24
Cleaning up HeadersJMSIT and added verify to JMSTopicPublisherTest

ajs6f:
https://www.pivotaltracker.com/story/show/83330628

### DIFF
--- a/fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/HeadersJMSIT.java
+++ b/fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/HeadersJMSIT.java
@@ -80,7 +80,7 @@ public class HeadersJMSIT implements MessageListener {
     /**
      * Time to wait for a set of test messages, in milliseconds.
      */
-    private static final long TIMEOUT = 2000;
+    private static final long TIMEOUT = 20000;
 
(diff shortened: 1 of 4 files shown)


fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/HeadersJMSIT.java
```diff
@@ -79,108 +105,74 @@
...
-        assertTrue(
-                "Found no message with correct identifer and correct event type!",
-                success);
     }
 
-    @Test
-    public void testRemoval() throws RepositoryException, InterruptedException,
-                             JMSException {
+    private void awaitMessageOrFail(final String id, final String eventType) {
+        await().pollInterval(ONE_SECOND).until(new Callable<Boolean>() {
```
awoods: If the `pollInterval` is 1sec and the timeout (which starts before the polling) is 2secs, it would seem that effectively the wait will just be a single second. Should we adjust either the polling interval or the timeout so the polling gets at least two tries?

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/41e3fb87b09a9393ee37d9b6cb935bd70823a6d5

PR #669, opened 2014-11-25
aliasing any jcr-namespaced properties to fedora:

acoburn:
https://www.pivotaltracker.com/n/projects/684825/stories/83411334

any `jcr:` properties are transformed to `fedora:` before sending to the JMS broker.

This potentially causes **any** `jcr:` property to populate the `fedora:` namespace, which may not be desired.

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/GetNamespacedProperties.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/GetNamespacedProperties.java
@@ -53,7 +53,7 @@ public class GetNamespacedProperties implements Function<FedoraEvent, FedoraEven
         for (String property : evt.getProperties()) {
             final String[] parts = property.split(":", 2);
             if (parts.length == 2) {
-                final String prefix = parts[0];
+                final String prefix = parts[0].equals("jcr") ? "fedora" : parts[0];
                 try {
                     event.addProperty(namespaceRegistry.getURI(prefix) + parts[1]);
                 } catch (RepositoryException ex) {
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/GetNamespacedProperties.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/GetNamespacedProperties.java
@@ -23,6 +23,9 @@ import javax.jcr.NamespaceRegistry;
 import javax.jcr.RepositoryException;
 import javax.jcr.Session;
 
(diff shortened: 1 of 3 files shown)


acoburn:
This is now tracked in JIRA: https://jira.duraspace.org/browse/FCREPO-1405

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/GetNamespacedProperties.java
```diff
@@ -54,11 +57,21 @@ public FedoraEvent apply(final FedoraEvent evt) {
...
                 final String prefix = parts[0];
-                try {
-                    event.addProperty(namespaceRegistry.getURI(prefix) + parts[1]);
-                } catch (RepositoryException ex) {
-                    LOGGER.trace("Prefix could not be dereferenced using the namespace registry: {}", property);
-                    event.addProperty(property);
+                if ("jcr".equals(prefix)) {
+                    if (jcrProperties.contains(createProperty(REPOSITORY_NAMESPACE + parts[1]))) {
+                        event.addProperty(REPOSITORY_NAMESPACE + parts[1]);
+                    } else if (LOGGER.isDebugEnabled()) {
```
awoods: Please remove the `LOGGER.isDebugEnabled()` guard to make it easier to change the logging state at runtime.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/GetNamespacedProperties.java
```diff
@@ -54,11 +57,21 @@ public FedoraEvent apply(final FedoraEvent evt) {
...
+                        LOGGER.debug("Swallowing jcr property: {}", property);
+                    }
+                } else {
+                    try {
+                        event.addProperty(namespaceRegistry.getURI(prefix) + parts[1]);
+                    } catch (RepositoryException ex) {
+                        if (LOGGER.isDebugEnabled()) {
```
awoods: Please remove the `LOGGER.isDebugEnabled()` guard to make it easier to change the logging state at runtime.

ajs6f: There's really just no reason to use these kinds of tests. They are built into the logging framework. Just `LOGGER.debug()` whatever it is and the logging framework will decide whether or not to execute the action.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/be0e046cdcbc0653a2958c044c6ccd34e8e0f6c8

PR #678, opened 2014-12-11
Added integration test for creating blank nodes.

peichman-umd:
Attempts to create 5000 new LDP resources with RDF content. The RDF
graph for each one has 13 blank nodes.

Resolves: https://jira.duraspace.org/browse/FCREPO-1257


awoods:
See comments in https://jira.duraspace.org/browse/FCREPO-1257

PR #688, opened 2015-01-20
Correct -kernel and -kernel-impl Javadoc links  to removed Fedora classe...

osmandin:
...s

https://jira.duraspace.org/browse/FCREPO-1263


ajs6f:
Looks nice. We'll have to keep our eyes peeled for any further examples of this kind of annoyance.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TombstoneImpl.java
```diff
@@ -33,7 +33,7 @@
     private static final Logger LOGGER = getLogger(TombstoneImpl.class);
 
     /**
-     * Construct a FedoraObject from an existing JCR Node
+     * Construct a {@link org.fcrepo.kernel.models.Container} from an existing JCR Node
```
awoods: This should be:
`{@link org.fcrepo.kernel.models.FedoraResource}`

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/4bf47ae22675e0fad9f3f0aa8f3a921a6e071736

PR #690, opened 2015-01-21
Replace DummyTransactionManager with JBoss Transaction Manager

awoods:
See: https://docs.jboss.org/author/display/MODE40/Getting+Started#GettingStarted-Usingatransactionmanager

Resolves: https://jira.duraspace.org/browse/FCREPO-1293


acoburn:
:+1: This change worked for me. (Tested on Tomcat 7 and Tomcat 8 with ~10K transactions)

PR #692, opened 2015-01-22
Create minted hierarchy for blank nodes

osmandin:
https://jira.duraspace.org/browse/FCREPO-1258

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
@@ -104,7 +104,7 @@ public class JcrRdfTools {
 
     private static final Model m = createDefaultModel();
 
-    private static final UUIDPathMinter pidMinter =  new UUIDPathMinter(10,1);
+    private static final UUIDPathMinter pidMinter =  new UUIDPathMinter();
 
     /**
      * Constructor with even more context.
@@ -174,7 +174,7 @@ public class JcrRdfTools {
      * @throws RepositoryException
      */
     public Value createValue(final ValueFactory valueFactory, final RDFNode data, final int type)
-            throws RepositoryException {
+        throws RepositoryException {
         assert (valueFactory != null);
 
 
@@ -406,7 +406,7 @@ public class JcrRdfTools {
         final AnonId id = resource.asResource().getId();
 
         if (!skolemizedBnodeMap.containsKey(id)) {
-            final String path = "/.well-known/genid/" + pidMinter.mintPid();
+            final String path = skolemizedId() + pidMinter.mintPid();
             final Node orCreateNode = jcrTools.findOrCreateNode(session, path);
             orCreateNode.addMixin(FEDORA_BLANKNODE);
             final Resource skolemizedSubject = nodeToResource(idTranslator).convert(orCreateNode);
@@ -414,6 +414,10 @@ public class JcrRdfTools {
         }
 
         return skolemizedBnodeMap.get(id);
+    }
+
+    private static String skolemizedId() {
+        return "/.well-known/genid/";
     }
 
 }
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
@@ -409,6 +409,7 @@ public class JcrRdfTools {
             final String path = skolemizedId() + pidMinter.mintPid();
             final Node orCreateNode = jcrTools.findOrCreateNode(session, path);
             orCreateNode.addMixin(FEDORA_BLANKNODE);
+            orCreateNode.addMixin(FEDORA_PAIRTREE);
             final Resource skolemizedSubject = nodeToResource(idTranslator).convert(orCreateNode);
             skolemizedBnodeMap.put(id, skolemizedSubject);
         }
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
@@ -50,6 +50,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.hp.hpl.jena.rdf.model.AnonId;
 import com.hp.hpl.jena.rdf.model.Model;
 import com.hp.hpl.jena.rdf.model.Statement;
+import org.fcrepo.kernel.identifiers.PidMinter;
 import org.fcrepo.kernel.impl.services.AbstractService;
 import org.fcrepo.kernel.models.FedoraResource;
 import org.fcrepo.kernel.RdfLexicon;
@@ -105,7 +106,7 @@ public class JcrRdfTools {
 
     private static final Model m = createDefaultModel();
 
-    private static final UUIDPathMinter pidMinter =  new UUIDPathMinter();
+    private static final PidMinter pidMinter =  new UUIDPathMinter();
 
     /**
      * Constructor with even more context.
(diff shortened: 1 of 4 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
```diff
@@ -104,6 +104,7 @@
 
     private static final Model m = createDefaultModel();
 
+    private static final UUIDPathMinter pidMinter =  new UUIDPathMinter(10,1);
```
awoods: The default value here should probably be:
`new UUIDPathMinter(2, 4);`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
```diff
@@ -173,7 +174,7 @@ public Value createValue(final Node node,
      * @throws RepositoryException
      */
     public Value createValue(final ValueFactory valueFactory, final RDFNode data, final int type)
-        throws RepositoryException {
+            throws RepositoryException {
```
awoods: Remove inadvertent formatting change.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
ajs6f: The default values should be, in any event, labelled constants.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
```diff
@@ -414,7 +416,4 @@ private Resource getSkolemizedResource(final IdentifierConverter<Resource, Fedor
         return skolemizedBnodeMap.get(id);
     }
 
-    private static String skolemizedId() {
```
awoods: Keep this method, as it reduces unnecessary code changes and encapsulates the ".well-known/genid/" location.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
```diff
@@ -405,7 +406,8 @@ private Resource getSkolemizedResource(final IdentifierConverter<Resource, Fedor
         final AnonId id = resource.asResource().getId();
 
         if (!skolemizedBnodeMap.containsKey(id)) {
-            final Node orCreateNode = jcrTools.findOrCreateNode(session, skolemizedId());
+            final String path = "/.well-known/genid/" + pidMinter.mintPid();
+            final Node orCreateNode = jcrTools.findOrCreateNode(session, path);
```
awoods: Ensure that interim path elements are marked with mixin type: FEDORA_PAIRTREE.
See: https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java#L67

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
```diff
@@ -173,7 +174,7 @@ public Value createValue(final Node node,
      * @throws RepositoryException
      */
     public Value createValue(final ValueFactory valueFactory, final RDFNode data, final int type)
```
awoods: Please add/update tests in `JcrRdfToolsTest.java`.

osmandin: I tried adding test but gave up after not finding a way to get hold of the blank nodes created in JcrRdfToolsTest. Are you looking for existence of pairtree mixin on the hierarchy nodes? Thanks.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
```diff
@@ -64,7 +64,13 @@ protected Node findNode(final Session session, final String path) {
         }
     }
 
-    private static void tagHierarchyWithPairtreeMixin(final Node baseNode, final Node createdNode)
+    /**
+     *
+     * @param baseNode
```
awoods: Based on reading these javadocs, I, as a API consumer, have no better understanding of what this method does. 
Please add some detail.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
```diff
@@ -104,6 +105,7 @@
 
     private static final Model m = createDefaultModel();
 
+    private static final UUIDPathMinter pidMinter =  new UUIDPathMinter();
```
ajs6f: Should this field not be of a an interface type?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
```diff
@@ -415,6 +428,7 @@ private Resource getSkolemizedResource(final IdentifierConverter<Resource, Fedor
     }
 
     private static String skolemizedId() {
-        return "/.well-known/genid/" + randomUUID();
+        return "/.well-known/genid/";
```
ajs6f: This method isn't producing an actual identifier, but a prefix. Maybe the name can change to reflect that?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
```diff
@@ -405,8 +407,19 @@ private Resource getSkolemizedResource(final IdentifierConverter<Resource, Fedor
         final AnonId id = resource.asResource().getId();
 
         if (!skolemizedBnodeMap.containsKey(id)) {
-            final Node orCreateNode = jcrTools.findOrCreateNode(session, skolemizedId());
+            final String pid = pidMinter.mintPid();
```
awoods: Why are you not able to move `pidMinter.mintPid()` back into the method `skolemizedId()`, then:
`final String path = skolemizedId();`, and
only call `jcrTools.findOrCreateNode(session, path)` once instead of twice?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
awoods: I would think you could follow the pattern set by:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/rdf/JcrRdfToolsTest.java#L312 or
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/rdf/JcrRdfToolsTest.java#L329

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
awoods: Agreed.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
osmandin: Because that would tag /wellknown/genid as well, and then all the blank nodes would appear on the main page (instead of under /wellknown/genid). I too wanted jcrTools.findOrCreate() for parent path /wellknown/genid to be called once, but can't think of a way to implement this alternatively in a clean fashion (this method does not explicitly get called in a loop). One option I did originally have in mind was to not use the existing method in AbstractService#tagHierarchyWithPairtreeMixin(), and string tokenize the path, but this approach doesn't seem good to me (split/join on the way in and then splitting again; not using the existing functionality, etc).

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
awoods: Tests are still needed. See comment above ^^.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/b6435477522c6752cc72feec7fc07ed58f10e07e

PR #695, opened 2015-01-23
add fedorabinary check

yinlinchen:
The workflow is check whether the child node is a FedoraBinary, get description, and added to the triple. The test is passed, but failed in Fedora Repository HTTP API Module...

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/PropertiesRdfContext.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/PropertiesRdfContext.java
@@ -26,6 +26,7 @@ import javax.jcr.RepositoryException;
 
 import com.hp.hpl.jena.rdf.model.Resource;
 
+import org.fcrepo.kernel.models.FedoraBinary;
 import org.fcrepo.kernel.models.FedoraResource;
 import org.fcrepo.kernel.identifiers.IdentifierConverter;
 import org.fcrepo.kernel.impl.rdf.impl.mappings.PropertyToTriple;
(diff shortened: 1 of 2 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/PropertiesRdfContext.java
```diff
@@ -66,7 +68,14 @@ public PropertiesRdfContext(final FedoraResource resource,
     private Iterator<Triple> triplesFromProperties(final FedoraResource n)
         throws RepositoryException {
         LOGGER.trace("Creating triples for node: {}", n);
-        final Iterator<Property> allProperties = n.getNode().getProperties();
+        final Iterator<Property> allProperties;
+        if (n instanceof FedoraBinary) {
```
awoods: In the case when the FedoraResource is a FedoraBinary, you should be able to concatenate the properties from both the description and the arg node itself. The end.

yinlinchen: I see, so the logic should be first get properties from n.getNode().getProperties() and then if the node is a binary, append description.getNode().getProperties(). 
Can you tell me how to append allProperties?

awoods: Iterators.concat() should do the trick.
http://guava-libraries.googlecode.com/svn-history/r46/trunk/javadoc/com/google/common/collect/Iterators.html

yinlinchen:
All test passed. FCREPO-1290 fixed.

PR #702, opened 2015-01-29
updated travis-ci configuration

acoburn:
This adds travis-ci support to the Java8 branch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: java
 jdk:
-  - oraclejdk7
+  - oraclejdk8
 before_install:  
  - "echo $JAVA_OPTS"
  - "export JAVA_OPTS=-Xmx512m"


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: java
 jdk:
-  - oraclejdk7
+  - oraclejdk8
 before_install:  
  - "echo $JAVA_OPTS"
  - "export JAVA_OPTS=-Xmx512m"
```
ajs6f: Why drop the heap adjustment?

acoburn: @ajs6f I only changed the jdk specification (oraclejdk7 -> oraclejdk8) -- the heap size was unchanged.

ajs6f: @acoburn I'm a dope! I misread the hyphens at the start of those lines as minus signs indicating deletions!

PR #703, opened 2015-01-30
Address copyright date to 2015

yinlinchen:
- change 2014 to 2015 in pom.xml
- run mvn license:format, checkstyle, mvn clean install.
- check the diff

Resolves: https://jira.duraspace.org/browse/FCREPO-1297


awoods:
I would have expected two updates to the top level pom.xml.

yinlinchen:
updated.

awoods:
I am not seeing a second commit...

yinlinchen:
I use -f again, won't use that anymore.  Can you see my change now?

https://github.com/yinlinchen/fcrepo4/blob/fcrepo-1297/pom.xml

PR #705, opened 2015-01-31
Add default properties to pom.xml

claussni:
A couple of system properties are referenced throughout the POMs and code but never declared in any POM. These properties can be set in Maven command line executions: ```mvn ... -Dfcrepo.test.port=8080```

The properties should be declared in the man POM using some default values. They can than be overridden by command line arguments.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,6 +59,9 @@
     <jacoco.out.it.file>jacoco-it.exec</jacoco.out.it.file>
     <sonar.jacoco.reportPath>${jacoco.outputDir}/${jacoco.out.unit.file}</sonar.jacoco.reportPath>
     <sonar.jacoco.itReportPath>${jacoco.outputDir}/${jacoco.out.it.file}</sonar.jacoco.itReportPath>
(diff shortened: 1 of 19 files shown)


pom.xml
```diff
@@ -59,6 +59,10 @@
     <jacoco.out.it.file>jacoco-it.exec</jacoco.out.it.file>
     <sonar.jacoco.reportPath>${jacoco.outputDir}/${jacoco.out.unit.file}</sonar.jacoco.reportPath>
     <sonar.jacoco.itReportPath>${jacoco.outputDir}/${jacoco.out.it.file}</sonar.jacoco.itReportPath>
+    <!-- default properties that can be altered on the command line -->
+    <fcrepo.test.context.path/>
```
awoods: Unless there is a specific reason to include these two jacoco arg defaults, I would suggest leaving them out of the PR.
I am concerned that our Jenkins/Sonar reporting will become broken due to them.
- http://jenkins.fcrepo.org/
- http://sonar.fcrepo.org/

PR #706, opened 2015-02-03
Fcrepo 1290

yinlinchen:
Add return nonRdfSource properties in response and test case template


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/PropertiesRdfContext.java
```diff
@@ -56,19 +56,25 @@
      */
 
     public PropertiesRdfContext(final FedoraResource resource,
-                                final IdentifierConverter<Resource, FedoraResource> idTranslator)
```
awoods: As a side note, please do not include unrelated formatting updates in pull-requests.

yinlinchen: line 59? It was formatted by Eclipse using the auto style check using the https://github.com/fcrepo4/fcrepo4/tree/master/src/site/eclipse and I run mvn checkstyle:check before submit PR.

awoods: In any case, please do not include non-related formatting updates.

PR #707, opened 2015-02-03
override hashcode()

yinlinchen:
When override equal(), should override hashcode()

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/PreferTag.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/PreferTag.java
@@ -146,4 +146,13 @@ public class PreferTag implements Comparable<PreferTag> {
             return false;
         }
     }
+
+    @Override
+    public int hashCode() {
+        if (getTag() == null) {
+            return 0;
+        }
+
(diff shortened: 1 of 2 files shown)


awoods:
Please add a test that demonstrates the "why" of adding the hashCode() method.

yinlinchen:
According to the Java Language Specification, there is a contract between equals(Object) and hashCode():

    If two objects are equal according to the equals(Object) method, then calling the hashCode method on each of the two objects must produce the same integer result. It is not required that if two objects are unequal according to the equals(java.lang.Object) method, then calling the hashCode method on each of the two objects must produce distinct integer results. However, the programmer should be aware that producing distinct integer results for unequal objects may improve the performance of hashtables. 

In order to comply with this contract, those methods should be either both inherited, or both overridden.

awoods:
Every functional update should be accompanied by one or more unit and integration tests.
Please add a test that demonstrates the "why" of adding the hashCode() method.

yinlinchen:
will add testHashCode()

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/c2e1af0a87da42941ec4c837abc2715154994873

PR #710, opened 2015-02-11
Address Reduce Sonar major issue - "Use Index Of Char"

yinlinchen:
-  Use String.indexOf(char) when checking for the index of a single character; it executes faster.

Resolves: https://jira.duraspace.org/browse/FCREPO-1328


awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/4f3dee0ee9b3ab4b3cbc0bb9fa9deaff65049c73

PR #711, opened 2015-02-12
Add version label check for valid characters

osmandin:
https://jira.duraspace.org/browse/FCREPO-1246

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersioning.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersioning.java
@@ -62,6 +62,9 @@ import org.springframework.context.annotation.Scope;
 
 import com.google.common.annotations.VisibleForTesting;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 /**
  * @author cabeer
  * @since 9/25/14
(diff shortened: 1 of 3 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersioning.java
```diff
@@ -191,4 +195,14 @@ protected FedoraResource resource() {
     protected Session session() {
         return session;
     }
+
+    /**
+     * Determines if a label contains valid characters
+     * @param label label
+     * @return whether label contains valid characters
+     */
+    public boolean invalidLabel(final String label) {
```
awoods: This method should be "private".
Also, I agree with the javadocs on this method, that it is more clear if the method returns whether the arg label is valid. Additionally, it would make the logic above easier to read, i.e. no double-negative. Instead it would read:
`if (!isBlank(slug) && validLabel(slug)) {`

I suggest renaming the method and returning `!matcher.find()`

awoods: In looking at this closer, the label validation should be enforced even if the user does not make the request via the REST API. Therefore, you should move your updates deeper into the application, i.e. https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/VersionServiceImpl.java#L162

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/bb6072d83da9897bf4a20f357353387dd05efb09

PR #714, opened 2015-02-12
Change generic exceptions to more specific exceptions

osmandin:
https://jira.duraspace.org/browse/FCREPO-1317?jql=key%20in%20(FCREPO-1317)

http://sonar.fcrepo.org/drilldown/issues/1?&rule=squid%3AS00112&rule_sev=MAJOR&severity=MAJOR#

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/spring/ModeShapeRepositoryFactoryBean.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/spring/ModeShapeRepositoryFactoryBean.java
@@ -24,6 +24,7 @@ import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
 import javax.inject.Inject;
 
+import org.fcrepo.kernel.exception.RepositoryRuntimeException;
 import org.modeshape.jcr.JcrRepository;
 import org.modeshape.jcr.ModeShapeEngine;
 import org.modeshape.jcr.NoSuchRepositoryException;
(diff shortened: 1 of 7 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/spring/ModeShapeRepositoryFactoryBean.java
```diff
@@ -60,21 +61,24 @@
...
+
+            // next line ensures that repository starts before the factory is used.
+            final org.modeshape.common.collection.Problems problems =
+                    repository.getStartupProblems();
+            for (final org.modeshape.common.collection.Problem p : problems) {
+                LOGGER.error("ModeShape Start Problem: {}", p.getMessageString());
+                // TODO determine problems that should be runtime errors
+            }
+        } catch (Exception e) {
+            throw new RepositoryRuntimeException(e);
```
ajs6f: I may be being really too picky here, but to my understanding, this is not a `RepositoryRuntimeException` precisely because the repository is not running. We _could_ put in a new `RepositoryFailedStartup` kind of thing, or maybe MODE already has such thing, or maybe I'm just being too picky...

ajs6f:
This is good stuff and I'm all in favor, but I do want to record here (for my own failing memory) that we probably ought to be subclassing `RepositoryRuntimeException` and using the subclasses a little more. Which in no way is any vote against this PR.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/spring/ModeShapeRepositoryFactoryBean.java
awoods: Sounds very reasonable. But since @osmandin does not appear to agree, let's move this through for now.

PR #716, opened 2015-02-12
Remove nested try-catch blocks

osmandin:
https://jira.duraspace.org/browse/FCREPO-1330

http://sonar.fcrepo.org/drilldown/issues/1?&rule=squid%3AS1141&rule_sev=MAJOR&severity=MAJOR#

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
@@ -264,33 +264,17 @@ public class HttpResourceConverter extends IdentifierConverter<Resource,FedoraRe
      */
     private Node getFrozenNodeByLabel(final String baseResourcePath, final String label) {
         try {
-            try {
-                final Node frozenNode = session.getNodeByIdentifier(label);
+            final Node n = getNode(baseResourcePath, label);
 
-            /*
-             * We found a node whose identifier is the "label" for the version.  Now
-             * we must do due dilligence to make sure it's a frozen node representing
-             * a version of the subject node.
-             */
-                final Property p = frozenNode.getProperty("jcr:frozenUuid");
-                if (p != null) {
-                    final Node subjectNode = session.getNode(baseResourcePath);
-                    if (p.getString().equals(subjectNode.getIdentifier())) {
-                        return frozenNode;
-                    }
-                }
-            /*
+            if (n != null) {
+                return n;
+            }
+
+             /*
              * Though a node with an id of the label was found, it wasn't the
              * node we were looking for, so fall through and look for a labeled
              * node.
              */
-            } catch (final ItemNotFoundException ex) {
-            /*
-             * the label wasn't a uuid of a frozen node but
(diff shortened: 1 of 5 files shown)


fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
```diff
@@ -305,6 +289,32 @@ private Node getFrozenNodeByLabel(final String baseResourcePath, final String la
         }
     }
 
+    private Node getNode(final String baseResourcePath, final String label) throws RepositoryException {
```
ajs6f: Just a future-aimed note that after Java 8,this is the kind of work that should be done by returning an [Optional](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html), not a value which might be null.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
```diff
@@ -614,6 +592,39 @@ public Node getNodeVersion(final String label) {
 
     }
 
+    private Node getFrozenNode(final String label) throws RepositoryException {
```
ajs6f: Ditto.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/utils/JcrPropertyStatementListener.java
```diff
@@ -160,6 +156,14 @@ public void removedStatement(final Statement s) {
 
     }
 
+    private void removeMixin(final FedoraResource resource, final Resource mixinResource, final Resource s) {
```
ajs6f: Maybe instead of in a method on this class, the try-catch should be moved back into `JcrRdfTools`?

awoods: :+1: ...and add a log.warn() that removing the mixin failed.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/JcrRdfTools.java
```diff
@@ -296,10 +296,14 @@ public void removeMixin(final FedoraResource resource,
                             final Resource mixinResource,
                             final Map<String, String> nsPrefixMap) throws RepositoryException {
 
-        final Node node = resource.getNode();
-        final String mixinName = getPropertyNameFromPredicate(node, mixinResource, nsPrefixMap);
-        if (repositoryHasType(session, mixinName) && node.isNodeType(mixinName)) {
-            node.removeMixin(mixinName);
+        try {
```
awoods: Since `JcrRdfTools.removeMixin()` already throws a `RepositoyException`, the `try|catch` you have added can be safely removed.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/3fa9652d4ea9d60a85154dee4d718e3c031170f9

PR #719, opened 2015-02-13
Fix sonar warnings related to null array return, null variable dereferen...

osmandin:
...cing, indirectly invoked but otherwise unused variable, etc.

https://jira.duraspace.org/browse/FCREPO-1335

All addressed except # 1 in the ticket description.

But the following are addressed in this PR:

Unused private method	1 file:	http://sonar.fcrepo.org/drilldown/issues/1?&rule=squid%3AUnusedPrivateMethod&rule_sev=MAJOR&severity=MAJOR
Singular Field	1 file: http://sonar.fcrepo.org/drilldown/issues/1?&rule=pmd%3ASingularField&rule_sev=MAJOR&severity=MAJOR
Correctness - Nullcheck of value previously dereferenced	1 file: http://sonar.fcrepo.org/drilldown/issues/1?&rule=findbugs%3ARCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE&rule_sev=MAJOR&severity=MAJOR
Empty arrays and collections should be returned instead of null	1 file: http://sonar.fcrepo.org/drilldown/issues/1?&rule=squid%3AS1168&rule_sev=MAJOR&severity=MAJOR

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersions.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersions.java
@@ -98,8 +98,11 @@ public class FedoraVersions extends ContentExposingResource {
         this.pathIntoVersion = pathIntoVersion;
     }
 
+    /**
+     * Initialization method. Clients don't invoke this.
+     */
     @PostConstruct
-    private void postConstruct() {
+    public void postConstruct() {
         this.path = externalPath + "/" + FCR_VERSIONS + "/" + pathListIntoVersion;
         this.label = pathListIntoVersion.split("/", 2)[0];
     }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersions.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersions.java
@@ -98,11 +98,8 @@ public class FedoraVersions extends ContentExposingResource {
         this.pathIntoVersion = pathIntoVersion;
     }
 
-    /**
-     * Initialization method. Clients don't invoke this.
-     */
     @PostConstruct
-    public void postConstruct() {
+    private void postConstruct() {
         this.path = externalPath + "/" + FCR_VERSIONS + "/" + pathListIntoVersion;
         this.label = pathListIntoVersion.split("/", 2)[0];
     }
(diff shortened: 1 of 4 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersions.java
```diff
@@ -98,8 +98,11 @@ public FedoraVersions(final String path, final String label, final String pathIn
         this.pathIntoVersion = pathIntoVersion;
     }
 
+    /**
+     * Initialization method. Clients don't invoke this.
+     */
     @PostConstruct
-    private void postConstruct() {
+    public void postConstruct() {
```
awoods: The @PostConstruct annotation works if the method is `private` or `public`. I would suggest leaving this as `private` and I can flag the Sonar issue as a `false positive`.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/TypeRdfContext.java
```diff
@@ -66,10 +66,12 @@ private void concatRdfTypes() throws RepositoryException {
         final NodeType primaryNodeType = resource().getNode().getPrimaryNodeType();
         nodeTypesB.add(primaryNodeType);
```
awoods: If `primaryNodeType` is null, an NPE will be thrown on line 67.
I would suggest checking if `primaryNodeType` is null between lines 66 and 67.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/utils/infinispan/CacheLoaderChunkInputStream.java
```diff
@@ -194,6 +194,6 @@ private void endOfStream() {
         if (blobCache.contains(chunkKey)) {
             return blobCache.load(chunkKey).getValue();
         }
-        return null;
+        return new byte[]{};
```
awoods: Returning `null` in this `private` method is probably cleaner and more efficient than Sonar's suggestion.

ajs6f:
Just as a general note, I would be cautious about spending a lot of time fixing Sonar issues before the move to Java 8. A lot of this code just disappears entirely then, so this isn't really a great way to spend time.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/7cd7cc4964d4840b092c997bbd1af43979da50b2

PR #729, opened 2015-02-18
Fix Sonar major issues

yinlinchen:
- fixed string buffer for string appends
- fixed static inner class
- fixed Locale

Resolves: https://jira.duraspace.org/browse/FCREPO-1337

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
@@ -364,7 +364,7 @@ public class HttpResourceConverter extends IdentifierConverter<Resource,FedoraRe
         if (path != null) {
 
             if (resource instanceof NonRdfSourceDescription) {
-                path = path + "/" + FCR_METADATA;
+                path += "/" + FCR_METADATA;
             }
 
             if (path.endsWith(JCR_CONTENT)) {
@@ -424,7 +424,7 @@ public class HttpResourceConverter extends IdentifierConverter<Resource,FedoraRe
     /**
      * Translate the current transaction into the identifier
      */
-    class TransactionIdentifierConverter extends Converter<String, String> {
+    static class TransactionIdentifierConverter extends Converter<String, String> {
         public static final String TX_PREFIX = "tx:";
 
         private final Session session;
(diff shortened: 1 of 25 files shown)


fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
```diff
@@ -354,7 +354,7 @@ private String doBackwardPathOnly(final FedoraResource resource) {
         if (path != null) {
 
             if (resource instanceof NonRdfSourceDescription) {
-                path = path + "/" + FCR_METADATA;
+                path += "/" + FCR_METADATA;
```
awoods: The Sonar recommendation is to use a StringBuffer instead of String concatenation. Your update is a different syntax for the same String concatenation.
In any case, the performance impact of String concatenation is tolerable here, and no update is needed.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
```diff
@@ -130,7 +131,7 @@ public static ViewHelpers getInstance() {
             date = getVersionDate(graph, triple.getObject());
             String key = isNullOrEmpty(date) ? format.format(new Date()) : date;
             while (map.containsKey(key)) {
-                key = key + "1";
+                key += "1";
```
awoods: Ditto from above.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
```diff
@@ -122,7 +123,7 @@ public static ViewHelpers getInstance() {
         final Node subject, final Resource predicate) {
         final Iterator<Triple> versions = getObjects(graph, subject, predicate);
         final Map<String, Node> map = new TreeMap<>();
-        final SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+        final SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", Locale.US);
```
awoods: `Locale.US` will probably not make sense for the rest of the world that uses Fedora. In this case, allowing `SimpleDateFormat` to use its default `Locale` is probably fine.

PR #737, opened 2015-02-24
Upgrade ModeShape to version 4.2.0.Final

osmandin:
IT tests had to be updated possibly due to MODE-2414


awoods:
https://jira.duraspace.org/browse/FCREPO-1374

PR #739, opened 2015-02-27
Change problematic SPARQL update  statement namespace prefix URI

osmandin:
https://jira.duraspace.org/browse/FCREPO-1381

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
@@ -93,7 +93,7 @@ public class FedoraResourceImpl extends JcrTools implements FedoraJcrTypes, Fedo
 
     protected Node node;
 
-    private static final Pattern uriPattern = compile("[a-z]*:<[a-z]*://.*[^/]*>");
+    private static final Pattern uriPattern = compile("PREFIX [a-z]*:<[a-z]*://[^#(PREFIX)]*>");
 
     /**
      * Construct a {@link org.fcrepo.kernel.models.FedoraResource} from an existing JCR Node
@@ -376,19 +376,15 @@ public class FedoraResourceImpl extends JcrTools implements FedoraJcrTypes, Fedo
     }
 
     private static String changePrefixUri(final String statement) {
-        String modifiedStatement = statement;
-        final String s[] = statement.split("(?i)INSERT");
-        final String prefix[] = s[0].split("(?i)PREFIX");
+        final Matcher m = uriPattern.matcher(statement.trim());
+        final StringBuffer sb = new StringBuffer();
 
-        for (String p: prefix) {
-            final Matcher m = uriPattern.matcher(p.trim());
-            if (m.matches()) {
-                final String t = p.replace(">", "/>");
-                modifiedStatement = modifiedStatement.replace(p, t);
-                LOGGER.info("Changed prefix uri from {} to {}", p, t);
-            }
+        while (m.find()) {
+            m.appendReplacement(sb, m.group().replaceFirst(">", "/>"));
         }
-        return modifiedStatement;
+
+        m.appendTail(sb);
+        return sb.length() > 0 ? sb.toString() : statement;
     }
 
     @Override
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
@@ -93,7 +93,7 @@ public class FedoraResourceImpl extends JcrTools implements FedoraJcrTypes, Fedo
 
     protected Node node;
 
-    private static final Pattern uriPattern = compile("[a-z]*:<[a-z]*://[a-zA-Z.0-9]*>");
+    private static final Pattern uriPattern = compile("[a-z]*:<[a-z]*://.*[^/]*>");
 
     /**
      * Construct a {@link org.fcrepo.kernel.models.FedoraResource} from an existing JCR Node
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
@@ -24,6 +24,7 @@ import static com.google.common.collect.Iterators.transform;
 import static com.google.common.collect.Lists.newArrayList;
 import static com.hp.hpl.jena.update.UpdateAction.execute;
 import static com.hp.hpl.jena.update.UpdateFactory.create;
+import static java.util.regex.Pattern.compile;
 import static org.apache.commons.codec.digest.DigestUtils.shaHex;
 import static org.fcrepo.kernel.impl.identifiers.NodeResourceConverter.nodeConverter;
 import static org.fcrepo.kernel.impl.utils.FedoraTypesUtils.isFrozenNode;
@@ -40,6 +41,8 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.Iterator;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.jcr.AccessDeniedException;
 import javax.jcr.ItemNotFoundException;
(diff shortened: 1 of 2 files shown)


awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/f257425b72e939a0d159ff9b16de5fb1792fa213

cbeer:
Wait a moment. What's happening in this patch? I may not understand the regex happening, but it looks like we're "correcting" incoming namespace URIs, even if the client sent completely valid data. I don't know what the SPARQL spec has to say about it (@ajs6f?), but RDF/XML says:

> The RDF URI reference represented by a QName is determined by appending the local name part of the QName after the namespace name (URI reference) part of the QName.

The problem, as I understood it from the email, is the namespace prefix (which are essentially arbitrary..) wasn't being correctly calculated on serialization, so I'm not sure why changing the SPARQL Update behavior is the right place to fix this.

awoods:
@cbeer, the issue is that if a prefix is defined without a final or intermediate delimiter (/ or #), then Jena registers a statement such as:
`PREFIX dsc:<http://myurl.org>`
as:
`ns001 = http://`
...which clearly is not desired.
This is probably a bug in Jena, and should likely be resolved at that level. But in the meantime, @osmandin 's PR "fixes" such client-provided namespaces by adding a trailing slash.
@cbeer ? @osmandin ? can you think of a better solution? Does this sound like a Jena issue? If so, @osmandin , could you create a ticket and/or patch for Jena?

ajs6f:
I don't see how that is not a Jena issue.

awoods:
@osmandin, can you track this down and create a Jena ticket and/or patch?
In the meantime, is the update in this current PR concerning?

cbeer:
I agree that it's a problem with Jena if they're not handling un-terminated domain names correctly and should be fixed there. I'd be really concerned with this fix we're setting ourselves at odds with the RDF spec. If I send

```
PREFIX dsc: <http://myurl.org>

INSERT { <> dsc:test "abc" }
```

My reading of the RDF/XML spec says that should expand to `http://myurl.orgabc`, not `http://myurl.org/abc` (yes, even though, arguably, no user in their right mind will want the former). Even if Jena is giving us bad namespace information, at least we're storing and returning the right values (per the spec).

cbeer:
I also initially thought we were adding the trailing slash to any URI, but it seems to just be bare domain names.  I still don't understand how that regexp works, but if this broke, e.g.

```
PREFIX ex: <http://some/uri>

INSERT { <> ex:test "abc" } # http://some/uritest
```

we would be at serious risk of breaking real-world data.

ajs6f:
And there are such URIS-- the3 JCR namespace is one.

awoods:
@osmandin, @ajs6f, @cbeer, can you move this forward? I am on-travel next week, starting at 7am tomorrow.

ajs6f:
Make either @cbeer or I the reviewer for this ticket.

cbeer:
I vote @ajs6f. In the meantime, should we revert f257425b72e939a0d159ff9b16de5fb1792fa213? I'd hate for someone to come to rely on that behavior to correct user error.

awoods:
Reverted with: https://github.com/fcrepo4/fcrepo4/commit/af48df0c6c564d46c1cea8777e4d88c1a1900dab

osmandin:
@cbeer @ajs6f @awoods 

Just to clarify, the regex would've converted <http://some/uri> into <http://some/uri/>.

cbeer:
@osmandin that was not the behavior I observed

osmandin:
@cbeer I re-tested the PR, and it seems to be creating the "correct" namespace by amending the URIs as intended. Did you already have that "esc" namespace defined? Do you recall what was the behavior you observed? 

The reason I ask is that if it wasn't working for you, we should probably update (only) the relevant IT (in master) by incorporating the URI you tried. This will help test any Jena patch. 

Steps I tried to verify the old intended fix(w/ d1dbb0d or github.com/osmandin/fcrepo4/commits/1381):

1. Create container "espresso"

2. Create file namespaces.rdf

PREFIX esc:<http://abc.org>
INSERT {
<>
esc:package "ppp";
}
WHERE { }


3. Update:

curl -i -X PATCH "http://localhost:8080/rest/espresso" -H "Content-Type: application/sparql-update" --data-binary "@namespace.sparql"

4. Observe in the UI resulting namespace PREFIX esc: <http://abc.org/> (in Update Properties) and property esc: package ppp (in Properties)

If you checkout master, instead, there should be the unexpected "ns001:abc.orgpackage".

awoods:
Probably neither behavior you note, @osmandin, is correct. The desired behavior in your scenario should be to find an unmodified namespace declaration of: `esc: http://abc.org`

osmandin:
@awoods I think both namespaces are equal (at least according to the IRI spec), but, yes, we shouldn't modify a string if we don't have to. I'll take a look at Jena again.

awoods:
@osmandin, I suspect you will find that if you request a sparql-update with two defined prefixes, like:
PREFIX esc0:<http://abc.org>
PREFIX esc1:<http://abc.org/>

They will both be registered.
Or rather, if you define the prefix:
PREFIX mix-mine: <http://www.jcp.org/jcr/mix/1.0/>

"mix-mine" will also be registered, although "mix: http://www.jcp.org/jcr/mix/1.0" already exists.

awoods:
There does not seem to be a good solution here, since even if we do register a namespace that does not end with a delimiter (/ or #) that namespace will never be found in subsequent logic.
For example, if the following is registered:
`PREFIX esc0:http://abc.org`
subsequent terms, such as:
`http://abc.orgtitle`
will not be detected as having the namespace: `http://abc.org`.
I would suggest we throw an exception when we detect the scenario of trying to register the namespace: `http://`.
Thoughts, @osmandin? @ajs6f? @cbeer?

cbeer:
-1. Namespaces are arbitrary string concatenations. We should infer no semantic meaning behind them and no client should be sad if we don't use their namespace prefixes (in fact, we reserve the right to ignore their suggestions)

awoods:
@cbeer, in this situation, what is your suggestion?

escowles:
If this only impacts namespace declarations without a / or # in them, and the updates work and it's just the prefix that's not used, then I think we should leave this as-is.  All of the updates so far look like they would break SPARQL Update that works now, so that doesn't seem like a good tradeoff to me.

cbeer:
I agree with @escowles. If you /really/ wanted to do something, you'd ignore the offending namespaces entirely for purposes of serializing into JCR properties and accept whatever the namespace minter wants to offer.

awoods:
It seems like buggy behavior when a user defines:
`PREFIX esc0: http://abc.org`
and the result is, an ignored `esc0` and new declaration:
`ns001 = http://`

At a minimum, I would second the suggestion to not register `http://` if it is found as a [predicate.getNameSpace()](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/converters/PropertyConverter.java#L116)

osmandin:
@awoods Your latest suggestion seems to be compatible with what @cbeer is saying. But we should probably also specify and logger.info() the reason the namespace would be ignored.

As a side note, @escowles, the proposed updates, now closed, won't break the SPARQL update (if by breaking you mean that an update won't work technically. There were an IT test that demonstrated the viability of the proposed change).

osmandin:
@awoods Not registering http:// is easy, but there are other cases as well, I think. For example, the relevant Jena method would  return http://myurl.org/u as http://myurl.org/, whereas the Jena namespace map contains http://myurl.org/u. So, http://myurl.org/u results in ns001:up namespace in JCR. 

If we make the policy that if the namespace obtained by predicate.getNameSpace() is not in the Jena namespace map, then other test classes that rely on ns001 (e.g. https://github.com/osmandin/fcrepo4/blob/master/fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/rdf/converters/PropertyConverterTest.java#L128) being present would break. 

So, it seems to me that we need a policy, rather than just ignoring "http://"

If you think that's not an issue, and just ignoring http:// is fine, then I'll go ahead.

awoods:
@osmandin, it is starting to sound like the best action is no action, in which case, we should minimally add documentation to the wiki detailing what are good namespace declarations, and what declarations will produce unexpected, system-generated prefixes.

osmandin:
@awoods, I have added a wiki page, and also PR 754 to remove the ignored IT tests that was meant to test any coding changes. 

If it's helpful to the user, one could also add a logger.info warning in PropertyConverter if an unexpected namespace prefix is returned by Jena.

awoods:
A log message would be great.

PR #740, opened 2015-03-01
Enable LDP indirect containers to update referenced container

awoods:
* Handle 'insertedContainerProperty' that has been converted to a reference property
* Provide current session to RdfContexts when constructed in FedoraResourceImpl

Resolves: https://jira.duraspace.org/browse/FCREPO-1383

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -2175,6 +2175,90 @@ public class FedoraLdpIT extends AbstractResourceIT {
     }
 
     @Test
+    public void testLdpIndirectContainerInteraction() throws Exception {
+
+        // Create resource (object)
+        final String resourceId = getRandomUniquePid();
+        final HttpResponse rResponse = createObject(resourceId);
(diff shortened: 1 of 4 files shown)


cbeer:
@awoods I'm confused about:

> Handle 'insertedContainerProperty' that has been converted to a reference property

Are we trying to deal with a scenario when the value of the `ldp:insertedContentRelation` is within the repository namespace, or something else?  Maybe I'm just missing the connection between the `LdpContainerRdfContext` change and the test you added..

awoods:
The basic issue is that the property provided in `ldp:insertedContentRelation` is not a `REFERENCE` type property, usually. The `insertedContainerProperty` https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/LdpContainerRdfContext.java#L140 (e.g. info:proxy/for) is actually persisted on the resource as "info:proxy/for_ref"

Therefore, when the "info:proxy/for" property is requested, https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/LdpContainerRdfContext.java#L187 , it is not found.

If you remove the updates from `LdpContainerRdfContext` and just run the new integration test, you will see that the implementation in `master` does not currently support IndirectContainers.

PR #741, opened 2015-03-03
Fcrepo 1290

yinlinchen:


fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/rdf/impl/PropertiesRdfContextTest.java
```diff
@@ -0,0 +1,118 @@
...
+
+
+    /*
+     * (non-Javadoc) test the nonRdfSource response are included both NonRdfSourceDescription and RdfSourceDescription
+     */
+    @Test
+    public void testFedoraBinaryProperties() throws RepositoryException {
+        final Model results = new PropertiesRdfContext(mockBinary, idTranslator).asModel();
+
+        assertTrue("Response contains both NonRdfSourceDescription and RdfSourceDescription", results
```
awoods: This test looks good. Please split this `assertTrue` into two assertions, instead of having `results.contains(...) && results.contains(...)`

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/rdf/impl/PropertiesRdfContextTest.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /*
+     * (non-Javadoc) test the FedoraResource response are included only RdfSourceDescription
+     */
+    @Test
+    public void testFedoraResourceProperties() throws RepositoryException {
+
+        final Model results = new PropertiesRdfContext(mockResource, idTranslator).asModel();
+        assertTrue("Response contains RdfSourceDescription", results
```
awoods: Use the same two assertions from the test above, but make the second assertion be: `assertFalse(...)`

fcrepo-kernel-impl/src/test/java/org/fcrepo/integration/kernel/impl/FedoraResourceImplIT.java
```diff
@@ -647,6 +647,24 @@ public void testGetChildrenWithBinary() {
...
     @Test
+    public void testGetChildrenWithBinaryCustomProperties() throws RepositoryException {
+        final String pid = getRandomPid();
+        final Container container = containerService.findOrCreate(session, "/" + pid);
+        final FedoraResource resource = binaryService.findOrCreate(session, "/" + pid + "/a");
+        final javax.jcr.Node node = resource.getNode();
+        node.setProperty("dc:title", "this-is-some-title");
+
+        final Graph graph =
+                container.getChildren().next().getTriples(subjects, PropertiesRdfContext.class).asModel().getGraph();
```
ajs6f: Do you want to rely on `next()` for this or do you want to retrieve the child by name?

fcrepo-kernel-impl/src/test/java/org/fcrepo/integration/kernel/impl/FedoraResourceImplIT.java
```diff
@@ -647,6 +647,24 @@ public void testGetChildrenWithBinary() {
...
+
+        //jcr property
+        final Node s = createGraphSubjectNode(container.getChildren().next());
```
ajs6f: Ditto.

fcrepo-kernel-impl/src/test/java/org/fcrepo/integration/kernel/impl/FedoraResourceImplIT.java
```diff
@@ -647,6 +647,24 @@ public void testGetChildrenWithBinary() {
...
+        final Node p = DC_TITLE.asNode();
+        final Node o = createLiteral("this-is-some-title");
+        assertTrue(graph.contains(s, p, o));
```
ajs6f: Leave an error message here for use in case of failure that explains why we _should_ see the triple included.

fcrepo-kernel-impl/src/test/java/org/fcrepo/integration/kernel/impl/FedoraResourceImplIT.java
```diff
@@ -647,6 +647,24 @@ public void testGetChildrenWithBinary() {
     }
 
     @Test
+    public void testGetChildrenWithBinaryCustomProperties() throws RepositoryException {
+        final String pid = getRandomPid();
+        final Container container = containerService.findOrCreate(session, "/" + pid);
+        final FedoraResource resource = binaryService.findOrCreate(session, "/" + pid + "/a");
+        final javax.jcr.Node node = resource.getNode();
+        node.setProperty("dc:title", "this-is-some-title");
```
ajs6f: You don't actually need the `javax.jcr.Node` but the once, so I would just do `resource.getNode().setProperty("dc:title", "this-is-some-title");`, but it's a matter of style.

fcrepo-kernel-impl/src/test/java/org/fcrepo/integration/kernel/impl/FedoraResourceImplIT.java
yinlinchen: This test is to make sure if the child is a nonRdfSource with a dc:title, the request does include the child's title. I don't need to retrieve the child by name.

ajs6f: I know. You're assuming that the child will be the _first_ child of the container, and it now is, but it may or may not always be-- you're assuming that we will never change the code to add other children in some other way in between your method calls and that the JCR iteration order won't ever change. I'm not saying your test doesn't work now, I'm saying it's a bit brittle unless you specify which child to retrieve.

fcrepo-kernel-impl/src/test/java/org/fcrepo/integration/kernel/impl/FedoraResourceImplIT.java
ajs6f: You might want to add a check here to see that the child did indeed come back as a non-RDF source.

fcrepo-kernel-impl/src/test/java/org/fcrepo/integration/kernel/impl/FedoraResourceImplIT.java
yinlinchen: I am not assuming that the child will be the first child. A parent can have multiple children and there maybe just one  child with dc:title among them. I was not think of retrieve the child by name because the change I made in  PropertiesRdfContext.java was Iterators...

ajs6f: You _are_ assuming that. You are only testing one child, the first one to be returned by `next()`. That's the "first child". If that happens not to be the binary-child you created then the test will fail even though the software may be operating correctly. It would be stronger to retrieve all the children and filter to the one that has the name you gave your binary-child, then check _it_ for the title you think it should have. Let's leave it at that.

awoods: Running this new test in `FedoraResourceImplIT` against the master branch successfully passes... therefore it does not demonstrate a fix to the issue of:
https://jira.duraspace.org/browse/FCREPO-1290

I suggest removing the updates to this class, and instead adding the integration test provided here:
https://github.com/yinlinchen/fcrepo4/pull/4

ajs6f: That's interesting. It seems to imply that the problem is somewhere in whatever is producing triples from resource-children. I think that's `ChildrenRdfContext`?

yinlinchen: The test was assuming there is only one child and that's not stronger enough. I will update this test to retrieve all the children and filter to the one that has the name and check it. Thanks for the comments.

ajs6f: Wait-- did you not see @awoods' comment?

yinlinchen: Didn't see it, but see it now.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/b94778562793ea0163a6f7aeab66d6544766b08b
and
https://github.com/fcrepo4/fcrepo4/commit/bea17d0c1f950dc6892bb33476acbc6fe92e15dd

PR #744, opened 2015-03-06
Surgical multi-relationship removal - Addresses FCREPO-1362

whikloj:
See: [FCREPO-1362](https://jira.duraspace.org/browse/FCREPO-1362)

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
@@ -57,7 +57,6 @@ import com.google.common.base.Converter;
 import com.google.common.base.Function;
 import com.google.common.base.Predicate;
 import com.google.common.collect.Iterators;
-import com.hp.hpl.jena.rdf.model.RDFNode;
 import com.hp.hpl.jena.rdf.model.Resource;
 
 import org.fcrepo.kernel.FedoraJcrTypes;
@@ -68,8 +67,6 @@ import org.fcrepo.kernel.exception.MalformedRdfException;
 import org.fcrepo.kernel.exception.PathNotFoundRuntimeException;
 import org.fcrepo.kernel.exception.RepositoryRuntimeException;
 import org.fcrepo.kernel.identifiers.IdentifierConverter;
-import org.fcrepo.kernel.impl.rdf.converters.ValueConverter;
-import org.fcrepo.kernel.impl.rdf.impl.DefaultIdentifierTranslator;
 import org.fcrepo.kernel.impl.utils.JcrPropertyStatementListener;
 import org.fcrepo.kernel.services.functions.JcrPropertyFunctions;
 import org.fcrepo.kernel.utils.iterators.GraphDifferencingIterator;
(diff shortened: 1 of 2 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
```diff
@@ -257,7 +260,22 @@ public void delete() {
             final Iterator<Property> inboundProperties = Iterators.concat(references, weakReferences);
 
             while (inboundProperties.hasNext()) {
-                inboundProperties.next().remove();
+                final Property prop = inboundProperties.next();
+                final List<Value> newVals = new ArrayList<Value>();
```
ajs6f: Use the diamond operator: `new ArrayList<>()`.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
```diff
@@ -257,7 +260,22 @@ public void delete() {
...
+                final Iterator<Value> propIt = JcrPropertyFunctions.property2values.apply(prop);
```
ajs6f: Use a static import.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
```diff
@@ -257,7 +260,22 @@ public void delete() {
...
+                while (propIt.hasNext()) {
+                    final Value v = propIt.next();
+                    final Node n = this.getSession().getNodeByIdentifier(v.getString());
```
ajs6f: You can drop the `this` for just `getSession()`.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
```diff
@@ -257,7 +260,22 @@ public void delete() {
             final Iterator<Property> inboundProperties = Iterators.concat(references, weakReferences);
 
             while (inboundProperties.hasNext()) {
-                inboundProperties.next().remove();
+                final Property prop = inboundProperties.next();
+                final List<Value> newVals = new ArrayList<Value>();
+                final Iterator<Value> propIt = JcrPropertyFunctions.property2values.apply(prop);
+                while (propIt.hasNext()) {
```
ajs6f: Given how few references there are likely to be, you might consider rolling up the iterator into a collection and then `remove`ing the one you don't want.

whikloj: Is that of any benefit? I don't think I can use remove(Object) so I am still iterating over the list. But I might be missing something.

ajs6f: Oh, shoot, you're right. You can't. But you can still do better than this via [Guava's `removeIf`](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/Iterators.html#removeIf%28java.util.Iterator,%20com.google.common.base.Predicate%29).

whikloj: Ok, that is interesting, but I'm not sure that using a Predicate doesn't make the code more confusing.
Also the **property2values** function returns an UnmodifiableIterator, so the **removeIf** fails to modify it.
I put this attempt in a [separate branch](https://github.com/whikloj/fcrepo4/blob/FCREPO-1362-2/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java#L262-L283) so you can see what I've done, incase I missed something.

ajs6f: Oh, I didn't know that `property2values` did that-- that's lame. It's not worth worrying about this any more anyway- this all goes away with Java 8's Streams, so this is limited-lifetime code. I'm sorry I took up this much time with it.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/4a77e57d18687b6e4f653e0d88765d007e284584

PR #746, opened 2015-03-10
Make header and body markup consistent with BS standards

kevinglover:
This is a minor change that will allow for better presentation. The nav element should be exclusive of the container element for flow to be consistent with other sites using bootstrap.


awoods:
@kevinglover, could you please create a JIRA ticket describing the issue and fix.
https://jira.duraspace.org/browse/FCREPO
Thanks for this PR!

awoods:
Addresses: https://jira.duraspace.org/browse/FCREPO-1400

PR #747, opened 2015-03-12
Fail on Sparql update for subjects outside repository domain

whikloj:
Addresses [FCREPO-1361](https://jira.duraspace.org/browse/FCREPO-1361)


ajs6f:
Because you're PRing from your repo and not from origin, I can't make line comments, but if I could, I would suggest logging the exception before throwing it. This is so that in debugging situations, it's clear where the problem happened. Remember that exceptions thrown out of that listener don't actually get reported out until quite a ways downstream.

whikloj:
Gah. Sorry about that, I was confused as you had commented inline to my changes to the ontology repo, but the restrictions on it maybe different. I have added in some debug statements.

ajs6f:
No prob-- it's because I don't have write access to your clone of fcrepo4. Not to be a picky person, but maybe ERROR is more appropriate for this than DEBUG?

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/2e5265c8f78e111e1d0030207ad32be105bb26cd

PR #748, opened 2015-03-13
BNodes are BNodes and should be treated as such

ajs6f:
DO NOT MERGE YET


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/Deskolemizer.java
```diff
@@ -0,0 +1,115 @@
...
+                public Resource get() {
+                    return context.createResource();
+                }
+            }));
+
+    private static final Logger log = getLogger(Deskolemizer.class);
+
+    /**
+     * @param idTranslator
+     * @param context
```
awoods: Please change `context` to `model`
...and add meaning descriptions.

ajs6f: "Meaning descriptions"?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/Deskolemizer.java
```diff
@@ -0,0 +1,115 @@
...
+        if (isSkolem(n)) {
+            log.debug("Replacing {} with bnode.", n);
+            return bnodeSubstitutions.getUnchecked(n.asResource());
+        }
+        return n;
+    }
+
+    private boolean isSkolem(final RDFNode n) {
+        return n.isURIResource() &&
+                (n.asResource().getURI().indexOf('?') == -1) &&
```
awoods: Please add comments detailing the "why" of these criteria.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/Deskolemizer.java
```diff
@@ -0,0 +1,115 @@
...
+                idTranslator.inDomain(n.asResource()) &&
+                !idTranslator.asString(n.asResource()).contains("/fcr:") &&
+                idTranslator.convert(n.asResource()).hasType("fedora:Skolem");
```
awoods: Please use `FedoraJcrTypes.FEDORA_SKOLEMNODE` instead of the String literal.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
```diff
@@ -457,18 +473,55 @@ public Boolean isNew() {
...
+                    }
+                }, inputModel.listStatements())));
+
+        for (final Resource skolemNode : skolemizer.get()) {
+            LOGGER.debug("Checking for skolem node: {}", skolemNode);
+            final String skolemPath = idTranslator.asString(skolemNode);
+            if (!hashAndSkolemCreator.exists(getSession(), skolemPath)) {
+                LOGGER.debug("Creating skolem node at: {}", skolemPath);
+                try {
+                    hashAndSkolemCreator.findOrCreate(getSession(), skolemPath).getNode().addMixin("fedora:Skolem");
```
awoods: Please use `FedoraJcrTypes.FEDORA_SKOLEMNODE` instead of the String literal.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/utils/JcrPropertyStatementListener.java
```diff
@@ -97,9 +102,30 @@ public void addedStatement(final Statement input) {
...
+            final Statement s = skolemizer.apply(hashURIDetector.apply(input));
+
+            // create all needed skolem nodes
+            for (final Resource skolemNode : skolemizer.get()) {
+                LOGGER.debug("Checking for skolem node: {}", skolemNode);
+                final String path = idTranslator.asString(skolemNode);
+                LOGGER.debug("under path: {}", path);
+                if (!skolemAndHashCreator.exists(session, path)) {
+                    LOGGER.debug("Creating skolem node at: {}", path);
+                    skolemAndHashCreator.findOrCreate(session, path).getNode().addMixin("fedora:Skolem");
```
awoods: Please use `FedoraJcrTypes.FEDORA_SKOLEMNODE` instead of the String literal.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/Deskolemizer.java
ajs6f: Good catch.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/rdf/DeskolemizerTest.java
```diff
@@ -0,0 +1,176 @@
...
+    @Mock
+    private FedoraResource mockNonSkolemFedoraResource;
+
+    private static final Model testModel = createDefaultModel();
+
+    private static final Node skolemNode = createURI("info:skolem");
+
+    @Before
+    public void setUp() {
+        when(mockSkolemFedoraResource.hasType("fedora:Skolem")).thenReturn(true);
```
awoods: Please use `FedoraJcrTypes.FEDORA_SKOLEMNODE` instead of the String literal.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/rdf/DeskolemizerTest.java
```diff
@@ -0,0 +1,176 @@
...
+        when(mockNonSkolemFedoraResource.hasType("fedora:Skolem")).thenReturn(false);
```
awoods: Please use `FedoraJcrTypes.FEDORA_SKOLEMNODE` instead of the String literal.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/rdf/DeskolemizerTest.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    @Test
+    public void tripleWithNoBNodeOperations() {
+        when(mockIdTranslator.inDomain(any(Resource.class))).thenReturn(true);
+        when(mockIdTranslator.asString(any(Resource.class))).thenReturn("non-fcr-URI");
+        when(mockIdTranslator.convert(any(Resource.class))).thenReturn(mockSkolemFedoraResource);
+        when(mockSkolemFedoraResource.hasType("fedora:Skolem")).thenReturn(false);
```
awoods: Please use `FedoraJcrTypes.FEDORA_SKOLEMNODE` instead of the String literal.

fcrepo-kernel-impl/src/test/java/org/fcrepo/integration/kernel/impl/FedoraResourceImplIT.java
```diff
@@ -578,7 +584,8 @@ public boolean apply(final javax.jcr.Property property) {
 
         final javax.jcr.Node skolemizedNode = session.getNodeByIdentifier(values[0].getString());
 
-        assertTrue(skolemizedNode.getPath().contains("/.well-known/genid/"));
+        assertTrue(Lists.transform(asList(skolemizedNode.getMixinNodeTypes()), toStringFunction()).contains(
+                "fedora:Skolem"));
```
awoods: Please use `FedoraJcrTypes.FEDORA_SKOLEMNODE` instead of the String literal.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/Skolemizer.java
```diff
@@ -0,0 +1,130 @@
...
+ * instance of this class should be used only with a contextual topic of one single resource and only for one
+ * document's scope of RDF about that resource.
+ *
+ * @author ajs6f
+ */
+public class Skolemizer implements Function<Statement, Statement>, Supplier<Set<Resource>> {
+
+    private final RDFVisitor nodeSkolemizer;
+
+    public static final Resource SKOLEM_TYPE = createResource(REPOSITORY_NAMESPACE + "skolem");
```
awoods: Remove unused constant `SKOLEM_TYPE`.

ajs6f: I think you're looking at a previous commit. That's not there anymore.

---
A. Soroka
The University of Virginia Library

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -192,7 +193,7 @@ public Triple apply(final Statement input) {
         }
         servletResponse.addHeader("Vary", "Accept, Range, Accept-Encoding, Accept-Language");
 
-        return Response.ok(rdfStream).build();
+        return ok(rdfStream.withThisContext(rdfStream.transform(new Deskolemizer(idTranslator, null)))).build();
```
awoods: Does the logic here, at the fcrepo-http-api level, imply that blank nodes are not completely handled a level deeper within the kernel... for clients that interact with Fedora objects directly as opposed through HTTP?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -295,7 +296,9 @@ public RdfStream apply(final FedoraResource child) {
         }
 
 
-        return rdfStream;
+        return rdfStream.withThisContext(rdfStream.transform(new Deskolemizer(idTranslator, resource == null ? null
```
awoods: ditto from above

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
ajs6f: Hm. Very good point. Let me see if I can move this deeper, out of HTTP. I think it very likely that I can.

ajs6f: I'm coming around to see this differently-- I now don't think this stuff is out of place in the HTTP layer. This is on the reasoning that bnodes are scoped entirely to the _document_ in which they appear, and there is no notion of document in our code below the HTTP layer… I think I could make this look better in the HTTP layer, but I really do think now that it belongs here.
I think that clients interacting at a deeper layer need to handle deskolemization themselves, _because only those clients themselves will know what the appropriate scope is in which to deskolemize_.
If this is to be handled in the kernel, what is the notion of scope in play? What is a "document" in the kernel? We operate only over streams of triples in the kernel…

awoods: Thanks for the further thought, @ajs6f. Let's go with your assumption until we have a pressing need to do otherwise.

ajs6f: In that case, I will rebase and maybe tighten up how this works in the HTTP layer and we should be able to move forward.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -141,6 +146,10 @@
 
     protected abstract String externalPath();
 
+    protected Deskolemizer deskolemizer() {
+        return deskolemizer == null ? deskolemizer = new Deskolemizer(translator(), null) : deskolemizer;
```
awoods: Please unpack this single statement into its intelligible, multi-line parts.

ajs6f: You should make such style considerations clear and get them into Checkstyle rules, if you intend to enforce them.

awoods: That would be great if we could encode subjective measures of "readability" into checkstyles. I try not to encroach on other's style preferences, but when a style impedes personal readability, I raise the flag.
Thanks for the niggly update.

ajs6f: If you are objecting to the ternary operator, then that should be encoded in Checkstyle. If, on the other hand, your qualm is genuinely subjective, it's not a valid criticism.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
```diff
@@ -49,12 +49,9 @@
     protected abstract Session session();
 
     protected IdentifierConverter<Resource, FedoraResource> translator() {
-        if (idTranslator == null) {
-            idTranslator = new HttpResourceConverter(session(),
-                    uriInfo.getBaseUriBuilder().clone().path(FedoraLdp.class));
-        }
-
-        return idTranslator;
+        return idTranslator == null ? idTranslator =
```
awoods: For the reasons above (and the fact that no logic is being changed here), please revert this update.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/Deskolemizer.java
```diff
@@ -0,0 +1,136 @@
...
+import com.hp.hpl.jena.rdf.model.Statement;
+
+/**
+ * Deskolemization is abstractly a function from RDF nodes to RDF nodes, but here we implement it, purely for
+ * convenience of operation, as a function from triples to triples. This object should be used to translate only one
+ * document's scope of RDF.
+ *
+ * @author ajs6f
+ */
+public class Deskolemizer implements Function<Triple, Triple> {
```
awoods: Shouldn't `Deskolemizer` implement `Deskolemizer`?
...which would strongly argue for a different name of the implementing class, e.g. `DeskolemizerImpl`.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/identifiers/Deskolemizer.java
```diff
@@ -0,0 +1,30 @@
...
+import com.hp.hpl.jena.graph.Triple;
+
+/**
+ * Deskolemization is abstractly a function from RDF nodes to RDF nodes, but here we implement it, purely for
+ * convenience of operation, as a function from triples to triples. An implementing object should be used to
+ * translate only one document's scope of RDF.
+ *
+ * @author ajs6f
+ */
+public interface Deskolemizer extends Function<Triple, Triple> {
```
awoods: ...alternatively, this `interface` could be removed, as it is not used.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/Deskolemizer.java
ajs6f: Yes, this is an oversight. As far as the name goes, I don't know how much difference it makes. The only reason that there is a `Deskolemizer` interface is because the workings of this process tangle up into the kernel API, not because we have any reasonable expectation of more than one implementation of deskolemization. There is no clean way to introduce blank nodes, and this code is crap, but it's as good as I think we can expect to do. If you think it clearer, I can certainly change the name here to `DeskolemizerImpl`.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/identifiers/Deskolemizer.java
ajs6f: Yes, I think it's probably cleaner to just remove it, since the type being passed around in the kernel is now `Service<Container>`.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/Deskolemizer.java
awoods: I would either suggest:
1. Change the implementation name to `DeskolemizerImpl` and have it implement `Deskolemizer`, or
2. Simply remove the `Deskolemizer` interface.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/utils/JcrPropertyStatementListener.java
```diff
@@ -50,33 +57,32 @@
...
-     * @param idTranslator the id translator
-     * @param session the session
-     */
-    public JcrPropertyStatementListener(final IdentifierConverter<Resource, FedoraResource> idTranslator,
-                                        final Session session) {
-        this(idTranslator, new JcrRdfTools(idTranslator, session));
-    }
+    private final Skolemizer skolemizer;
+
+    private final Service<Container> skolemAndHashCreator;
```
awoods: Given its type and use of `skolemAndHashCreator` in this class, I would find it more clear if the variable were named `containerService`.

ajs6f: I disagree. `containerService` is just a restatement of the type of the param. `skolemAndHashCreator` actually explains how the param is used. I'd like to hear from someone else.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/Skolemizer.java
```diff
@@ -0,0 +1,126 @@
...
+    private final Set<Resource> skolemNodes = new HashSet<>();
+
+    private static final Logger log = getLogger(Skolemizer.class);
+
+    /**
+     * @param topic The URI of the contextual topic of the RDF supplied to this code. Blank nodes will be skolemized
+     *        to identifiers rooted at this URI.
+     */
+    public Skolemizer(final Resource topic) {
+        final String prefix = topic + "/" + randomUUID().toString().replace('-', '/');
```
awoods: I have not probed more deeply yet, but I am concerned that `Resource.toString()` here will create a String that is not compatible with JCR path creation performed in `JcrPropertyStatementListener.addedStatement()`.
https://github.com/apache/jena/blob/master/jena-core/src/main/java/com/hp/hpl/jena/rdf/model/Resource.java#L95-L104

Hopefully this is a non-issue.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -141,6 +146,10 @@
 
     protected abstract String externalPath();
 
+    protected Deskolemizer deskolemizer() {
```
awoods: It does not appear that this method is used outside of this class. Is there a reason for not making it `private`?

ajs6f: Sure, why not?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/utils/FedoraTypesUtils.java
```diff
@@ -299,31 +301,14 @@ public static boolean isReferenceProperty(final Node node, final String property
...
-                existingAncestorPath.append(pathSegment);
-                if (i != (len - 1)) {
-                existingAncestorPath.append("/");
-                }
-            } else {
-                if (i != 0) {
-                    existingAncestorPath.deleteCharAt(existingAncestorPath.length() - 1);
-                }
-                break;
+        String potentialPath = path.startsWith("/") ? path : "/" + path;
```
awoods: I am not sure I see the relevance of this update in this PR... but I guess it is an improvement <shrug/>

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/ceacc9832890906c8dee8f4533f5b33679d9b875

PR #753, opened 2015-03-23
Ignore wrong namespace prefix from Jena for JCR namespace registration

osmandin:
fcrepo-1381


osmandin:
Closing it in favor of documentation.

PR #757, opened 2015-03-27
Return 501 upon POST for LDPR

osmandin:
fcrepo-1396


awoods:
See: https://github.com/osmandin/fcrepo4/pull/4

We will also want to catch exceptions that come from an invalid Link header giving a clear message and 4xx code to the client.

awoods:
@osmandin, we should also include `checkLinkForLdpResourceCreation()` as a check on the `PUT` method: `createOrReplaceObjectRdf()`

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/c2bd525097b8e0cd553e542c63ae218ec8823115

PR #760, opened 2015-03-31
Adding userID and userAgent to events

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-1420

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
@@ -16,7 +16,9 @@
 package org.fcrepo.http.api;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.gson.JsonObject;
 import com.hp.hpl.jena.rdf.model.Resource;
+import org.apache.commons.lang.StringUtils;
 import org.fcrepo.http.commons.AbstractResource;
 import org.fcrepo.http.commons.api.rdf.HttpResourceConverter;
 import org.fcrepo.kernel.models.FedoraResource;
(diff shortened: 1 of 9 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
```diff
@@ -74,13 +75,15 @@ public FedoraResource getResourceFromPath(final String externalPath) {
     /**
      * Set the baseURL for JMS events.
      * @param uriInfo the uri info
+     * @param headers HTTP headers
      **/
-    protected void setUpJMSBaseURIs(final UriInfo uriInfo) {
+    protected void setUpJMSBaseURIs(final UriInfo uriInfo, final HttpHeaders headers) {
```
ajs6f: Is this method name still accurate?

fcrepo-jms/src/main/java/org/fcrepo/jms/headers/DefaultMessageFactory.java
```diff
@@ -79,36 +93,23 @@ private void setBaseURL(final FedoraEvent event) {
...
-                LOGGER.debug("MessageFactory baseURL: {}", baseURL);
+                this.userAgent = json.get("userAgent").getAsString();
+                LOGGER.debug("MessageFactory baseURL: {}, userAgent: {}", baseURL, userAgent);
 
             } else {
-                LOGGER.warn("MessageFactory baseURL is empty!");
+                LOGGER.warn("MessageFactory event UserData is empty!");
             }
 
         } catch ( final Exception ex ) {
```
ajs6f: Is there a narrower exception we can catch here?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
escowles: I can rename it to something like setUpJMSInfo.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
```diff
@@ -74,13 +75,15 @@ public FedoraResource getResourceFromPath(final String externalPath) {
...
         try {
             final URI baseURL = uriInfo.getBaseUri();
             LOGGER.debug("setting baseURL = " + baseURL.toString());
             final ObservationManager obs = session().getWorkspace().getObservationManager();
-            final String json = "{\"baseURL\":\"" + baseURL.toString() + "\"}";
+            final String json = "{\"baseURL\":\"" + baseURL.toString() + "\","
+                    + "\"userAgent\":\"" + headers.getHeaderString("user-agent") + "\"}";
```
barmintor: Assuming this should be conditional on the presence of the header?

fcrepo-jms/src/main/java/org/fcrepo/jms/headers/DefaultMessageFactory.java
escowles: JsonParser.parse throws JsonParseException, but the accessors throw IllegalStateExceptions -- so we could narrow to RuntimeException.

ajs6f: I suppose that's a bit better. At least it won't trigger one of those
annoying Sonar warnings.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
escowles: Yes, and it would be better to build a JSON object instead of doing the string quoting ourselves.

PR #763, opened 2015-04-03
Collapsing node and fcr:content events to prevent duplicate messages

escowles:
Updating event machinery to collapse events about content nodes and the fcr:content property so adding/removing content only produces a single event

Fixes https://jira.duraspace.org/browse/FCREPO-1440


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/eventmappings/AllNodeEventsOneEvent.java
```diff
@@ -59,7 +61,9 @@
         @Override
         public String apply(final Event ev) {
             try {
-                final String id = ev.getIdentifier();
+                // build id from nodepath+user to collapse multiple nodes from adding/removing content nodes
+                final String id = ev.getPath().replaceAll("/jcr:content/","/").replaceAll(propPattern,"")
+                        + "-" + ev.getUserID();
```
awoods: In doing some manual testing, it looks like NonRdfSources named "/book/ucsd:page" are being incorrectly sorted down to "/book".

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/3de050c781632dedf8e182ac099406017a1e22b6

PR #766, opened 2015-04-08
adding fixity events II

yulgit1:


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraBinaryImpl.java
```diff
@@ -270,14 +276,20 @@ public String getFilename() {
     }
 
     @Override
-    public RdfStream getFixity(final IdentifierConverter<Resource, FedoraResource> idTranslator) {
-        return getFixity(idTranslator, getContentDigest(), getContentSize());
+    public RdfStream getFixity(final IdentifierConverter<Resource, FedoraResource> idTranslator,
+                               final String baseURL,
+                               final String userAgent,
+                               final EventBus eventBus) {
```
ajs6f: @awoods These kinds of shenanigans in parameter lists are why we should think about bringing the core model classes into container management. I had to do something similar with the blank nodes thing.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraBinaryImpl.java
```diff
@@ -270,14 +276,20 @@ public String getFilename() {
...
     }
 
     @Override
     public RdfStream getFixity(final IdentifierConverter<Resource, FedoraResource> idTranslator,
                                final URI digestUri,
-                               final long size) {
+                               final long size,
+                               final String baseURL,
+                               final String userAgent,
+                               final EventBus eventBus) {
```
escowles: Passing this from the REST API into FedoraBinaryImpl seems clunky to me -- maybe @awoods or @ajs6f know why injecting it didn't work for me or @yulgit1 ?

ajs6f: Because instances of the core model types are not managed by the injection container. We could do that, but it would be a massive change in the core. @awoods and I have talked about it, but made no move.

ajs6f: I had to do something similar in the blank node PR. It is not just clunky, it's basically wrong.

awoods: @ajs6f, once the blanknode PR gets in, maybe putting core model types under container management would be an interesting project for you to tackle?

ajs6f: I'm not sure I'll have the time soon. Let's talk about it in the next week.

awoods:
Closing this PR in favor of the approaches defined in these two tickets:
- https://jira.duraspace.org/browse/FCREPO-1737
- https://jira.duraspace.org/browse/FCREPO-1738

PR #767, opened 2015-04-09
File event cleanup

escowles:
* fixing fcr:content/jcr:content mixup
* tagging content events with fedora:hasContent property

Fixes https://jira.duraspace.org/browse/FCREPO-1445

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/eventmappings/AllNodeEventsOneEvent.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/eventmappings/AllNodeEventsOneEvent.java
@@ -124,7 +124,7 @@ public class AllNodeEventsOneEvent implements InternalExternalEventMapper {
 
         private void addProperty( final FedoraEvent fedoraEvent, final Event ev ) {
             try {
-                if (ev.getPath().indexOf(JCR_CONTENT) != -1) {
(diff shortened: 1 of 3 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/eventmappings/AllNodeEventsOneEvent.java
```diff
@@ -123,6 +124,9 @@ public void remove() {
 
         private void addProperty( final FedoraEvent fedoraEvent, final Event ev ) {
             try {
+                if (ev.getPath().indexOf(JCR_CONTENT) != -1) {
```
ajs6f: Why not simply `String.contains()` here?

awoods:
Why did the integration test in:
https://github.com/fcrepo4/fcrepo4/commit/3de050c781632dedf8e182ac099406017a1e22b6

not catch this?

escowles:
@awoods I think the IT doesn't sufficiently capture the way the REST API creates binary content to trigger emitting both nodes.  I'll see if I can trigger the bug in the current IT, and if not then I'll add a new IT to fcrepo-webapp (where we have both the REST API and JMS available).

awoods:
@escowles, thanks. We should hold off on merging this PR until you come back with more IT updates.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/67494888385d1541f034f0340fa6298c84dcf6e5

PR #768, opened 2015-04-14
1411

osmandin:
DO NOT MERGE. Just for testing. Please ignore style issues.

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
@@ -59,6 +59,8 @@ import com.google.common.base.Predicate;
 import com.google.common.collect.Iterators;
 import com.hp.hpl.jena.rdf.model.Resource;
 
+import com.hp.hpl.jena.rdf.model.Statement;
+import com.hp.hpl.jena.rdf.model.StmtIterator;
 import org.fcrepo.kernel.FedoraJcrTypes;
 import org.fcrepo.kernel.models.NonRdfSourceDescription;
 import org.fcrepo.kernel.models.FedoraBinary;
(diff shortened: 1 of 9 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
```diff
@@ -465,10 +467,23 @@ public Boolean isNew() {
...
         final Model inputModel, final RdfStream originalTriples) throws MalformedRdfException {
 
+        LOGGER.info("Resource path: {}", getPath());
+
+        final StmtIterator stmtIterator = inputModel.listStatements();
+
+        while (stmtIterator.hasNext()) {
+            final Statement s = stmtIterator.next();
+            final Resource subject = s.getSubject();
+            if (!subject.toString().contains(getPath())) {
```
ajs6f: This test is badly wrong.

- It is comparing a JCR value (the result of `getPath()`) with an RDF value (`subject.toString()`). That kind of comparison _must_ be mediated by an `IdentifierConverter`. Use something more like `equals(idTranslator.convert(subject))`, catching exceptions, or pass in the request URI and do a comparison to that.
- It does not account for blank nodes.
- It does not account for hash-URIs

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
```diff
@@ -465,10 +467,23 @@ public Boolean isNew() {
     public void replaceProperties(final IdentifierConverter<Resource, FedoraResource> idTranslator,
         final Model inputModel, final RdfStream originalTriples) throws MalformedRdfException {
 
+        LOGGER.info("Resource path: {}", getPath());
+
+        final StmtIterator stmtIterator = inputModel.listStatements();
```
ajs6f: if you are going to use  a loop here at all, use `listSubjects()` instead.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
```diff
@@ -465,10 +467,23 @@ public Boolean isNew() {
...
+
+        while (stmtIterator.hasNext()) {
+            final Statement s = stmtIterator.next();
+            final Resource subject = s.getSubject();
+            if (!subject.toString().contains(getPath())) {
+                LOGGER.info("Input model subject not target of resource: {}", subject.toString());
+                throw new UnsupportedOperationException("Non targetted resource " + subject.toString());
```
ajs6f: Let's subclass this to something more specific, a la `NonLocalTriplesException` or the like.

PR #769, opened 2015-04-15
Adding SuppressByMixinFilter to suppress events related to nodes with any of a given set of mixins

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-1456

### DIFF
--- a/fcrepo-kernel-impl/pom.xml
+++ b/fcrepo-kernel-impl/pom.xml
@@ -122,8 +122,6 @@
     <dependency>
       <groupId>com.jayway.awaitility</groupId>
       <artifactId>awaitility</artifactId>
-      <version>${awaitility.version}</version>
-      <scope>test</scope>
     </dependency>
   </dependencies>
 
--- a/fcrepo-kernel-impl/pom.xml
+++ b/fcrepo-kernel-impl/pom.xml
@@ -119,6 +119,12 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.jayway.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <version>1.6.2</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/fcrepo-kernel-impl/pom.xml
+++ b/fcrepo-kernel-impl/pom.xml
@@ -122,7 +122,7 @@
     <dependency>
       <groupId>com.jayway.awaitility</groupId>
       <artifactId>awaitility</artifactId>
-      <version>1.6.2</version>
+      <version>${awaitility.version}</version>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/DefaultFilter.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/DefaultFilter.java
@@ -38,7 +38,6 @@ import org.fcrepo.kernel.observer.EventFilter;
 import org.slf4j.Logger;
 
 import java.util.Collection;
-import java.util.List;
 
 /**
  * {@link EventFilter} that passes only events emitted from nodes with a Fedora
@@ -79,10 +78,7 @@ public class DefaultFilter implements EventFilter {
     @Override
     public boolean apply(final Event event) {
         try {
-            final org.modeshape.jcr.api.observation.Event modeEvent = getJcr21Event(event);
-
-            final List<NodeType> nodeTypes = ImmutableList.copyOf(modeEvent.getMixinNodeTypes());
-            final Collection<String> mixinTypes = ImmutableSet.copyOf(transform(nodeTypes, nodetype2string));
+            final Collection<String> mixinTypes = getMixinTypes(event);
             return mixinTypes.contains(FEDORA_RESOURCE)
                     || mixinTypes.contains(FEDORA_BINARY)
                     || mixinTypes.contains(FEDORA_NON_RDF_SOURCE_DESCRIPTION)
@@ -95,9 +91,13 @@ public class DefaultFilter implements EventFilter {
         }
     }
 
-    private static org.modeshape.jcr.api.observation.Event getJcr21Event(final Event event) {
+    protected static Collection<String> getMixinTypes(final Event event)
+            throws PathNotFoundException, RepositoryException {
         try {
-            return (org.modeshape.jcr.api.observation.Event) event;
+            final org.modeshape.jcr.api.observation.Event modeEvent =
+                    (org.modeshape.jcr.api.observation.Event) event;
+            return ImmutableSet.copyOf(transform(ImmutableList.copyOf(modeEvent.getMixinNodeTypes()),
+                    nodetype2string));
         } catch (final ClassCastException e) {
             throw new ClassCastException(event + " is not a Modeshape Event");
         }
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/DefaultFilter.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/DefaultFilter.java
@@ -15,8 +15,10 @@
  */
 package org.fcrepo.kernel.impl.observer;
 
+import static com.google.common.base.Functions.toStringFunction;
 import static com.google.common.base.Throwables.propagate;
-import static com.google.common.collect.Iterables.transform;
+import static com.google.common.collect.Collections2.transform;
+import static com.google.common.collect.Sets.newHashSet;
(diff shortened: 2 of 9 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/DefaultFilter.java
```diff
@@ -95,9 +91,13 @@ public boolean apply(final Event event) {
...
     }
 
-    private static org.modeshape.jcr.api.observation.Event getJcr21Event(final Event event) {
+    protected static Collection<String> getMixinTypes(final Event event)
+            throws PathNotFoundException, RepositoryException {
         try {
-            return (org.modeshape.jcr.api.observation.Event) event;
+            final org.modeshape.jcr.api.observation.Event modeEvent =
+                    (org.modeshape.jcr.api.observation.Event) event;
+            return ImmutableSet.copyOf(transform(ImmutableList.copyOf(modeEvent.getMixinNodeTypes()),
```
ajs6f: Why a copy to a `Set` here?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/SuppressByMixinFilter.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public boolean apply(final Event event) {
+        try {
+            return super.apply(event) && Collections.disjoint(getMixinTypes(event), suppressedMixins);
+        } catch (final PathNotFoundException e) {
+            LOGGER.trace("Dropping event from outside our assigned workspace:\n", e);
+            return false;
+        } catch (final RepositoryException e) {
+            throw propagate(e);
```
ajs6f: Maybe use `RepositoryRuntimeException` here?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/DefaultFilter.java
```diff
@@ -95,9 +91,13 @@ public boolean apply(final Event event) {
...
+                    nodetype2string));
```
ajs6f: Maybe `Functions.toStringFunction()` would work here?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/DefaultFilter.java
escowles: That's what the code was doing before -- but we could just remove that extra step and return the List.

ajs6f: I see no advantage to the extra copy.

ajs6f: Also, perhaps just `Arrays.asList()` instead of a copy to an `ImmutableList`?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/DefaultFilter.java
```diff
@@ -79,10 +78,7 @@ public DefaultFilter() {
     @Override
     public boolean apply(final Event event) {
         try {
-            final org.modeshape.jcr.api.observation.Event modeEvent = getJcr21Event(event);
-
-            final List<NodeType> nodeTypes = ImmutableList.copyOf(modeEvent.getMixinNodeTypes());
-            final Collection<String> mixinTypes = ImmutableSet.copyOf(transform(nodeTypes, nodetype2string));
+            final Collection<String> mixinTypes = getMixinTypes(event);
             return mixinTypes.contains(FEDORA_RESOURCE)
```
ajs6f: Maybe put these three types in a collection and use `Collections.disjoint()` as below.

ajs6f:
I have one minor qualm-- do we want an IT for this?

escowles:
That's a good idea -- it'll give me a place to document the spring bean setup to use this.

ajs6f:
Right on!

fcrepo-kernel-impl/src/test/java/org/fcrepo/integration/kernel/impl/observer/SuppressByMixinFilterIT.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.Set;
+
+/**
+ * <p>SuppressByMixinFilterIT class.</p>
+ *
+ * @author escowles
+ * @since 2015-04-15
+ */
+@ContextConfiguration({"/spring-test/eventing-suppress.xml", "/spring-test/repo.xml"})
+public class SuppressByMixinFilterIT extends AbstractIT {
```
ajs6f: Perhaps we can use awaitility here as we do for other event-driven tests in fcrepo-jms?

fcrepo-kernel-impl/src/test/java/org/fcrepo/integration/kernel/impl/observer/SuppressByMixinFilterIT.java
```diff
@@ -0,0 +1,120 @@
...
+        se.save();
+        se.logout();
+
+        try {
+            Thread.sleep(500);
+        } catch (final InterruptedException e) {
+            e.printStackTrace();
+        }
+
+        assertEquals("Event not suppressed!", (Integer)0, eventBusMessageCount);
```
ajs6f: This seems a bit brittle-- see comment above.

fcrepo-kernel-impl/src/test/java/org/fcrepo/integration/kernel/impl/observer/SuppressByMixinFilterIT.java
```diff
@@ -0,0 +1,120 @@
...
+        se.save();
+        se.logout();
+
+        try {
+            Thread.sleep(500);
+        } catch (final InterruptedException e) {
+            e.printStackTrace();
+        }
+
+        assertEquals("Wrong number of events!", (Integer)1, eventBusMessageCount);
```
ajs6f: Ditto.

fcrepo-kernel-impl/src/test/resources/spring-test/eventing-suppress.xml
```diff
@@ -0,0 +1,33 @@
...
+    event bus (fedoraInternalEventBus) -->
+
+    <context:annotation-config/>
+
+    <!-- listener that moves JCR Events to the Fedora internal event bus -->
+    <bean class="org.fcrepo.kernel.impl.observer.SimpleObserver"/>
+
+    <!-- used by bean above to filter which events get put on the bus -->
+    <bean name="fedoraEventFilter" class="org.fcrepo.kernel.impl.observer.SuppressByMixinFilter">
+      <constructor-arg>
```
ajs6f: Maybe use the Spring `c:` constructor-arg namespace here for brevity?

escowles: I wasn't able to find any examples of how to use the c: constructor-arg namespace to refer to a set instead of a literal value.  Is there a way to do this?

ajs6f: Yeah, make a separate set (via `util:set`) and use `-ref` in the `c:` element. That also gives you the chance to name the set something clear, like "typesWithNoEventing" or something.

ajs6f:
@escowles Did you want me to review and merge this?

escowles:
@ajs6f yes, please!

fcrepo-kernel-impl/pom.xml
```diff
@@ -119,6 +119,12 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
ajs6f: `fcrepo-jms` also uses awaitility, so maybe we should move the versioning info up into `fcrepo4`'s pom?

fcrepo-kernel-impl/src/test/resources/spring-test/eventing-suppress.xml
```diff
@@ -0,0 +1,34 @@
...
+
+    <!-- Context that suports event management, including the internal
+    event bus (fedoraInternalEventBus) -->
+
+    <context:annotation-config/>
+
+    <!-- listener that moves JCR Events to the Fedora internal event bus -->
+    <bean class="org.fcrepo.kernel.impl.observer.SimpleObserver"/>
+
+    <!-- used by bean above to filter which events get put on the bus -->
```
ajs6f: "bean _below_", right?

escowles:
@ajs6f I've moved the awaitility version to the top-level pom.xml and updated the spring comments.

PR #770, opened 2015-04-15
Remove serialization label triple

osmandin:
fcrepo-1471

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/url/HttpApiResources.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/url/HttpApiResources.java
@@ -123,8 +123,6 @@ public class HttpApiResources implements UriAwareResourceModelFactory {
 
                 //RDF the serialization
                 final Resource formatRDF = createResource(REPOSITORY_NAMESPACE + key);
-
-                model.add(formatRDF, RDFS_LABEL, key);
                 model.add(format, dcFormat, formatRDF);
             }
         }


fcrepo-http-api/src/main/java/org/fcrepo/http/api/url/HttpApiResources.java
```diff
@@ -123,8 +123,6 @@ private void addNodeStatements(final FedoraResource resource, final UriInfo uriI
 
                 //RDF the serialization
                 final Resource formatRDF = createResource(REPOSITORY_NAMESPACE + key);
-
-                model.add(formatRDF, RDFS_LABEL, key);
                 model.add(format, dcFormat, formatRDF);
```
awoods: Does this line also pose a problem?
`model.add(format, dcFormat, formatRDF);`

ajs6f: That triple is about the "export" endpoint. Unless it's actually blocking @osmandin right now, I'm inclined to ignore it and wait 'til our later moves to genuine HTTP conneg to rip it out at that time.

awoods: My question was if the triple on the next line is also showing up in RDF responses.

ajs6f: I know. I'm suggesting that if it doesn't matter for @osmandin then we should probably ignore it and not spend time investigating it, because it is going to go away anyway after HTTP conneg comes through.

awoods: In the absence of a response from @osmandin, I will assume the second triple is not a problem.

PR #774, opened 2015-04-17
Allow override of JMS baseUrl ...

whikloj:
...with system property fcrepo.jms.baseUrl with format host:port

Added Unit test

Addresses [FCREPO-1259](https://jira.duraspace.org/browse/FCREPO-1259)

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
@@ -104,7 +104,7 @@ abstract public class FedoraBaseResource extends AbstractResource {
     protected String getBaseUrlProperty() {
         final String propBaseURL = System.getProperty("fcrepo.jms.baseUrl", "");
         if (propBaseURL.length() > 0 && propBaseURL.startsWith("http")) {
-            return uriInfo.getBaseUriBuilder().clone().uri(propBaseURL).toString();
+            return uriInfo.getBaseUriBuilder().uri(propBaseURL).toString();
         }
         return "";
     }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
@@ -103,14 +103,8 @@ abstract public class FedoraBaseResource extends AbstractResource {
      */
     protected String getBaseUrlProperty() {
         final String propBaseURL = System.getProperty("fcrepo.jms.baseUrl", "");
-        if (propBaseURL.length() > 0 && propBaseURL.indexOf(':') > -1) {
-            final String[] parts = propBaseURL.split(":");
-            if (parts.length == 3) {
-                return uriInfo.getBaseUriBuilder().clone().scheme(parts[0]).host(parts[1].substring(2)).port(
-                    Integer.valueOf(parts[2])).toString();
-            } else {
-                return uriInfo.getBaseUriBuilder().clone().scheme(parts[0]).host(parts[1].substring(2)).toString();
-            }
+        if (propBaseURL.length() > 0 && propBaseURL.startsWith("http")) {
+            return uriInfo.getBaseUriBuilder().clone().uri(propBaseURL).toString();
         }
         return "";
     }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
@@ -32,8 +32,6 @@ import javax.jcr.observation.ObservationManager;
 import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.UriInfo;
 
-import java.net.URI;
-
 import static org.slf4j.LoggerFactory.getLogger;
 
 /**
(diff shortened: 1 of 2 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
```diff
@@ -94,4 +95,22 @@ protected void setUpJMSInfo(final UriInfo uriInfo, final HttpHeaders headers) {
...
+     * exists.
+     * 
+     * @return String the base Url
+     */
+    protected String getBaseUrlProperty() {
+        final String propBaseURL = System.getProperty("fcrepo.jms.baseUrl", "");
+        if (propBaseURL.length() > 0) {
+            if (propBaseURL.indexOf(':') > -1) {
+                final String[] parts = propBaseURL.split(":");
+                return uriInfo.getBaseUriBuilder().clone().host(parts[0]).port(Integer.valueOf(parts[1])).toString();
```
escowles: Isn't parts[0] going to be "http" here?

acoburn: could also be "https"

escowles: Right, but not the hostname, which I would expect to be in parts[1]

whikloj: Ok, I see what happened there. My expectation was host is host, and not protocol. But that could definitely depend on interpretation. I'll put in a check.

whikloj: Before I do this, is it your expectation that if I am running Fcrepo at "http://blah.com:8080" I might want the JMS messages to reference "https://blah.com:8443"? I couldn't think of why you would, but now I'm thinking it might be easier to pass the whole property to UriBuilder.uri(String s)?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
```diff
@@ -94,4 +95,23 @@ protected void setUpJMSInfo(final UriInfo uriInfo, final HttpHeaders headers) {
...
+        final String propBaseURL = System.getProperty("fcrepo.jms.baseUrl", "");
+        if (propBaseURL.length() > 0 && propBaseURL.indexOf(':') > -1) {
+            final String[] parts = propBaseURL.split(":");
+            if (parts.length == 3) {
+                return uriInfo.getBaseUriBuilder().clone().scheme(parts[0]).host(parts[1].substring(2)).port(
+                    Integer.valueOf(parts[2])).toString();
+            } else {
+                return uriInfo.getBaseUriBuilder().clone().scheme(parts[0]).host(parts[1].substring(2)).toString();
+            }
+        }
```
ajs6f: What is the point of breaking up this URI and then reassembling it?

escowles: IMHO, we should return the fcrepo.jms.baseUrl as-is without parsing and rebuilding.  The use case for this is users wanting to override the default generated baseURL, and I would expect it to be very frustrating if the value was altered in any way.

ajs6f: Agreed.

whikloj: Honestly at this point, I'm not sure there is any point. I can replace it all with:
return uriInfo.getBaseUriBuilder().clone().uri(propBaseURL).toString();

ajs6f: Right on, brother.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
```diff
@@ -94,4 +95,17 @@ protected void setUpJMSInfo(final UriInfo uriInfo, final HttpHeaders headers) {
...
+    /**
+     * Produce a baseURL for JMS events using the system property fcrepo.jms.baseUrl of the form http[s]://host[:port],
+     * if it exists.
+     *
+     * @return String the base Url
+     */
+    protected String getBaseUrlProperty() {
+        final String propBaseURL = System.getProperty("fcrepo.jms.baseUrl", "");
+        if (propBaseURL.length() > 0 && propBaseURL.startsWith("http")) {
+            return uriInfo.getBaseUriBuilder().clone().uri(propBaseURL).toString();
```
ajs6f: Why the invocation of `clone()`?

whikloj: Possibly a mistaken understanding of the UriBuilder, but my assumption from the Javadocs and the usage in [IdentiferConverter](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java#L54) was to preserve the original state. Perhaps its not needed?

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/8779cf0581cbd4f7fb07e741bf7a8ac27af1b68d

PR #778, opened 2015-04-20
Fix for behaviour described in fcrepo4-777

TheStokoe:
Modify federated file system connector to use initial sha1 as the basis for comparison as per the fixity check behaviour described in https://wiki.duraspace.org/display/FEDORA40/How+to+audit+fixity+in+a+filesystem+federation


acoburn:
Closing this as the filesystem federation has been removed from the core of Fedora.

PR #779, opened 2015-04-21
Catching tombstone errors when checking for skolem nodes

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-1483


ajs6f:
Nice catch, @escowles.

PR #781, opened 2015-04-23
Adding FedoraResource.setURIProperty method

escowles:
Adds kernel API for setting URI properties without conversion to reference types, to support the audit use case of preserving audit events that link to repository resources after those resources have been deleted.  See https://github.com/fcrepo4-labs/fcrepo-audit/commit/3d7d6d348424df73940bf992b2770cdbc92dc3d5#commitcomment-10865344

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
@@ -263,6 +263,14 @@ public class FedoraResourceImpl extends JcrTools implements FedoraJcrTypes, Fedo
         }
     }
 
+    /**
+     * Set the given property value for this resource as a URI, without translating any URIs that
+     * appear to be references to repository resources.  Using untranslated URIs to refer to
+     * repository resources will disable referential integrity checking, but also allows referring
+     * to resources that do not exist, have been deleted, etc.
+     * @param relPath the given path
+     * @param value the URI value
+     */
     @Override
     public void setURIProperty(final String relPath, final URI value) {
         try {
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
@@ -37,6 +37,7 @@ import static org.slf4j.LoggerFactory.getLogger;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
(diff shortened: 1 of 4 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
```diff
@@ -262,6 +264,15 @@ public Property getProperty(final String relPath) {
     }
 
     @Override
+    public void setURIProperty(final String relPath, final URI value) {
```
awoods: Can you add a Javadoc note like the one you have on this commit that indicates that no references are created with this method. That would be helpful in the future to let people understand the specialness of this method.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/FedoraResourceImplTest.java
```diff
@@ -395,6 +398,14 @@ public void testGetProperty() throws RepositoryException {
     }
 
     @Test
+    public void testSetURIProperty() throws URISyntaxException, RepositoryException {
+        final String prop = "premis:hasEventRelatedObject";
+        final String uri = "http://localhost:8080/rest/1";
```
awoods: The argument name and javadoc indicates that the URI is a relative-path:
https://github.com/fcrepo4/fcrepo4/pull/781/files#diff-688c65561dd248cf2a0096c799be5d18R89

This integration test is not a relative-path, and furthermore, the host and port (in a production scenario) may change or need idTranslation.

To be clear, are you expecting Resource.setURIProperty() to use full URLs or relative-paths? If full URLs, then are we concerned about context translation?

escowles: The relPath argument is the property name, not the value.  I copied the parameters javadoc from the getProperty() method for consistency, but I could update this to be clearer that it's the property name in abbreviated format "premis:hasEventRelatedObject", not the full URI.

awoods: I also wonder about passing in values as full URLs that do not pass through an idTranslator. Is this a concern?

ajs6f: It's a concern if those values are URIs of things in the repo.

escowles: The URIs are references to things in the repo -- identifiers of resources related to audit events.  The intention with this approach is to build those URIs in the audit processor, and store them in the repository as URIs, so they are not deleted if the resource is deleted.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/c5e15ddf2bb5c5a1c8cd4916af10e1d00a291151

PR #782, opened 2015-04-23
Updating DefaultIdentifierTranslator to make it usable by audit service

escowles:
* Making RESOURCE_NAMESPACE into a private field because nothing is using it
* Adding constructor for custom resource namespaces

Addresses https://jira.duraspace.org/browse/FCREPO-1489

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/DefaultIdentifierTranslator.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/DefaultIdentifierTranslator.java
@@ -51,10 +51,7 @@ public class DefaultIdentifierTranslator extends IdentifierConverter<Resource, F
 
     private static final NodeResourceConverter nodeResourceConverter = new NodeResourceConverter();
 
-    /**
-     * Default namespace to use for node URIs
-     */
-    public static final String RESOURCE_NAMESPACE = "info:fedora/";
+    private final String resourceNamespace;
     private final Session session;
 
     /**
@@ -62,7 +59,15 @@ public class DefaultIdentifierTranslator extends IdentifierConverter<Resource, F
      * @param session the session
      */
     public DefaultIdentifierTranslator(final Session session) {
+        this(session, "info:fedora/");
+    }
+
+    /**
+     * Construct the graph with the provided resource namespace.
+    **/
+    public DefaultIdentifierTranslator(final Session session, final String resourceNamespace) {
         this.session = session;
+        this.resourceNamespace = resourceNamespace;
         setTranslationChain();
     }
 
@@ -109,7 +114,7 @@ public class DefaultIdentifierTranslator extends IdentifierConverter<Resource, F
 
     @Override
     public boolean inDomain(final Resource subject) {
-        return subject.isURIResource() && subject.getURI().startsWith(RESOURCE_NAMESPACE);
+        return subject.isURIResource() && subject.getURI().startsWith(resourceNamespace);
     }
 
     @Override
@@ -121,7 +126,7 @@ public class DefaultIdentifierTranslator extends IdentifierConverter<Resource, F
         } else {
             relativePath = absPath;
         }
-        return createResource(RESOURCE_NAMESPACE + reverse.convert(relativePath));
+        return createResource(resourceNamespace + reverse.convert(relativePath));
     }
 
     @Override
@@ -130,7 +135,7 @@ public class DefaultIdentifierTranslator extends IdentifierConverter<Resource, F
             return null;
         }
 
-        final String path = subject.getURI().substring(RESOURCE_NAMESPACE.length() - 1);
+        final String path = subject.getURI().substring(resourceNamespace.length() - 1);
 
         final String absPath = forward.convert(path);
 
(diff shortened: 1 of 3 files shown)


ajs6f:
With what kind of namespaces do you expect people to use this stateful class?

awoods:
@ajs6f, see: https://github.com/fcrepo4-labs/fcrepo-audit/pull/3/files#diff-3ff409abe3384fddfe4bc7d36c6ce055R261

ajs6f:
Thank you, @awoods, but that explains very little to me. Is there some general theory of operation for this class? Knowing the quality of @escowles's work, I'm not worried-- but I would like to understand why he considered it better to alter this class than to create a new `IdentifierConverter` impl.

escowles:
@ajs6f I started out doing just that -- and then realized that I only needed to change the hard-coded REPOSITORY_NAMESPACE variable to the baseURL, and I could use the DefaultIdentifierConverter.

awoods:
@ajs6f, @escowles, I am inclined to merge this PR unless there are concerns.

ajs6f:
I'd rather see this diverge into an `AuditIdentiferConverter` or something of the sort.

ajs6f:
What is the default value for the namespace at play?

awoods:
In the context of Audit, the "default" value is dynamic: `baseURL + "/"`, as you can see here:
https://github.com/fcrepo4-labs/fcrepo-audit/pull/3/files#diff-3ff409abe3384fddfe4bc7d36c6ce055R261

ajs6f:
It is in no way remotely clear to me how that line of code documents the use of the class under comment. At the very least, please let's include some verbal explanation of the new constructor and its parameters. "Default namespace to use for node URIs" and "Construct the graph with the provided resource namespace." doesn't begin to explain the matter. I'm confident that the thought behind this change is valuable, but I really don't understand yet why this is presented as a change to this class and not the basis of a new class.

awoods:
Thanks for the perspective, @ajs6f. I will hold off until tomorrow... hoping @escowles nails it in the morning.

ajs6f:
I will be available tomorrow morning via IRC or Hangout or what-have-ye. I don't want to hold up the functionality that @escowles is moving forward. But I do have the unpleasant feeling that we aren't all on the same page with respect to the construction of identifiers, and that's such a crucial element of repository work that I'm willing to be annoying about this.

escowles:
@ajs6f My initial approach was to create a new implementation in the audit package called `AuditIdentifierConverter`.  I looked at the `DefaultIdentifierTranslator` class and quickly realized I could just extend it -- if the DEFAULT_NAMESPACE variable wasn't final.  So I made that change, and started looking to see what I would need to customize, and wound up with only a constructor.  So I decided it was a little silly to extend a class just to add a new constructor signature, when I was already having to change the original class.  I think just adding the constructor to `DefaultIdentifierTranslator` is the simplest way forward, and will make it more usable from any other non-web contexts.

Though I'm willing to be convince that I should be using `HttpResourceConvertor` instead:

https://github.com/fcrepo4-labs/fcrepo-audit/pull/3/files#diff-3ff409abe3384fddfe4bc7d36c6ce055R261

The task is to update the properties on a newly-created audit event node.  Creating the triples is the easy part, but I wasn't able to get an `HttpResourceConvertor` instance to successfully consider the resource's URI to be in-domain.  I tracked that as far as:

https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java#L155

Where the URI matched the template, but there wasn't a `path` value.  Since I'm not operating in a web context, I can't create a UriBuilder with the JAX-RS context.  So I suspect there is some other way to create a UriBuilder that would satisfy the requirements.  Having that requirement documented in the constructor javadoc would be good for future users of this class, too.

awoods:
That all makes good sense, @escowles. I believe @ajs6f is wanting some javadoc-umentation describing something about the use and assumptions of the identifiers that would be created by the new constructor on `DefaultIdentifierTranslator`. My feeling is that a common, documented understanding is desired, not necessarily a different implementation.

escowles:
@awoods, I've updated the Javadoc comments to better explain the purpose of the class, differentiate the constructors from each other, and link to HttpResourceConverter.

awoods:
Thanks, @escowles. The docs look good. Hopefully @ajs6f will provide comments shortly.

ajs6f:
In fact, now that I understand what the desire is I don't agree that this is a good change to the class (instead of a new class) but I don't think the difference is worth troubling about.

awoods:
@ajs6f, would your suggestion be to create a new class that extends `DefaultIdentifierTranslator`?

ajs6f:
The other way around-- to create a class for what @escowles very reasonably wants to do, something like `PrefixingIdentifierTranslator`, and let `DefaultIdentifierTranslator` inherit from that.

escowles:
OK, @ajs6f @awoods, I've created a new PrefixingIdentifierTranslator and made DefaultIdentifierTranslator extend from that.

ajs6f:
Thanks. As a TODO to myself, much of what happens with chains of translators inside these converters should be factored out into some abstract class, or post Java 8, default implementation.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/2c3eb0ffc6156e35c750a095ad04a2d6b9948d07

PR #783, opened 2015-04-24
Lessen leakage of JCR dependent code that extends beyond the kernel.

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-1487

### DIFF
--- a/fcrepo-connector-file/pom.xml
+++ b/fcrepo-connector-file/pom.xml
@@ -25,7 +25,6 @@
     <dependency>
       <groupId>org.modeshape</groupId>
       <artifactId>modeshape-jcr</artifactId>
-      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.glassfish.hk2.external</groupId>
--- a/fcrepo-connector-file/pom.xml
+++ b/fcrepo-connector-file/pom.xml
@@ -21,6 +21,12 @@
       <groupId>javax.jcr</groupId>
       <artifactId>jcr</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>org.modeshape</groupId>
+      <artifactId>modeshape-jcr</artifactId>
+      <scope>provided</scope>
+    </dependency>
     <dependency>
       <groupId>org.glassfish.hk2.external</groupId>
       <artifactId>javax.inject</artifactId>
--- a/fcrepo-connector-file/src/main/java/org/fcrepo/connector/file/FedoraFileSystemConnector.java
+++ b/fcrepo-connector-file/src/main/java/org/fcrepo/connector/file/FedoraFileSystemConnector.java
@@ -39,6 +39,7 @@ import java.util.Map;
 
 import com.google.common.annotations.VisibleForTesting;
 
+import org.fcrepo.kernel.exception.RepositoryRuntimeException;
 import org.infinispan.schematic.document.Document;
 import org.modeshape.connector.filesystem.ExternalJsonSidecarExtraPropertyStore;
 import org.modeshape.connector.filesystem.FileSystemConnector;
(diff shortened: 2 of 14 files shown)


fcrepo-connector-file/pom.xml
```diff
@@ -21,6 +21,12 @@
       <groupId>javax.jcr</groupId>
       <artifactId>jcr</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>org.modeshape</groupId>
+      <artifactId>modeshape-jcr</artifactId>
+      <scope>provided</scope>
```
ajs6f: I'm a little confused about this scope-- is the thought that the _container_ will provide this artifact (which clearly ain't the case...)?

fcrepo-connector-file/src/main/java/org/fcrepo/connector/file/FedoraFileSystemConnector.java
```diff
@@ -83,16 +84,20 @@
 
     @Override
     public void initialize(final NamespaceRegistry registry,
-                           final NodeTypeManager nodeTypeManager) throws RepositoryException, IOException {
-        super.initialize(registry, nodeTypeManager);
+                           final NodeTypeManager nodeTypeManager) throws IOException {
+        try {
+            super.initialize(registry, nodeTypeManager);
+        } catch (RepositoryException e) {
+            throw new RepositoryRuntimeException("Error initializing FedoraFileSystemConnector!", e);
```
ajs6f: No problem with this code, just a note to ourselves: we do so much `throw new RepositoryRuntimeException(something)` that I'm starting to wonder if we shouldn't write some static methods to the purpose a la Guavas `Throwables.propagate(...)`.

fcrepo-jms/src/main/java/org/fcrepo/jms/headers/DefaultMessageFactory.java
```diff
@@ -117,6 +117,30 @@ public Message getMessage(final FedoraEvent jcrEvent,
         return message;
     }
 
+    private String getUserData(final FedoraEvent jcrEvent) {
```
ajs6f: Should this exception action be in `FedoraEvent`?

fcrepo-kernel-impl/pom.xml
```diff
@@ -27,6 +27,7 @@
     <dependency>
       <groupId>org.modeshape</groupId>
       <artifactId>modeshape-jcr</artifactId>
+      <scope>provided</scope>
```
ajs6f: I have the same confusion about this scope as above. I guess I just don't understand, given our disengagement with J2EE or OSGi, what we actually mean by `provided`...

fcrepo-jms/src/main/java/org/fcrepo/jms/headers/DefaultMessageFactory.java
awoods: Agreed... which will also stop FedoraEvent from throwing its current `javax.jcr.RepositoryException`s

ajs6f: More and more, I think we should adopt a natural habit of rethrowing `RepositoryException` as soon and as low as we can. (Perhaps factored into a static utility method.) There's really nothing we expect to do with it other than eventually report it at our own API. This is to be accounted to the facts that we are composing middleware, not user-facing software, and that `RespositoryException` is generic.

fcrepo-connector-file/pom.xml
awoods: The issue is that a few existing pom.xml files depend on `modeshape-jcr` and only some of them actually define that dependency. Scope `provided` ensures that the given dependency does not propagate transitively. However, thanks to you raising this flag, I see that this PR only works because `fcrepo-http-commons/pom.xml` is still such a perpetrator of transitively leaking the `modeshape-jcr` dependency to the other modules... otherwise, the dependency is never actually `provided` by the container (or fcrepo-webapp), which causes build failure.

ajs6f: Right, right. In theory, we want the only transitive dependency to be "pure" JCR API, not impl APIs, right?

fcrepo-jms/src/main/java/org/fcrepo/jms/headers/DefaultMessageFactory.java
awoods: By this, I assume you are suggesting rethrowing `javax.jcr.RepositoryException` as another, `org.fcrepo` exception?

ajs6f: Yes, sorry, `RepositoryRuntimeException

fcrepo-connector-file/pom.xml
awoods: Right... and ideally, most of the Fedora modules would not even depend on "pure" JCR. I would like to see as few modules as possible care anything about JCR.

ajs6f: Good point. That is a higher and better ideal. We should factor as much JCR as we can down through the modules to `kernel-impl`.

awoods:
As noted in the second commit, 
** Please do not commit/merge this PR without first squashing these commits **

fcrepo-kernel/src/main/java/org/fcrepo/kernel/observer/FedoraEvent.java
```diff
@@ -119,13 +118,17 @@ public String getPath() throws RepositoryException {
...
      * @param e JCR Event
     **/
-    public static String getPath(final Event e) throws RepositoryException {
-        if (e.getType() == PROPERTY_ADDED   ||
-            e.getType() == PROPERTY_CHANGED ||
-            e.getType() == PROPERTY_REMOVED) {
-            return e.getPath().substring(0, e.getPath().lastIndexOf("/"));
+    public static String getPath(final Event e) {
+        try {
+            if (e.getType() == PROPERTY_ADDED   ||
```
ajs6f: It would be better for this test to use a constant collection of `PROPERTY_ADDED`, `PROPERTY_CHANGED`, `PROPERTY_REMOVED` and `Collection.contains()`.

awoods: I am going to suggest not including non-related clean-up/refactoring to this PR.

ajs6f: Okay. Maybe write a TODO in here?

PR #785, opened 2015-04-25
Account for blank nodes in HTML UI

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-1496

### DIFF
--- a/fcrepo-http-api/src/main/resources/views/common-metadata.vsl
+++ b/fcrepo-http-api/src/main/resources/views/common-metadata.vsl
@@ -11,7 +11,11 @@
     <dd>
         <ol id="childList">
             #foreach($quad in $helpers.getObjects($rdf, $topic, $rdfLexicon.CONTAINS))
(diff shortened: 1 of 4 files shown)


ajs6f:
Perhaps we can have an IT to prevent regression?

awoods:
I would agree if we had an HTML integration test framework that worked.

ajs6f:
The problem was that an exception was being thrown. We don't need to check the actual HTML, just make a request and get a 200.

acoburn:
@ajs6f are you suggesting that we should wait on merging this until an integration test is written?

ajs6f:
I would.

acoburn:
ok, I'll try to fit this in today.

ajs6f:
Thanks. I think it's worth doing because otherwise, we risk regression as the general bnode problem continues to evolve. We obviously have no test for this now.

awoods:
@acoburn, let me know once you start to work on this integration test... and I will do the same if I start to work on it.

acoburn:
@awoods sounds good

acoburn:
@awoods shall we close this PR? The related ticket is marked as "Closed: Won't Fix"

PR #786, opened 2015-04-26
DO NOT MERGE

awoods:
Incremental step towards allowing updates on blank nodes

Related ticket: https://jira.duraspace.org/browse/FCREPO-1385

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -306,7 +306,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
             httpTripleUtil.addHttpComponentModelsForResourceToStream(rdfStream, resource(), uriInfo, translator());
         }
 
-        return rdfStream.map(deskolemizer());
+        return rdfStream;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -306,7 +306,7 @@ public RdfStream apply(final FedoraResource child) {
             httpTripleUtil.addHttpComponentModelsForResourceToStream(rdfStream, resource(), uriInfo, translator());
         }
 
-        return rdfStream.map(deskolemizer());
+        return rdfStream;
```
ajs6f: If you make this change, then at what point is the necessary deskolemization actually going to happen?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
```diff
@@ -404,9 +403,7 @@ public void updateProperties(final IdentifierConverter<Resource, FedoraResource>
         final Model maybeModel = idTranslator.reverse().convert(this).getModel();
         final Model context = maybeModel == null ? createDefaultModel() : maybeModel;
 
-        final Deskolemizer deskolemize = new Deskolemizer(idTranslator, context);
-
-        final Model model = originalTriples.withThisContext(originalTriples.transform(deskolemize)).asModel();
+        final Model model = originalTriples.asModel();
```
ajs6f: If you make this change, the update SPARQL will be running against RDF with skolem nodes in it. Attempts to deal with bnodes will always fail.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
```diff
@@ -404,9 +403,7 @@ public void updateProperties(final IdentifierConverter<Resource, FedoraResource>
         final Model maybeModel = idTranslator.reverse().convert(this).getModel();
         final Model context = maybeModel == null ? createDefaultModel() : maybeModel;
 
-        final Deskolemizer deskolemize = new Deskolemizer(idTranslator, context);
-
-        final Model model = originalTriples.withThisContext(originalTriples.transform(deskolemize)).asModel();
```
ajs6f: This would have been simpler as `final Model model = originalTriples.map(deskolemize).asModel();`

acoburn:
@awoods should this PR also be closed?

PR #792, opened 2015-05-13
remove org.modeshape exports from fcrepo-connector-file bundle

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1534


fcrepo-connector-file/src/main/java/org/fcrepo/connector/file/FedoraFileSystemConnector.java
```diff
@@ -84,20 +82,16 @@
 
     @Override
     public void initialize(final NamespaceRegistry registry,
-                           final NodeTypeManager nodeTypeManager) throws IOException {
-        try {
```
awoods: As you can see from the previous commit on this file, we want to stop the jcr exception `RepositoryException` from leaking out of this class:
https://github.com/acoburn/fcrepo4/commit/de826b28a2d8c0fb81ea4cc9db0a6bab5f90643b#diff-172c1a74effcaea58305292a8256a1b9

fcrepo-connector-file/src/main/java/org/fcrepo/connector/file/ExternalJsonSidecarExtraPropertyStore.java
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * An implementation of ExtraPropertyStore, based on
+ * org.modeshape.connector.filesystem.JsonSidecarExtraPropertyStore that stores the
+ * properties in a separate configured directory than the filesystem federation itself.
+ *
+ * @author Mike Durbin
+ * @author acoburn
+ */
+public class ExternalJsonSidecarExtraPropertyStore implements ExtraPropertiesStore {
```
awoods: We seem to have lost the history of this class. Can you instead perform a `git mv` on `org.modeshape.connector.filesystem.ExternalJsonSidecarExtraPropertyStore` to this new location?

fcrepo-connector-file/src/main/java/org/fcrepo/connector/file/ExternalJsonSidecarExtraPropertyStore.java
```diff
@@ -0,0 +1,171 @@
...
+                            .getAbsoluteFile().toPath().relativize(propertyFileInFederation);
+            file = propertyStoreRoot.toPath().resolve(relativePath).toFile();
+        }
+        if (!file.getParentFile().exists() && !file.getParentFile().mkdirs()) {
+            throw new RepositoryRuntimeException("Unable to create directories " + file.getParentFile() + ".");
+        }
+        return file;
+    }
+
+    @Override
```
awoods: For the methods that you copied from ModeShape's codebase, could you provide links to the ModeShape GitHub source so we can check for updates in the future.

fcrepo-connector-file/src/main/java/org/fcrepo/connector/file/ExternalJsonSidecarExtraPropertyStore.java
```diff
@@ -0,0 +1,171 @@
...
+    public boolean removeProperties(final String id) {
+        final File file = sidecarFile(id);
+        if (!file.exists()) {
+            return false;
+        }
+        file.delete();
```
awoods: For example, I would suggest `return file.delete()` here, but if the ModeShape original source has the code as written here, I would not suggest an update.

fcrepo-connector-file/src/main/java/org/fcrepo/connector/file/FedoraFileSystemConnector.java
```diff
@@ -165,6 +159,12 @@ protected boolean shouldCacheProperties() {
         return extraPropertiesStore() != null && (!isReadonly() || this.propertiesDirectory != null);
     }
 
+
+    @Override
+    public boolean isRoot(final String id) {
```
awoods: Why is providing a pass-through implementation of `isRoot()` and `isContentNode()` required here?

acoburn: Because these functions are protected in the modeshape impl. In order to be used by the `ExternalJsonSidecarExtraPropertyStore` class, they must be public. While the `ExternalJsonSidecarExtraPropertyStore` class was part of the org.modeshape package, it could access those functions.

fcrepo-connector-file/src/main/java/org/fcrepo/connector/file/ExternalJsonSidecarExtraPropertyStore.java
acoburn: The modeshape impl is like this: https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/connector/filesystem/JsonSidecarExtraPropertyStore.java#L139

The only changes are the addition of `final` keywords and some checkstyle-related formatting.

fcrepo-connector-file/src/main/java/org/fcrepo/connector/file/ExternalJsonSidecarExtraPropertyStore.java
acoburn: The full URLs are too long for the checkstyle rules. Would you suggest using github's url shortening service?

fcrepo-connector-file/src/main/java/org/fcrepo/connector/file/ExternalJsonSidecarExtraPropertyStore.java
acoburn: The challenge here is that git tracks file-renames based on content similarity, and the content of the two files is almost entirely different. Even when I use `git mv` and then make these modifications, a subsequent `git log --follow path/to/file` doesn't show the earlier revisions of the class. The only way to preserve the history is to rename the file in one commit and then make the changes in another. (and not squash those commits). That would involve starting over with this PR, which isn't a big deal.

awoods: I was trying to see what code you added new in this PR to this class, hence the search for history.
It is probably fine as is... re-doing the PR is likely not worth the benefit at this point.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/dff2d5f0be464981563a5c2acaad3c3422dfe732

PR #796, opened 2015-05-19
Minor, update javadocs

awoods:
Relates to: https://jira.duraspace.org/browse/FCREPO-1549

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/PrefixingIdentifierTranslator.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/PrefixingIdentifierTranslator.java
@@ -39,7 +39,7 @@ import java.util.List;
 /**
  * A very simple {@link IdentifierConverter} which translates JCR paths into Fedora subjects with
  * a configurable resource namespace (e.g., a baseURL).  When a REST API context is available for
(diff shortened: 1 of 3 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/PrefixingIdentifierTranslator.java
```diff
@@ -39,7 +39,7 @@
 /**
  * A very simple {@link IdentifierConverter} which translates JCR paths into Fedora subjects with
  * a configurable resource namespace (e.g., a baseURL).  When a REST API context is available for
- * constructing URIs, {@link HttpResourceConverter} should be used instead.
+ * constructing URIs, org.fcrepo.http.commons.api.rdf.HttpResourceConverter should be used instead.
```
escowles: Shouldn't this keep the {@link ...}?

awoods: I wanted the same thing, but `fcrepo-kernel-impl` does not have dependency visibility into `fcrepo-http-commons`.
Hence the @link syntax does not work.

escowles: OK, that makes sense, so this is good to go.

PR #806, opened 2015-06-01
clean up OSGi exports

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1562

Also: org.fcrepo.kernel.spring.ModeShapeRepositoryFactoryBean doesn't exist. It should be org.fcrepo.kernel.impl.spring.ModeShapeRepositoryFactoryBean

### DIFF
--- a/fcrepo-http-api/pom.xml
+++ b/fcrepo-http-api/pom.xml
@@ -232,7 +232,8 @@
               org.fcrepo.http.commons.domain,
               org.fcrepo.http.commons.domain.ldp,
               org.fcrepo.http.commons.responses,
-              org.fcrepo.http.commons.session
+              org.fcrepo.http.commons.session,
+              org.fcrepo.connector.file
             </Import-Package>
           </instructions>
         </configuration>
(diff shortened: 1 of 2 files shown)


fcrepo-http-api/pom.xml
```diff
@@ -232,9 +232,7 @@
               org.fcrepo.http.commons.domain,
               org.fcrepo.http.commons.domain.ldp,
               org.fcrepo.http.commons.responses,
-              org.fcrepo.http.commons.session,
-              org.fcrepo.connector.filesystem,
```
awoods: Instead of removing this, should it instead be `org.fcrepo.connector.file`?

acoburn: Yes, it should be `org.fcrepo.connector.file`. Just pushed the relevant change.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/a5b7037d1ec05033296b42ffbad3e2b30db00754

PR #807, opened 2015-06-01
Stop spewing UUIDs

yinlinchen:
Resolves: https://jira.duraspace.org/browse/FCREPO-1539


ajs6f:
It's clear from this PR that we will no longer be testing to prove that we _do_ spew UUIDs, but will we no longer actually be spewing them? I see no changes to the actual RDF generation machinery here...

yinlinchen:
What file are you refer to? I thought remove JCR_UUID in
fcrepo-kernel/src/main/java/org/fcrepo/kernel/FedoraJcrTypes.java is the
change need to be done for this ticket...

ajs6f:
No, you need to make sure that the triple in question actually stops being published. I don't know where it is created, but it's probably somewhere near `PropertiesRdfContext`.

awoods:
Superseded by: https://github.com/fcrepo4/fcrepo4/pull/827

PR #808, opened 2015-06-02
POST/PATCH must advertise constraints if triples not persisted

whikloj:
Addresses [FCREPO-1418](https://jira.duraspace.org/browse/FCREPO-1418)

### DIFF
--- a/fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/ConstraintViolationExceptionMapperTest.java
+++ b/fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/ConstraintViolationExceptionMapperTest.java
@@ -1,0 +1,66 @@
+/**
+ * Copyright 2015 DuraSpace, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.http.commons.exceptionhandlers;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import javax.jcr.RepositoryException;
+import javax.ws.rs.core.Link;
+import javax.ws.rs.core.UriInfo;
+
+import org.fcrepo.kernel.exception.ConstraintViolationException;
+
+import static org.fcrepo.http.commons.test.util.TestHelpers.getUriInfoImpl;
+import static org.fcrepo.kernel.RdfLexicon.CONSTRAINED_BY;
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import static org.mockito.MockitoAnnotations.initMocks;
+
+/**
+ * ConstraintViolationExceptionMapperTest
+ *
+ * @author whikloj
+ * @since 2015-06-22
+ */
+public class ConstraintViolationExceptionMapperTest {
+
+    private UriInfo mockInfo;
+
+    @Mock
+    private ConstraintViolationExceptionMapper mapper;
+
+    @Before
+    public void setUp() throws RepositoryException, URISyntaxException {
+        initMocks(this);
+        this.mockInfo = getUriInfoImpl();
+    }
+
+    @Test
+    public void testBuildConstraintLink() {
+        final URI testLink = URI.create("http://localhost/static/constraints/ConstraintViolationException.rdf");
+
+        final ConstraintViolationException ex = new ConstraintViolationException("This is an error.");
+
+        final Link link = ConstraintViolationExceptionMapper.buildConstraintLink(ex, mockInfo);
+
+        assertEquals(link.getRel(), CONSTRAINED_BY.getURI());
+        assertEquals(link.getUri(), testLink);
+    }
+}
--- a/fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/ConstraintViolationExceptionMapperTest.java
+++ b/fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/ConstraintViolationExceptionMapperTest.java
@@ -30,8 +30,9 @@ import static org.junit.Assert.assertEquals;
 
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
 import org.mockito.Mock;
-import static org.mockito.MockitoAnnotations.initMocks;
+import org.mockito.runners.MockitoJUnitRunner;
 
 /**
  * ConstraintViolationExceptionMapperTest
@@ -39,6 +40,7 @@ import static org.mockito.MockitoAnnotations.initMocks;
  * @author whikloj
  * @since 2015-06-22
  */
+@RunWith(MockitoJUnitRunner.class)
 public class ConstraintViolationExceptionMapperTest {
 
     private UriInfo mockInfo;
@@ -48,7 +50,6 @@ public class ConstraintViolationExceptionMapperTest {
 
     @Before
     public void setUp() throws RepositoryException, URISyntaxException {
-        initMocks(this);
         this.mockInfo = getUriInfoImpl();
     }
 
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
@@ -65,6 +65,7 @@ import org.fcrepo.kernel.FedoraJcrTypes;
 import org.fcrepo.kernel.models.NonRdfSourceDescription;
 import org.fcrepo.kernel.models.FedoraBinary;
 import org.fcrepo.kernel.models.FedoraResource;
+import org.fcrepo.kernel.exception.ConstraintViolationException;
 import org.fcrepo.kernel.exception.MalformedRdfException;
 import org.fcrepo.kernel.exception.PathNotFoundRuntimeException;
 import org.fcrepo.kernel.exception.RepositoryRuntimeException;
@@ -493,6 +494,8 @@ public class FedoraResourceImpl extends JcrTools implements FedoraJcrTypes, Fedo
         try {
             new RdfRemover(idTranslator, getSession(), replacementStream
                     .withThisContext(differencer)).consume();
+        } catch (final ConstraintViolationException e) {
+            throw e;
         } catch (final MalformedRdfException e) {
             exceptions.append(e.getMessage());
             exceptions.append("\n");
@@ -501,6 +504,8 @@ public class FedoraResourceImpl extends JcrTools implements FedoraJcrTypes, Fedo
         try {
             new RdfAdder(idTranslator, getSession(), replacementStream
                     .withThisContext(differencer.notCommon())).consume();
+        } catch (final ConstraintViolationException e) {
+            throw e;
         } catch (final MalformedRdfException e) {
             exceptions.append(e.getMessage());
         }
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/utils/iterators/PersistingRdfStreamConsumer.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/utils/iterators/PersistingRdfStreamConsumer.java
@@ -21,8 +21,11 @@ import static org.fcrepo.kernel.impl.rdf.ManagedRdf.isManagedMixin;
 import static org.slf4j.LoggerFactory.getLogger;
 
 import com.google.common.base.Joiner;
+
 import org.fcrepo.kernel.models.FedoraResource;
+import org.fcrepo.kernel.exception.ConstraintViolationException;
 import org.fcrepo.kernel.exception.MalformedRdfException;
+import org.fcrepo.kernel.exception.OutOfDomainSubjectException;
 import org.fcrepo.kernel.exception.RepositoryRuntimeException;
 import org.fcrepo.kernel.identifiers.IdentifierConverter;
 
@@ -90,7 +93,7 @@ public abstract class PersistingRdfStreamConsumer implements RdfStreamConsumer {
                             t);
                 } else {
                     LOGGER.error("subject ({}) is not in repository domain.", t.getSubject().toString());
-                    throw new MalformedRdfException(String.format(
+                    throw new OutOfDomainSubjectException(String.format(
                         "RDF Stream contains subject(s) (%s) not in the domain of this repository.", t.getSubject()));
                 }
                 return result;
@@ -112,6 +115,8 @@ public abstract class PersistingRdfStreamConsumer implements RdfStreamConsumer {
 
             try {
                 operateOnTriple(t);
+            } catch (final ConstraintViolationException e) {
+                throw e;
             } catch (final MalformedRdfException e) {
                 exceptions.add(e.getMessage());
             }
@@ -146,6 +151,8 @@ public abstract class PersistingRdfStreamConsumer implements RdfStreamConsumer {
                         t);
                 operateOnProperty(t, subjectNode);
             }
+        } catch (final ConstraintViolationException e) {
+            throw e;
         } catch (final RepositoryException | RepositoryRuntimeException e) {
             throw new MalformedRdfException(e.getMessage(), e);
         }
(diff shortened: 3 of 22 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
```diff
@@ -493,6 +494,8 @@ public void replaceProperties(final IdentifierConverter<Resource, FedoraResource
         try {
             new RdfRemover(idTranslator, getSession(), replacementStream
                     .withThisContext(differencer)).consume();
+        } catch (final ConstraintViolationException e) {
```
ajs6f: Do we want to immediately rethrow a `ConstraintViolationException` or do we want to accumulate them to give a more full report?

fcrepo-webapp/src/main/webapp/static/constraints/owl2html.xsl
```diff
@@ -0,0 +1,198 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ajs6f: Is this being used "live" to generate HTML, or at we at least caching it?

ajs6f:
On the whole this looks really nice.

@whikloj ++

Just a few comments in-line, and I do want to make sure I understand how we are publishing the RDF/HTML and if we are doing it efficiently.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/exception/ConstraintViolationException.java
```diff
@@ -0,0 +1,82 @@
...
+     */
+    public ConstraintViolationException(final String msg, final Throwable rootCause) {
+        super(msg, rootCause);
+    }
+
+    /**
+     * Generates a link header for a constraint based on UriInfo and the name of the Exception class.
+     * @return Link
+     */
+    public static Link buildConstraintLink(final ConstraintViolationException e, final UriInfo uriInfo) {
```
ajs6f: Since the expected clients of this method are actually subclasses of this class, you could just make it an instance method, which would shorten the code a bit.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/exception/ConstraintViolationException.java
```diff
@@ -0,0 +1,82 @@
...
+        String constraintURI = "";
+        if (uriInfo != null) {
+            constraintURI =
+                uriInfo.getBaseUri().getScheme() + "://" + uriInfo.getBaseUri().getAuthority() + CONSTRAINT_DIR +
```
ajs6f: Not a huge thing, but maybe use a `String.format()` here and factor out `uriInfo.getBaseUri()` and `e.getClass().toString()` for clarity?

fcrepo-kernel/src/main/java/org/fcrepo/kernel/exception/ConstraintViolationException.java
```diff
@@ -0,0 +1,82 @@
...
+    public ConstraintViolationException(final String msg, final Throwable rootCause) {
+        super(msg, rootCause);
+    }
+
+    /**
+     * Generates a link header for a constraint based on UriInfo and the name of the Exception class.
+     * @return Link
+     */
+    public static Link buildConstraintLink(final ConstraintViolationException e, final UriInfo uriInfo) {
+        String constraintURI = "";
```
ajs6f: Instead of a mutable variable and `if-then`, you could do `final String constraintURI = uriInfo == null ? "" : $the-long-string-concatentation`. You could also do something crafty with `Optional`, but that's probably less clear.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
whikloj: You mean like we are doing directly below this with the MalformedRdfException? 
I was a little concerned about how we are reporting if we violate multiple constraints. In which case maybe our constraints should be a single file and the exception message would have to identify which constraint? Not sure...

fcrepo-webapp/src/main/webapp/static/constraints/owl2html.xsl
whikloj: Live, I was not even sure how to get it working so I didn't get any further into the guts of how one would do that.

ajs6f: Okay, that's cool, but let's make a ticket to get some caching or memoization in there.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraResourceImpl.java
ajs6f: Yeah, that's a very good point. I think you've got it right now. Most folks are not going to throw in RDF with many many violations, anyway.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/exception/ConstraintViolationException.java
whikloj: I seem to have to pass the uriInfo into this method regardless. If I don't, then it becomes null. Is this an aspect of @Context or am I doing something wrong?

whikloj: I am a moron, I had @Context in all the ExceptionMappers and *not* in the Exceptions. I think I got this.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/exception/ConstraintViolationException.java
```diff
@@ -0,0 +1,75 @@
...
+import javax.ws.rs.core.Link;
+import javax.ws.rs.core.UriInfo;
+
+/**
+ * A constraint has been violated.
+ *
+ * @author whikloj
+ * @since 2015-05-29
+ */
+public class ConstraintViolationException extends MalformedRdfException {
```
awoods: It seems like there are situations in this PR where the `ConstraintViolationException` has nothing to do with malformed RDF. Is there a particular reason why it is necessary that `ConstraintViolationException` extends `MalformedRdfException`?

fcrepo-kernel/src/main/java/org/fcrepo/kernel/exception/ManagedTypeException.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package org.fcrepo.kernel.exception;
+
+/**
+ * Certain mixin types are managed by the repository only.
+ *
+ * @author whikloj
+ * @since 2015-06-02
+ */
+public class ManagedTypeException extends ConstraintViolationException {
```
awoods: Is there a difference between this class and `ServerManagedPropertyException`?
If not, I would suggest removing this class.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/exception/ConstraintViolationException.java
whikloj: Most of these exceptions are due to errors in the RDF passed to Fedora. It is possible that the error is because the client was unaware of our constraint, but it is an error in the RDF as we see it.

It also made it easier to fit into the existing architecture as there are some ```try {} catch {}``` that handle ```MalformedRdfException``` that did not therefore need to be changed.

It can however be made a direct subclass of ```RepositoryRuntimeException``` if you prefer.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/exception/ManagedTypeException.java
whikloj: This was the part of this ticket I said I was not sure of in [the comments of the JIRA ticket](https://jira.duraspace.org/browse/FCREPO-1418?focusedCommentId=42814&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-42814). 
This was being disregarded previously [here](https://github.com/fcrepo4/fcrepo4/pull/808/files#diff-ef4f510e2f7370114164c1907486eb9dR149), and I am not entirely sure why or if this is needed at all.

That being said I have no problem removing this and using ServerManagedPropertyException in its place.

awoods: I would suggest removing `ManagedTypeException`, and replacing the one use of it in production code with `ServerManagedPropertyException`.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/exception/ConstraintViolationException.java
awoods: It is fine. We can leave the inheritance as is.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/exception/ManagedTypeException.java
ajs6f: I don't know if you've used it, but there clearly is a potential difference. `ManagedTypeException` could not be used to indicate a failure for the presence of managed _properties_. `ServerManagedPropertyException` could. I see them as clearly different.

awoods: This is the only place where `ManagedTypeException` is used:
https://github.com/fcrepo4/fcrepo4/pull/808/files#diff-ef4f510e2f7370114164c1907486eb9dR149

And it is not for that purpose.

ajs6f: How the heck is that not for the purpose? That is exactly for the purpose. A user tries to manipulate a repository-managed type and the exception is thrown.

awoods: You are correct. I misread your comment above. The two exception types seem redundant to me, but if you like them, then sure, let's keep them. At a minimum, for clarity I would suggest aligning their names to either:
* `ManagedTypeException` and `ManagedPropertyException`, or
* `ServerManagedTypeException` and `ServerManagedPropertyException`

ajs6f: I like the second choice-- it emphasizes that the problem is not that _someone_ is managing this stuff, but specifically that the _server_ is.

whikloj: Ok, so to make sure I have this right. I'm going to:
* Change ```ManagedTypeException``` to ```ServerManagedTypeException```
* Change [this line](https://github.com/fcrepo4/fcrepo4/pull/808/files#diff-ef4f510e2f7370114164c1907486eb9dR149) to throw a ```ServerManagedPropertyException```

Is that what I am hearing?

awoods: @whikloj, I believe the only suggestion is to change the class name as you noted:
* `ManagedTypeException` to `ServerManagedTypeException`

The line you mention does not need any change. Do you agree, @ajs6f?

ajs6f: I agree-- that line is fine.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/exception/ConstraintViolationException.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public ConstraintViolationException(final String msg, final Throwable rootCause) {
+        super(msg, rootCause);
+    }
+
+    /**
+     * Generates a link header for a constraint based on UriInfo and the name of the Exception class.
+     * @return Link
+     */
+    public Link buildConstraintLink(final UriInfo uriInfo) {
```
awoods: Please add a unit test for this class/method.

awoods:
Please add a new integration test method to FedoraLdpIT that creates an error case and verifies the created link header.

awoods:
It would be helpful to have an integration test that verifies:
# the existence of the static RDF pages are available in fcrepo-webapp (see: https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-webapp/src/test/java/org/fcrepo/integration/SanityCheckIT.java)
# the existence of the static RDF pages in the codebase, with the proper name, that correlates to the respective Exception classes.

fcrepo-webapp/src/main/webapp/static/constraints/owl2html.xsl
awoods: Additionally, this stylesheet tries to do much more than is needed by the constraint RDF documents. See attached screenshot for sections that should be removed from this xsl.
screenshot from 2015-06-22 16 32 32

fcrepo-kernel/src/main/java/org/fcrepo/kernel/exception/ConstraintViolationException.java
whikloj: How do you mock up a URI? I keep getting a target invocation error. Everything is working great then it executes the buildConstraintLink twice and tosses some exception to the JUnit.

awoods: See: https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraFixityTest.java#L70-L71

whikloj:
Regarding this SanityCheck, I think that it:
1. Would need the issue about Grizzly not serving static pages to be resolved first.
2. Because there are RDF files for a subset of Exceptions (subclasses of ConstraintViolationException), I'm not sure how you would locate those Exceptions.

As always, if you want to give me some **Java education** on a way to do this, I can add it to the PR. :wink:

fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/ConstraintViolationExceptionMapperTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private UriInfo mockInfo;
+
+    @Mock
+    private ConstraintViolationExceptionMapper mapper;
+
+    @Before
+    public void setUp() throws RepositoryException, URISyntaxException {
+        initMocks(this);
+        this.mockInfo = getUriInfoImpl();
```
ajs6f: You might want to use @MockitoJUnitRunner, it's a bit crisper to my eye.

fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/ConstraintViolationExceptionMapperTest.java
```diff
@@ -0,0 +1,66 @@
...
+public class ConstraintViolationExceptionMapperTest {
+
+    private UriInfo mockInfo;
+
+    @Mock
+    private ConstraintViolationExceptionMapper mapper;
+
+    @Before
+    public void setUp() throws RepositoryException, URISyntaxException {
+        initMocks(this);
```
ajs6f: You might want to use @MockitoJUnitRunner, it's a bit crisper to my eye.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/utils/iterators/PersistingRdfStreamConsumer.java
```diff
@@ -90,7 +94,7 @@ public boolean apply(final Triple t) {
                             t);
                 } else {
                     LOGGER.error("subject ({}) is not in repository domain.", t.getSubject().toString());
-                    throw new MalformedRdfException(String.format(
+                    throw new OutOfDomainSubjectException(String.format(
```
ajs6f: This invocation keeps occurring with what appears to be the same string. Maybe make it a constructor?

awoods:
See: https://github.com/whikloj/fcrepo4/pull/2
...will look at the SanityIT.java later.

awoods:
See: https://github.com/whikloj/fcrepo4/pull/3 for the SanityIT.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/exception/ConstraintViolationException.java
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.kernel.exception;
+
+import javax.ws.rs.core.UriInfo;
```
awoods: We do not want a `javax.ws.rs` dependency in `fcrepo-kernel`. You should be able to remove the following method:
https://github.com/fcrepo4/fcrepo4/pull/808/files#diff-c14ecfb343036db02d3070d59bd47b0eR64
...and perform all of the `buildConstraintLink` logic in `ConstraintExceptionMapper`, no?

fcrepo-kernel/src/main/java/org/fcrepo/kernel/exception/OutOfDomainSubjectException.java
```diff
@@ -0,0 +1,46 @@
...
+        super(String.format("RDF Stream contains subject(s) (%s) not in the domain of this repository.", subject));
+    }
+
+    /**
+     * Ordinary constructor.
+     *
+     * @param msg the message
+     * @param rootCause the root cause
+     */
+    public OutOfDomainSubjectException(final String msg, final Throwable rootCause) {
```
awoods: Method never used. Please remove.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/447e56a9546900f62a2f1ad7a75077280072e773

PR #812, opened 2015-06-04
FedoraEvent now mints internal UUID for downstream use (FCREPO-1507)

escowles:
* Reversing kernel/mint dependency so FedoraEvent can use UUIDPathMinter to mint IDs.
* First step in https://jira.duraspace.org/browse/FCREPO-1507

### DIFF
--- a/fcrepo-kernel/src/main/java/org/fcrepo/kernel/observer/FedoraEvent.java
+++ b/fcrepo-kernel/src/main/java/org/fcrepo/kernel/observer/FedoraEvent.java
@@ -34,6 +34,9 @@ import javax.jcr.observation.Event;
 import org.fcrepo.kernel.exception.RepositoryRuntimeException;
 import org.fcrepo.kernel.utils.EventType;
 
+import org.fcrepo.mint.PidMinter;
+import org.fcrepo.mint.UUIDPathMinter;
+
 import com.google.common.base.Function;
 import com.google.common.base.Joiner;
 import com.google.common.collect.Iterables;
@@ -48,9 +51,12 @@ import com.google.common.collect.Iterables;
 public class FedoraEvent {
 
     private Event e;
+    private String eventID;
 
     private Set<Integer> eventTypes = new HashSet<>();
     private Set<String> eventProperties = new HashSet<>();
+
+    private static final PidMinter pidMinter = new UUIDPathMinter();
 
     /**
      * Wrap a JCR Event with our FedoraEvent decorators
@@ -59,6 +65,7 @@ public class FedoraEvent {
      */
     public FedoraEvent(final Event e) {
         checkArgument(e != null, "null cannot support a FedoraEvent!");
+        eventID = pidMinter.mintPid();
         this.e = e;
     }
 
@@ -70,6 +77,7 @@ public class FedoraEvent {
      */
     public FedoraEvent(final FedoraEvent e) {
         checkArgument(e != null, "null cannot support a FedoraEvent!");
+        eventID = e.getEventID();
         this.e = e.e;
     }
 
@@ -182,6 +190,13 @@ public class FedoraEvent {
         } catch (RepositoryException e1) {
             throw new RepositoryRuntimeException("Error getting event date!", e1);
         }
+    }
+
+    /**
+     * Get the event ID.
+    **/
+    public String getEventID() {
+        return eventID;
     }
 
     @Override
--- a/fcrepo-kernel/src/main/java/org/fcrepo/kernel/observer/FedoraEvent.java
+++ b/fcrepo-kernel/src/main/java/org/fcrepo/kernel/observer/FedoraEvent.java
@@ -51,7 +51,7 @@ import com.google.common.collect.Iterables;
 public class FedoraEvent {
 
     private Event e;
-    private String eventID;
+    private final String eventID;
 
     private Set<Integer> eventTypes = new HashSet<>();
     private Set<String> eventProperties = new HashSet<>();
@@ -149,17 +149,6 @@ public class FedoraEvent {
     }
 
     /**
-     * @return the node identifer of the underlying JCR {@link Event}s
-     */
-    public String getIdentifier() {
-        try {
-            return e.getIdentifier();
-        } catch (RepositoryException e1) {
-            throw new RepositoryRuntimeException("Error getting event identifier!", e1);
-        }
-    }
-
-    /**
      * @return the info map of the underlying JCR {@link Event}s
      */
     public Map<Object, Object> getInfo() {
@@ -194,6 +183,7 @@ public class FedoraEvent {
 
     /**
      * Get the event ID.
+     * @return Event identifier to use for building event URIs (e.g., in an external triplestore).
     **/
     public String getEventID() {
         return eventID;
--- a/fcrepo-mint/src/main/java/org/fcrepo/mint/HttpPidMinter.java
+++ b/fcrepo-mint/src/main/java/org/fcrepo/mint/HttpPidMinter.java
@@ -182,10 +182,10 @@ public class HttpPidMinter implements PidMinter {
             return responseToPid( EntityUtils.toString(resp.getEntity()) );
         } catch ( IOException ex ) {
             LOGGER.warn("Error minting pid from {}: {}", url, ex);
-            throw new RuntimeException("Error minting pid", ex);
+            throw new PidMinterException("Error minting pid", ex);
         } catch ( Exception ex ) {
             LOGGER.warn("Error processing minter response", ex);
-            throw new RuntimeException("Error processing minter response", ex);
+            throw new PidMinterException("Error processing minter response", ex);
         }
     }
 }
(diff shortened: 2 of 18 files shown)


fcrepo-kernel/src/main/java/org/fcrepo/kernel/observer/FedoraEvent.java
```diff
@@ -184,6 +192,13 @@ public long getDate() {
         }
     }
 
+    /**
+     * Get the event ID.
```
awoods: Use `@return` javadoc annotation.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/observer/FedoraEvent.java
```diff
@@ -48,10 +51,13 @@
 public class FedoraEvent {
 
     private Event e;
+    private String eventID;
```
awoods: `eventID` can be `final`

fcrepo-kernel/src/main/java/org/fcrepo/kernel/observer/FedoraEvent.java
```diff
@@ -59,6 +65,7 @@
      */
     public FedoraEvent(final Event e) {
         checkArgument(e != null, "null cannot support a FedoraEvent!");
+        eventID = pidMinter.mintPid();
```
ajs6f: Do the docs for minters explain that they will be used for both resource IDs and event IDs? Some people may not want those ID spaces conflated.

fcrepo-mint/src/main/java/org/fcrepo/mint/HttpPidMinter.java
```diff
@@ -184,10 +182,10 @@ public String mintPid() {
             return responseToPid( EntityUtils.toString(resp.getEntity()) );
         } catch ( IOException ex ) {
             LOGGER.warn("Error minting pid from {}: {}", url, ex);
-            throw new RepositoryRuntimeException("Error minting pid", ex);
+            throw new RuntimeException("Error minting pid", ex);
```
ajs6f: Can we at least have a semantically-transparent subclass of `RuntimeException`, not just a naked `RuntimeException`? E.g. `PidMinteratingException`?

fcrepo-kernel/src/main/java/org/fcrepo/kernel/observer/FedoraEvent.java
escowles: Maybe we could inject a PidMinter instance instead of using the default constructor, allowing people to configure a minter specifically for events if they wanted to have a different id scheme than for nodes?

ajs6f: Yes, or they might want a different ID scheme for some other reason. Injection (with an appropriate default) is a good move here, I think.

awoods: Injection would be nice... although FedoraEvent objects are not Spring-managed.

ajs6f: Then perhaps we need to provide a factory type into which we can inject.

awoods: We could add a `PidMinter` parameter to the `FedoraEvent` constructor and inject the pidMinter instance into the `AllNodeEventsOneEvent`, spring-managed bean.
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/eventmappings/AllNodeEventsOneEvent.java#L50

escowles: Or we could do something like HttpClient useSystemProperties -- in this case, that would probably mean one property to specify the class to use, and then an expectation that the PidMinter instances would implement a useSystemProperties method, with a customizable suffix so you could use the same class for events and nodes with different parameters.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/observer/FedoraEvent.java
```diff
@@ -184,6 +192,13 @@ public long getDate() {
         }
     }
 
+    /**
+     * Get the event ID.
+    **/
+    public String getEventID() {
```
awoods: As a note, there is already a `getIdentifier()` method on this class... which appears to be only used in test code. It would be nice to remove the ambiguity of two get-ID methods if possible.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/observer/FedoraEvent.java
ajs6f: Oh, I don't like that idea, @escowles . That's a whole new mechanism for configuration and wiring we'd be introducing.

ajs6f: @awoods ' idea will work, but only if we can guarantee that the only place that `FedoraEvent`s will be produced will be implementations of `InternalExternalEventMapper` (including, but _not_ limited to, `AllNodeEventsOneEvent`). Otherwise we need an actual factory. I hate Spring. In CDI, this problem just wouldn't arise.

awoods: How would CDI change the picture in this case?

ajs6f: CDI can inject parameters into a constructor for _any_ class, which automatically becomes managed. So a class that needs a `FedoraEvent` could just use the constructor and CDI, behind the scenes, could inject an appropriate identifier minter.

escowles: Do we need to resolve the PidMinter injection/customization for this ticket?  This really seems like a thorny issue that could be its own ticket.

ajs6f: Agreed. But at least let's make that ticket, eh? Otherwise we'll forget.

escowles: I've created https://jira.duraspace.org/browse/FCREPO-1568 for allowing customization of the Event ID minting behavior.

awoods: @ajs6f, in your CDI scenario above, would `new FedoraEvent(...)` be called directly? In other words, how are instances created? It seems like calling `new` would circumvent any container management.

ajs6f: Yeah, that's a good point, and I misspoke. What you'd do is have a producer method in the class using `FedoraEvent`. The call to `new` would occur there and the result would be a managed bean.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/observer/FedoraEvent.java
escowles: getIdentifier() returns the underlying JCR event's identifier -- i.e., the path of the node that's being acted on.  I'll update the Javadoc to make this more clear.

awoods: ...however, if `getIdentifier()` is never actually used, maybe we get rid of it?

ajs6f: +1 to cutting off dangling threads.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/observer/FedoraEvent.java
awoods: /play nyan

ajs6f: Fedora technology is, indeed, at the level of Nyan cat. That's what is inside every Fedora repository, powering it.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/observer/FedoraEvent.java
escowles: I realized I was confusing getIdentifier() with getPath() (the output of getPath() is used in the identifier JMS header...).  So I'll go ahead and remove the getIdentifier method, and rework the handful of tests that use it.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/75916010e5dfbdc74e4e415837e3366babbbc834

PR #816, opened 2015-06-19
Add test that should pass... but does not :(

awoods:
This PR should not be merged into master. It should be incorporated (along with associated functional code) into PR-811.
An integration test on migration-utils revealed that PR-811 introduced an error:
http://jenkins.fcrepo.org/job/migration-utils/org.fcrepo.migration$migration-utils/115/testReport/junit/org.fcrepo.migration.handlers/BasicObjectVersionHandlerIT/org_fcrepo_migration_handlers_BasicObjectVersionHandlerIT/

PR-811 has been reverted from the master branch:
https://github.com/fcrepo4/fcrepo4/commit/50a04b588bc4369af9ef12a5be4a6073d665344d

Rework of the following ticket is required: https://jira.duraspace.org/browse/FCREPO-1409

### DIFF
--- a/fcrepo-kernel-impl/src/test/java/org/fcrepo/integration/kernel/impl/FedoraResourceImplIT.java
+++ b/fcrepo-kernel-impl/src/test/java/org/fcrepo/integration/kernel/impl/FedoraResourceImplIT.java
@@ -465,6 +465,19 @@ public class FedoraResourceImplIT extends AbstractIT {
     }
 
     @Test
+    public void testValidSparqlUpdateValidationAltSyntax() throws RepositoryException {
+        final String pid = UUID.randomUUID().toString();
+        final FedoraResource object = containerService.findOrCreate(session, pid);
+        object.updateProperties(subjects,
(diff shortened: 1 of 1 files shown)


ajs6f:
Does this connect to a ticket?

osmandin:
@awoods I'd like to confirm whether the statement in your PR is valid or not. When I change the underlying check logic (which did have a bug because it was counting on WHERE appearing after INSERT) and pass the statement as OK, I get an exception from Jena:

com.hp.hpl.jena.query.QueryParseException: Encountered " <INSERT_DATA> "INSERT DATA "" at line 1, column 88.
Was expecting one of:
    <EOF> 
    ";" ...

awoods:
I am checking it now, but in the meantime you can run the migration-utils against your original PR to see the error. If you enable TRACE on migration-utils, you will see the exact sparql-update request that fails.

awoods:
Closing this PR in favor of an updated PR to @osmandin's 1409 branch.
https://github.com/osmandin/fcrepo4/pull/5

PR #820, opened 2015-06-25
FCREPO-1599

ajs6f:
Minor use of Java 8 syntax and types in kernel observation code.


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/NOOPFilter.java
```diff
@@ -33,12 +31,7 @@
 public class NOOPFilter implements EventFilter {
 
     @Override
-    public Predicate<Event> getFilter(final Session session) {
-        return this;
-    }
-
-    @Override
-    public boolean apply(final Event event) {
+    public boolean test(final Event event) {
```
acoburn: Is this class actually used anywhere? Can it be entirely removed?

ajs6f: I don't think it is, I think it was left as an example. I have no brief for it. @awoods ?

awoods: This class has served its purpose... and deserves to be laid to rest. Drop the monkey on it.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/eventmappings/AllNodeEventsOneEvent.java
```diff
@@ -35,14 +36,12 @@
...
-import com.google.common.base.Function;
 import com.google.common.collect.Multimap;
 
 /**
- * Maps all JCR {@link Event}s concerning one JCR node to one
- * {@link FedoraEvent}. Adds the types of those JCR events together to calculate
- * the final type of the emitted FedoraEvent. TODO stop aggregating events in
- * the heap and make this a purely iterative algorithm, if possible
+ * Maps all JCR {@link Event}s concerning one JCR node to one {@link FedoraEvent}. Adds the types of those JCR events
+ * together to calculate the final type of the emitted FedoraEvent. <br/>
```
acoburn: This <br/> tag causes an error with `mvn javadoc:jar` (which is why travis fails). If you could remove it, I think this PR is good to go.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/NOOPFilter.java
acoburn: @ajs6f feel free to remove this class in this PR; otherwise, removing it can be part of a separate PR.

ajs6f: I'll add a commit with the excision. Monkeys? That's all Fedora needs. Monkeys in the codebase.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/eventmappings/AllNodeEventsOneEvent.java
ajs6f: Really? That's illegal now? Oh, well. I'll take it out. I wonder how you're supposed to do that kind of formatting now...

ajs6f:
Made changes: kill NOOPFilter and AllNodeEventsOneEvent Javadoc fix.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/aa8748ba908ca9203efab7fb7b398388aafb3113

PR #826, opened 2015-06-26
Minor code cleanup

ajs6f:
https://jira.duraspace.org/browse/FCREPO-1606

### DIFF
--- a/fcrepo-transform/src/main/java/org/fcrepo/transform/TransformationFactory.java
+++ b/fcrepo-transform/src/main/java/org/fcrepo/transform/TransformationFactory.java
@@ -44,8 +44,8 @@ public class TransformationFactory {
      * @throws SecurityException if security exception occurred
      */
     public TransformationFactory() {
-        mimeToTransform.put(contentTypeSPARQLQuery, is -> new SparqlQueryTransform(is));
-        mimeToTransform.put(APPLICATION_RDF_LDPATH, is -> new LDPathTransform(is));
+        mimeToTransform.put(contentTypeSPARQLQuery, SparqlQueryTransform::new);
+        mimeToTransform.put(APPLICATION_RDF_LDPATH, LDPathTransform::new);
     }
 
     /**
--- a/fcrepo-transform/src/main/java/org/fcrepo/transform/TransformationFactory.java
+++ b/fcrepo-transform/src/main/java/org/fcrepo/transform/TransformationFactory.java
@@ -23,6 +23,7 @@ import javax.ws.rs.core.MediaType;
 import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Function;
 
 import static org.apache.jena.riot.WebContent.contentTypeSPARQLQuery;
 import static org.fcrepo.transform.transformations.LDPathTransform.APPLICATION_RDF_LDPATH;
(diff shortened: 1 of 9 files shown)


fcrepo-transform/src/main/java/org/fcrepo/transform/TransformationFactory.java
```diff
@@ -31,18 +32,20 @@
...
+    private final Map<String, Function<InputStream, Transformation>> mimeToTransform = new HashMap<>();
 
     /**
      * Get a new TransformationFactory with the default classes
      * @throws SecurityException if security exception occurred
      */
     public TransformationFactory() {
-        mimeToTransform.put(contentTypeSPARQLQuery, new SparqlQueryTransform(null));
-        mimeToTransform.put(APPLICATION_RDF_LDPATH, new LDPathTransform(null));
+        mimeToTransform.put(contentTypeSPARQLQuery, is -> new SparqlQueryTransform(is));
```
awoods: Same comment as in earlier PR:
`mimeToTransform.put(contentTypeSPARQLQuery, SparqlQueryTransform::new);`

fcrepo-transform/src/main/java/org/fcrepo/transform/TransformationFactory.java
```diff
@@ -56,7 +59,7 @@ public TransformationFactory() {
     public <T> Transformation<T> getTransform(final MediaType contentType, final InputStream inputStream) {
         final String mimeType = contentType.toString();
         if (mimeToTransform.containsKey(mimeType)) {
-            return (Transformation<T>) mimeToTransform.get(contentType.toString()).newTransform(inputStream);
+            return mimeToTransform.get(contentType.toString()).apply(inputStream);
```
awoods: This seems incorrect to me. The previous logic returned a new transformation with the arg `inputStream` query. The new logic returns a new transformation trying to apply the arg `inputStream` as if it were an `RdfStream`.

ajs6f: No, it doesn't. The tests wouldn't even pass if it did that. `RdfStream` has no relevance here. Look at the new type of `mimeToTransform`: `Map<String, Function<InputStream, Transformation>>`. Give it a `String`, it will give you a `Function<InputStream, Transformation>`. The old type let you give it a `String` and gave you a `Transform` which was only used to produce a new `Transform` via `newTransform(InputStream)` (which no longer exists). `newTransform(InputStream)` was semantically equivalent to `Function<InputStream, Transformation>`, but the new way is not idiosyncratic. The semantics are the same.

awoods: What was perplexing me is the fact that `Transformation` has an `apply(RdfStream)` method that overrides `Function::apply`. However, there is another lambda function at play that has its own `apply` method.
I see the layers now.

ajs6f: It's currying. The whole type is: `String` -> `InputStream` -> `RdfStream` ->  `Something` (whatever the `Transformation` goes to).

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/1a4cf678bf4adf29c5d656d3fb730fed7f5c1a55

PR #827, opened 2015-07-02
Eliminate the publishing of UUID triples

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-1539

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/PropertiesRdfContext.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/PropertiesRdfContext.java
@@ -16,9 +16,11 @@
 package org.fcrepo.kernel.impl.rdf.impl;
 
 import static org.fcrepo.kernel.impl.utils.FedoraTypesUtils.isInternalProperty;
+import static org.fcrepo.kernel.utils.UncheckedPredicate.uncheck;
 import static org.slf4j.LoggerFactory.getLogger;
 
 import java.util.Iterator;
+import java.util.function.Predicate;
 
 import javax.jcr.Property;
 import javax.jcr.RepositoryException;
@@ -46,6 +48,8 @@ public class PropertiesRdfContext extends NodeRdfContext {
     private final PropertyToTriple property2triple;
 
     private static final Logger LOGGER = getLogger(PropertiesRdfContext.class);
+
+    private static final Predicate<Property> IS_NOT_UUID = uncheck(p -> !p.getName().equals("jcr:uuid"));
 
     /**
      * Default constructor.
@@ -76,7 +80,7 @@ public class PropertiesRdfContext extends NodeRdfContext {
         }
 
         final UnmodifiableIterator<Property> properties =
-                Iterators.filter(allProperties, isInternalProperty.negate()::test);
+                Iterators.filter(allProperties, isInternalProperty.negate().and(IS_NOT_UUID)::test);
         return Iterators.concat(Iterators.transform(properties, property2triple));
 
     }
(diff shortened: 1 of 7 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/PropertiesRdfContext.java
```diff
@@ -76,7 +82,7 @@ public PropertiesRdfContext(final FedoraResource resource,
         }
 
         final UnmodifiableIterator<Property> properties =
-                Iterators.filter(allProperties, isInternalProperty.negate()::test);
+                Iterators.filter(allProperties, and(isInternalProperty.negate()::test, not(IS_UUID)));
```
ajs6f: After the Java 8 PRs, perhaps better `isInternalProperty.and(IS_UUID.negate())::test`, or even change `IS_UUID` to `IS_NOT_UUID` and remove the negation here.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/PropertiesRdfContext.java
```diff
@@ -23,6 +26,7 @@
 import javax.jcr.Property;
 import javax.jcr.RepositoryException;
 
+import com.google.common.base.Predicate;
```
ajs6f: After the Java 8 PRs, this is better `java.util.function.Predicate`.

fcrepo-kernel-impl/src/test/java/org/fcrepo/integration/kernel/impl/FedoraResourceImplIT.java
```diff
@@ -202,12 +202,13 @@ public void testObjectGraph() throws RepositoryException {
         final Graph graph = object.getTriples(subjects, PropertiesRdfContext.class).asModel().getGraph();
 
         // jcr property
-        final Node s = createGraphSubjectNode(object);
+        Node s = createGraphSubjectNode(object);
         Node p = createURI(REPOSITORY_NAMESPACE + "uuid");
         Node o = createLiteral(object.getNode().getIdentifier());
-        assertTrue(graph.contains(s, p, o));
+        assertFalse(graph.contains(s, p, o));
```
ajs6f: Should this be `assertFalse(graph.contains(s, p, ANY));`? Does it matter what the object of this triple is? We just don't want to see anything with this predicate, right?

fcrepo-transform/src/test/java/org/fcrepo/integration/SparqlQueryTransformIT.java
```diff
@@ -67,7 +69,29 @@
     public void shouldDoStuff() throws RepositoryException {
```
ajs6f: We should probably name this test something.

fcrepo-transform/src/test/java/org/fcrepo/integration/SparqlQueryTransformIT.java
```diff
@@ -67,7 +69,29 @@
...
         final Session session = repo.login();
```
ajs6f: This should be in a `try-finally`.

fcrepo-transform/src/test/java/org/fcrepo/integration/SparqlQueryTransformIT.java
```diff
@@ -67,7 +69,29 @@
...
+        try (final QueryExecution qexec = testObj.apply(stream)) {
+            assert (qexec != null);
+            final ResultSet results = qexec.execSelect();
+            assert (results != null);
+            assertTrue(results.hasNext());
+        }
+    }
+
+    @Test
+    public void shouldDoStuffNoUUID() throws RepositoryException {
```
ajs6f: Not a great name for a test.

PR #835, opened 2015-07-12
Cleanup in fcrepo-http-api

ajs6f:
https://jira.duraspace.org/browse/FCREPO-1631


awoods:
It looks like Travis is failing with javadoc errors:
(mvn javadoc:test-aggregate)
https://s3.amazonaws.com/archive.travis-ci.org/jobs/70637161/log.txt

ajs6f:
Fixed, I think. Try again.

awoods:
Now Travis is saying there is a checkstyle error:
https://s3.amazonaws.com/archive.travis-ci.org/jobs/70794854/log.txt

Before submitting a PR, I would suggest running the following three commands:
* mvn clean install
* mvn javadoc:jar
* mvn javadoc:test-aggregate

ajs6f:
If that is the normal build check set, you need to actually key that into Maven. It's not hard to do that. Most people are going to expect that `mvn clean install` will be enough, and they're right.

awoods:
@ajs6f, the history here is that we used to have the javadoc plugin activated by default in the top-level pom.xml. Then, since we did not want to generate javadocs on each build, we moved the plugin into the release profile. Then, when we moved to Java8, javadoc errors became strictly enforced. 
The way our pom.xml is currently set up is to only run javadocs at release-time, therefore only showing errors at release-time... which does not make for a smooth release.
Therefore, we enabled Travis to run the javadoc commands that I mentioned above to strike a middle ground: users not having to build javadocs on every build, but the issues being caught before release.

I am happy to make changes if you have a suggestion, however.

ajs6f:
We moved the Javadoc step out because it wasn't needed to build and took time. Now it _is_ needed to build, so either we should stop strictly enforcing the Javadocs rules or it should be run for `verify`.

awoods:
Sounds good. Please create a ticket to re-add javadoc in the build process.

ajs6f:
https://jira.duraspace.org/browse/FCREPO-1634

awoods:
Resolved with: 
https://github.com/fcrepo4/fcrepo4/commit/43c3b3a4641fa5a4d629eec0da8bde878b8bae69 and
https://github.com/fcrepo4/fcrepo4/commit/31c1f11647bbd1be55b4aeda9803dd2417a7e129 and
https://github.com/fcrepo4/fcrepo4/commit/7141787b9cb139471ba081a42062509be8b24343

PR #836, opened 2015-07-13
https://jira.duraspace.org/browse/FCREPO-1411

ajs6f:
https://jira.duraspace.org/browse/FCREPO-1411

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -16,11 +16,11 @@
 package org.fcrepo.http.api;
 
 
-import static com.google.common.base.Predicates.not;
 import static com.google.common.collect.Iterators.concat;
 import static com.google.common.collect.Iterators.filter;
 import static com.google.common.collect.Iterators.transform;
 import static com.hp.hpl.jena.rdf.model.ModelFactory.createDefaultModel;
+import static com.hp.hpl.jena.vocabulary.RDF.type;
 import static javax.ws.rs.core.HttpHeaders.CACHE_CONTROL;
 import static javax.ws.rs.core.MediaType.APPLICATION_OCTET_STREAM_TYPE;
 import static javax.ws.rs.core.Response.ok;
@@ -105,7 +105,6 @@ import com.google.common.collect.Iterators;
 import com.hp.hpl.jena.graph.Triple;
 import com.hp.hpl.jena.rdf.model.Model;
 import com.hp.hpl.jena.rdf.model.Statement;
-import com.hp.hpl.jena.vocabulary.RDF;
 
 /**
  * An abstract class that sits between AbstractResource and any resource that
@@ -113,6 +112,7 @@ import com.hp.hpl.jena.vocabulary.RDF;
  * content.
  *
  * @author Mike Durbin
+ * @author ajs6f
  */
 public abstract class ContentExposingResource extends FedoraBaseResource {
 
@@ -135,6 +135,9 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
     protected FedoraResource resource;
 
     private static final long MAX_BUFFER_SIZE = 10240000;
+
+    private static final Predicate<Triple> IS_MANAGED_TYPE = t -> t.getPredicate().equals(type.asNode()) &&
+            isManagedNamespace.apply(t.getObject().getNameSpace());
 
     protected abstract String externalPath();
 
@@ -205,13 +208,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
         if (ldpPreferences.prefersServerManaged()) {
             tripleFilter = x -> true;
         } else {
-            tripleFilter = new Predicate<Triple>() {
-                @Override
-                public boolean test(final Triple input) {
-                    return input.getPredicate().equals(RDF.type.asNode())
-                            && isManagedNamespace.apply(input.getObject().getNameSpace());
-                }
-            }.negate().and(not(isManagedTriple)::apply);
+            tripleFilter = IS_MANAGED_TYPE.or(isManagedTriple::apply).negate();
         }
 
         if (ldpPreferences.prefersServerManaged()) {
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -16,8 +16,6 @@
 package org.fcrepo.http.api;
 
 
-import static com.google.common.base.Predicates.alwaysTrue;
-import static com.google.common.base.Predicates.and;
 import static com.google.common.base.Predicates.not;
 import static com.google.common.collect.Iterators.concat;
 import static com.google.common.collect.Iterators.filter;
@@ -41,6 +39,7 @@ import static org.fcrepo.kernel.RdfLexicon.DIRECT_CONTAINER;
 import static org.fcrepo.kernel.RdfLexicon.INDIRECT_CONTAINER;
 import static org.fcrepo.kernel.RdfLexicon.LDP_NAMESPACE;
 import static org.fcrepo.kernel.RdfLexicon.isManagedNamespace;
+import static org.fcrepo.kernel.impl.rdf.ManagedRdf.isManagedTriple;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -48,6 +47,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Date;
 import java.util.Iterator;
+import java.util.function.Predicate;
 
 import javax.inject.Inject;
 import javax.jcr.AccessDeniedException;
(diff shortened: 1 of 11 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -213,24 +203,24 @@ protected RdfStream getResourceTriples() {
...
-            tripleFilter = and(not(ManagedRdf.isManagedTriple), not(new Predicate<Triple>() {
+            tripleFilter = new Predicate<Triple>() {
                 @Override
-                public boolean apply(final Triple input) {
+                public boolean test(final Triple input) {
                     return input.getPredicate().equals(RDF.type.asNode())
                             && isManagedNamespace.apply(input.getObject().getNameSpace());
                 }
-            }));
+            }.negate().and(not(isManagedTriple)::apply);
```
acoburn: why not use `Predicate::or(isManagedTriple).negate()` here?

ajs6f: Because I stared at this too long to understand it or care any more. Send me a PR that seems clearer to you and I will happily merge it.

ajs6f: This whole mess of "isManagedFoot" and "isManagedEar" and specific type-tests needs to have the stuffing kicked out of it and beaten into something that makes some kind of sense.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -602,10 +585,16 @@ protected void replaceResourceWithStream(final FedoraResource resource,
...
     protected void patchResourcewithSparql(final FedoraResource resource,
-                                           final String requestBody,
-                                           final RdfStream resourceTriples)
-            throws MalformedRdfException, AccessDeniedException {
-        resource.updateProperties(translator(), requestBody, resourceTriples);
+            final String requestBody,
+            final RdfStream resourceTriples)
+                    throws MalformedRdfException, AccessDeniedException {
+        if (resource instanceof NonRdfSourceDescription) {
+            // update the description instead
```
awoods: This comment seems misleading. `// update the description instead`
The logic here is updating the described resource, i.e. the binary, not the description. 
According to your intent, is the comment wrong or the logic?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -592,7 +593,7 @@ public void testPatchBinaryDescription() throws Exception {
         patch.addHeader("Content-Type", "application/sparql-update");
         final BasicHttpEntity e = new BasicHttpEntity();
         e.setContent(new ByteArrayInputStream(
-                ("INSERT { <" + location +
+                ("INSERT { <" + serverAddress + pid + "/x" +
```
awoods: This integration test update would indicate that the logic:
https://github.com/fcrepo4/fcrepo4/pull/836/files#diff-d372035f89c8d0846920e20bd13e0145R591

...Should be `if (resource instanceof NonRdfSource)`

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
ajs6f: You're quite right-- that was from an earlier version. I'll fix the comment.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
ajs6f: No. Obviously, the test passes, so that can't be the case.

The simple sense is that descriptions contain triples about the described thing, so only triples with the described thing as their subject are legal.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
awoods: If we are putting all of the description on the binary, then what is the use of the NonRdfSourceDescription? just a URI to GET/PATCH on that under the covers actually interacts with the binary resource properties?
Also, this seems to be in conflict with the following ticket, which is moving in the direction of making the NonRdfSourceDescription a regular container: https://jira.duraspace.org/browse/FCREPO-1590

ajs6f: I agree that `fcr:metadata` makes little sense and I have argued in the past to get rid of it. I don't see that there is any conflict with that ticket. You're welcome to offer a different design.

awoods: We need to make sure that if we are going with the design in this ticket, that it reconcile with:
https://jira.duraspace.org/browse/FCREPO-1540 and
https://jira.duraspace.org/browse/FCREPO-1590

ajs6f: I never saw FCREPO-1540 and I completely disagree with it. FCREPO-1590 is unplanned and unassigned and unless you have in hand a design that will somehow better fit with the non-existent design for 1590, I have no interest in redoing this (1411).

ajs6f: [This](https://jira.duraspace.org/browse/FCREPO-1635) is the correct choice.

awoods: It is surprising that you never saw FCREPO-1540, as it was included in the last three Tech Meetings:
https://wiki.duraspace.org/display/FF/2015-07-02+-+Fedora+Tech+Meeting
https://wiki.duraspace.org/display/FF/2015-07-09+-+Fedora+Tech+Meeting
https://wiki.duraspace.org/display/FF/2015-07-16+-+Fedora+Tech+Meeting

@escowles, are you ok with the approach in this PR?

ajs6f: No, just because it's on a list of tickets does not mean that it was discussed. It was not. You should know, since you were at least some of those meetings.

escowles: I don't have a strong opinion about whether the triples should be on the NonRDFSource or the description.  But I do remember discussing them (before 7/1) and agreeing that it was going to be on the description.

awoods: @escowles, I also wanted to make sure that you did not see the approach in this PR conflicting with:
https://jira.duraspace.org/browse/FCREPO-1590

escowles: I think there are a couple of different priorities here, and different people care about different things, leading to some of this always-shifting discussion.  Some of the priorities are:

1. Full-featured descriptions of binaries (e.g., Skolemized bnodes, hash URI resources, child containers, etc.)
2. Binary descriptions tightly bound to the binary (e.g., deleted when the binary is deleted).
3. Request URI == RDF subject URI of the triples returned

My understanding was that the way we were going to handle these competing priorities was to keep fcr:metadata, make sure all binary properties used it as the RDF subject, and then add container/bnode/hashURI features to it.

awoods: Thanks, @escowles. I share your understanding. 
1. Given the priorities as you listed, do you see an architectural conflict with the approach presented in this PR?
2. @ajs6f, do you agree with the above noted priorities? and do you see an architectural conflict between the priorities and the approach presented in this PR?
3. Would storing properties on the binary vs. description resource be a more constructive approach?

ajs6f: We need to take this to a better forum than comments on an outdated diff.
Otherwise, your understanding is not supported by any of the conversation so far. For example, neither of the tickets about subjects of triples in `fcr:metadata` (1540 or 1635) demand that the subjects of the triples in question be `fcr:metadata`, as @escowles describes. They both demand that the subjects in question be the _NonRDFSource, instead_.
In other words, we don't have a coherent policy here that we can implement and we need to discuss this more publicly to get one.

awoods: +1 to public forum.
Yes, I misread @escowles comment: `My understanding was that the way we were going to handle these competing priorities was to keep fcr:metadata, make sure all binary properties used it as the RDF subject`.
My understanding was that the properties were to be stored on `fcr:metadata` but have the subject of the binary.

ajs6f: Right, but that's not what any of the issues say and it's not what I did here. Whether it's what we want to do or not, let's find out in a larger discussion.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/6e755b2768b5a4e7496119efbcdddef1f009c163

PR #840, opened 2015-07-14
Respond with 400 status on request with empty path segments

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-1537

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
@@ -112,6 +112,10 @@ public abstract class AbstractResourceIT {
         return new HttpGet(serverAddress + id);
     }
 
+    protected static HttpHead headObjMethod(final String id) {
+        return new HttpHead(serverAddress + id);
+    }
+
     protected static HttpDelete deleteObjMethod(final String id) {
         return new HttpDelete(serverAddress + id);
     }
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -194,7 +194,7 @@ public class FedoraLdpIT extends AbstractResourceIT {
     public void testHeadObject() throws IOException {
         final String id = getRandomUniqueId();
         createObject(id).close();
-        final HttpHead headObjMethod = new HttpHead(serverAddress + id);
+        final HttpHead headObjMethod = headObjMethod(id);
         assertEquals(OK.getStatusCode(), getStatus(headObjMethod));
     }
 
@@ -202,7 +202,7 @@ public class FedoraLdpIT extends AbstractResourceIT {
     public void testHeadDefaultContainer() throws IOException {
         final String id = getRandomUniqueId();
         createObject(id).close();
-        final HttpHead headObjMethod = new HttpHead(serverAddress + id);
+        final HttpHead headObjMethod = headObjMethod(id);
         try (final CloseableHttpResponse response = execute(headObjMethod)) {
             assertTrue("Didn't find LDP container link header!", getLinkHeaders(response).contains(
                     BASIC_CONTAINER_LINK_HEADER));
@@ -216,7 +216,7 @@ public class FedoraLdpIT extends AbstractResourceIT {
         createObjectAndClose(id);
         addMixin(id, BASIC_CONTAINER.getURI());
 
-        final HttpHead headObjMethod = new HttpHead(serverAddress + id);
+        final HttpHead headObjMethod = headObjMethod(id);
         try (final CloseableHttpResponse response = execute(headObjMethod)) {
             final Collection<String> links = getLinkHeaders(response);
             assertTrue("Didn't find LDP container link header!", links.contains(BASIC_CONTAINER_LINK_HEADER));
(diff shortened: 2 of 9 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
```diff
@@ -112,6 +112,10 @@ protected static HttpGet getObjMethod(final String id) {
         return new HttpGet(serverAddress + id);
     }
 
+    protected static HttpHead headObjMethod(final String id) {
+        return new HttpHead(serverAddress + id);
+    }
+
```
ajs6f: Did you factor this into the subclasses of `AbstractResourceIT` every time a `HEAD` request occurs?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -1732,6 +1732,12 @@ public void testJsonLdProfile() throws IOException {
         assertEquals("Should be two values!", 2, titles.findValues("@value").size());
     }
 
+    @Test
+    public void testEmptyPath() {
```
ajs6f: This isn't an empty path. Maybe `testPathWithEmptySegment` or at least a comment indicating the eact meaning here?

fcrepo-kernel/src/main/java/org/fcrepo/kernel/exception/InvalidResourceIdentifierException.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package org.fcrepo.kernel.exception;
+
+/**
+ * This exception is used for invalid resource identifiers, such as when a resource path has empty segments.
+ *
+ * @author awoods
+ * @since July 14, 2015
+ */
+public class InvalidResourceIdentifierException extends RepositoryRuntimeException {
```
ajs6f: So this is specifically distinct from when the identifier is valid, but there is no such resource? You might want to add that as a comment. Also, you might want to remark that the notion of validity will depend on the `IdentifierConverter` chain in use.

fcrepo-kernel/src/main/java/org/fcrepo/kernel/utils/NamespaceTools.java
```diff
@@ -90,6 +94,12 @@ public static void validatePath(final Session session, final String path) {
         final String relPath = path.replaceAll("^/+", "").replaceAll("/+$", "");
         final String[] pathSegments = relPath.split("/");
         for (final String segment : pathSegments) {
+            // Empty path elements are not valid!
```
ajs6f: No, this is the wrong place for this test. Putting it here assumes that HTTP URIs are translated wholemeal into JCR paths, which is the exact assumption that the `IdentifierConverter` system was designed to break. This test belongs in `HttpResourceConverter` or near it.

ajs6f: Also, is this looping test better than `relPath.contains("//")`?

awoods: I would take it even further in saying that this test should exist deeper in the core such that it is performed regardless of whether the request is over HTTP or native Java.

I also, completely agree in principle that this check breaks the design and intent of the `IdentifierConverter`. However, I believe consistency outweighs design principle in this case given the following:
- NamespaceTools#validatePath() is called in two places: 
* https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java#L129
* https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/NodeServiceImpl.java#L57

#validatePath() called from HttpResourceConverter#doForward() is already doing something similar to what you are asking for.
#validatePath() called from NodeServiceImpl#exists() is currently used in 6 different places in the RESTful services: FedoraLdp and FedoraNodes. These calls from RESTful services to NodeServiceImpl#exists() are the source of the real issue of breaking the `IdentifierConverter` design.

I think the future refactoring of the #exists() calls to be more in line with the `IdentifierConverter` design will be much simpler and cleaner if we keep the validation logic consolidated versus scattering it throughout the codebase in the short term.

Your other comments can be trivially addressed once we get this question sorted.

ajs6f: No, those locations are correct usage of this method, and the method is fine. You are implementing the new test in the wrong place. It should not be in this method. This method processes JCR paths, but you have been asked to process an HTTP URI. The new test should be in `HttpResourceConverter` and nowhere else, and it should not be used anywhere else. You should also refactor `FedoraLdp` and `FedoraNodes` to avoid using this method, and instead to use a validation strategy based on the use of an `IdentifierConverter`.

Consistency in error is no virtue.

awoods: It is difficult to agree or disagree with your statements without knowing exactly to which `those locations` or `this method` are referring. Would you please edit your comment to swap in the values for the references?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
awoods: yes

ajs6f: Awesometastic.

ajs6f:
Squash-merged via: 
https://github.com/fcrepo4/fcrepo4/commit/01305e29d102a49c2059483eb3b8af5a6439c3f3

PR #829, opened 2015-07-05
FCREPO-1505 (https://jira.duraspace.org/browse/FCREPO-1505)

jrgriffiniii:
Extending the FedoraLdp#createObject method for handling the creation of child objects for pairtree child resources; Implementing the FedoraLdpIT#testIngestOnPairTree method for integration testing.

PR #841, opened 2015-07-15
FCREPO-1627 (https://jira.duraspace.org/browse/FCREPO-1627)

jrgriffiniii:
Raising a ClientErrorException with a 403 status code for attempts to create graphs using PUT requests for resources under pairtree Nodes within FedoraLdp#createOrReplaceObjectRdf; Implementing FedoraLdpIT#testPutOnPairtree for the case in which clients attempt to PUT graphs under pairtree Nodes


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -271,6 +271,8 @@ public Response createOrReplaceObjectRdf(
 
         if (httpConfiguration.putRequiresIfMatch() && StringUtils.isBlank(ifMatch) && !resource.isNew()) {
             throw new ClientErrorException("An If-Match header is required", 428);
+        } else if (resource.hasType(FEDORA_PAIRTREE)) {
+            throw new ClientErrorException("Graphs cannot be created for resources under pairtree nodes", FORBIDDEN);
```
ajs6f: Do we want something more semantically specific here?

awoods: @ajs6f, are you suggesting a new, more specific exception type?

ajs6f: Ye, something like `CannotModifySystemResourceException1` except with a snappier title.

jrgriffiniii: I would please propose the following in response to this:

* ForbiddenResourceModificationException
* UnsupportedResourceModificationException
* IllegalResourceModification
* BadResourceModification

My preference lies primarily with the term "Forbidden" (the mapping to the HTTP 403 response description feels appropriate), but I could certainly defend "Unsupported" as well.

"Illegal" and "Bad" concern me given that they may easily be applicable within situations in which, I would speculate, a 405 response is appropriate.  "Cannot" also concerns me in this regard (as well as it being identified as undesirable).

ajs6f: `ForbiddenResourceModificationException` is fine with me.

awoods: :+1:

awoods:
This looks good to me. @ajs6f, let me know if your review is done, then I will squash and merge tonight.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -271,6 +273,9 @@ public Response createOrReplaceObjectRdf(
 
         if (httpConfiguration.putRequiresIfMatch() && StringUtils.isBlank(ifMatch) && !resource.isNew()) {
             throw new ClientErrorException("An If-Match header is required", 428);
+        } else if (resource.hasType(FEDORA_PAIRTREE)) {
+            throw new ForbiddenResourceModificationException("Graphs cannot be created for resources " +
+                                                              "under pairtree nodes");
```
ajs6f: I'm a little confused here, but shouldn't this read `"Graphs cannot be created for resources that are pairtree nodes"` and wouldn't it be better to have one string on one line?

awoods: Personally, I would prefer wording along the lines of, '"Resources cannot be created under pairtree elements."'

ajs6f: I'm not worried about the word "graph" and if you like it better without it, fine by me. I'm confused about whether at this moment in the workflow the client is trying to create anything under `resource` or is trying to create `resource` or what. I just want to make sure a) that this test is coming at the right time in the workflow and b) that it is throwing out an exception with the right message.

awoods: POST without a slug creates a resource as a child of the POST target container.
e.g. POST /rest/test
results in:
/rest/test/new-resource

ajs6f: Okay, cool then. How about `"Resources cannot be manually created under pairtree elements!"`

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -271,6 +273,9 @@ public Response createOrReplaceObjectRdf(
 
         if (httpConfiguration.putRequiresIfMatch() && StringUtils.isBlank(ifMatch) && !resource.isNew()) {
             throw new ClientErrorException("An If-Match header is required", 428);
+        } else if (resource.hasType(FEDORA_PAIRTREE)) {
+            throw new ForbiddenResourceModificationException("Graphs cannot be created for resources " +
```
awoods: We should probably have symmetry between this line (277) and the previous ticket:
https://github.com/fcrepo4/fcrepo4/pull/829/files#diff-4a14fb4dcbde6f7d32274c43ca723c16R421

I would suggest changing the `ClientErrorException` from the previous ticket to the new `ForbiddenResourceModificationException`.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -708,6 +708,25 @@ public void testPutMalformedRDFOnObject() throws IOException {
...
+     * @throws IOException in case of IOException
+     */
+    @Test
+    public void testPutOnPairtree() throws IOException {
+
+        final String objName = getLocation(postObjMethod());
+        final String pairtreeName = objName.substring(serverAddress.length(), objName.lastIndexOf('/'));
+        final String subjectURI = serverAddress + pairtreeName;
+
+        final HttpPut put = putObjMethod(pairtreeName);
```
awoods: As noted in the JIRA ticket, this test should be something like:
`final HttpPut put = putObjMethod(pairtreeName + "/test");`

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -272,6 +273,9 @@ public Response createOrReplaceObjectRdf(
 
         if (httpConfiguration.putRequiresIfMatch() && StringUtils.isBlank(ifMatch) && !resource.isNew()) {
             throw new ClientErrorException("An If-Match header is required", 428);
+        } else if (resource.hasType(FEDORA_PAIRTREE)) {
```
awoods: This test should happen before the resource is already created at line 267.

jrgriffiniii:
After failed attempts to rebase the fcrepo-1627 branch, this pull request is deprecated by #901.

PR #842, opened 2015-07-16
Adding appropriate inbound triples based on LDP containment properties

escowles:
* adding IT to cover typical PCDM use case of proxies

Fixes https://jira.duraspace.org/browse/FCREPO-1497

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -1065,8 +1065,8 @@ public class FedoraLdpIT extends AbstractResourceIT {
         final String uuid = getRandomUniqueId();
         final String pid1 = uuid + "/parent";
         final String pid2 = uuid + "/child";
-        createObject(pid1);
-        createObject(pid2);
+        createObjectAndClose(pid1);
+        createObjectAndClose(pid2);
 
         final String memberRelation = "http://pcdm.org/models#hasMember";
 
@@ -1091,50 +1091,50 @@ public class FedoraLdpIT extends AbstractResourceIT {
         // retrieve the parent and verify the outbound triples exist
         final HttpGet getParent =  new HttpGet(serverAddress + pid1);
         getParent.addHeader("Accept", "application/n-triples");
-        final GraphStore parentGraph = getGraphStore(getParent);
-
-        assertTrue(parentGraph.contains(Node.ANY,
-                createURI(serverAddress + pid1),
-                createURI(memberRelation),
-                createURI(serverAddress + pid2)));
+        try (final CloseableGraphStore parentGraph = getGraphStore(getParent)) {
+            assertTrue(parentGraph.contains(Node.ANY,
+                    createURI(serverAddress + pid1),
+                    createURI(memberRelation),
+                    createURI(serverAddress + pid2)));
+        }
 
         // retrieve the members container and verify the LDP triples exist
         final HttpGet getContainer =  new HttpGet(serverAddress + pid1 + "/members");
         getContainer.addHeader("Prefer", "return=representation;include=\"http://www.w3.org/ns/ldp#PreferMembership\"");
         getContainer.addHeader("Accept", "application/n-triples");
-        final GraphStore containerGraph = getGraphStore(getContainer);
+        try (final CloseableGraphStore containerGraph = getGraphStore(getContainer)) {
+            assertTrue(containerGraph.contains(Node.ANY,
+                    createURI(serverAddress + pid1 + "/members"),
+                    createURI("http://www.w3.org/ns/ldp#hasMemberRelation"),
+                    createURI(memberRelation)));
 
-        assertTrue(containerGraph.contains(Node.ANY,
-                createURI(serverAddress + pid1 + "/members"),
-                createURI("http://www.w3.org/ns/ldp#hasMemberRelation"),
-                createURI(memberRelation)));
+            assertTrue(containerGraph.contains(Node.ANY,
+                    createURI(serverAddress + pid1 + "/members"),
+                    createURI("http://www.w3.org/ns/ldp#insertedContentRelation"),
+                    createURI("http://www.openarchives.org/ore/terms/proxyFor")));
 
-        assertTrue(containerGraph.contains(Node.ANY,
-                createURI(serverAddress + pid1 + "/members"),
-                createURI("http://www.w3.org/ns/ldp#insertedContentRelation"),
-                createURI("http://www.openarchives.org/ore/terms/proxyFor")));
-
-        assertTrue(containerGraph.contains(Node.ANY,
-                createURI(serverAddress + pid1 + "/members"),
-                createURI("http://www.w3.org/ns/ldp#membershipResource"),
-                createURI(serverAddress + pid1)));
+            assertTrue(containerGraph.contains(Node.ANY,
+                    createURI(serverAddress + pid1 + "/members"),
+                    createURI("http://www.w3.org/ns/ldp#membershipResource"),
+                    createURI(serverAddress + pid1)));
+        }
 
 
         // retrieve the member and verify inbound triples exist
         final HttpGet getMember =  new HttpGet(serverAddress + pid2);
         getMember.addHeader("Prefer", "return=representation; include=\"" + INBOUND_REFERENCES.toString() + "\"");
         getMember.addHeader("Accept", "application/n-triples");
-        final GraphStore memberGraph = getGraphStore(getMember);
+        try (final CloseableGraphStore memberGraph = getGraphStore(getMember)) {
+            assertTrue(memberGraph.contains(Node.ANY,
+                    Node.ANY,
+                    createURI("http://www.openarchives.org/ore/terms/proxyFor"),
+                    createURI(serverAddress + pid2)));
 
-        assertTrue(memberGraph.contains(Node.ANY,
-                Node.ANY,
-                createURI("http://www.openarchives.org/ore/terms/proxyFor"),
-                createURI(serverAddress + pid2)));
-
-        assertTrue(memberGraph.contains(Node.ANY,
-                createURI(serverAddress + pid1),
-                createURI(memberRelation),
-                createURI(serverAddress + pid2)));
+            assertTrue(memberGraph.contains(Node.ANY,
+                    createURI(serverAddress + pid1),
+                    createURI(memberRelation),
+                    createURI(serverAddress + pid2)));
+        }
     }
 
     @Test
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -1061,7 +1061,84 @@ public class FedoraLdpIT extends AbstractResourceIT {
     }
 
     @Test
-    public void testGetObjectGraphLacksUUID() throws IOException {
+    public void testGetObjectReferencesIndirect() throws Exception {
+        final String uuid = getRandomUniquePid();
+        final String pid1 = uuid + "/parent";
+        final String pid2 = uuid + "/child";
+        createObject(pid1);
+        createObject(pid2);
+
+        final String memberRelation = "http://pcdm.org/models#hasMember";
+
+        // create an indirect container
+        final HttpPut createContainer = new HttpPut(serverAddress + pid1 + "/members");
+        createContainer.addHeader("Content-Type", "text/turtle");
+        final String membersRDF = "<> a <http://www.w3.org/ns/ldp#IndirectContainer>; "
+            + "<http://www.w3.org/ns/ldp#hasMemberRelation> <" + memberRelation + ">; "
+            + "<http://www.w3.org/ns/ldp#insertedContentRelation> <http://www.openarchives.org/ore/terms/proxyFor>; "
+            + "<http://www.w3.org/ns/ldp#membershipResource> <" + serverAddress + pid1 + "> . ";
+        createContainer.setEntity(new StringEntity(membersRDF));
+        assertEquals(CREATED.getStatusCode(), getStatus(createContainer));
+
+        // create a proxy in the indirect container
+        final HttpPost createProxy = new HttpPost(serverAddress + pid1 + "/members");
+        createProxy.addHeader("Content-Type", "text/turtle");
+        final String proxyRDF = "<> <http://www.openarchives.org/ore/terms/proxyFor> <" + serverAddress + pid2 + ">;"
+            + " <http://www.openarchives.org/ore/terms/proxyIn> <" + serverAddress + pid1 + "> .";
+        createProxy.setEntity(new StringEntity(proxyRDF));
+        assertEquals(CREATED.getStatusCode(), getStatus(createProxy));
+
+        // retrieve the parent and verify the outbound triples exist
+        final HttpGet getParent =  new HttpGet(serverAddress + pid1);
+        getParent.addHeader("Accept", "application/n-triples");
+        final GraphStore parentGraph = getGraphStore(getParent);
+
+        assertTrue(parentGraph.contains(Node.ANY,
+                NodeFactory.createURI(serverAddress + pid1),
+                NodeFactory.createURI(memberRelation),
+                NodeFactory.createURI(serverAddress + pid2)));
+
+        // retrieve the members container and verify the LDP triples exist
+        final HttpGet getContainer =  new HttpGet(serverAddress + pid1 + "/members");
+        getContainer.addHeader("Prefer", "return=representation;include=\"http://www.w3.org/ns/ldp#PreferMembership\"");
+        getContainer.addHeader("Accept", "application/n-triples");
+        final GraphStore containerGraph = getGraphStore(getContainer);
+
+        assertTrue(containerGraph.contains(Node.ANY,
+                NodeFactory.createURI(serverAddress + pid1 + "/members"),
+                NodeFactory.createURI("http://www.w3.org/ns/ldp#hasMemberRelation"),
+                NodeFactory.createURI(memberRelation)));
+
+        assertTrue(containerGraph.contains(Node.ANY,
+                NodeFactory.createURI(serverAddress + pid1 + "/members"),
+                NodeFactory.createURI("http://www.w3.org/ns/ldp#insertedContentRelation"),
+                NodeFactory.createURI("http://www.openarchives.org/ore/terms/proxyFor")));
+
+        assertTrue(containerGraph.contains(Node.ANY,
+                NodeFactory.createURI(serverAddress + pid1 + "/members"),
+                NodeFactory.createURI("http://www.w3.org/ns/ldp#membershipResource"),
+                NodeFactory.createURI(serverAddress + pid1)));
+
+
+        // retrieve the member and verify inbound triples exist
+        final HttpGet getMember =  new HttpGet(serverAddress + pid2);
+        getMember.addHeader("Prefer", "return=representation; include=\"" + INBOUND_REFERENCES.toString() + "\"");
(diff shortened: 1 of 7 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -1061,7 +1061,84 @@ public void testGetObjectReferences() throws IOException {
...
+        assertTrue(parentGraph.contains(Node.ANY,
+                createURI(serverAddress + pid1),
+                createURI(memberRelation),
+                createURI(serverAddress + pid2)));
+
+        // retrieve the members container and verify the LDP triples exist
+        final HttpGet getContainer =  new HttpGet(serverAddress + pid1 + "/members");
+        getContainer.addHeader("Prefer", "return=representation;include=\"http://www.w3.org/ns/ldp#PreferMembership\"");
+        getContainer.addHeader("Accept", "application/n-triples");
+        final GraphStore containerGraph = getGraphStore(getContainer);
```
ajs6f: Use a `CloseableGraphStore` with `try-with-resource` here.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -1061,7 +1061,84 @@ public void testGetObjectReferences() throws IOException {
...
+        createProxy.addHeader("Content-Type", "text/turtle");
+        final String proxyRDF = "<> <http://www.openarchives.org/ore/terms/proxyFor> <" + serverAddress + pid2 + ">;"
+            + " <http://www.openarchives.org/ore/terms/proxyIn> <" + serverAddress + pid1 + "> .";
+        createProxy.setEntity(new StringEntity(proxyRDF));
+        assertEquals(CREATED.getStatusCode(), getStatus(createProxy));
+
+        // retrieve the parent and verify the outbound triples exist
+        final HttpGet getParent =  new HttpGet(serverAddress + pid1);
+        getParent.addHeader("Accept", "application/n-triples");
+        final GraphStore parentGraph = getGraphStore(getParent);
```
ajs6f: Use a `CloseableGraphStore` with `try-with-resource` here.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -1061,7 +1061,84 @@ public void testGetObjectReferences() throws IOException {
     }
 
     @Test
-    public void testGetObjectGraphLacksUUID() throws IOException {
+    public void testGetObjectReferencesIndirect() throws Exception {
+        final String uuid = getRandomUniqueId();
+        final String pid1 = uuid + "/parent";
+        final String pid2 = uuid + "/child";
+        createObject(pid1);
```
ajs6f: Use `createObjectAndClose()` if you don't need to keep the response open.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -1061,7 +1061,84 @@ public void testGetObjectReferences() throws IOException {
...
+                createURI(serverAddress + pid1 + "/members"),
+                createURI("http://www.w3.org/ns/ldp#membershipResource"),
+                createURI(serverAddress + pid1)));
+
+
+        // retrieve the member and verify inbound triples exist
+        final HttpGet getMember =  new HttpGet(serverAddress + pid2);
+        getMember.addHeader("Prefer", "return=representation; include=\"" + INBOUND_REFERENCES.toString() + "\"");
+        getMember.addHeader("Accept", "application/n-triples");
+        final GraphStore memberGraph = getGraphStore(getMember);
```
ajs6f: Use a `CloseableGraphStore` with `try-with-resource` here.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/LdpContainerRdfContext.java
```diff
@@ -203,9 +203,13 @@ public FedoraResourceTripleFunction(final String insertedContainerProperty,
                 return Iterators.transform(values, new Function<Value, Triple>() {
                     @Override
                     public Triple apply(final Value input) {
-                        final RDFNode membershipResource = new ValueConverter(session(), translator())
-                                .convert(input);
-                        return create(subject(), memberRelation, membershipResource.asNode());
+                        try {
+                            final RDFNode membershipResource = new ValueConverter(resource().getNode().getSession(),
```
ajs6f: Why is the session in the transformation not the right one here?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/ReferencesRdfContext.java
```diff
@@ -48,21 +57,36 @@ public ReferencesRdfContext(final FedoraResource resource,
...
+        putReferencesIntoContext(resource.getNode().getWeakReferences());
+        putReferencesIntoContext(resource.getNode().getReferences());
     }
 
-    private Iterator<Triple> putWeakReferencePropertiesIntoContext() throws RepositoryException {
-        final Iterator<Property> properties = resource().getNode().getWeakReferences();
+    private void putReferencesIntoContext(final Iterator<Property> properties) throws RepositoryException {
+        while (properties.hasNext()) {
+            final Property p = properties.next();
+            concat(property2triple.apply(p));
```
ajs6f: This is going to result in _many_ more object creations. It's really not great to `concat` one triple at a time if you can possibly handle them in bunches.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/ReferencesRdfContext.java
```diff
@@ -48,21 +57,36 @@ public ReferencesRdfContext(final FedoraResource resource,
...
 
-        return Iterators.concat(Iterators.transform(properties, property2triple));
+            for ( final PropertyIterator it = p.getParent().getProperties(); it.hasNext(); ) {
+                final Property potentialProxy = it.nextProperty();
```
ajs6f: Why not leave this inside the `for`?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/ReferencesRdfContext.java
```diff
@@ -48,21 +57,36 @@ public ReferencesRdfContext(final FedoraResource resource,
...
+        }
     }
-
-    private Iterator<Triple> putStrongReferencePropertiesIntoContext() throws RepositoryException {
-        final Iterator<Property> properties = resource().getNode().getReferences();
-
-        return Iterators.concat(Iterators.transform(properties, property2triple));
-
+    private void putProxyReferencesIntoContext(final Value v) throws RepositoryException {
+        if (v.getType() == PATH) {
```
ajs6f: Don't we have dedicated code for this kind of dereferencing?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/LdpContainerRdfContext.java
awoods: I suspect the `session()` is not getting set in the first place. Is there danger in updating:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/NodeRdfContext.java#L52
...to pass in to super() `resource.getNode().getSession()`?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/ReferencesRdfContext.java
```diff
@@ -48,21 +57,36 @@ public ReferencesRdfContext(final FedoraResource resource,
         throws RepositoryException {
         super(resource, idTranslator);
         property2triple = new PropertyToTriple(resource.getNode().getSession(), idTranslator);
-        concat(putStrongReferencePropertiesIntoContext());
-        concat(putWeakReferencePropertiesIntoContext());
+        session = resource.getNode().getSession();
```
awoods: The grandparent class of `ReferencesRdfContext` already has a `Session` data member:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel/src/main/java/org/fcrepo/kernel/utils/iterators/RdfStream.java#L52

It is probably not set in this scenario either. 
The comment above about passing in a Session at:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/NodeRdfContext.java#L52 
...could also resolve the issue of needing to produce a local session in this class.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/LdpContainerRdfContext.java
escowles: That seems like a good suggestion to me -- I don't see the session getting set anywhere, and doing it in the NodeRdfContext would fix this awkward workaround, and might let us clean up similar workarounds elsewhere.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/ReferencesRdfContext.java
```diff
@@ -48,21 +57,36 @@ public ReferencesRdfContext(final FedoraResource resource,
...
+        putReferencesIntoContext(resource.getNode().getWeakReferences());
+        putReferencesIntoContext(resource.getNode().getReferences());
     }
 
-    private Iterator<Triple> putWeakReferencePropertiesIntoContext() throws RepositoryException {
-        final Iterator<Property> properties = resource().getNode().getWeakReferences();
+    private void putReferencesIntoContext(final Iterator<Property> properties) throws RepositoryException {
+        while (properties.hasNext()) {
```
awoods: This approach seems to lose the previous advantage of generating triples on-demand, since the update runs through the iterators during object creation.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/LdpContainerRdfContext.java
ajs6f: I dimly, dimly remember this being the intent for that field; that it would be a convenient handle for the `Session` in context without constantly dealing with JCR's annoying `RepositoryException` addiction.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/ReferencesRdfContext.java
ajs6f: @awoods  is dead right here. Stay lazy, stay happy.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/ReferencesRdfContext.java
awoods: This could help:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/converters/ValueConverter.java#L78
...as well as possibly this:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/utils/FedoraTypesUtils.java#L232

Although when dealing with "references" on Fedora, we also need to be conscious of these:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/utils/FedoraTypesUtils.java#L211 and
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/utils/FedoraTypesUtils.java#L220

ajs6f: Do we have the work that @cbeer did to set up that reference-property scheme documented somewhere? I have to admit I don't really understand it.

awoods: nope

ajs6f: AWESOME SAUCE.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/ReferencesRdfContext.java
```diff
@@ -48,21 +57,36 @@ public ReferencesRdfContext(final FedoraResource resource,
...
+            final Property p = properties.next();
+            concat(property2triple.apply(p));
 
-        return Iterators.concat(Iterators.transform(properties, property2triple));
+            for ( final PropertyIterator it = p.getParent().getProperties(); it.hasNext(); ) {
```
awoods: Minor note: some inline comments would be helpful.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/ReferencesRdfContext.java
```diff
@@ -48,21 +57,36 @@ public ReferencesRdfContext(final FedoraResource resource,
...
-
+    private void putProxyReferencesIntoContext(final Value v) throws RepositoryException {
+        if (v.getType() == PATH) {
+            putProxyReferencesIntoContext(session.getNode(v.getString()));
+        } else if (v.getType() == REFERENCE || v.getType() == WEAKREFERENCE) {
+            putProxyReferencesIntoContext(session.getNodeByIdentifier(v.getString()));
+        }
+    }
+    private void putProxyReferencesIntoContext(final Node n) throws RepositoryException {
+        concat(new LdpContainerRdfContext(nodeConverter.convert(n), translator()));
```
awoods: Performing manual testing (see comment in ticket: https://jira.duraspace.org/browse/FCREPO-1497) I was not able to demonstrate the desired behavior. 
That fact withstanding, I am vaguely concerned about potentially adding "LdpContainerRdfContexts" for all references coming off of incoming-reference nodes... but will check more deeply once I can see the desired behavior of this ticket.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/ReferencesRdfContext.java
escowles: I *think* the key thing here is to convert the logic to a Function or something similar.  But there are several places where there could be nothing, or there could be several references, it might be easier if it was just calling concat() instead of returning a triple iterator.  Is there something like Iterators.transform() that just calls a Callable instance?

ajs6f: Yes, you want to ensconce your actual computation in an abstraction that can remain on the heap-- a function. Are you saying that there are times when your computation, given a `Property`, may not return a triple? May just say, "Nothing to do here-- next `Property`, please!"

escowles: Yes, basically it's iterating over the properties of the nodes that refer to it.  For each of them, it might return nothing if the reference isn't from a proxy, or it might return several triples if that proxy is in several containers.

Now that I put it that way, it sounds like I need to break down the logic into a chain of filtering and transforming steps.

ajs6f: Ok, no problem. Look at what is done in `PropertiesRdfContext`. The function mapped over an iteration of properties there goes from a property to an iterator of triples. (This is just flatmapping.) So if there are multiple triples from a property, no problem. If there is only one, no problem (iterator of one). And if there are none, no problem (empty iterator). This is a handy trick for this kind of situation-- rely on the fact that the "optional" monad is an example of a "collectionish" monad.

ajs6f: Yes, chains of filtering, transforming, and concat'ing steps is the native idiom for this RDF context stuff. Think of the agglomeration of all these "context" classes as representing the transformation of JCR contents into RDF, not the process of _running_ that transformation, just the transformation itself. Hopefully, @acoburn will get around to making it all readable by using Java 8 idioms. :)

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/LdpContainerRdfContext.java
```diff
@@ -203,8 +203,8 @@ public FedoraResourceTripleFunction(final String insertedContainerProperty,
                 return Iterators.transform(values, new Function<Value, Triple>() {
                     @Override
                     public Triple apply(final Value input) {
-                        final RDFNode membershipResource = new ValueConverter(session(), translator())
-                                .convert(input);
+                        final RDFNode membershipResource = new ValueConverter(session(),
```
awoods: I do not see a change here. Please rollback this extraneous update.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/ReferencesRdfContext.java
```diff
@@ -48,21 +64,53 @@ public ReferencesRdfContext(final FedoraResource resource,
...
-        return Iterators.concat(Iterators.transform(properties, property2triple));
-
-    }
+    /* References from LDP indirect containers are generated dynamically by LdpContainerRdfContext, so they won't
+       show up in getReferences()/getWeakReferences().  Instead, we should check referrers to see if they are
+       members of an IndirectContainer and generate the appropriate inbound references. */
+    private Function<Property, Iterator<Value>> potentialProxies = new Function<Property, Iterator<Value>>() {
+        @Override
+        public Iterator<Value> apply(final Property p) {
+            final Set<Value> values = new HashSet<Value>();
```
awoods: Minor: Change to:
`final Set<Value> values = new HashSet<>();`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/ReferencesRdfContext.java
```diff
@@ -48,21 +64,53 @@ public ReferencesRdfContext(final FedoraResource resource,
...
+            try {
+                for ( final PropertyIterator it = p.getParent().getProperties(); it.hasNext(); ) {
```
awoods: This `for` block seems to be duplicating the capabilities found in the `PropertyValueIterator`.
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/mappings/PropertyValueIterator.java

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/8845a22fdeffbec07bb0d0abd35ec9fd3854e788

PR #791, opened 2015-05-11
Allow updating mimeType and filename properties

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-1532

PR #843, opened 2015-07-16
Allow updating mimeType and filename properties

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-1532
Supersedes #791


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/utils/FedoraTypesUtils.java
```diff
@@ -107,12 +107,16 @@
     });
 
     /**
+     * Check if a property is a reference property.
```
awoods: This comment seems like a copy/paste error.
Also, this function `isSuppressedProperty` should be `private`.

awoods: I need to squash these commits anyways. I will go ahead and make the two minor updates mentioned in the comment above.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/aadf58de7a4a8d11372c220987bd5da4169a3fea

PR #844, opened 2015-07-17
Fix for FCREPO-1640

ajs6f:


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/rdf/impl/TypeRdfContext.java
```diff
@@ -58,71 +61,36 @@ public TypeRdfContext(final FedoraResource resource,
 
         //include rdf:type for primaryType, mixins, and their supertypes
         concatRdfTypes();
+        if (resource() instanceof NonRdfSource) {
+            // gather versionability info from the parent
+            if (resource().getNode().getParent().isNodeType("mix:versionable")) {
```
awoods: Instead of relying on `resource().getNode().getParent()`, it may be more maintainable if you use: 
`((NonRdfSource) resource()).getDescription().getNode().isNodeType(...)`
...which raises the question of versionability of NonRdfSource properties. Do we care if the description is versionable if in fact the properties (stored on the NonRdfSource) are not themselves versionable?

ajs6f: _Can_ the description be versionable if the described isn't? Do we offer that?

ajs6f: I'm not sure whether your suggestion is more maintainable or not, because we don't want to conflate description and versioning lifecycle. IOW, I'm looking at the parent _not_ because it is the description, but _specifically_ because it is the parent. If it wasn't the description and the description lived elsewhere (which we can easily imagine supporting) I would still be looking at the parent. I wouldn't go and trace through to the description. That's because it is the parent that controls the versionability, not the description. It just happens that for the moment, in this case, they are the same.

awoods: Fair enough. 
My question becomes, are we loosing versionability of NonRdfSource properties by storing them on the NonRdfSource? In other words, why is the check for "mix:versionable" not being performed on the NonRdfSource?: `resource().getNode().isNodeType(...)`

ajs6f: It's not being checked there because it's not being stored there. As to why, I do not know. You would have to go investigate that code. I assumed it was for a reasonable reason, but after all, this is Fedora.

I don't think we're losing anything. I don't think we currently _can_ version `NonRdfSource` properties independently of the bitstream. Can you demonstrate that we can?

awoods: Previous to: https://jira.duraspace.org/browse/FCREPO-1411
Properties were being stored on the NonRdfSourceDescription... which can be versionable. But since we are now storing the properties on the NonRdfSource itself, we should not care if the description is versionable. 
We should care if the NonRdfSource is versionable, no?

ajs6f: The `NonRdfSourceDescription` is the parent of the `NonRdfSource`. When it is versionable, so is the `NonRdfSource`. In fact, the tests wouldn't pass if this weren't so (see `FedoraVersionsIT.testDatastreamAutoMixinAndRevert()`). I still don't understand what you claim has been lost here.

ajs6f: Please supply a use case for the condition that the bitstream is versionable but its properties are not.

awoods: To make the question as simple as possible, "Why are we checking if the parent if versionable instead of whether the NonRdfSource is versionable"?

ajs6f: See above: "It's not being checked there because it's not being stored there. As to why, I do not know. You would have to go investigate that code. I assumed it was for a reasonable reason, but after all, this is Fedora."

awoods: I am suggesting that the versionable property has in the past been stored on the description because it was the description that held the properties... which were being versioned.
Now, the properties are stored on the NonRdfSource, so it would therefore seem to make sense that we no longer care if the description is versioned, but rather if the NonRdfSource is versioned.

ajs6f: _Both the description and the bitstream were versioned_ and _both still are_. You have yet to explain to me why we would want only one of the two to be versioned if the other is not. The fact that the JCR versioning functionality is engaged through the upper and not the lower node seems to me to be utterly incidental. If you would like to offer a PR in which it moves, I would be happy to merge it.

awoods: Thanks for the clarification about both the description and the NonRdfSource being versioned. Your code makes it look like the triple stating that the NonRdfSource is versioned is derived from whether the NonRdfSource's parent is versioned. Am I misreading this:
https://github.com/fcrepo4/fcrepo4/pull/844/files#diff-e7befd57ec900402b6ffad300b697af1R66

As far as any desire to have one or the other of the description or NonRdfSource versioned and not the other, I have none. I am sorry if I gave an indication otherwise.

ajs6f: It is indeed the parent's status that produces the child's triple. This is because they will always be the same (to my understanding). Do you think an inline comment is appropriate here?

awoods: You are correct, @ajs6f. Versioning appears to work as expected on NonRdfSources and their descriptions.

awoods: An additional inline comment would be bonus.

ajs6f: Will do (will comment, that is). I no longer know what "work as expected" even means any more, so I feel like we're in good shape here.

ajs6f:
Closed in favor of:

https://github.com/fcrepo4/fcrepo4/pull/855

PR #847, opened 2015-07-20
Reintroduce javadoc plugin into build process

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-1634

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1007,6 +1007,21 @@
           </execution>
         </executions>
       </plugin>
+
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -1007,6 +1007,21 @@
           </execution>
         </executions>
       </plugin>
+
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-javadoc-plugin</artifactId>
+        <executions>
+          <execution>
```
ajs6f: At what phase will this execute?

awoods: package

ajs6f: Cool.

PR #850, opened 2015-07-22
remove unnecessary VisibleForTesting annotation

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1647

### DIFF
--- a/fcrepo-connector-file/src/main/java/org/fcrepo/connector/file/FedoraFileSystemConnector.java
+++ b/fcrepo-connector-file/src/main/java/org/fcrepo/connector/file/FedoraFileSystemConnector.java
@@ -37,8 +37,6 @@ import java.util.HashMap;
 import java.io.IOException;
 import java.util.Map;
 
-import com.google.common.annotations.VisibleForTesting;
-
 import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
 import org.infinispan.schematic.document.Document;
 import org.modeshape.connector.filesystem.FileSystemConnector;
@@ -324,12 +322,10 @@ public class FedoraFileSystemConnector extends FileSystemConnector {
 
     /* Overriding so unit test can mock. */
     @Override
-    @VisibleForTesting
     protected File fileFor( final String id ) {
         return super.fileFor(id);
     }
     @Override
-    @VisibleForTesting
     protected DocumentReader readDocument( final Document document ) {
         return super.readDocument(document);
     }


fcrepo-connector-file/src/main/java/org/fcrepo/connector/file/FedoraFileSystemConnector.java
```diff
@@ -324,12 +322,10 @@ protected void touchParent( final String id ) {
 
     /* Overriding so unit test can mock. */
     @Override
-    @VisibleForTesting
```
awoods: These annotations are actually helpful... although there may be another way of accomplishing the same goal.

acoburn: If they are helpful, that's different, but may I ask _how_ are they helpful? My understanding is that this annotation will relax the visibility of private methods during testing. All of our uses of it, though, are as decorators on protected and public functions.

ajs6f: That annotation is purely informative. It's not within the power of any annotation to break visibility.

awoods: Yes, I find them helpful from an informational perspective. `public` and `protected` methods mean increased responsibility. It is useful to know that some methods have been relaxed from `private` for the sake of unit testing. 
Instead of simply removing these annotations, I would suggest replacing them with a prominent comment/javadoc that communicates the same idea.

ajs6f: I'm not sure I even want to remove the annotation. What is the advantage of so doing and adding Javadocs? Is it the Guava dependency that bugs you, @acoburn ?

acoburn: The issue is with https://jira.duraspace.org/browse/FCREPO-1628 and OSGi in general: the modeshape-jcr bundle depends on a certain cassandra driver which depends on an older version of guava (16.0). When bundles are resolved, I get a version dependency conflict (one dep chain needs v.16 and another dep chain needs v.18) and so I can't get the fcrepo-kernel-modeshape bundle to start up properly, and downgrading to guava 16 for fcrepo isn't an option.

ajs6f: Okay, well, if it's blocking OSGi, that's a good reason to remove it in favor of docs. Why on Earth is MODE loading Cassandra drivers?

awoods: ...and can we prevent such loading?

acoburn: Modeshape allows you to optionally store binary files in Cassandra, so it lists this as a dependency. In principle, modeshape's MANIFEST.MF file should specify the dependency as optional, but we are not so lucky. That is, the dependency is unnecessary from the perspective of fcrepo-kernel-modeshape, but from the modeshape-jcr module, it is a hard dependency.

The confusing thing, though, is that the datastax module shows the guava import can be anything in the version range [14-19) https://github.com/datastax/java-driver/blob/2.0/driver-core/pom.xml#L192, but when I actually try to start that bundle in Karaf, it claims to require only version 16.

And no, I can't prevent the cassandra drivers from needing to be loaded. Nor have I had any luck with setting the resolution of that bundle to optional. The problem appears to be that it is a transitive dependency, and so it's rather out of our control unless we start doing things like creating a fat jar, shading, etc, etc (which I'd prefer to avoid).

awoods: OSGi trumps @VisibleForTesting.
...but please leave a comment or the like in the code to communicate the same idea.

acoburn: I'm actually inclined to close this PR, leaving the @VisibleForTesting annotation in place. The issue I'm trying to resolve at present relates to version mismatches 18.0 != 16.0.1, but @VisibleForTesting is available in both versions, so even if we downgraded to guava 16.0.1, we could still use this annotation.

PR #853, opened 2015-07-23
Fix for FCREPO-1648

ajs6f:


awoods:
Minor: update pom.xml test setting

Resolves: https://jira.duraspace.org/browse/FCREPO-1648

PR #856, opened 2015-07-27
remove dependency on fcrepo-mint

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1642

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
@@ -19,6 +19,7 @@ import static com.google.common.base.MoreObjects.toStringHelper;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.Sets.union;
 import static java.util.Collections.singleton;
+import static java.util.UUID.randomUUID;
 import static javax.jcr.observation.Event.PROPERTY_ADDED;
 import static javax.jcr.observation.Event.PROPERTY_CHANGED;
 import static javax.jcr.observation.Event.PROPERTY_REMOVED;
@@ -27,14 +28,12 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
-import java.util.function.Supplier;
 
 import javax.jcr.RepositoryException;
 import javax.jcr.observation.Event;
 
 import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
 import org.fcrepo.kernel.api.utils.EventType;
-import org.fcrepo.mint.UUIDPathMinter;
 
 import com.google.common.base.Function;
 import com.google.common.base.Joiner;
@@ -55,8 +54,6 @@ public class FedoraEvent {
     private Set<Integer> eventTypes = new HashSet<>();
     private Set<String> eventProperties = new HashSet<>();
 
-    private static final Supplier<String> pidMinter = new UUIDPathMinter();
-
     /**
      * Wrap a JCR Event with our FedoraEvent decorators
      *
@@ -64,7 +61,7 @@ public class FedoraEvent {
      */
     public FedoraEvent(final Event e) {
         checkArgument(e != null, "null cannot support a FedoraEvent!");
-        eventID = pidMinter.get();
+        eventID = randomUUID().toString();
         this.e = e;
     }
 
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
@@ -54,7 +54,8 @@ public class FedoraEvent {
 
     private Set<Integer> eventTypes = new HashSet<>();
     private Set<String> eventProperties = new HashSet<>();
-    private UniqueValueSupplier pidMinter = new DefaultPathMinter();
+
+    private final static UniqueValueSupplier pidMinter = new DefaultPathMinter();
 
     /**
      * Wrap a JCR Event with our FedoraEvent decorators
@@ -202,5 +203,5 @@ public class FedoraEvent {
             getDate()).add("Info:", getInfo()).toString();
     }
 
-    private class DefaultPathMinter implements HierarchicalIdentifierSupplier { }
+    private static class DefaultPathMinter implements HierarchicalIdentifierSupplier { }
 }
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
@@ -55,7 +55,7 @@ public class FedoraEvent {
     private Set<Integer> eventTypes = new HashSet<>();
     private Set<String> eventProperties = new HashSet<>();
 
-    private final static UniqueValueSupplier pidMinter = new DefaultPathMinter();
+    private static final UniqueValueSupplier pidMinter = new DefaultPathMinter();
 
     /**
      * Wrap a JCR Event with our FedoraEvent decorators
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
@@ -28,7 +28,6 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
-import javax.inject.Inject;
 import javax.jcr.RepositoryException;
 import javax.jcr.observation.Event;
 
@@ -55,8 +54,6 @@ public class FedoraEvent {
 
     private Set<Integer> eventTypes = new HashSet<>();
     private Set<String> eventProperties = new HashSet<>();
-
-    @Inject
     private UniqueValueSupplier pidMinter = new DefaultPathMinter();
 
     /**
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
@@ -19,7 +19,6 @@ import static com.google.common.base.MoreObjects.toStringHelper;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.Sets.union;
 import static java.util.Collections.singleton;
-import static java.util.UUID.randomUUID;
 import static javax.jcr.observation.Event.PROPERTY_ADDED;
 import static javax.jcr.observation.Event.PROPERTY_CHANGED;
 import static javax.jcr.observation.Event.PROPERTY_REMOVED;
@@ -29,10 +28,12 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
+import javax.inject.Inject;
 import javax.jcr.RepositoryException;
 import javax.jcr.observation.Event;
 
 import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
+import org.fcrepo.kernel.api.services.functions.UniqueValueSupplier;
 import org.fcrepo.kernel.api.utils.EventType;
 
 import com.google.common.base.Function;
@@ -54,6 +55,9 @@ public class FedoraEvent {
     private Set<Integer> eventTypes = new HashSet<>();
     private Set<String> eventProperties = new HashSet<>();
 
+    @Inject
+    private UniqueValueSupplier pidMinter = new DefaultPathMinter();
+
     /**
      * Wrap a JCR Event with our FedoraEvent decorators
      *
@@ -61,7 +65,7 @@ public class FedoraEvent {
      */
     public FedoraEvent(final Event e) {
         checkArgument(e != null, "null cannot support a FedoraEvent!");
-        eventID = randomUUID().toString();
+        eventID = pidMinter.get();
         this.e = e;
     }
 
@@ -199,4 +203,6 @@ public class FedoraEvent {
             Joiner.on(',').join(eventProperties)).add("Path:", getPath()).add("Date: ",
             getDate()).add("Info:", getInfo()).toString();
     }
+
+    private class DefaultPathMinter implements UniqueValueSupplier { }
 }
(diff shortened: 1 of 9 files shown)


escowles:
This replicates the current behavior, but there's also an outstanding ticket for making the event ids configurable: https://jira.duraspace.org/browse/FCREPO-1568

Can we inject a Supplier<String>, allowing us to wire up whatever kind of ID minter was desired?  This would also remove the dependency on fcrepo-mint.

ajs6f:
@escowles How do you want to use a `Supplier`? To supply a minter or to supply ids? Keep in mind that the basic `Supplier` type doesn't guarantee that you get the same thing each time, so supplying ids makes sense, but not supplying minters.

escowles:
@ajs6f I think it's stripping out the angle brackets: I meant to say Supplier&lt;String&gt; -- so a supplier of ids.  I think it makes sense to use the same approach we're using for the main repository ids in https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/AbstractResource.java#L101

ajs6f:
Holy crap! That's no good. `Supplier` doesn't guarantee that it _won't_ give you the same thing over and over again. I hope I didn't do that in `AbstractResource`, but I think I might have absent-mindedly done that. If we want to use `Supplier` to produce a stream of unique things, we have to subtype it.

escowles:
So maybe we should add a PidMinter interface to fcrepo-kernel-api, and then inject fcrepo-mint instances in the fcrepo-http-commons repository config?

ajs6f:
I don't think we should be as specific as `PidMinter` because we don't need to be. `UniqueValueSupplier` would do it. Then we could toast fcrepo-mint, right? We could offer a sidecar extension with the HTTP minter in it.

escowles:
`UniqueValueSupplier<String>` sounds good to me.  Do you mean remove fcrepo-mint altogether?  Do the UUID impls. get moved to `fcrepo-kernel-modeshape` or somewhere else?

ajs6f:
Hey, man, this is java 8! The UUID impl becomes default method bodies on the interface itself.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
```diff
@@ -64,7 +61,7 @@
      */
     public FedoraEvent(final Event e) {
         checkArgument(e != null, "null cannot support a FedoraEvent!");
-        eventID = pidMinter.get();
+        eventID = randomUUID().toString();
```
awoods: Instead of hard-coding `randomUUID()` here, is there a reason why an `UniqueValueSupplier<String>`, as suggested in the code review comments, was not injected?

As a note, `randomUUID().toString()` is not equivalent to the previous `UUIDPathMinter.get()`. `UUIDPathMinter` provides a pairtree path with a UUID leaf element, whereas `randomUUID` simply provides a UUID.
The reason a full pairtree path is important is because the `InternalAuditor` uses the `FedoraEvent#eventID` as the identifier name at which an audit record resource is created:
https://github.com/fcrepo4-labs/fcrepo-audit/blob/master/src/main/java/org/fcrepo/audit/InternalAuditor.java#L211

The pairtree path provided by the `UUIDPathMinter` gets around the performance limitation of too many children of a given resource.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
```diff
@@ -64,7 +61,7 @@
      */
     public FedoraEvent(final Event e) {
         checkArgument(e != null, "null cannot support a FedoraEvent!");
```
ajs6f: This should be Java 8 `Objects::requireNonNull`.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
```diff
@@ -27,14 +28,12 @@
...
 
 import javax.jcr.RepositoryException;
 import javax.jcr.observation.Event;
 
 import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
 import org.fcrepo.kernel.api.utils.EventType;
-import org.fcrepo.mint.UUIDPathMinter;
 
 import com.google.common.base.Function;
 import com.google.common.base.Joiner;
```
ajs6f: These Guava types should be replaced with java 8 types `Function` and `StringJoiner` (or potentially `String::join` or `Collectors::joining`).

acoburn: I completely agree. This PR, however, was focused entirely on removing the dependency on fcrepo-mint, so doing all the java8 refactoring seemed out of scope.

awoods: Agreed, @acoburn... but thanks for the tips, @ajs6f!

acoburn:
Updated PR with an injectable `UniqueValueSupplier` that subclasses `Supplier<String>`. Now neither of the `fcrepo-kernel-*` modules depend on `fcrepo-mint`.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/functions/UniqueValueSupplier.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.stream.IntStream;
+import java.util.StringJoiner;
+
+/**
+ * Unique value minter that creates hierarchical IDs from a UUID
+ *
+ * @author awoods
+ * @author acoburn
+ */
+public interface UniqueValueSupplier extends Supplier<String> {
```
ajs6f: This is way more specific than the name `UniqueValueSupplier` would imply. How about `HierarchicalIdentifierSupplier` < `UniqueValueSupplier`, with `UniqueValueSupplier` simply guaranteeing the uniqueness of results (no default impl) and `HierarchicalIdentifierSupplier` bringing the "minter that creates hierarchical IDs from a UUID" semantic with the default impl?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/functions/UniqueValueSupplier.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Mint a unique identifier as a UUID
+     *
+     * @return uuid
+     */
+    default public String get() {
+        final int defaultLength = 2;
+        final int defaultCount = 4;
```
ajs6f: Is there some reason these are local variables?

acoburn: can you have private fields on an interface?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/functions/UniqueValueSupplier.java
acoburn: That works for me. I can add the additional interface.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/functions/UniqueValueSupplier.java
ajs6f: Why `private`? These are defaults the effects of which are entirely obvious-- why not `(public) static`?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/functions/UniqueValueSupplier.java
ajs6f: Right on.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/functions/UniqueValueSupplier.java
acoburn: That makes sense -- I had been trying not to leak implementation details, but in this case, that's probably a good thing to advertise.

ajs6f: Let it all hang out. There is no place for shame or modesty in Fedora.

acoburn:
added an interface for `HierarchicalIdentifierSupplier` < `UniqueValueSupplier` and moved the default impl to `HierarchicalIdentifierSupplier`, as @ajs6f suggested

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/functions/UniqueValueSupplier.java
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.kernel.api.services.functions;
+
+import java.util.function.Supplier;
+
+/**
+ * Unique value minter that creates unique IDs
```
ajs6f: This is now the wrong comment. This type just guarantees the uniqueness of its supplied values. It has nothing to do with minting IDs.

acoburn: The comment has been updated.

awoods:
This looks good. @escowles, you will note that the issue of injecting a `pidMinter` implementation into `FedoraEvent`s is still outstanding, for the same reasons mentioned in the ticket:
https://jira.duraspace.org/browse/FCREPO-1568

If @escowles or @ajs6f are comfortable with this current PR as well, I will go ahead and squash/merge.

escowles:
:+1:  -- looks like a good improvement, and we can figure out the injection separately.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
```diff
@@ -54,8 +54,7 @@
 
     private Set<Integer> eventTypes = new HashSet<>();
     private Set<String> eventProperties = new HashSet<>();
-
-    private static final Supplier<String> pidMinter = new UUIDPathMinter();
+    private UniqueValueSupplier pidMinter = new DefaultPathMinter();
```
acoburn: shouldn't this be `static final`?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/JcrRdfTools.java
```diff
@@ -108,7 +108,7 @@
 
     private static final Model m = createDefaultModel();
 
-    private static final Supplier<String> pidMinter =  new UUIDPathMinter();
+    private final UniqueValueSupplier pidMinter =  new DefaultPathMinter();
```
acoburn: And shouldn't this be `static`?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
awoods: yes

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/JcrRdfTools.java
awoods: yes

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/functions/UniqueValueSupplier.java
```diff
@@ -0,0 +1,27 @@
...
+package org.fcrepo.kernel.api.services.functions;
+
+import java.util.function.Supplier;
+
+/**
+ * A {@link java.util.function.Supplier} interface that guarantees the uniqueness of its provided values.
+ *
+ * @author acoburn
+ */
+public interface UniqueValueSupplier extends Supplier<String> {
```
awoods: Since this PR keeps staying open, I keep looking at it...
Why is this new class `UniqueValueSupplier` in the `org.fcrepo.kernel.api.services.functions` package?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/functions/HierarchicalIdentifierSupplier.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.stream.IntStream;
+import java.util.StringJoiner;
+
+/**
+ * Unique value minter that creates hierarchical IDs from a UUID
+ *
+ * @author awoods
+ * @author acoburn
+ */
+public interface HierarchicalIdentifierSupplier extends UniqueValueSupplier {
```
awoods: Why is this new class `HierarchicalIdentifierSupplier` in the `org.fcrepo.kernel.api.services.functions` package?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/functions/UniqueValueSupplier.java
acoburn: I put them in `o.f.k.a.services.functions` because they are different types of `Supplier` interfaces, which, itself provides a functional interface. That's the only reason.... If you'd prefer they go in `o.f.k.a.identifiers`, that's fine by me.

awoods: Tying it back to `Supplier` is rationale enough for me.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/FedoraEvent.java
awoods: If you push one more commit with the addition of `static` we can close this PR.

acoburn: Commit pushed (actually two). The nested classes had to be made static in order to use them on a static field.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/f7daf307912d78cc57098a74c2b04c2daf5d54e6

PR #859, opened 2015-07-27
remove redundant javadoc commands

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1657

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ before_install:
 
 script:
  - mvn install -B -V
- - mvn javadoc:jar
- - mvn javadoc:test-aggregate
 
 notifications:
   irc: "irc.freenode.org#fcrepo"


ajs6f:
Is this going to prevent Travis from running Javadocs?

acoburn:
@ajs6f my understanding is that travis will continue to run javadocs. If you look at the [travis output](https://s3.amazonaws.com/archive.travis-ci.org/jobs/72877788/log.txt), you will see that it first runs:

    mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V

(which just downloads dependencies). After that it runs:

    mvn install -B -V

And if you look in the `fcrepo-kernel-api` section, you will see some javadoc warnings, suggesting to me that it is most certainly running `javadoc:jar`

ajs6f:
Oky, but is `javadoc:jar` enough, or do we need `javadoc:test-aggregate` too?

acoburn:
That I don't actually know. I can see where javadoc:jar is being executed: https://github.com/fcrepo4/fcrepo4/blob/master/pom.xml#L1032-L1044

But in the report section: https://github.com/fcrepo4/fcrepo4/blob/master/pom.xml#L1054-L1060 I don't see anything about test-aggregate.

If that is, indeed, the case, that command (`mvn javadoc:test-aggregate`) should not be dropped from the travis.yml file

ajs6f:
Hm. Anyone know who added it or why? @awoods ?

awoods:
`javadoc:test-aggregate` gets run by either the `release:prepare` or site generation steps of the release process. We want to know if there are issues with `javadoc:test-aggregate` prior to pushing the release button.
The `javadoc:test-jar` goal performs the same function:
https://github.com/fcrepo4/fcrepo4/blob/master/pom.xml#L1040

Therefore, it should be safe to remove `javadoc:test-aggregate` from travis.

PR #863, opened 2015-07-30
Simple cleanup using Java 8 types and syntax for RDF generating code

ajs6f:


escowles:
This looks good to me (not sure why one of the Travis builds failed, it builds fine for me), and I've added to it with some more comments and a couple minor tweaks: https://github.com/fcrepo4/fcrepo4/pull/865

ajs6f:
I demand higher reliability from the CI product for which I pay nothing!

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/utils/iterators/RdfStream.java
```diff
@@ -359,6 +359,11 @@ public Triple apply(final T prototriple) {
         };
     }
 
+    protected static <From, To> Iterator<To> flatMap(final Iterator<From> i,
```
acoburn: Rather than reimplementing a `flatmap` here (next we'll be implementing `map` and `filter`), why not user `StreamSupport` and the `Iterable` interface to convert the `Iterator` into a `Stream`?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/LdpContainerRdfContext.java
```diff
@@ -76,7 +74,10 @@ public LdpContainerRdfContext(final FedoraResource resource,
             throws RepositoryException {
         super(resource, idTranslator);
         final Iterator<Property> memberReferences = resource.getNode().getReferences(LDP_MEMBER_RESOURCE);
-        final Iterator<Property> properties = Iterators.filter(memberReferences, isContainer );
+        final Iterator<Property> properties = Iterators.filter(memberReferences, UncheckedPredicate.uncheck(p -> {
```
acoburn: Shouldn't this just be `uncheck`, since that function has already been statically imported?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/utils/iterators/RdfStream.java
ajs6f: `Iterators::transform` and `Iterators::filter` already exist.

1. `Iterable` implies that we can produce multiple iterators by calling `Iterable::iterator` several times. We don't want to make that claim.
2.  `Stream` is considerably heavier than `Iterator`. I'm very much in favor of `Stream` becoming more common in our code, but when we need only a very tight set of operations (and the majority of uses of `flatMap` are either alone or with a `filter` or two) it's not clear to me that it's worth the overhead in code or execution.

I'm not saying we shouldn't ever make `Stream`s at a low level, but I am saying that we shouldn't always make `Stream`s at a low level. Fundamentally, `Stream` represents a computation, but `Iterator` represents a flow of elements, and that's really what we have here.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/LdpContainerRdfContext.java
ajs6f: No, that's `UncheckedFunction::uncheck`. Try statically importing it and you will see that the compiler gets confused and can't do the type inference.

acoburn: Oh, of course. My mistake.

acoburn:
This looks good to me. @awoods any comments?

awoods:
See: https://jira.duraspace.org/browse/FCREPO-1663

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/BlankNodeRdfContext.java
```diff
@@ -1,122 +0,0 @@
...
-import static org.fcrepo.kernel.modeshape.utils.FedoraTypesUtils.isBlankNode;
-
-/**
- * Embed all blank nodes in the RDF stream
- *
- * @author cabeer
- * @author ajs6f
- * @since 10/9/14
- */
-public class BlankNodeRdfContext extends NodeRdfContext {
```
awoods: The history for this class has been lost. Instead, you should use:
`git mv`

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/utils/iterators/RdfStream.java
```diff
@@ -57,7 +57,7 @@
 
     /**
      * Constructor that begins the stream with proffered triples.
-     *
```
awoods: Mixing non-functional format changes with substantive commits makes code review more painful and code archaeology more difficult.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/BlankNodeRdfContext.java
acoburn: I do not believe that `git mv` will preserve the history of this class. Git uses its own heuristic for determining whether a file has been deleted and added somewhere else (e.g. a simple "move") or whether a newly added file is a truly different thing (e.g. a delete and an unrelated addition). If you look at the implementation of `SkolemNodeRdfContext`, it shares very few similar lines with `BlankNodeRdfContext`. So, even with `git mv`, that history is not necessarily going to be saved. The only sure way to keep that history is to rename the file but keep the implementation basically intact (as much as possible) in one commit. In a subsequent commit (that does not get squashed) the implementation can be changed to reflect the Java8 idioms as expressed here.

Personally, I'm not convinced that doing this simply to preserve the history (first a rename with minimal changes to the file followed by a subsequent, more substantive reimplementation) is worth the effort.

awoods: I would be very surprised (although happy to be proven wrong) to see that `git mv` did not save the history in this case. Have you tested your theory, @acoburn?

acoburn: I have encountered similar issues at other times, when a file is renamed _and_ changed substantially. The issue is this: git tracks content, not files.

awoods: Thanks for the link, @acoburn.
http://article.gmane.org/gmane.comp.version-control.git/217

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/SkolemNodeRdfContext.java
```diff
@@ -0,0 +1,79 @@
...
+        final Iterator<Property> properties = resource().getNode().getProperties();
+
+        final Iterator<Property> references = Iterators.filter(properties,
+                uncheck(p -> REFERENCE_TYPES.contains(p.getType())));
+
+        final Iterator<Node> nodes = Iterators.transform(new PropertyValueIterator(references),
+                UncheckedFunction.<Value, Node>uncheck(v -> v.getType() == PATH ?
+                        session().getNode(v.getString()) : session().getNodeByIdentifier(v.getString()))::apply);
+
+        return Iterators.filter(nodes, isBlankNode::test);
```
awoods: In a similar fashion to this class itself, the `isBlankNode` predicate should be renamed to `isSkolemNode` in `FedoraTypesUtils`.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/89d2826552799f0252a277621e2808916f5deadb

PR #864, opened 2015-07-31
Fcrepo 1640

ajs6f:

PR #869, opened 2015-08-05
Update HTML UI logic to support single-subject Binary resources

awoods:
Related to: https://jira.duraspace.org/browse/FCREPO-1640

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -30,6 +30,7 @@ import static javax.ws.rs.core.Response.Status.PARTIAL_CONTENT;
 import static javax.ws.rs.core.Response.Status.REQUESTED_RANGE_NOT_SATISFIABLE;
 import static org.apache.commons.lang.StringUtils.isBlank;
 import static org.apache.jena.riot.RDFLanguages.contentTypeToLang;
+
 import static org.fcrepo.kernel.api.FedoraJcrTypes.LDP_BASIC_CONTAINER;
 import static org.fcrepo.kernel.api.FedoraJcrTypes.LDP_DIRECT_CONTAINER;
 import static org.fcrepo.kernel.api.FedoraJcrTypes.LDP_INDIRECT_CONTAINER;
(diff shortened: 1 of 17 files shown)


acoburn:
I can't claim any significant experience with velocity templates, but I can confirm that the extra triples on binary resources aren't displayed in the HTML view after this patch is applied. However, I still see a "Other Resources" label at the bottom of the page, which is confusing. When I add a binary through the web interface, it creates a new resource with a `fedora:primaryType` equal to `nt:resource`, so the template file used is, therefore, `nt-resource.vsl`, but it appears that this PR is about updating the `nt-file.vsl` template. Shouldn't `nt-resource.vsl` also be updated?

Also, when I add a second version, as described in the ticket (update via `curl -X POST ...`), the template for the old version is, seemingly, all wrong. At least, it doesn't show any values in the main section of the page -- only in the collapsible "Other Resources" section. That node has a `fedora:primaryType = nt:frozenNode`, but there are is no `nt-frozenNode.vsl` template file. Does it then default to `node.vsl`? Either way (in the `fcr:versions/node/fcr:metadata` view), I see triples whose subject is not the resource being described. That is, there are no triples for `</node/fcr:versions/abcdefg>` (other than `fedora:hasFixityService`), but there are plenty of triples for `</node/fcr:versions/abcdefg/fcr:metadata>`

I think this needs more work.

awoods:
@acoburn, the confusion seems to be coming from the fact that you tested [PR-869](https://github.com/fcrepo4/fcrepo4/pull/869) in isolation. PR-869 is not against `master` but instead against the main PR, [PR-864](https://github.com/fcrepo4/fcrepo4/pull/864).
PR-864 makes Fedora single-subject on binaries, whereas PR-869 makes the UI work on top of those underlying updates.
PR-864 has been reviewed, but the UI changes on top has no reviewer... hence the request for your help.
Sorry for the confusion.

acoburn:
@awoods unless I am mistaken, I did test #869 against #864. As for the single-subject issue on binary objects, that is, in fact fixed with this PR. My point, though, may be out of scope for (but not unrelated to) this ticket, namely that the same issue applies to the fcr:version interface, and that I still see a *Other Resources* header at the bottom of the page (with nothing below it) -- not a bug, just a confusing UI element.

awoods:
@ajs6f, I hope you are well and enjoying late summer. When you have a moment, could you please rebase this PR on master?

ajs6f:
Did you not see the question I just asked you in IRC?

awoods:
This PR is now being tracked on https://github.com/fcrepo4/fcrepo4/pull/864

PR #870, opened 2015-08-05
HTML UI: Ensure resource creation URIs have no double '/' chars

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-1677

### DIFF
--- a/fcrepo-http-api/src/main/resources/views/common.js
+++ b/fcrepo-http-api/src/main/resources/views/common.js
@@ -2,18 +2,21 @@ $(document).ready(function() {
   $("#new_id").attr("placeholder", "(auto-generated identifier)");
 });
 
+function removeTrailingSlash(origURI) {
+    var trimmed = origURI;
+    while (trimmed.lastIndexOf('/') == trimmed.length - 1) {
+        trimmed = trimmed.substr(0, trimmed.length - 1);
+    }
+    return trimmed;
+}
+
 function addChild()
 {
     var id = $("#new_id").val().trim();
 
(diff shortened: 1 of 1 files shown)


fcrepo-http-api/src/main/resources/views/common.js
```diff
@@ -8,7 +8,12 @@ function addChild()
 
     var mixin = $("#new_mixin").val();
 
-    var newURI = $('#main').attr('resource') + (!id ? "" : "/" + id);
+    var base = $('#main').attr('resource');
+
+    // Remove trailing '/' if it exists
+    var cleanedBase = base.lastIndexOf('/') == base.length - 1 ? base.substr(0, base.length - 1) : base;
```
acoburn: Would it ever be possible that the `resource` attribute would end in two "/" characters? This JS code will remove a single trailing slash, but not more than one.

awoods: I do not believe the `resource` can currently find itself in a double '/' situation... but I can add a `while-ends-with-slash` function to prevent such an issue in the future.

acoburn: If you're confident that resource won't end in multiple slashes, we can leave it as-is; otherwise, adding the additional function makes sense to me.

awoods: Updated in new commit.

acoburn:
Resolved via 51545e1a83d1759be21397759ecdfe26ce7fcf0e

PR #871, opened 2015-08-06
Catch empty body PUT request error

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-1667

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -65,6 +65,8 @@ import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Request;
 import javax.ws.rs.core.Response;
 
+import com.fasterxml.jackson.core.JsonParseException;
+import org.apache.jena.atlas.RuntimeIOException;
 import org.fcrepo.http.commons.api.rdf.HttpTripleUtil;
 import org.fcrepo.http.commons.domain.MultiPrefer;
 import org.fcrepo.http.commons.domain.PreferTag;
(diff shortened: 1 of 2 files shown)


acoburn:
@awoods can you rebase this PR against `fcrepo4/master`?

awoods:
@acoburn, rebased.

ajs6f:
Travis appears to have failed. Also, are we confident that all places that Jena might throw a `RuntimeIOException` are places from which we want to map as done here?

awoods:
Alternatives to mapping the Jena `RuntimeIOException` are to:
- inspect the `cause` of the `RuntimeIOException` looking for `JsonParseException`, or
- specifically wrap this statement: https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java#L581

However, the Jena exception seems likely to be the result of a `BAD_REQUEST` scenario.
Regarding Travis, the branch builds locally. Probably a sporadic test error.

ajs6f:
I'd rather take wrap the statement _and_ check for the cause. `RuntimeIOException` is really generic. We could easily catch it and throw something more meaningful.

awoods:
Sure. The downside, of course, is that there may be other places in the code that would benefit from a Jena `RuntimeIOExceptionMapper`.

ajs6f:
Can we be sure what to produce (HTTP) for the generic case?

awoods:
Can you clarify that question?

ajs6f:
It's the same

awoods:
I would answer that question with, "yes", since this particular scenario is the first time the `RuntimeIOException` has surfaced. 

However, if you would feel more comfortable with: 
- wrapping the specific location in `ContentExposingResource`, followed by 
- inspecting the `RuntimeIOException` cause, followed by 
- throwing a Fedora-specific exception (`MalformedRdfException` ??) that is caught by its own exception-mapper, I would be happy to add a new commit.

ajs6f:
I honestly would. The semantics of `RuntimeIOException` are just what you'd think (just an unchecked `IOException`). I think rethrowing something (sure, `MalformedRdfException`, when that's appropriate) inside our code is cleaner. Especially if it turns out that there are several ways for the problem to happen (besides a purely empty body) and we want to distinguish them.

awoods:
New commit responds to comments.
@ajs6f, let me know if this looks good to you. If so, I will squash the two commits.

ajs6f:
Yeah, I like that much better. Thanks!

PR #873, opened 2015-08-07
Replace guava's Function and Predicate with Java8 idioms

acoburn:
https://jira.duraspace.org/browse/FCREPO-1681


awoods:
Do you feel like it is out of scope to also remove:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/utils/EventType.java#L18

acoburn:
I went back and forth on that one and decided to leave it in. My rationale was that to do the same thing without `forMap`, the code will become somewhat longer:

    final EventType type = translation.get(i);
    if (type == null) {
        throw new IllegalArgumentException("...");
    }
    return type;

And I felt that the current use of `forMap` made the code more clear and concise. If our goal was to _remove_ our dependency on guava, that would be one thing, but that wasn't my goal with this PR. (I really tried to confine the changes to the `Function` and `Predicate` classes, and this is from the `Functions` class (note the 's')).

PR #874, opened 2015-08-08
Create parent POM module

acoburn:
https://jira.duraspace.org/browse/FCREPO-1684

This is a first stab at cleaning up the maven configuration. The idea is that this will eliminate a lot of the redundant maven configuration across modules that may not inherit directly from the `fcrepo` maven POM. As a guiding principle, I created the `fcrepo-parent` to be the sort of POM artifact that I wish existed when configuring `fcrepo-camel` and `fcrepo-camel-toolbox`. I have verified separately that `fcrepo-camel` builds correctly when using `fcrepo-parent` as a parent POM (i.e. after eliminating most of the existing plugin configuration in `fcrepo-camel/pom.xml`).

 I have some open questions related to this PR:

  * I removed a number of dependencies that were not used (directly or transitively), including: javax.mail, enunciate(?), commons-compress, jaxb, javassist. I also removed joda-time as a direct dependency: it is not used directly and is only imported transitively via modeshape-jcr.
  * I left the `enunciate` and `jmeter` plugins in the fcrepo pom file, though I'm not sure they are being used. I also didn't touch the `lifecycle-mapping` plugin: I don't use eclipse, but others do, so I assume it is relevant for them.
  * Plugin configuration, for the most part, was moved to the fcrepo-parent POM, but anything that didn't seem truly applicable to a wide range of projects stayed in the fcrepo POM. I left the enforcer plugin configuration in the fcrepo POM, since those rules may not apply to other projects.
  * I am not particularly familiar with the `mvn site` and the release/deployment-related phases, so I would appreciate it if @awoods could verify that this configuration actually does what I intend it to do: that is, to make deployment, site generation and releases easier and more consistent across projects.


fcrepo-kernel-modeshape/pom.xml
```diff
@@ -8,7 +8,7 @@
   </parent>
 
   <artifactId>fcrepo-kernel-modeshape</artifactId>
-  <name>Fedora Repository Kernel Implementation based on Modeshape</name>
+  <name>Fedora Repository Kernel Implementation (Modeshape)</name>
   <description>The Fedora Commons repository kernel: Provides the basic abstractions at the heart of the Fedora
```
ajs6f: Shouldn't this description say something about Modeshape?

acoburn: Yes, it should.

ajs6f:
I think the `jmeter` stuff is toast-worthy, at this point. The `enunciate` plugin was pretty cool, although we never got it working really well. It might be worth taking another stab at that. The `lifecycle-mapping` stuff shouldn't be there. People who use Eclipse (like me) should do that stuff in their own preferences and not in the `pom.xml`s.

awoods:
Please rebase on master.

acoburn:
@awoods this branch is up to date with regard to master

pom.xml
```diff
@@ -162,47 +143,34 @@
         <version>${guava.version}</version>
       </dependency>
       <dependency>
-        <groupId>org.apache.abdera</groupId>
-        <artifactId>abdera-parser</artifactId>
```
awoods: Who did you determine that `abdera-parser` is not needed?

pom.xml
```diff
@@ -474,12 +428,6 @@
         </exclusions>
       </dependency>
       <dependency>
-        <groupId>org.javassist</groupId>
```
awoods: Who did you determine that `javassist` is not needed? I vaguely recall seeing this in relation to the Sonar reporting... but could be wrong.

pom.xml
ajs6f: Abdera had something to do with RSS, right?

pom.xml
```diff
@@ -680,12 +478,6 @@
           </configuration>
           <executions>
             <execution>
-              <id>integration-test</id>
```
awoods: Why was the `integration-test` goal removed?
https://maven.apache.org/surefire/maven-failsafe-plugin/usage.html

fcrepo-parent/pom.xml
```diff
@@ -0,0 +1,566 @@
...
+          <configuration>
+            <skipDeploy>true</skipDeploy>
+          </configuration>
+          <dependencies>
+            <dependency>
+              <!-- allows markdown syntax for site generation. To use it
+                place files below src/site/markdown/[filename].md -->
+              <groupId>org.apache.maven.doxia</groupId>
+              <artifactId>doxia-module-markdown</artifactId>
+              <version>1.6</version>
```
awoods: Please add version property for: doxia-module-markdown

pom.xml
```diff
@@ -900,37 +625,26 @@
...
       </plugin>
+
       <plugin>
-        <artifactId>maven-compiler-plugin</artifactId>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-javadoc-plugin</artifactId>
       </plugin>
+
       <plugin>
         <artifactId>maven-enforcer-plugin</artifactId>
```
awoods: The `maven-enforcer-plugin` seems valuable across all Fedora-related projects. What was the rationale for not moving it up to the parent?

pom.xml
```diff
@@ -982,309 +696,17 @@
...
-          <outputName>surefire-report</outputName>
-          <aggregate>true</aggregate>
-          <reportsDirectories>
-            <reportsDirectory>${project.build.directory}/surefire-reports/</reportsDirectory>
-            <reportsDirectory>${project.build.directory}/failsafe-reports/</reportsDirectory>
-          </reportsDirectories>
-        </configuration>
-      </plugin>
-      <plugin>
-        <artifactId>maven-changelog-plugin</artifactId>
```
awoods: Please keep `maven-changelog-plugin` (but remove `displayFileDetailUrl`)
- http://maven.apache.org/plugins/maven-changelog-plugin/examples/modifying-scm-links.html
- http://docs.fcrepo.org/site/4.2.0/changelog.html

pom.xml
```diff
@@ -982,309 +696,17 @@
...
-      <releases>
-        <enabled>true</enabled>
-      </releases>
-      <snapshots>
-        <enabled>false</enabled>
-      </snapshots>
-    </pluginRepository>
-  </pluginRepositories>
 
   <scm>
```
awoods: Since one of the issues I have seen in external modules is a missing `scm` section, it would be ideal if we could move this up to the `parent` using property placeholders.
https://jira.duraspace.org/browse/FCREPO-1558

fcrepo-parent/pom.xml
```diff
@@ -0,0 +1,566 @@
...
+        <property>
+          <name>performRelease</name>
+          <value>true</value>
+        </property>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-javadoc-plugin</artifactId>
```
awoods: It is not clear that this configuration of `maven-javadoc-plugin` will pick up the necessary config:
https://github.com/fcrepo4/fcrepo4/pull/874/files#diff-600376dffeb79835ede4a0b285078036L1222

fcrepo-parent/pom.xml
```diff
@@ -0,0 +1,566 @@
...
+              <report>plugin-updates-report</report>
+              <report>property-updates-report</report>
+            </reports>
+          </reportSet>
+        </reportSets>
+      </plugin>
+    </plugins>
+  </reporting>
+
+  <repositories>
```
awoods: We may be able to drop this entire `repositories` section. I just deleted my local `.m2` directory, commented out this entire section, and ran a successful `mvn clean install`.

pom.xml
awoods: Yes. https://abdera.apache.org/
I am not concerned about it going away, but more interested in the technique for determining it is not needed.

acoburn: My basic technique was this: what packages does the artifact export? And are they used in the code anywhere?

Sometimes, there is a transitive dependency conflict and it is necessary to explicitly load a particular version so that the class loader can resolve dependencies properly, but this did not appear to be the case with abdera.

fcrepo-parent/pom.xml
awoods: ...although, please add:
```
  <repositories>
    <repository>
      <id>snapshots-repo</id>
      <url>http://oss.sonatype.org/content/repositories/snapshots</url>
      <releases>
        <enabled>false</enabled>
      </releases>
      <snapshots>
        <enabled>true</enabled>
      </snapshots>
    </repository>
  </repositories>
```

pom.xml
acoburn: Sonar doesn't seem to care about the lack of `javassist`, so I'm going to leave this out for now.

awoods: FYI, sonar for `fcrepo4` has been disabled until this ticket is completed:
https://jira.duraspace.org/browse/FCREPO-1670

pom.xml
acoburn: I can move the plugin version and artifactId into the parent, but the specific executions and related configuration I left in the main fcrepo4 pom (moving the config up a level only produced additional warning output for me)

fcrepo-parent/pom.xml
acoburn: Seems like it does for me: `mvn help:effective-pom -P release-sign-artifacts`

fcrepo-parent/pom.xml
acoburn: I actually think that the sonatype parent defines this: https://github.com/sonatype/oss-parents/blob/master/oss-parent/pom.xml#L32-L44

pom.xml
acoburn: I am going to leave the enforcer plugin defn in the fcrepo4/pom.xml file because moving any part of it to the parent only results in error messages for me.

pom.xml
```diff
@@ -780,86 +516,77 @@
...
+      <plugin>
+        <artifactId>maven-failsafe-plugin</artifactId>
+        <configuration>
+          <argLine>${jacoco.agent.it.arg}</argLine>
+        </configuration>
+        <executions>
+          <execution>
+            <id>default-integration-test</id>
+            <goals>
+              <goal>integration-test</goal>
```
awoods: The recommendation from the Maven docs seems to suggest our original configuration of the plugin, i.e. having both `integration-test` and `verify`.
https://maven.apache.org/surefire/maven-failsafe-plugin/usage.html

awoods:
The updates look good. In doing some testing I have run into issues with the following:
```bash
cd fcrepo-message-consumer
mvn javadoc:test-aggregate
```
```bash
cd fcrepo-module-auth-xacml
mvn javadoc:jar
```
```bash
cd fcrepo4-swordserver
mvn javadoc:test-aggregate
```

acoburn:
@awoods: I have added some additional PRs for the projects listed, related to this change. See the Jira ticket for details.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/99397f6e282ae524832c13749456c6ec1f3a3d9d

PR #877, opened 2015-08-09
upgrade to commons.lang3

acoburn:
https://jira.duraspace.org/browse/FCREPO-1688


acoburn:
N.B. this will most certainly cause conflicts with #874, so depending on which one is merged first, the other will need to be rebased.

PR #878, opened 2015-08-09
unit tests for fcrepo-connector-file

acoburn:
https://jira.duraspace.org/browse/FCREPO-1689

### DIFF
--- a/fcrepo-connector-file/src/test/java/org/fcrepo/connector/file/ExternalJsonSidecarExtraPropertyStoreTest.java
+++ b/fcrepo-connector-file/src/test/java/org/fcrepo/connector/file/ExternalJsonSidecarExtraPropertyStoreTest.java
@@ -15,17 +15,37 @@
  */
 package org.fcrepo.connector.file;
 
+import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
+import org.infinispan.schematic.document.Document;
+import org.infinispan.schematic.document.EditableDocument;
+import org.infinispan.schematic.document.Json;
+import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
 import org.mockito.runners.MockitoJUnitRunner;
 import org.modeshape.jcr.cache.document.DocumentTranslator;
+import org.modeshape.jcr.value.Name;
+import org.modeshape.jcr.value.Property;
+import org.modeshape.jcr.value.basic.BasicName;
+import org.modeshape.jcr.value.basic.BasicSingleValueProperty;
 
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
 
 import static java.nio.file.Files.createTempDirectory;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyObject;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 /**
@@ -42,11 +62,9 @@ public class ExternalJsonSidecarExtraPropertyStoreTest {
 
     private static final String FEDERATION_ROOT = "/federation-root";
     private static final String FILE_PATH = "/federation-root/file";
-    @Test
-    public void testSidecarFile() throws IOException {
-        final File tmp = createTempDirectory("filesystem-federation").toFile();
-        tmp.deleteOnExit();
 
+    @Before
+    public void setUp() {
         when(mockConnector.fileFor("/")).thenReturn(new File(FEDERATION_ROOT));
         when(mockConnector.isContentNode("/")).thenReturn(false);
         when(mockConnector.isRoot("/")).thenReturn(true);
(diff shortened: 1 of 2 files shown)


fcrepo-connector-file/src/test/java/org/fcrepo/connector/file/ExternalJsonSidecarExtraPropertyStoreTest.java
```diff
@@ -54,12 +72,269 @@ public void testSidecarFile() throws IOException {
...
+        final Name key3 = new BasicName("info", "three");
+        final Name key4 = new BasicName("info", "four");
+        final Property prop1 = new BasicSingleValueProperty(langProp, "eins");
+        final Property prop2 = new BasicSingleValueProperty(langProp, "zwei");
+        final Property prop3 = new BasicSingleValueProperty(langProp, "drei");
+
+        properties.put(key1, prop1);
+        properties.put(key2, prop2);
+        properties.put(key3, prop3);
+        properties.put(key4, null);
```
ajs6f: What's the point of this `null` value? Is this just to cover some line in the class?

fcrepo-connector-file/src/test/java/org/fcrepo/connector/file/ExternalJsonSidecarExtraPropertyStoreTest.java
```diff
@@ -54,12 +72,269 @@ public void testSidecarFile() throws IOException {
...
+        assertTrue(store.contains("/file"));
+
+        final Name langProp = new BasicName("lang", "de");
+        final Name key1 = new BasicName("info", "one");
+        final Name key2 = new BasicName("info", "two");
+        final Name key3 = new BasicName("info", "three");
+        final Name key4 = new BasicName("info", "four");
+        final Property prop1 = new BasicSingleValueProperty(langProp, "eins");
+        final Property prop2 = new BasicSingleValueProperty(langProp, "zwei");
+        final Property prop3 = new BasicSingleValueProperty(langProp, "drei");
```
ajs6f: Should all these names and properties be constants? They seem to reappear.

fcrepo-connector-file/src/test/java/org/fcrepo/connector/file/FedoraFileSystemConnectorTest.java
```diff
@@ -165,6 +166,28 @@ public void testGetDocumentByIdDatastream() {
...
 
     @Test
+    public void testGetDocumentByIdObject() {
+        when(mockTranslator.getPrimaryTypeName(any(Document.class)))
+                .thenReturn(NT_FOLDER);
+        when(mockNameFactory.create(anyString())).thenReturn(
+                new BasicName("", tmpFile.getName()));
+
+        final Document doc = connector.getDocumentById("/" + tmpFile.getName());
+        assertNotNull(doc);
```
ajs6f: Can we have any stronger assertion here?

fcrepo-connector-file/src/test/java/org/fcrepo/connector/file/ExternalJsonSidecarExtraPropertyStoreTest.java
acoburn: exactly -- if there is an item in a map whose value is null, that entry is skipped when applied to the external json store

fcrepo-connector-file/src/test/java/org/fcrepo/connector/file/ExternalJsonSidecarExtraPropertyStoreTest.java
acoburn: yes, they probably should be

fcrepo-connector-file/src/test/java/org/fcrepo/connector/file/ExternalJsonSidecarExtraPropertyStoreTest.java
ajs6f: Why would such a value exist to begin with? IOW, do we need that test in the code at all?

fcrepo-connector-file/src/test/java/org/fcrepo/connector/file/FedoraFileSystemConnectorTest.java
acoburn: Quite possibly. There I was simply mirroring the existing test cases you see above and below (`testGetDocumentByIdDatastream`, `testGetDocumentByIdContent`, etc)

fcrepo-connector-file/src/test/java/org/fcrepo/connector/file/ExternalJsonSidecarExtraPropertyStoreTest.java
acoburn: The test is there simply to assert the logic of the [implementation](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-connector-file/src/main/java/org/fcrepo/connector/file/ExternalJsonSidecarExtraPropertyStore.java#L203-L208)

ajs6f: @awoods isn't that your code? Do we need all those `null` tests or can we just throw an exception for malformed input?

acoburn: That's actually modeshape code, modified ever so slightly. The modeshape version, however is a package-protected class, so in order to access it, the `ExternalJsonSidecarExtraPropertyStore` class was reimplemented in the fcrepo namespace. See this commit: https://github.com/fcrepo4/fcrepo4/commit/dff2d5f0be464981563a5c2acaad3c3422dfe732

ajs6f: Yeah, I remember that. I'm asking if we need to copy that stuff so exactly.

awoods: It is probably easier to maintain copied code if it closely resembles its source. But throwing an `InvalidArgumentException` or the like is fine too... either way.

acoburn: The thing about the `null` values is that when a property's value is set to `null`, the object corresponding to that key in the backing JSON document is removed. If the value is non-null, the corresponding property is updated. Unless there are objections, I'd be more inclined to keep the implementation as-is, since we know it works at present.

awoods: :+1:

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/51d92027bc6afbf0053e1e50787682ba33af496b

PR #881, opened 2015-08-11
Fix for FCREPO-1678

ajs6f:


acoburn:
I am noticing that a request for a `NonRdfSourceDescription` returns the same `ETag` value as the `NonRdfSource`. This isn't an issue for caching proxies, as the resources correspond to different URLs. I also notice, however, that when the `NonRdfSourceDescription` changes (e.g. via `sparql-update`), the `ETag` header for the `NonRdfSource` also changes. Is that the intended behavior? The bits on the `NonRdfSource` surely have not changed, so why the new `ETag` value? This is certainly better than the current behavior -- where a `sparql-update` on the description changes the `ETag` on the `Binary`, but not the description.

ajs6f:
That is the correct behavior. The description and the bitstream have an independent existence in HTTP-land, although we had rather they did not. (You will recall that after much thought and argument, it wasn't clear how to merge them.) But they do not have an independent existence as resources in the repository. They each live exactly as long as the other. The properties in question _are properties of the resource persisted as the bitstream_, not some nebulous and ill-defined "metadata node".

PR #882, opened 2015-08-12
move modeshape-specific code out of the kernel-api

acoburn:
Related to https://jira.duraspace.org/browse/FCREPO-1694


ajs6f:
This is good, but we need to go even further and get the JCR stuff out of the API code. But this is a good start.

acoburn:
We totally need to go further, this is just a start...

ajs6f:
I think I will see if I can send a PR to this going a bit further. Maybe we can ping-pong it a bit. If not I will just merge this by itself.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/converters/PropertyConverter.java
```diff
@@ -94,7 +94,10 @@ public static String getPropertyNameFromPredicate(final Node node,
                                                       final Resource predicate,
                                                       final Map<String, String> namespaceMapping)
             throws RepositoryException {
-        final NamespaceRegistry namespaceRegistry = getNamespaceRegistry.apply(node);
+
+        final NamespaceRegistry namespaceRegistry = uncheck((final Node n) ->
```
awoods: Was there are reason for not just using `NamespaceTools.getNamespaceRegistry(n)`?
https://github.com/fcrepo4/fcrepo4/pull/882/files#diff-a456c1fc2649a798904fe05c3ae2e5fdR46

acoburn:
This addresses the subtask: https://jira.duraspace.org/browse/FCREPO-1824

Now, the only remaining trace of Modeshape is in the `o.f.k.a.services.RepositoryService` class. The methods `backupRepository` and `restoreRepository` each return a `org.modeshape.jcr.api.Problems` object. I would like feedback on what to do with that.

fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/FedoraUserSecurityContext.java
```diff
@@ -15,7 +15,7 @@
  */
 package org.fcrepo.auth.common;
 
-import static org.fcrepo.kernel.api.FedoraJcrTypes.JCR_CONTENT;
+import static org.modeshape.jcr.api.JcrConstants.JCR_CONTENT;
```
awoods: The initial intent of this work was to remove dependencies from other modules on the "fcrepo-kernel-modeshape" module. This particular refactoring seems to be moving in the opposite direction.

acoburn: I disagree. The core issue is this: isolate all modeshape-related code in the `fcrepo-kernel-modeshape` module. And in order to achieve this, there are two primary issues to resolve.

First, this means removing all modeshape-specific code from `fcrepo-kernel-api` where it *most certainly* does not belong. That is the primary goal of this particular PR. And IMO it is the right place to begin.

The second part of the issue relates to removing specific dependencies on `fcrepo-kernel-modeshape`. The primary issue here relates to how `RdfStream::getTriples` is used (See related issue at https://jira.duraspace.org/browse/FCREPO-1312). That is, a class that calls `RdfStream::getTriples` must pass in a class from the `org.fcrepo.kernel.modeshape.rdf.impl` package. See, for example here: http://git.io/v8DUq, where the fcrepo-http-api module is completely tied to the fcrepo-kernel-modeshape module. That sort of interdependence will need to change; but it should be a separate effort (separate from this PR).

In this particular case: fcrepo-auth-common is already entirely bound to modeshape, so using a JCR-related constant from `org.modeshape.jcr.api` instead of `org.fcrepo.kernel.api` makes perfect sense to me.

ajs6f: I agree with @acoburn on this, and I ask him to please create a ticket for that refactoring to get dependence on `org.fcrepo.kernel.modeshape.rdf.impl ` out of the kernel. That was a gross oversight for which I think I can mostly be blamed. @cbeer wrote that, but I convinced him that it was a good idea. It was a good idea, but not good enough. We need to completely rethink the RDF generation subsystem in light of these things:

1. The new Java 8 Streams API
2. Our desire to refactor JCR into invisibility.
3. Our need to produce a specification that can be reimplemented in a practical way.

Let me further suggest. along the lines of @acoburn 's argument, that this module actually be renamed ` fcrepo-auth-common-modeshape`.

awoods: The argument here makes sense. Can you also describe how this current PR moves us in the direction of the "parent" ticket:
https://jira.duraspace.org/browse/FCREPO-1694 : "*Ensure that no non-deployable module compile-depends on fcrepo-kernel-modeshape*"
...or maybe the intent/title of the above ticket is simply misleading.

ajs6f: It is confused. I had forgotten about the relationship between authN and MODE. We can alter the title to be more precise and therefore accurate. Did I miss anything other than authN?

awoods: `find . -name 'pom.xml' -exec grep -H "fcrepo-kernel-modeshape" "{}" \;`
results in:
```
./fcrepo-kernel-modeshape/pom.xml:  <artifactId>fcrepo-kernel-modeshape</artifactId>
./fcrepo-connector-file/pom.xml:      <artifactId>fcrepo-kernel-modeshape</artifactId>
./fcrepo-auth-common/pom.xml:      <artifactId>fcrepo-kernel-modeshape</artifactId>
./fcrepo-http-api/pom.xml:      <artifactId>fcrepo-kernel-modeshape</artifactId>
./fcrepo-http-api/pom.xml:      <artifactId>fcrepo-kernel-modeshape</artifactId>
./pom.xml:    <module>fcrepo-kernel-modeshape</module>
./fcrepo-webapp/pom.xml:      <artifactId>fcrepo-kernel-modeshape</artifactId>
./fcrepo-webapp/target/console-work/META-INF/maven/org.fcrepo/fcrepo-webapp/pom.xml:      <artifactId>fcrepo-kernel-modeshape</artifactId>
./fcrepo-http-commons/pom.xml:      <artifactId>fcrepo-kernel-modeshape</artifactId>
./fcrepo-boms/fcrepo4-bom/pom.xml:        <artifactId>fcrepo-kernel-modeshape</artifactId>
./fcrepo-jms/pom.xml:      <artifactId>fcrepo-kernel-modeshape</artifactId>
```

acoburn: `fcrepo-jms` has a test-dependency, which doesn't matter. It would be expected that `fcrepo-webapp` depends on the modeshape impl. Both `fcrepo-connector-file` and `fcrepo-auth-common` have a special relationship to modeshape and should remain as they are. The only concerning dependencies are `fcrepo-http-api` and `fcrepo-http-commons`.

acoburn: Also, I believe that fcrepo-audit depends on fcrepo-kernel-modeshape (in the same way fcrepo-http-api does).

awoods:
Regarding: https://github.com/fcrepo4/fcrepo4/pull/882#issuecomment-154422402
We should probably change the backup/restore feature to transact in RDF instead of JCR/XML... that would likely resolve the ModeShape dependency in the code and for clients.

awoods:
Resolved with: 
https://github.com/fcrepo4/fcrepo4/commit/f04635cb9757560e77e7639a2cb98569968ea0db and
https://github.com/fcrepo4/fcrepo4/commit/42e6fe4224aeaf28c105387b89ed97cde5e31b27

fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/FedoraUserSecurityContext.java
ajs6f: Is that a test dependency?

ajs6f:
That is repo-wide backup/restore?

awoods:
yes

ajs6f:
We'll need more than RDF, right? The bitstreams need to come with. I hate to keep banging the same drum, but maybe we should look at @cwilper 's Semantic Packages again.

awoods:
I think both options need to be available:
1. RDF and binaries
2. RDF only

I suspect the community would be more attracted to BagIt. How do you view the comparison between BagIt and SCP?

ajs6f:
I view BagIt as a second-class choice. There's no instant translation of RDF into BagIt and the client libraries are awful. That having been said, Semantic Packages were never implemented, so there would be extra work there. It might be better to think of another choice-- RDF with links to binaries, instead of trying to package everything together.

escowles:
For better or worse, BagIt is what everybody uses (that I've heard detail what they use, anyway).  I think the mapping should be pretty easy: the RDF would be an extra tag file, the child nodes and binaries would be the payload, and the binary descriptions would be converted to sidecar RDF files.

ajs6f:
Right, the idea of sidecar RDF files isn't very attractive, and I don't see that pushing the RDF into a tag file actually works all that well. But to be honest, I'm not going to do this work and I don't think that the "whole repo" backup is all that important anyway, so whatever gets the box checked most quickly is fine with me.

acoburn:
Just to be clear, the question of BagIt vs. SCP vs. something else is actually orthogonal to the question of modeshape dependencies in the kernel-api. The problem with the `org.fcrepo.kernel.api.services.RepositoryService` class is that the `backup` and `restore` methods return a `org.modeshape.jcr.api.Problems` object. This object is effectively the same as a `Collection<Throwable>`, and by writing a small shim in the `RepositoryServiceImpl` class, the modeshape-specific class could be removed from the kernel-api. Does that sound reasonable?

ajs6f:
Yes, this was a tangent. +1 to shimming over `Problems`.

DiegoPino:
Sorry for crushing in here. But I have been doing some research related to this stuff for islandora. I found this people, doing https://researchobject.github.io/specifications/bundle/ looks more semantic ready than adapting bagIT

Their approach is very aligned to LDP and have even some tools. More over, they have very similar approaches on many things to what you are doing.

http://www.researchobject.org/specifications/

Just a contribution (maybe just adding noise to the conversation),but i really don't die for BagIT nor for a BagIT semantic monster.

awoods:
@acoburn :+1: for shimmy-shimmy

ajs6f:
@awoods Do we have a separate issue available to track discussion about the backup/restore serialization?

awoods:
@ajs6f, yes: https://github.com/fcrepo4/fcrepo4/issues/946
Also, @DiegoPino, thanks for you link to RO-Bundle. It appears to be BagIt-meets-RDF. I obviously have not gone too deeply into it, but one concern would be around any potential restrictions in namespaces, predicates, and mandatory properties or structure that F4 may have a hard time supporting.
In any case, let's continue the discussion on this issue: https://github.com/fcrepo4/fcrepo4/issues/946

ajs6f:
@DiegoPino Please put a link to RO-Bundle in #946.

DiegoPino:
@ajs6f and @awoods: done in #946. Thanks

PR #887, opened 2015-08-18
Tear fcrepo-transform out

whikloj:
Addresses [FCREPO-1670](https://jira.duraspace.org/browse/FCREPO-1670)
- See also https://github.com/whikloj/fcrepo-transform


awoods:
I assume we will want to make sure the vagrant machinery (fcrepo-webapp-plus) is still including fcrepo-transform so Solr indexing will still work.
That implies moving fcrepo-transform into fcrepo4-exts.

whikloj:
Ok. Since I can't create a repo under fcrepo4-exts. I made the fcrepo-transform under my user. 

Which you can fork or whatever. Is that what you are referring to or did I miss the point?

acoburn:
Here you go: https://github.com/fcrepo4-exts/fcrepo-transform

acoburn:
You'll want to be sure the version history persists.

awoods:
@acoburn, indeed. It looks like the git history got lost during the extraction from the core codebase.

acoburn:
This might be helpful: http://gbayer.com/development/moving-files-from-one-git-repository-to-another-preserving-history/

whikloj:
Still don't think I can push to that repository, but I have fixed mine to maintain the version history.

acoburn:
@whikloj you should be able to issue a PR to that repo

whikloj:
@acoburn I think I would have to clone your empty repo and then fill it again. I think it would be easier if you forked my fcrepo-transform repo into fcrepo-exts.

whikloj:
alternatively, the instructions you gave me worked great if you want to fill the new empty repo in fcrepo4-exts/fcrepo-transform yourself.

whikloj:
@acoburn - or more precisely.
```
git clone https://github.com/fcrepo4/fcrepo4 fcrepo4-transform
git remote rm origin
git filter-branch --subdirectory-filter <directory 1> -- --all
git add remote origin https://github.com/fcrepo4-exts/fcrepo-transform
git branch --set-upstream-to=origin
```

awoods:
The following files also need to be removed (and the entire fcrepo-transform directory):
- https://github.com/whikloj/fcrepo4/blob/3002ad1729122306e630a9ee7c2c77cb1d0384f2/fcrepo-transform/src/main/java/org/fcrepo/transform/TransformationFactory.java
- https://github.com/whikloj/fcrepo4/blob/3002ad1729122306e630a9ee7c2c77cb1d0384f2/fcrepo-transform/src/test/java/org/fcrepo/transform/http/FedoraTransformTest.java

Also, we do not want any dependencies in fcrepo4 on projects in fcrepo4-exts. Therefore, you also need to remove `fcrepo-transform` from the following:
- https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-webapp/pom.xml#L74
- https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-boms/fcrepo4-bom/pom.xml#L91

acoburn:
Also, will fcrepo-transform remain in the org.fcrepo maven group? Or will it change to org.fcrepo.transform?

awoods:
@acoburn, yes, moving to the `org.fcrepo.transform` group makes sense.

whikloj:
so does the parent become the org.sonatype.oss like fcrepo-camel and fcrepo4?

acoburn:
@whikloj I'd keep the parent as it is for now.

acoburn:
@whikloj just an FYI, there is a ticket for creating an `fcrepo-parent` POM so that extension modules can inherit from that rather than from `org.sonatype.oss` (as is currently the case with fcrepo-camel and fcrepo-camel-toolbox): https://jira.duraspace.org/browse/FCREPO-1684

But I have to address some code review comments before we can finalize that ticket.

acoburn:
@whikloj can you rebase this PR with the latest updates from master?

whikloj:
@acoburn rebased

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/266bc0ab7f3bec42723f431adee4669b0b163e80

PR #890, opened 2015-08-21
Follow-on to https://jira.duraspace.org/browse/FCREPO-1667

awoods:
Remove use of Jena-atlas runtime exception, 
...which added a non-OSGi friendly dependency.

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -67,6 +67,7 @@ import javax.ws.rs.core.Response;
 
 import com.fasterxml.jackson.core.JsonParseException;
 import org.apache.jena.atlas.RuntimeIOException;
+import org.apache.jena.riot.RiotException;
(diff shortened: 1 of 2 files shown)


acoburn:
@awoods it's not that jena.atlas is not OSGi friendly, but rather that the transitive dependencies clash:

fcrepo-camel depends on clerezza to do some rdf (de-)serialization, and that in turn depends on a wrapped version of jena (version 2.11), which does not contain that particular o.a.jena.atlas class. So, when fcrepo4 is launched for the integration tests, the class loader can't find the jena.atlas class in question, since it doesn't exist in the wrapped clerezza bundles.

That is to say, this proposed change is fine, but perhaps we should address it with a change in the camel component -- though doing so is a bit more complex.

awoods:
@acoburn, I will leave it up to you if you would prefer handling this with the current PR or doing something else within fcrepo-camel.

awoods:
Personally, pushing the exception handling down from FedoraLdp into ContentExposingResource seems like an improvement in any case.

acoburn:
@awoods I'll take a look at how complicated a fix for this in the camel component. Either way, I agree with you that moving the exception handling to `ContentExposingResource` is a good thing.

PR #891, opened 2015-08-25
Filtering references that aren't inbound to the node in question

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-1701


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/ReferencesRdfContext.java
```diff
@@ -68,6 +69,10 @@ public ReferencesRdfContext(final FedoraResource resource,
         putReferencesIntoContext(resource.getNode());
     }
 
+    private final Predicate<Triple> INBOUND = t -> {
+        return t.getObject().getURI().equals(translator().reverse().convert(resource()).getURI());
+    };
```
ajs6f: Maybe the shorter `t.getObject().getURI().equals(uriFor(resource()).getURI())`?

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/731e92ca5ad30e83edae4ef6db972d9e8a1e72fd

PR #895, opened 2015-08-28
FCREPO-1713. Refactor EVERYONE user principal into the FAD.

peichman-umd:
Instead of a single static principal named EVERYONE in the ServletContainerAuthenticationProvider class, rely on each delegate to supply a Principal that it uses to represent the EVERYONE user through a getEveryonePrincipal() method.

Updated the integration tests with mocks for the EVERYONE principal for their mock FAD objects.

### DIFF
--- a/fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/FedoraUserSecurityContext.java
+++ b/fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/FedoraUserSecurityContext.java
@@ -32,7 +32,7 @@ import org.slf4j.LoggerFactory;
  * @author Gregory Jansen
  */
 public class FedoraUserSecurityContext implements SecurityContext,
-AdvancedAuthorizationProvider {
+        AdvancedAuthorizationProvider {
 
     private static final Logger LOGGER = LoggerFactory
             .getLogger(FedoraUserSecurityContext.class);
--- a/fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/FedoraUserSecurityContext.java
+++ b/fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/FedoraUserSecurityContext.java
@@ -15,15 +15,13 @@
  */
 package org.fcrepo.auth.common;
 
-import static org.fcrepo.auth.common.ServletContainerAuthenticationProvider.EVERYONE;
+import java.security.Principal;
 
 import org.modeshape.jcr.security.AdvancedAuthorizationProvider;
 import org.modeshape.jcr.security.SecurityContext;
 import org.modeshape.jcr.value.Path;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.security.Principal;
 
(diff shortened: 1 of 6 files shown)


fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/ServletContainerAuthenticationProvider.java
```diff
@@ -218,4 +195,13 @@ public void setFad(final FedoraAuthorizationDelegate fad) {
 
         return principals;
     }
+
+    /**
+     * Get the principal that the injected authorization delegate uses to represent the "EVERYONE" user.
+     *
+     * @return principal representing "EVERYONE"
+     */
+    public Principal getEveryonePrincipal() {
```
awoods: Is this `public` method ever called? If not, please remove.

fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/ServletContainerAuthenticationProvider.java
```diff
@@ -177,12 +154,12 @@ public ExecutionContext authenticate(final Credentials credentials,
         } else {
 
             sessionAttributes
-                    .put(FedoraAuthorizationDelegate.FEDORA_USER_PRINCIPAL,
-                            EVERYONE);
+            .put(FedoraAuthorizationDelegate.FEDORA_USER_PRINCIPAL,
```
awoods: If you are going to reformat this line, please go ahead and move the `.put(...)` up to line:156.

fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/FedoraUserSecurityContext.java
```diff
@@ -34,7 +32,7 @@
  * @author Gregory Jansen
  */
 public class FedoraUserSecurityContext implements SecurityContext,
-        AdvancedAuthorizationProvider {
+AdvancedAuthorizationProvider {
```
awoods: Generally, reformatting that has nothing to do with functional updates is frowned upon for the reason that it distracts the code review process, as well as makes code-history archaeology more difficult.

ajs6f: Just to be clear, this is entirely a concern of @awoods. No one else has ever so much as noticed it.

awoods: ...and @barmintor

peichman-umd: These formatting changes were put in there automatically by Eclipse. I would be happy to revert them.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/4629227735a6df3de25da2b8347e9cd190ef1c69

PR #899, opened 2015-09-01
added getTypes() method

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1717

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
@@ -18,6 +18,7 @@ package org.fcrepo.kernel.api.models;
 import java.net.URI;
 import java.util.Date;
 import java.util.Iterator;
+import java.util.List;
 
 import javax.jcr.AccessDeniedException;
 import javax.jcr.Node;
@@ -108,11 +109,23 @@ public interface FedoraResource {
     Date getLastModifiedDate();
 
     /**
-     * Check if this object uses a given mixin
+     * Check if this object uses a given RDF type
+     *
+     * <p>Note: the type parameter should be in prefixed short form, so ldp:Container or ex:Image
+     * are both acceptable types. This method does not assume any jcr to fedora prefix mappings are
+     * managed by the implementation, so hasType("jcr:frozenNode") is a valid use of this method.</p>
+     *
      * @param type the given type
-     * @return a collection of mixin names
+     * @return whether the object has the given type
      */
     boolean hasType(final String type);
+
+    /**
+     * Get the RDF:type values for this resource
+     * @return a list of types for this resource
+     */
+    List<URI> getTypes();
+
     /**
      * Update the provided properties with a SPARQL Update query. The updated
      * properties may be serialized to the JCR store.
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -409,6 +409,7 @@ public class FedoraResourceImpl extends JcrTools implements FedoraJcrTypes, Fedo
                 .map(uncheck(x -> x.getName()))
                 .distinct()
                 .map(nodeTypeNameToURI::apply)
+                .peek(x -> LOGGER.debug("node has rdf:type {}", x))
                 .collect(Collectors.toList());
 
         } catch (final PathNotFoundException e) {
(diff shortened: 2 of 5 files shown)


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -385,6 +390,41 @@ public boolean hasType(final String type) {
...
+            mixinTypes.stream()
+                .map(NodeType::getSupertypes)
+                .flatMap(Arrays::stream)
+                .forEach(nodeTypes::add);
+
+            return nodeTypes.stream()
+                .map(uncheck(x -> x.getName()))
+                .distinct()
+                .map(nodeTypeNameToURI::apply)
+                .collect(Collectors.toList());
```
awoods: A "debug" logging call would be helpful on the collected types.

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/FedoraResourceImplTest.java
```diff
@@ -150,6 +166,48 @@ public void testGetCreatedDate() throws RepositoryException {
...
+        when(mockPrimarySuperNodeType.getName()).thenReturn(
+                mockNodeTypePrefix + ":" + mockPrimarySuperNodeTypeName);
+
+        when(mockMixinNodeType.getSupertypes()).thenReturn(
+                new NodeType[] {mockMixinSuperNodeType, mockMixinSuperNodeType});
+        when(mockMixinSuperNodeType.getName()).thenReturn(
+                mockNodeTypePrefix + ":" + mockMixinSuperNodeTypeName);
+
+        final List<URI> types = testObj.getTypes();
+        assertTrue(types.contains(URI.create(REPOSITORY_NAMESPACE + "somePrimaryType")));
```
awoods: A minor point, but you could use the constants defined above instead of recreating the Strings.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
```diff
@@ -108,11 +109,23 @@
...
-     * @return a collection of mixin names
+     * @return whether the object has the given type
      */
     boolean hasType(final String type);
+
+    /**
+     * Get the RDF:type values for this resource
+     * @return a list of types for this resource
+     */
+    List<URI> getTypes();
```
ajs6f: With this a `List<URI>`, are we outlawing relative URIs for types?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
ajs6f: Is this duplicating stuff that is in `TypeRdfContext`? Do we need to DRY up?

ajs6f: Never mind. I see what you did there.

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/FedoraResourceImplTest.java
```diff
@@ -99,6 +103,18 @@
...
+    private NodeType mockPrimaryNodeType;
+
+    @Mock
+    private NodeType mockMixinNodeType;
+
+    @Mock
+    private NodeType mockPrimarySuperNodeType;
+
+    @Mock
+    private NodeType mockMixinSuperNodeType;
```
ajs6f: We have some kind of checkstyles thing that disallows multiple-field declarations. I think this is a good example of why we should change that.

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/FedoraResourceImplTest.java
ajs6f: Amen.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
awoods: What would be an example of a "relative URI" for types?

ajs6f: Any relative URI. "foo/bar".

acoburn: Since an `rdf:type` should always refer to an `rdfs:Class`, is it possible that a relative URI would ever represent a `Class`?

ajs6f: Sure, it's trivial. All I have to do is type some resource in the repo as an `rdfs:Class` and then use a relative URI. It turns out that it doesn't matter, though. At least since Java 7, `java.net.URI` can hold a relative URI. So no worries.

acoburn:
I added some debug logging in the getTypes() implementation and cleaned up the unit test.

awoods:
This looks good.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/b03c4ea93d0471f2c46ba2c626d3e2427429a41d

PR #902, opened 2015-09-03
Could use Metaspace to replace PermGen

yinlinchen:


pom.xml
```diff
@@ -564,7 +564,7 @@
             This line enables code-coverage. However, it also blows the build's permgen.
             Fix me!
           -->
-          <argLine>-XX:-UseSplitVerifier ${jacoco.agent.unit.arg}</argLine>
+          <argLine>-XX:-UseSplitVerifier -XX:MaxMetaspaceSize=512m ${jacoco.agent.unit.arg}</argLine>
```
awoods: We can probably also remove `UseSplitVerifier`
http://stackoverflow.com/questions/31566672/java-8-what-is-the-equivalent-of-usesplitverifier

pom.xml
```diff
@@ -545,7 +545,7 @@
       <plugin>
         <artifactId>maven-failsafe-plugin</artifactId>
         <configuration>
-          <argLine>${jacoco.agent.it.arg}</argLine>
+          <argLine>-XX:MaxMetaspaceSize=128m ${jacoco.agent.it.arg}</argLine>
```
awoods: Why `128m` here and `512m` below and `1024m` above?

yinlinchen: saw these from old pom.xml:  https://github.com/fcrepo4/fcrepo4/pull/883/files

awoods:
Please create a JIRA ticket for tracking this update.

PR #907, opened 2015-09-09
Add logic to handle patch requests that reference the root...

awoods:
- Removing "jcr:content" from authorization checks

Partial resolution of: https://jira.duraspace.org/browse/FCREPO-1725

### DIFF
--- a/fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/FedoraUserSecurityContext.java
+++ b/fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/FedoraUserSecurityContext.java
@@ -37,6 +37,9 @@ public class FedoraUserSecurityContext implements SecurityContext,
     private static final Logger LOGGER = LoggerFactory
             .getLogger(FedoraUserSecurityContext.class);
 
+    // Defined here because FedoraJcrTypes is in fcrepo-http-api, which would create a cyclic dependency
+    protected static final String JCR_CONTENT = "jcr:content";
+
     protected Principal userPrincipal = null;
 
     protected FedoraAuthorizationDelegate fad = null;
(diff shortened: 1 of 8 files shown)


fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/FedoraUserSecurityContext.java
```diff
@@ -37,6 +37,9 @@
     private static final Logger LOGGER = LoggerFactory
             .getLogger(FedoraUserSecurityContext.class);
 
+    // Defined here because FedoraJcrTypes is in fcrepo-http-api, which would create a cyclic dependency
+    protected static final String JCR_CONTENT = "jcr:content";
```
acoburn: `FedoraJcrTypes` is in fcrepo-kernel-api, and if `jcr:content` were defined there, I'm not sure that I see how a cyclic dependency would be created.

awoods: Try it, it is true.

acoburn: I disagree. See my pull request: https://github.com/awoods/fcrepo4/pull/4

awoods: Thanks for the PR.
Would you agree that your PR creates a dependency from fcrepo-auth-common on to fcrepo-http-api? This new dependency is coming transitively from some other artifact... but exists, nonetheless. If you explicitly add the dependency into fcrepo-auth-common's pom.xml, Maven complains.

Are you comfortable relying on this transitive coupling?
Am I missing something?

acoburn: We're talking about different things.

This PR creates a dependency on `fcrepo-kernel-api`, not `fcrepo-http-api`. If the dependency to `fcrepo-kernel-api` is added to maven (which it should be: see my additional commit), there is no complaint. This was the whole point, a few weeks ago, in removing `fcrepo-mint` from the kernel-api: that way the kernel-api is a the very bottom of food-chain, as it were, and *any* other module should freely be able to use it.

ajs6f: Yes, if there is any problem with any module at all using the API, that is a massive architecture failure. the API should be essentially independent of _everything else_. I'm not saying it is, I'm saying that such is the condition for which we must be working.

awoods: Thanks, @acoburn. You are correct, I was transposing `fcrepo-http-api` for `fcrepo-kernel-api`.

acoburn:
Merged via 249977cae0036673fe2e293097380947c7052f63

PR #911, opened 2015-09-23
Users can now delete large literal property values.

bseeger:
- Large literals that are stored in the binary store are now removeable
via standard means.

Resolves: https://jira.duraspace.org/browse/FCREPO-1561

### DIFF
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NodePropertiesTools.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NodePropertiesTools.java
@@ -28,6 +28,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import org.apache.commons.lang3.StringUtils;
+
 import javax.jcr.Node;
 import javax.jcr.Property;
 import javax.jcr.PropertyType;
@@ -200,19 +202,21 @@ public class NodePropertiesTools {
         if (node.hasProperty(propertyName)) {
 
             final Property property = node.getProperty(propertyName);
+            final String strValueToRemove = valueToRemove.getString();
 
             if (property.isMultiple()) {
                 final AtomicBoolean remove = new AtomicBoolean();
                 final Value[] newValues = stream(node.getProperty(propertyName).getValues()).filter(uncheck(v -> {
                     final String strVal = v.getString();
 
-                    if (strVal.equals(valueToRemove.getString())) {
+                    LOGGER.debug("v is '{}', valueToRemove is '{}'", v, strValueToRemove );
+                    if (strVal.equals(strValueToRemove)) {
                         remove.set(true);
                         return false;
                     }
 
                     LOGGER.warn("Failed to find value for property '{}'. Unable to compare value (type: {}) " +
-                       "'{}' to requested value '{}'", propertyName, v.getType(), strVal, valueToRemove);
+                       "'{}' to requested value '{}'", propertyName, v.getType(), strVal, strValueToRemove);
                     return true;
                 })).toArray(Value[]::new);
 
@@ -222,31 +226,28 @@ public class NodePropertiesTools {
                         LOGGER.debug("Removing property '{}'", propertyName);
                         property.remove();
                     } else {
-                        LOGGER.debug("Removing value '{}' from property '{}'", valueToRemove, propertyName);
+                        LOGGER.debug("Removing value '{}' from property '{}'", strValueToRemove, propertyName);
                         property.setValue(newValues);
                     }
                 } else {
-                    LOGGER.warn("Value not removed from property name '{}' (value '{}')", propertyName,
-                            valueToRemove);
+                    LOGGER.debug("Value not removed from property name '{}' (value '{}')", propertyName,
+                            strValueToRemove);
+                    throw new RepositoryException ("Property '" + propertyName + "': Unable to remove value '" +
+                            StringUtils.substring(strValueToRemove, 0, 50) + "...'");
                 }
-
             } else {
 
-                String strVal = "";
-                try {
-                    strVal = property.getValue().getString();
+                final String strPropVal = property.getValue().getString();
 
-                    if (property.getValue().equals(valueToRemove)) {
-                        LOGGER.debug("single value: Removing value from property {}", propertyName);
-                        property.remove();
-                    } else {
-                        LOGGER.debug("Value not removed from property name '{}' (value '{}')", propertyName,
-                                valueToRemove);
-                    }
-                } catch (RepositoryException e) {
-                    LOGGER.warn("Failed to remove value from property '{}'. Unable to compare value '{}' " +
-                        "(type: {}) '{}' to requested value {}", propertyName, property.getValue().getType(),
-                        strVal, valueToRemove);
+                LOGGER.debug("Removing string '{}'", strValueToRemove);
+                if (StringUtils.equals(strPropVal, strValueToRemove)) {
+                    LOGGER.debug("single value: Removing value from property '{}'", propertyName);
+                    property.remove();
+                } else {
+                    LOGGER.debug("Value not removed from property name '{}' (property value: '{}';compare value: '{}')",
+                            propertyName, strPropVal, strValueToRemove);
+                    throw new RepositoryException("Property '" + propertyName + "': Unable to remove value '" +
+                            StringUtils.substring(strValueToRemove, 0, 50) + "'");
                 }
             }
         }
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NodePropertiesTools.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NodePropertiesTools.java
@@ -215,8 +215,6 @@ public class NodePropertiesTools {
                         return false;
                     }
 
-                    LOGGER.warn("Failed to find value for property '{}'. Unable to compare value (type: {}) " +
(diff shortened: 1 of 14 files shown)


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NodePropertiesTools.java
```diff
@@ -203,9 +203,18 @@ public void removeNodeProperty(final Node node, final String propertyName, final
             if (property.isMultiple()) {
                 final AtomicBoolean remove = new AtomicBoolean();
                 final Value[] newValues = stream(node.getProperty(propertyName).getValues()).filter(v -> {
-                    if (v.equals(valueToRemove)) {
-                        remove.set(true);
-                        return false;
+                    String strVal = "";
```
ajs6f: Better `final String strVal;`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NodePropertiesTools.java
```diff
@@ -203,9 +203,18 @@ public void removeNodeProperty(final Node node, final String propertyName, final
...
+                    try {
+                        strVal = v.getString();
+
+                        if (strVal.equals(valueToRemove.getString())) {
+                            remove.set(true);
+                            return false;
+                        }
+
+                    } catch (RepositoryException e) {
```
ajs6f: Consider `UncheckedPredicate` instead.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NodePropertiesTools.java
```diff
@@ -213,18 +222,34 @@ public void removeNodeProperty(final Node node, final String propertyName, final
...
+
+                String strVal = "";
+                try {
+                    strVal = property.getValue().getString();
+
+                    if (property.getValue().equals(valueToRemove)) {
+                        LOGGER.debug("single value: Removing value from property {}", propertyName);
+                        property.remove();
+                    } else {
+                        LOGGER.info("Value not removed from property name '{}' (value '{}')", propertyName,
```
ajs6f: Is this `INFO` when the message above is `DEBUG`?

bseeger: I put in 'info', so that something might come out in the log, else it's a silent fail.  Suggestions on how to handle the 'else' in this case welcome.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NodePropertiesTools.java
bseeger: Okay, I'll change that.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NodePropertiesTools.java
ajs6f: Just `LOGGER::debug`, is what I meant.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NodePropertiesTools.java
bseeger: Thanks! I'll put in the UncheckedPredicate.  I don't know if it should continue processing or not. At the very least it would be nice to warn the user that their remove failed, but, being a newbie, I'm not sure what the best mechanism to make that happen would be.

ajs6f: If the remove fails without effect in the API, the repository will not be in the situation in which the user tried to put it, but this will happen quietly. The user will continue blithely with his or her life, while the repository marches off in a different and strange direction. At some later time, the user will discover to his or her grimace- and heartburn-inducing chagrin that the giant literal he or she thought banished remains in place and is now causing confusion, dismay, existential angst, and unexpected fixity results.

I think that failure should stop computation and throw an exception back into the API.

bseeger: My new commit doesn't address this.  (Just an FYI).  I will do another one that does address this very soon.

bseeger: I've been revisiting my logic and want to make sure I understand the expected behavior.  Here are some cases: 

1) user requests remove of property value
      a) value doesn't exist   (silent completion?)  - ie user requests removal of "foo" when there is no property value "foo"
              Do we tell the user it didn't actually delete anything?? Doesn't seem necessary.
      b) value does exist and is removed successfully
      c) value does exist and is not removed
              create RepositoryException and notify user that remove failed. 

This sounds okay? Am I missing anything?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NodePropertiesTools.java
```diff
@@ -202,29 +203,50 @@ public void removeNodeProperty(final Node node, final String propertyName, final
...
+                    LOGGER.warn("Value not removed from property name '{}' (value '{}')", propertyName,
+                            valueToRemove);
                 }
 
             } else {
-                if (property.getValue().equals(valueToRemove)) {
-                    LOGGER.debug("Removing value from property {}", propertyName);
-                    property.remove();
+
+                String strVal = "";
```
ajs6f: Again, better `final String strVal;`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NodePropertiesTools.java
ajs6f: Yes, I think that those are the semantics we want. I don’t think any reasonable user will want feedback for case (a).

---
A. Soroka
The University of Virginia Library

bseeger:
New check in.  This is much cleaner (I hope!).  I'm not sure how to test the single-value case.  It does pass all the repository tests, and I know one tests a remove a singe-value property.  Is that enough?  (I did test the multi-value case as much as I could.) 

Sort of an aside but worth mentioning -- from testing, I noticed that it will not get into this removeNodeProperty method if you try to remove a value on a property and that value doesn't exist.  That check is happening somewhere above this code.  Overall it's a silent fail, but begs the question of "what if I mistyped the value and really do want to delete "Foo", and I'm not really paying attention to the fact that it failed because I typed DELETE WHERE { <> ex:test "foo" } instead." Sort of like you were saying in an above comment - the user goes on their way thinking it succeeded.  Separate bug?  

The original bug is resolved by my changes though.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NodePropertiesTools.java
```diff
@@ -194,37 +197,57 @@ public void removeReferencePlaceholders(final IdentifierConverter<Resource,Fedor
...
+                final Value[] newValues = stream(node.getProperty(propertyName).getValues()).filter(uncheck(v -> {
+                    final String strVal = v.getString();
+
+                    LOGGER.debug("v is '{}', valueToRemove is '{}'", v, strValueToRemove );
+                    if (strVal.equals(strValueToRemove)) {
                         remove.set(true);
                         return false;
                     }
+
+                    LOGGER.warn("Failed to find value for property '{}'. Unable to compare value (type: {}) " +
```
awoods: Is this LOGGER statement correct? We are inspecting a multi-valued property here. If the value to be removed (`strValueToRemove`) does not equal `strVal`, all that means is that the value to remove is another one in the list.
For example, the multi-valued property may have the following values: `"a", "b", "c"`. If this request is trying to remove "b", then "a" and "c" will fall through to this LOGGER statement.
That is a perfectly expected case, not a "warning".

bseeger: You're right, I'll remove that.  It would be confusing 'noise'.

awoods:
This looks good... and works as desired. Thank you!
Please add an integration test that would demonstrate the issue prior to your patch, and with the patch shows that the issue has been resolved.
For this new test, it is probably reasonable to just add a new @Test method to:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/test/java/org/fcrepo/integration/kernel/modeshape/FedoraResourceImplIT.java

Since that test class uses the following configuration which specifies `"minimumBinarySizeInBytes" : 40`, it should be easy to reproduce the error in your test method:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/test/resources/config/testing/repository.json

There are several examples of using sparql-update to update properties starting here:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/test/java/org/fcrepo/integration/kernel/modeshape/FedoraResourceImplIT.java#L457

bseeger:
This is an incremental checking, which I expected to fail, so please ignore this checkin for now.  Aaron Coburn is helping by looking at this code.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/8cceaebd0121c34e6b9c156f4d119398d58a1ad4

PR #912, opened 2015-09-24
use premis:hasEventOutcome for fixity results, and don't set it as immutable

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1746


awoods:
:+1: looks good.

PR #913, opened 2015-09-26
replace regex with jena request parser

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1706

> Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.

This removes the (incorrect) use of regular expressions in checking predicate values.

### DIFF
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -45,6 +45,7 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.Iterator;
 import java.util.List;
+import java.util.StringJoiner;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -435,8 +436,14 @@ public class FedoraResourceImpl extends JcrTools implements FedoraJcrTypes, Fedo
         final UpdateRequest request = create(sparqlUpdateStatement,
                 idTranslator.reverse().convert(this).toString());
 
-        for (final IllegalArgumentException ex : checkInvalidPredicates(request)) {
-            throw ex;
+        final Collection<IllegalArgumentException> errors = checkInvalidPredicates(request);
+
+        if (!errors.isEmpty()) {
+            final StringJoiner errorMessages = new StringJoiner(", ");
+            for (final IllegalArgumentException ex : errors) {
+                errorMessages.add(ex.getMessage());
+            }
+            throw new IllegalArgumentException(errorMessages.toString());
         }
 
         final JcrPropertyStatementListener listener = new JcrPropertyStatementListener(
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -40,11 +40,11 @@ import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Optional;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -87,7 +87,6 @@ import org.modeshape.jcr.api.JcrTools;
 import org.slf4j.Logger;
 
 import com.hp.hpl.jena.rdf.model.Model;
-import com.hp.hpl.jena.sparql.core.Quad;
 import com.hp.hpl.jena.sparql.modify.request.UpdateData;
 import com.hp.hpl.jena.sparql.modify.request.UpdateDataDelete;
 import com.hp.hpl.jena.sparql.modify.request.UpdateDataInsert;
@@ -436,9 +435,9 @@ public class FedoraResourceImpl extends JcrTools implements FedoraJcrTypes, Fedo
         final UpdateRequest request = create(sparqlUpdateStatement,
                 idTranslator.reverse().convert(this).toString());
 
-        checkInvalidPredicates(request).ifPresent(x -> {
-            throw new IllegalArgumentException("SPARQL UDPATE statement contains invalid URI: " + x);
-        });
+        for (final IllegalArgumentException ex : checkInvalidPredicates(request)) {
+            throw ex;
+        }
 
         final JcrPropertyStatementListener listener = new JcrPropertyStatementListener(
                 idTranslator, getSession(), idTranslator.reverse().convert(this).asNode());
@@ -688,7 +687,7 @@ public class FedoraResourceImpl extends JcrTools implements FedoraJcrTypes, Fedo
      *
      * @see <a href="https://jira.duraspace.org/browse/FCREPO-1409"> FCREPO-1409 </a> for details.
      */
-    private Optional<Quad> checkInvalidPredicates(final UpdateRequest request) {
+    private Collection<IllegalArgumentException> checkInvalidPredicates(final UpdateRequest request) {
         return request.getOperations().stream()
                 .flatMap(x -> {
                     if (x instanceof UpdateModify) {
@@ -703,7 +702,8 @@ public class FedoraResourceImpl extends JcrTools implements FedoraJcrTypes, Fedo
                     }
                 })
                 .filter(x -> x.getPredicate().isURI() && x.getPredicate().getURI().endsWith("/"))
-                .findFirst();
+                .map(x -> new IllegalArgumentException("Invalid predicate ends with '/': " + x.getPredicate().getURI()))
+                .collect(Collectors.toList());
     }
 
     private Node getFrozenNode(final String label) throws RepositoryException {
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -24,7 +24,6 @@ import static com.google.common.collect.Lists.newArrayList;
 import static com.hp.hpl.jena.update.UpdateAction.execute;
 import static com.hp.hpl.jena.update.UpdateFactory.create;
 import static java.util.Arrays.asList;
-import static java.util.regex.Pattern.compile;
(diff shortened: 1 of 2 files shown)


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -720,6 +681,31 @@ public Node getNodeVersion(final String label) {
...
     }
 
+    /**
+     * Helps ensure that there are no terminating slashes in the predicate.
+     * A terminating slash means ModeShape has trouble extracting the localName, e.g., for
+     * http://myurl.org/.
+     *
+     * @see <a href="https://jira.duraspace.org/browse/FCREPO-1409"> FCREPO-1409 </a> for details.
+     */
+    private Optional<Quad> checkInvalidPredicates(final UpdateRequest request) {
```
ajs6f: Why is this returning/not-returning a `Quad`?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -720,6 +681,31 @@ public Node getNodeVersion(final String label) {
 
     }
 
+    /**
+     * Helps ensure that there are no terminating slashes in the predicate.
+     * A terminating slash means ModeShape has trouble extracting the localName, e.g., for
```
ajs6f: Do we know why MODE has trouble with this? Do we have a ticket to give them a patch for it?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
acoburn: Do you mean: why a `Quad` instead of a predicate `String`? The only reason it's returning an `Optional` like this (instead of, e.g., a `boolean`) is so that the `throw` expression can include the problematic part of the statement in the message.

acoburn: ... but from my perspective, it could be a `Quad` or something else. A `Quad` is easier, since that's what the source data consists of.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
acoburn: I have no idea why MODE has trouble with this. @osmandin worked that earlier ticket and he may know. My intention here was simply to use a jena parser (which can be more precise) rather than a regular expression (which was too greedy).

acoburn: Note: @osmandin has a very nice writeup of Modeshape's behavior in the comments for this ticket: https://jira.duraspace.org/browse/FCREPO-1409

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
ajs6f: No, I mean why a `Quad` or anything else instead of some kind of information about what went wrong.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
ajs6f: My point is that this seems like a limitation in MODE, and we should file a ticket on them.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
acoburn: I could instead make this function return nothing and push the `throw new IllegalArgumentExeption()` into this function. Would you find that a cleaner implementation?

ajs6f: Yes. Another, more mannered approach would be to return `Collection<? extends Exception>`, which would allow for a more sophisticated report.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
osmandin: That was my thought or inclination too when I made the original regex PR, a temporary workaround (in my mind, at least), until we could file a ticket with ModeShape.

So the PR by @acoburn could be used temporarily until ModeShape changes this. I can file a request or submit a PR to them.

ajs6f: ++ @osmandin

acoburn:
@ajs6f I replaced the `Optional` with a `Collection<IllegalArgumentException>`. However, I'm still not convinced that a list is all that appropriate here since the first incorrect predicate will trigger an exception. That is, why is it necessary to collect the 2nd, 3rd, etc error if the first one encountered will cause the request to fail? (That's why I had used an optional in the first place, since I could just use the `findFirst` method on the `Quad`-stream.

ajs6f:
Because with a `Collection` in hand, you can send back a response that contains advice about _all_ the problematic predicates. No one wants to send a message, get a response that says, "This part is bad", correct it, send another message, get a response that says, "Oh, this other part is also bad." etc.

acoburn:
But then it would be better to respond with a `Collection<String>`. Then all the invalid predicates can be collected into a single thrown exception. Is that what you mean? The current implementation makes no sense to me.

ajs6f:
No, an exception represents an error. A string is a message about the error. I'm saying that this method should return errors, not messages. It's up to other parts of the system to convert between the two. You can do other things with an `Exception` besides `throw` or `catch` it.

acoburn:
@ajs6f I pushed a new commit to this PR; is this more along the lines of what you mean?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -435,61 +431,32 @@ public void updateProperties(final IdentifierConverter<Resource, FedoraResource>
...
+        final UpdateRequest request = create(sparqlUpdateStatement,
+                idTranslator.reverse().convert(this).toString());
+
+        final Collection<IllegalArgumentException> errors = checkInvalidPredicates(request);
+
+        if (!errors.isEmpty()) {
+            final StringJoiner errorMessages = new StringJoiner(", ");
+            for (final IllegalArgumentException ex : errors) {
+                errorMessages.add(ex.getMessage());
+            }
```
ajs6f: No loop, just `errors.stream().map(Exception::getMessage).collect(joining(",\n"))`

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -720,6 +687,32 @@ public Node getNodeVersion(final String label) {
...
+     *
+     * @see <a href="https://jira.duraspace.org/browse/FCREPO-1409"> FCREPO-1409 </a> for details.
+     */
+    private Collection<IllegalArgumentException> checkInvalidPredicates(final UpdateRequest request) {
+        return request.getOperations().stream()
+                .flatMap(x -> {
+                    if (x instanceof UpdateModify) {
+                        final UpdateModify y = (UpdateModify)x;
+                        return Stream.concat(y.getInsertQuads().stream(), y.getDeleteQuads().stream());
+                    } else if (x instanceof UpdateDataInsert || x instanceof UpdateDataDelete) {
```
ajs6f: Why not just `x instanceof UpdateData`?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -720,6 +687,32 @@ public Node getNodeVersion(final String label) {
...
+                        return ((UpdateData)x).getQuads().stream();
+                    } else if (x instanceof UpdateDeleteWhere) {
+                        return ((UpdateDeleteWhere)x).getQuads().stream();
+                    } else {
+                        return Stream.empty();
+                    }
```
ajs6f: This whole thing (above) might look better without all the `instanceof` instead using `Update::visit`. It would get rid of all the casting, too, and the code would be appropriately sensitive to Jena introducing any new subtypes of `Update`.

acoburn: I looked into this (replacing the `instanceof` calls with an `UpdateVisitor`), and the implementation is not particularly simple. If we use the existing jena implementation (`UpdateEngineWorker`), we have to use an interface that is deprecated in 3.x (`GraphStore`) and removed from the `UpdateEngineWorker` interface in 3.x, so when we move to the next jena version, we would probably need to reimplement it. Alternately, we could create a new class that fully implements the `UpdateVisitor` interface, but I don't see how that's any better than what we currently have. And given that this is just a patch until MODE is fixed, I don't think it is worth it.

ajs6f: I was not suggesting anything to do with `UpdateEngineWorker`. That is way outside our bounds. I don't think implementing `UpdateVisitor` is actually a big deal at all, _if_, as I said in IRC, we get mileage out of it for several different kinds of validations. If it's just for this, then yes, there's no point in doing it for a temporary workaround. The point would be to use it for this, for system-managed types and predicates, etc.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
awoods: @osmandin, please link your new ModeShape ticket here:
https://jira.duraspace.org/browse/FCREPO-1706

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/6ea28b2d707d0145e20f4aa10ff724a03cd0cbb8

PR #914, opened 2015-10-01
Do not expose UUID of versions - HTML UI

yinlinchen:
screen shot 2015-10-01 at 11 51 24 am
Resolves: https://jira.duraspace.org/browse/FCREPO-1715


awoods:
Deeper implementation needed.

PR #917, opened 2015-10-12
Using userspace last-modified property

escowles:
* Last-modified property and ETag based on it do not change with inbound references.

Fixes https://jira.duraspace.org/browse/FCREPO-1742


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -366,6 +374,18 @@ public Date getLastModifiedDate() {
...
 
+    /**
+     * Set the last-modified date to the current date.
+     */
+    @VisibleForTesting
+    public void touch() {
+        try {
+            getNode().setProperty(FEDORA_LASTMODIFIED, Calendar.getInstance(getTimeZone("UTC")));
+        } catch (final RepositoryException e) {
+            LOGGER.info("Exception caught when trying to update lastModified date", e);
```
ajs6f: Maybe `LOGGER::error` here?

awoods: Also, please do not dump the full exception (i.e. `LOGGER.info("...", e)`). Instead, just log the `e.getMessage()`.

escowles:
I've made the logging changes suggested and updated the lastModified test to round the dates to the nearest second to avoid intermittent failures.

awoods:
@escowles, is there a reason why `FedoraResourceImpl.updateProperties()` does not call `touch()`?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -347,7 +353,9 @@ public Date getCreatedDate() {
     public Date getLastModifiedDate() {
 
         try {
-            if (hasProperty(JCR_LASTMODIFIED)) {
+            if (hasProperty(FEDORA_LASTMODIFIED)) {
+                return new Date(getProperty(FEDORA_LASTMODIFIED).getDate().getTimeInMillis());
```
awoods: Some logging could be nice to help debug where the date is coming from.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/ContainerServiceImpl.java
```diff
@@ -63,7 +63,12 @@ public Container findOrCreate(final Session session, final String path) {
                 initializeNewObjectProperties(node);
             }
 
-            return new ContainerImpl(node);
+            final ContainerImpl container = new ContainerImpl(node);
```
awoods: Are we only going to use `fedora:lastModified` on containers? not Binaries? With this ticket, we may run into the same issue with binaries:
https://jira.duraspace.org/browse/FCREPO-1665

Would it make sense to call `touch()` in `AbstractService.findOrCreateNode()`?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -1966,4 +1966,34 @@ public void testPutReferenceRoot() throws Exception {
         }
     }
 
-}
\ No newline at end of file
+    @Test
+    public void inboundLinksDoNotUpdateEtag() throws IOException {
```
awoods: It is minor, but the naming convention is to begin the test methods with "test", such as: `testInboundLinksDoNotUpdateEtag()`

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/integration/kernel/modeshape/FedoraResourceImplIT.java
```diff
@@ -179,7 +179,25 @@ public void testLastModified() throws RepositoryException {
...
         final Container obj2 = containerService.findOrCreate(session, "/" + pid);
-        assertFalse( obj2.getLastModifiedDate().before(obj2.getCreatedDate()) );
+        final Date created = roundDate(obj2.getCreatedDate());
+        final Date modified = roundDate(obj2.getLastModifiedDate());
+        assertFalse(modified + " should not be before " + created, modified.before(created));
+    }
+
+    @Test
+    public void testTouch() throws RepositoryException {
+        final String pid = UUID.randomUUID().toString();
```
awoods: Might as well use: `final String pid = getRandomPid();`

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/FedoraResourceImplTest.java
```diff
@@ -255,6 +255,33 @@ public void testGetLastModifiedDate() {
     }
 
     @Test
+    public void testTouch() {
+        // test existing JCR_LASTMODIFIED
+        final Calendar someDate = Calendar.getInstance();
+        someDate.add(Calendar.DATE, -1);
+        try {
```
awoods: Instead of these `try/catch` blocks (here and at line:272), is there any reason not to just add a `throws` clause on the `testTouch()` method declaration? It is not clear that we will every run into an exception case when setting up the mocks.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/ContainerServiceImpl.java
escowles: It's not quite as easy as that, because `AbstractService.findOrCreatNode()` works with Nodes instead of FedoraResources.  But it's easy enough to call touch() from both the Container and Binary services.

escowles:
`FedoraResourceImpl.updateProperties()` should have been calling `touch()`, but the `jcr:lastModified` property was still being output, so no tests were failing.  In other cases, there were two `fedora:lastModified` properties in the output.

I've updated `updateProperties()` to call `touch()` and suppressed `jcr:lastModified` so now only the `fedora:lastModified` property is being used in the RDF generation.

awoods:
I am seeing some immediate, unexpected behavior:
1. Whenever I create a resource, the `fedora:lastModified` date/time precedes the `fedora:created` value... that is strange
2. When I add a child resource to a container, the `fedora:lastModified` does not change... I would have expected it to change, no?

@escowles, do you feel confident in this implementation? I am inclined to wait on further testing until you feel like all of the scenarios work as you would expect.

escowles:
I'm not surprised there are some additional places where we need to call touch().  The general whack-a-mole nature of this PR makes me think a little time writing up the expected behavior would be time well spent.

acoburn:
Closing this in favor of #1033

PR #920, opened 2015-10-26
change strong ETags to weak ETags

nianma:


nianma:

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -569,7 +569,8 @@ public String getEtagValue() {
         final Date lastModifiedDate = getLastModifiedDate();
 
         if (lastModifiedDate != null) {
-            return shaHex(getPath() + lastModifiedDate.getTime());
+            return hasType(FEDORA_NON_RDF_SOURCE_DESCRIPTION) ?
```
ajs6f: Does this work for containers? IOW, are containers going to get the `W`?

nianma: I think all the resources can be divided into two groups, RDF resource and Non-RDF resource. Container belongs to RDF resource. Shouldn't we have weak Etag apply for Container?

awoods:
You have the right idea; however, if you add your own "W/" prefix, the syntax in the response header is incorrect (see spec: https://tools.ietf.org/html/rfc7232#section-2.3):
* yours:
`ETag: "W/bfc91b1cb116181ba29a70f7e4239fc2fd0e6fb3"`

* spec:
`ETag: W/"bfc91b1cb116181ba29a70f7e4239fc2fd0e6fb3"`

You should be able to remove your updates from `FedoraResourceImpl.java` and add similar logic in the following two places:
* https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java#L466 and
* https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java#L506

awoods:
You will also be able to roll-back your unit test and add a new integration test to `FedoraLdpIT.java`.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -467,7 +468,8 @@ protected static void addCacheControlHeaders(final HttpServletResponse servletRe
         final Date date = mutableResource.getLastModifiedDate();
 
         if (!etag.getValue().isEmpty()) {
-            servletResponse.addHeader("ETag", etag.toString());
+            servletResponse.addHeader("ETag", mutableResource.hasType(FEDORA_NON_RDF_SOURCE_DESCRIPTION) ?
+               etag.toString() : "W/" + etag.toString());
```
ajs6f: No need for `"W/" + etag.toString())` because `toString()` is implicit. Just `"W/" + etag` is fine. In fact I would prefer:
```
servletResponse.addHeader("ETag", mutableResource.hasType((FEDORA_NON_RDF_SOURCE_DESCRIPTION) ?
+              "" : "W/" ) + etag);
```

awoods:
Superseded by: https://github.com/fcrepo4/fcrepo4/pull/1040

PR #921, opened 2015-10-26
add prefix check before update properties

nianma:
I don't have time to create new type exception. Also I would prefer some design this first like naming...

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/FedoraInvalidPrefixExceptionMapper.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/FedoraInvalidPrefixExceptionMapper.java
@@ -1,0 +1,47 @@
+/**
+ * Copyright 2015 DuraSpace, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.http.commons.exceptionhandlers;
+
+import static javax.ws.rs.core.Response.status;
+import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
+import static org.slf4j.LoggerFactory.getLogger;
+
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+
+import org.fcrepo.kernel.api.exception.FedoraInvalidPrefixException;
+
+import org.slf4j.Logger;
+
+/**
+ * For invalid namespace exceptions on CRUD actions for nodes/datastreams
+ *
+ * @author whikloj
+ * @since September 12, 2014
+ */
+@Provider
+public class FedoraInvalidPrefixExceptionMapper implements ExceptionMapper<FedoraInvalidPrefixException> {
+
+    private static final Logger LOGGER = getLogger(FedoraInvalidPrefixExceptionMapper.class);
+
+    @Override
+    public Response toResponse(final FedoraInvalidPrefixException e) {
+        LOGGER.trace("FedoraInvalidPrefixExceptionMapper caught an exception: {}", e.getMessage());
+        return status(BAD_REQUEST).entity(e.getMessage()).build();
+    }
+
+}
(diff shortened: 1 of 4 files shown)


nianma:
https://jira.duraspace.org/browse/FCREPO-1766

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -436,6 +436,19 @@ public void updateProperties(final IdentifierConverter<Resource, FedoraResource>
 
         final Collection<IllegalArgumentException> errors = checkInvalidPredicates(request);
 
+        request.getPrefixMapping().getNsPrefixMap().forEach(
+                       (k,v) -> { try {
+                                   LOGGER.info("451: prefix mapping is key -> value: " + k + "->" + v + "\n");
```
awoods: Comments on your logging:
* When using slf4j (which you are), you should take advantage of templating. For example, the above line should be:
`LOGGER.info("451: prefix mapping is key -> value: {} -> {}", k, v);`
* This should be `debug` not `info`

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -436,6 +436,19 @@ public void updateProperties(final IdentifierConverter<Resource, FedoraResource>
...
+                  if ( Arrays.asList(getSession().getWorkspace().getNamespaceRegistry().getPrefixes()).contains(k)
```
awoods: The list of namespace prefixes should likely be created once, before this stream and reused.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -436,6 +436,19 @@ public void updateProperties(final IdentifierConverter<Resource, FedoraResource>
...
+                     && !v.equals(getSession().getWorkspace().getNamespaceRegistry().getURI(k))) {
```
awoods: Likewise, you could probably retrieve the namespace-registry before this stream, and reuse it here.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -436,6 +436,19 @@ public void updateProperties(final IdentifierConverter<Resource, FedoraResource>
...
+                                     LOGGER.info("********451: has exists\n");
```
awoods: Please remove this logging statement.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -436,6 +436,19 @@ public void updateProperties(final IdentifierConverter<Resource, FedoraResource>
...
+                                    throw new IllegalArgumentException( "prefix has existed");
```
awoods: Per @ajs6f's earlier comment, please create a new Exception type for this scenario.
For examples see: https://github.com/fcrepo4/fcrepo4/tree/master/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/exception

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -436,6 +436,19 @@ public void updateProperties(final IdentifierConverter<Resource, FedoraResource>
...
+
+                                   }
+                                  } catch (RepositoryException e) {
+                                    LOGGER.info("********catch " + e);
```
awoods: What is this error case? Do we really want to be swallowing the exception?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
ajs6f: That better not be a reference to [HTTP 451](https://en.wikipedia.org/wiki/HTTP_451).

awoods:
Please create an integration test that demonstrates the issue before your functional updates, and proves the fix after your functional updates.
You can probably add a test method to:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/test/java/org/fcrepo/integration/kernel/modeshape/FedoraResourceImplIT.java

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/FedoraInvalidPrefixExceptionMapper.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.ws.rs.ext.Provider;
+
+import org.fcrepo.kernel.api.exception.FedoraInvalidPrefixException;
+
+import org.slf4j.Logger;
+
+/**
+ * For invalid namespace exceptions on CRUD actions for nodes/datastreams
+ *
+ * @author whikloj
```
awoods: `@author` should be `nianma`

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/805712449d8677a218faa490901a449fec3011e0

PR #922, opened 2015-10-26
update max metaspace size

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1796

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -539,7 +539,7 @@
       <plugin>
         <artifactId>maven-failsafe-plugin</artifactId>
         <configuration>
-          <argLine>-XX:MaxMetaspaceSize=128m ${jacoco.agent.it.arg}</argLine>
+          <argLine>-XX:MaxMetaspaceSize=512m ${jacoco.agent.it.arg}</argLine>
         </configuration>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -539,7 +539,7 @@
       <plugin>
         <artifactId>maven-failsafe-plugin</artifactId>
         <configuration>
-          <argLine>-XX:MaxMetaspaceSize=128m ${jacoco.agent.it.arg}</argLine>
+          <argLine>-XX:MaxMetaspaceSize=512m ${jacoco.agent.it.arg}</argLine>
```
ajs6f: The last time I wanted to up the memory for `failsafe`, @fasseg was of the vociferous opinion that it wold be difficult for people to build Fedora with a half-gig for tests. I disagreed, but we left it alone then. You might want to ping the `-tech` list letting them know this is happening, in case anyone wants to react.

acoburn: I picked 512m because that's the value currently in use for the (surefire) unit tests: see line 561 below. That is, unless I'm misreading this, it seems fcrepo already requires .5 GB for the tests (and has for some time).

PR #931, opened 2015-10-30
Allow an admin user to specify a delegated user.

peichman-umd:
The delegation is specified in an HTTP Header. The header is configured using a separate HttpHeaderPrincipalProvider bean named delegatedPrincipalProvider. If it is configured and a matching header is present in the request, the ServletContainerAuthenticationProvider will replace the userPrincipal in the request with the first principal found in the header.

- Added a getFirstPrincipal() convenience method to the HttpHeaderPrincipalProvider.
- Added unit tests for the delegated user functionality of the ServletContainerAuthenticationProvider. Check that admins can delegate but regular users cannot.
- Added unit test for the getFirstPrincipal method of HttpHeaderPrincipalProvider.
- Added an integration test with an On-Behalf-Of header and a mock FAD.

Resolves: https://jira.duraspace.org/browse/FCREPO-1790

### DIFF
--- a/fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/ServletContainerAuthenticationProvider.java
+++ b/fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/ServletContainerAuthenticationProvider.java
@@ -39,7 +39,7 @@ import org.slf4j.LoggerFactory;
  * @author Gregory Jansen
  */
 public final class ServletContainerAuthenticationProvider implements
-        AuthenticationProvider {
+AuthenticationProvider {
 
     private static ServletContainerAuthenticationProvider instance = null;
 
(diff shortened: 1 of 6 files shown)


fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/ServletContainerAuthenticationProvider.java
```diff
@@ -39,7 +39,7 @@
  * @author Gregory Jansen
  */
 public final class ServletContainerAuthenticationProvider implements
-        AuthenticationProvider {
+AuthenticationProvider {
```
awoods: rollback formatting change

peichman-umd:
Needed by fcrepo4/fcrepo-module-auth-webac#49

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/b807b2a4a414adf49b712df4a474edb4b6015e35

PR #932, opened 2015-10-31
removes fcrepo-mint module (to be moved to fcrepo4-exts)

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1696

fcrepo-mint will be added to the fcrepo4-exts organization and be avaiilable as an extension module.


fcrepo-webapp/src/main/resources/spring/master.xml
```diff
@@ -3,14 +3,13 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="
     http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-3.0.xsd">
-    
+
     <!-- Master context for fcrepo4. -->
 
     <import resource="${fcrepo.spring.repo.configuration:classpath:/spring/repo.xml}"/>
     <import resource="${fcrepo.spring.rest.configuration:classpath:/spring/rest.xml}"/>
-    <import resource="${fcrepo.spring.minter.configuration:classpath:/spring/minter.xml}"/>
```
escowles: Removing the minter.xml config file results in falling back on the default impl. in the `HierarchicalIdentifierSupplier` interface.  This is fine and preserves the default behavior.

But it also makes it easier to customize the minting.  Instead of being able to customize the number and length of path prefixes with env. variables, you would now have to compile a separate package, create a minter config file to configure it, and then replace/override one of the core config files to include it.

Can we make it easier to override the default minting behavior?  We probably don't want to complicate the default impl, but maybe we can at least make it easier to include custom local config?  E.g., we could have an empty "local" config file to provide an easy place to put custom config.

acoburn: This is an important point: we should certainly make it easy to customize minting behavior. What would you think about adding the minter module and related configuration to `fcrepo-webapp-plus`?

escowles: Including the minter in `fcrepo-webapp-plus` would be good.  We could use an updated version of the minter.xml to use the default impl, but show how to swap in `UUIDPathMinter` or another minter.

awoods: It sounds like https://github.com/fcrepo4-exts/fcrepo-webapp-plus/pull/30 should be updated based on this discussion, no?

acoburn: Yes, it should. I'll add a PR to that project, too.

awoods: @acoburn, a related PR already exists. You may just want to update this one: https://github.com/fcrepo4-exts/fcrepo-webapp-plus/pull/30

acoburn: Yes, I'm working on updating that branch now.

PR #934, opened 2015-11-01
remove mixinTypes, primaryType and hasAccessRoles from RDF representation

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1765

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraExportIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraExportIT.java
@@ -20,10 +20,9 @@ import static javax.ws.rs.core.Response.Status.CONFLICT;
 import static javax.ws.rs.core.Response.Status.CREATED;
 import static javax.ws.rs.core.Response.Status.NO_CONTENT;
 import static javax.ws.rs.core.Response.Status.OK;
-import static com.hp.hpl.jena.graph.NodeFactory.createLiteral;
 import static com.hp.hpl.jena.graph.Node.ANY;
 import static com.hp.hpl.jena.graph.NodeFactory.createURI;
-import static org.fcrepo.kernel.api.FedoraJcrTypes.FEDORA_PAIRTREE;
+import static com.hp.hpl.jena.vocabulary.RDF.type;
 import static org.fcrepo.kernel.api.RdfLexicon.REPOSITORY_NAMESPACE;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -68,7 +67,7 @@ public class FedoraExportIT extends AbstractResourceIT {
         try (final CloseableGraphStore graphStore = getGraphStore(new HttpGet(serverAddress + pairtreeName))) {
             assertTrue("Resource \"" + objName + " " + pairtreeName + "\" must be pairtree.",
                     graphStore.contains(ANY, createURI(serverAddress + pairtreeName),
-                            createURI(REPOSITORY_NAMESPACE + "mixinTypes"), createLiteral(FEDORA_PAIRTREE)));
+                            type.asNode(), createURI(REPOSITORY_NAMESPACE + "Pairtree")));
         }
         testRoundtrip(pairtreeName);
     }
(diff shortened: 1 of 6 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraExportIT.java
```diff
@@ -68,7 +67,7 @@ public void shouldRoundTripOnePairtree() throws IOException {
         try (final CloseableGraphStore graphStore = getGraphStore(new HttpGet(serverAddress + pairtreeName))) {
             assertTrue("Resource \"" + objName + " " + pairtreeName + "\" must be pairtree.",
                     graphStore.contains(ANY, createURI(serverAddress + pairtreeName),
-                            createURI(REPOSITORY_NAMESPACE + "mixinTypes"), createLiteral(FEDORA_PAIRTREE)));
+                            type.asNode(), createURI(REPOSITORY_NAMESPACE + "Pairtree")));
```
ajs6f: Maybe this should be a constant.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -938,7 +937,7 @@ public void testGetRepositoryGraph() throws IOException {
         try (final CloseableGraphStore graph = getGraphStore(getObjMethod(""))) {
             logger.trace("Retrieved repository graph:\n" + graph);
             assertTrue("Should find the root type", graph.contains(ANY,
-                    ANY, HAS_PRIMARY_TYPE.asNode(), createLiteral(ROOT)));
+                    ANY, type.asNode(), createURI(MODE_NAMESPACE + "root")));
```
ajs6f: Why are we exposing _anything_ in the MODE namespace? Why not either a) just get rid of it and let root-ness be inferred from the lack of a parent, or b) offer our own type?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersionsIT.java
```diff
@@ -148,7 +148,7 @@ public void testAddAndRetrieveVersion() throws IOException {
             logger.debug("Got version profile:");
 
             assertTrue("Didn't find a version triple!", versionResults.contains(ANY,
-                    ANY, HAS_PRIMARY_TYPE.asNode(), createLiteral("nt:frozenNode")));
+                    ANY, type.asNode(), createURI(JCR_NT_NAMESPACE + "frozenNode")));
```
ajs6f: Same deal as above: why publish _anything_ in a JCR namespace?

acoburn: I'm OK with filtering out *all* JCR-namespaced properties.

ajs6f: :+1:

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
acoburn: There is no `fedora:Root` property in the ontology. AFAIK, `mode:root` is the only MODE-namespaced property. Should we add a `fedora:Root` property to the ontology and have the modeshape impl map `mode:root` to `fedora:Root`?

escowles: :+1: -- having a fedora:Root type seems like a good way to explicitly mark the root node.  I would also be fine with just omitting it (lacking a parent is also a good implicit marker of root-ness).

acoburn:
@ajs6f I think much of this code will be combined with what is currently https://github.com/fcrepo4/fcrepo4/pull/935. That combined PR will also clean much of the filter-sprawl that is currently spread across `fcrepo-kernel-api` and `fcrepo-kernel-modeshape`

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
ajs6f: We could add a type (not a property). I don't know that we even need to do that. The inference is pretty straightforward. If you are curious as to whether a thing is the repo root, check to see if it has a parent. If so, no, it isn't. If not, yes, it is.

acoburn: yes, I meant `rdfs:Class`. For now, though, just eliminating the `mode:root` property seems like a big improvement. If we want to explicitly add it to the ontology, that's good, too, but the key thing would be to eliminate the modeshape namespace.

ajs6f: Agreed. I say remove it, and if we discover a good reason to so do, we can always add a type.

acoburn:
Closed in favor of #935

PR #936, opened 2015-11-02
Filtering out server-managed triples about hash URI fragments

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-1800

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -141,6 +141,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
 
     private static final Predicate<Triple> IS_MANAGED_TYPE = t -> t.getPredicate().equals(type.asNode()) &&
             isManagedNamespace.test(t.getObject().getNameSpace());
+    private static final Predicate<Triple> IS_MANAGED_TRIPLE = IS_MANAGED_TYPE.or(isManagedTriple::test);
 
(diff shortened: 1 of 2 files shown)


ajs6f:
Can you explain a little about why this is the fix, @escowles ? (Actually, maybe an in-line comment or two would be good...?)

escowles:
@ajs6f The predicates for testing whether triples are server-managed, and the logic for including various kinds of triples based on Prefer headers is already in `ContentExposingResource`, so it seemed like the logical place to implement the decision that server-managed triples should never be included for hash URI resources.  I could definitely add a comment that we never want server-managed triples for hash URI resources, even if we want them for containers.

The other approach I considered was doing the filtering in the triple generating machinery.  But I'm wary of over-complicating that already somewhat baroque code, and the logic of the decision seemed to fit better in the HTTP layer.

ajs6f:
No, I'm sorry, @escowles , I meant at the micro-level: why does ` -rdfStream.concat(filter(getTriples(HashRdfContext.class), tripleFilter::test));
+rdfStream.concat(filter(getTriples(HashRdfContext.class), IS_MANAGED_TRIPLE.negate()::test));`

do the job?

escowles:
Oh, I see: we are avoiding the logic surrounding the server-managed triple Prefer header and always rejecting them.  I'll add a comment to that effect.

ajs6f:
Right, right. Thanks!

escowles:
@ajs6f Anything else, or is this ready to merge?

ajs6f:
No, I'm grruvy. Want me to merge?

escowles:
Yes, please!

PR #938, opened 2015-11-03
add guard for non-literal titles

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1811

### DIFF
--- a/fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/responses/ViewHelpersTest.java
+++ b/fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/responses/ViewHelpersTest.java
@@ -250,6 +250,7 @@ public class ViewHelpersTest {
     @Test
     public void shouldUseTheObjectUriIfATitleIsNotAvailable() {
         final Graph mem = createDefaultModel().getGraph();
+        mem.add(new Triple(createURI("a/b/c"), DC_TITLE.asNode(), createURI("d/e/f")));
 
         assertEquals("a/b/c", testObj.getObjectTitle(mem, createURI("a/b/c")));
 


fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
```diff
@@ -212,7 +212,10 @@ public String getObjectTitle(final Graph graph, final Node subject) {
             final Iterator<Triple> objects = getObjects(graph, subject, p);
 
             if (objects.hasNext()) {
-                return objects.next().getObject().getLiteral().getLexicalForm();
+                final Node object = objects.next().getObject();
+                if (object.isLiteral()) {
+                    return object.getLiteral().getLexicalForm();
+                }
```
ajs6f: If `!object.isLiteral()` is there a sensible return value?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
```diff
@@ -212,7 +212,10 @@ public String getObjectTitle(final Graph graph, final Node subject) {
             final Iterator<Triple> objects = getObjects(graph, subject, p);
```
ajs6f: Maybe just `GraphUtil::listObjects` instead of our own special code?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
acoburn: I believe the current default is to use the URI of the node.

ajs6f: Okay, that seems pretty reasonable.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
acoburn: Honestly, a lot of the code in this class could be improved, but this PR is limited to the particular bug mentioned in the ticket.

ajs6f: Make a ticket to fix up this class and I will try and get around to it.

acoburn: here's a ticket for cleaning up the class: https://jira.duraspace.org/browse/FCREPO-1813

acoburn:
@ajs6f after rebasing, there's not much left to this PR -- just a unit test verifying that non-literal titles don't  throw errors.

ajs6f:
That's a little troubling. That implies that my PR for `ViewHelpers` changed the semantics of that class without us noticing, doesn't it?

acoburn:
@ajs6f not really -- the `.filter(Node::isLiteral)` you added did exactly what the original PR did. Incidentally, that's exactly how I _wanted_ to implement it, but the semantics of the class made that difficult, hence #941

acoburn:
@ajs6f to add onto that, one issue with the `ViewHelpers` class was that there was no test for non-Literal titles. That's what this (one-line) change adds.

ajs6f:
Oh, well, what the hell...

PR #939, opened 2015-11-03
HttpHeaderInjection to include headers optional modules

whikloj:
Address [FCREPO-1795](https://jira.duraspace.org/browse/FCREPO-1795)

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/HttpHeaderInjection.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/HttpHeaderInjection.java
@@ -1,0 +1,80 @@
+/**
+ * Copyright 2015 DuraSpace, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.http.commons.api;
+
+import static org.slf4j.LoggerFactory.getLogger;
+
+import java.util.Map;
+import javax.servlet.http.HttpServletResponse;
+import javax.ws.rs.core.UriInfo;
+
+import org.fcrepo.kernel.api.models.FedoraResource;
+
+import org.slf4j.Logger;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.stereotype.Component;
+
+import com.google.common.collect.ListMultimap;
+
+/**
+ * Inject optional headers from external processes
+ *
+ * @author whikloj
+ * @since 2015-10-30
+ */
+@Component
+public class HttpHeaderInjection implements ApplicationContextAware {
+
+    private static final Logger LOGGER = getLogger(HttpHeaderInjection.class);
+
+    private ApplicationContext applicationContext;
+
+    @Override
+    public void setApplicationContext(final ApplicationContext applicationContext) {
+        this.applicationContext = applicationContext;
+    }
+
+    /**
+     * Add additional Http Headers
+     *
+     * @param servletResponse the response
+     * @param uriInfo the URI context
+     * @param resource the resource
+     */
+    public void addHttpHeaderToResponseStream(final HttpServletResponse servletResponse, final UriInfo uriInfo,
+            final FedoraResource resource) {
+        LOGGER.debug("Adding additional HTTP headers to stream");
+
+        for (final String h : applicationContext.getBeanDefinitionNames()) {
+            LOGGER.debug("Found bean name: {}", h);
+        }
+        getUriAwareHttpHeaderFactories().forEach((bean, factory) -> {
+            LOGGER.debug("Adding HTTP headers using: {}", bean);
+            final ListMultimap<String, String> h =
+                    factory.createHttpHeadersForResource(uriInfo, resource);
+
+            h.entries().forEach((final Map.Entry<String, String> entry) -> {
+                servletResponse.addHeader(entry.getKey(), entry.getValue());
+            });
+        });
+    }
+
+    private Map<String, UriAwareHttpHeaderFactory> getUriAwareHttpHeaderFactories() {
+        return applicationContext
+                .getBeansOfType(UriAwareHttpHeaderFactory.class);
+    }
+}
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/UriAwareHttpHeaderFactory.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/UriAwareHttpHeaderFactory.java
@@ -1,0 +1,42 @@
+/**
+ * Copyright 2015 DuraSpace, Inc.
+ *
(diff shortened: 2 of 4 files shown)


whikloj:
Needed by https://github.com/fcrepo4/fcrepo-module-auth-webac/pull/48

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/HttpHeaderInjection.java
```diff
@@ -0,0 +1,80 @@
...
+import com.google.common.collect.ListMultimap;
+
+/**
+ * Inject optional headers from external processes
+ *
+ * @author whikloj
+ * @since 2015-10-30
+ */
+@Component
+public class HttpHeaderInjection implements ApplicationContextAware {
```
ajs6f: Just a small thing, but wouldn't this read better as `HttpHeaderInjector`?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/HttpHeaderInjection.java
```diff
@@ -0,0 +1,80 @@
...
+            final ListMultimap<String, String> h =
+                    factory.createHttpHeadersForResource(uriInfo, resource);
+
+            h.entries().forEach((final Map.Entry<String, String> entry) -> {
+                servletResponse.addHeader(entry.getKey(), entry.getValue());
+            });
+        });
+    }
+
+    private Map<String, UriAwareHttpHeaderFactory> getUriAwareHttpHeaderFactories() {
```
ajs6f: Can't this be factored up along with `HttpTripleUtil::getUriAwareTripleFactories`?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/UriAwareHttpHeaderFactory.java
```diff
@@ -0,0 +1,42 @@
...
+public interface UriAwareHttpHeaderFactory {
+
+    /**
+     * Given a resource and session, update the JAX-RS response with any additional headers.
+     *
+     * @param uriInfo the UriInfo
+     * @param resource the resource
+     * @return
+     */
+    public ListMultimap<String, String> createHttpHeadersForResource(UriInfo uriInfo,
```
ajs6f: Does this need to be `ListMultimap` or can it just be `Multimap`?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/HttpHeaderInjection.java
```diff
@@ -0,0 +1,80 @@
...
+        for (final String h : applicationContext.getBeanDefinitionNames()) {
+            LOGGER.debug("Found bean name: {}", h);
+        }
+        getUriAwareHttpHeaderFactories().forEach((bean, factory) -> {
+            LOGGER.debug("Adding HTTP headers using: {}", bean);
+            final ListMultimap<String, String> h =
+                    factory.createHttpHeadersForResource(uriInfo, resource);
+
+            h.entries().forEach((final Map.Entry<String, String> entry) -> {
+                servletResponse.addHeader(entry.getKey(), entry.getValue());
```
ajs6f: Why `final Map.Entry<String, String>`? Should the type be inferred?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/HttpHeaderInjection.java
ajs6f: Prefer subclassing `ApplicationObjectSupport` to impl'ing `ApplicationContextAware`.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/HttpHeaderInjection.java
whikloj: To confirm the big words. You're suggesting a superclass for HttpTripleUtil and HttpHeaderInjection that has the getUriAwareFactories?

whikloj: ...and would I also need to super-"interface" the UriAwareHttpHeaderFactory and UriAwareResourceModelFactory interfaces. So I could return them from a common function?

whikloj: Okay, after looking into this I think not. Mostly because this function requires the interface class to grab and I'm not sure about how to implement this with a generic interface that is "extended" by each different implementation and each of these return completely different data, HttpTripleUtils returns RDF and HttpHeaderInjector returns a Map of HTTP Headers. So are different other than the fact I learned how to do this from existing code. 

If you can point to some help, I am willing to try and implement though.

whikloj:
Sorry @ajs6f , I switched back to implementing ApplicationContextAware to avoid the NullPointerExceptions occurring (seemingly) because this only pulls from the external WebAC module currently.

ajs6f:
It's just a plain generic. The bounds don't matter at all. You have one method: `T create(final FedoraResource resource,  final UriInfo uriInfo, final IdentifierConverter<Resource,FedoraResource> idTranslator)` in the factory types (each of which subtypes `UriAwareFactory<T>`)  and one method in the generic supertype to the injectors: `UriAwareFactory<T> getUriAwareFactories()`.

ajs6f:
If you have resolved the injection problems, why not switch back to `@Optional`? Why are you having NPEs?

whikloj:
I did switch back to [[@Optional]], I still get a NPE when it tries to perform the getBeansOfType. I assume because there are none?

As for the generics, would we filter for the type in each utility? Because I am guessing both the HttpTripleUtil and the HttpHeaderInjector would grab each others classes, but we would only want to execute our own to ensure we get the correct output (RDF vs headers).

ajs6f:
No, that shouldn't be giving you an NPE, at least not the way I understand it now. It should work the same way as does the injection of `httpTripleUtil`.

ajs6f:
You could filter, or I _think_ you could use the type of `T` available in the superclass.

whikloj:
Alright, the NPE is because I was not instantiating the ApplicationObjectSupport correctly. So that is solved.

ajs6f:
MAXIMIZE THE SOLVED!

whikloj:
Forgive my ignorance, but if ```UriAwareHttpHeaderFactory``` is an interface and ```UriAwareResourceModelFactory``` is an interface. Would I generate an "parent" interface with no methods defined?

ajs6f:
No, I'm saying you try to give each factory a supertype, e.g. `UriAwareFactory<R>`, where `R` is `Model` in one case and `Multimap<String, String>` in the other, and it contains one method `R create(final FedoraResource resource, final UriInfo uriInfo, final IdentifierConverter<Resource,FedoraResource> idTranslator)` so that `createHttpHeadersForResource` => `@Override Multimap<String, String> create(...)` and `createModelForResource` => `@Override Model create(...)`.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/UriAwareHttpHeaderFactory.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @author whikloj
+ * @since 2015-11-01
+ */
+public interface UriAwareHttpHeaderFactory {
+
+    /**
+     * Given a resource and session, update the JAX-RS response with any additional headers.
+     *
+     * @param uriInfo the UriInfo
```
awoods: Please provide even more meaningful javadocs: `uriInfo`, `resource`, and `return`.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/097714241c927bdc4ce9613bd6bec17dbb4d5b67

PR #940, opened 2015-11-03
Removing FedoraVersions instance variable that was hiding superclass instance

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-1702


ajs6f:
@escowles Is that IT change just more precision testing the same thing? No change in the semantic of the test?

escowles:
The IT change is making sure the subject of the versioned fcr:metadata response is the binary (new behavior), instead of .../fcr:metadata (the old behavior).

ajs6f:
Oh, cool. Want me to merge?

PR #935, opened 2015-11-01
suppress JCR types from RDF representation

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1541

PR #941, opened 2015-11-03
Code cleanup in ViewHelpers

ajs6f:
https://jira.duraspace.org/browse/FCREPO-1813

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
@@ -15,20 +15,25 @@
  */
 package org.fcrepo.http.commons.responses;
 
-import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.common.collect.Lists.newArrayList;
-import static com.hp.hpl.jena.graph.Node.ANY;
+import static com.hp.hpl.jena.graph.GraphUtil.listObjects;
 import static com.hp.hpl.jena.graph.NodeFactory.createURI;
 import static com.hp.hpl.jena.rdf.model.ResourceFactory.createResource;
 import static com.hp.hpl.jena.rdf.model.ResourceFactory.createProperty;
+import static java.util.Arrays.asList;
+import static java.util.Arrays.stream;
+import static java.util.Collections.emptyMap;
 import static java.util.stream.Collectors.joining;
+import static java.util.stream.Collectors.toMap;
 import static org.fcrepo.kernel.api.RdfLexicon.CREATED_DATE;
 import static org.fcrepo.kernel.api.FedoraJcrTypes.FCR_METADATA;
 import static org.fcrepo.kernel.api.RdfLexicon.DC_TITLE;
+import static org.fcrepo.kernel.api.RdfLexicon.HAS_PRIMARY_TYPE;
 import static org.fcrepo.kernel.api.RdfLexicon.DCTERMS_TITLE;
 import static org.fcrepo.kernel.api.RdfLexicon.HAS_VERSION_LABEL;
 import static org.fcrepo.kernel.api.RdfLexicon.RDFS_LABEL;
 import static org.fcrepo.kernel.api.RdfLexicon.SKOS_PREFLABEL;
+import static org.fcrepo.kernel.api.RdfLexicon.WRITABLE;
 import static org.fcrepo.kernel.api.RdfLexicon.HAS_VERSION;
 import static org.fcrepo.kernel.api.RdfLexicon.RDF_NAMESPACE;
 import static org.fcrepo.kernel.api.RdfLexicon.DC_NAMESPACE;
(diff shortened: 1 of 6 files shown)


fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
```diff
@@ -182,48 +159,32 @@ public String getVersionLabel(final Graph graph,
...
-        } else if (subject.isBlank()) {
-            return subject.getBlankNodeLabel();
-        } else {
-            return subject.toString();
-        }
-
+        final Optional<String> title = TITLE_PROPERTIES.stream().map(Property::asNode).flatMap(p -> listObjects(
+                graph, sub, p).toList().stream()).filter(Node::isLiteral).map(Node::getLiteral).map(
+                        LiteralLabel::toString).findFirst();
+        return title.orElse(sub.isURI() ? sub.getURI() : sub.isBlank() ? sub.getBlankNodeLabel() : sub.toString());
```
acoburn: personally, I find it easier to read if the expressions are on separate lines:

    list.stream()
           .map(func1)
           .flatMap(func2)
           .collect(func3);

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
```diff
@@ -262,10 +218,7 @@ public boolean isWritable(final Graph graph, final Node subject) {
      * @return whether the subject is frozen node
      */
     public boolean isFrozenNode(final Graph graph, final Node subject) {
-        final Iterator<Triple> objects = getObjects(graph, subject, RdfLexicon.HAS_PRIMARY_TYPE);
-        return objects.hasNext()
-                && objects.next().getObject()
-                .getLiteralValue().toString().equals("nt:frozenNode");
+        return getValue(graph, subject, HAS_PRIMARY_TYPE.asNode()).filter("nt:frozenNode"::equals).isPresent();
```
acoburn: Will this conflict with the changes in #935 ? After that is merged, the notion of a FrozenNode will be mostly hidden outside of the modeshape impl, but you will be able to test for the fedora:Version type.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
acoburn: Does this make #938 unnecessary? I'm fine with that, but there is a unit test in that PR (for titles containing URIs) that should be pulled in.

ajs6f: I'm definitely not of the opinion that line-per-call is more readable, but this is a matter of style.

ajs6f: Hm, does look like a conflict with #938. Can you rebase over this?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
ajs6f: Hm, I don't really know-- I guess this test should be changed to match what's in #935? Can you send a PR for that?

ajs6f: Is it just a change in type to `fedora:Version`?

acoburn: yes, it's a change to fedora:Version.

ajs6f: Okay, if I do that here, what test will fail?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
acoburn: w/r/t #938 it seems that merging this PR first and rebasing #938 over this works for me.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
acoburn: I actually suspect that there is no test for that, since I wasn't getting test failures with #935 
If we merge this PR first, I can amend #935 to test this -- it will probably involve changing that function somewhat.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
ajs6f: Okay, cool. I'll wait to pull the trigger on #938, which otherwise looks good to me

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
ajs6f: Okay, I will leave this alone, and wait to hear from you that you are satisfied with #935.

PR #942, opened 2015-11-04
Removing ContentLocation which just reiterated the file path and leaked JCR impl. details

escowles:
* Also using List instead of Set for predictable triple ordering (makes output RDF prettier)

Fixes https://jira.duraspace.org/browse/FCREPO-1669


acoburn:
nice. :+1:

PR #943, opened 2015-11-05
remove JCR properties from root node context

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1819

This also removes the descriptors for the Modeshape/Infinispan cluster configuration, which renders the `GetCacheManager` and `GetClusterConfiguration` completely unnecessary. So they are removed, too.


ajs6f:
Big time :+1:

acoburn:
@ajs6f a somewhat related question -- there are some lines (currently commented out) related to :

    FIXME: removing due to performance problems, esp. w/ many files on federated filesystem

     b.add(create(subject(), HAS_OBJECT_COUNT.asNode(), createLiteral(String
                     .valueOf(getRepositoryCount(repository)))));
     b.add(create(subject(), HAS_OBJECT_SIZE.asNode(), createLiteral(String
                     .valueOf(getRepositorySize(repository)))));

Should those be removed, too?

ajs6f:
@awoods @escowles @cbeer What do you think? People want that info, but it is damn hard, and I think we decided that with federation, nearly impossible.

escowles:
+1 to removing the object count/size here.  This would be a great thing to include in webapp-plus, so people who weren't using federation could add it in.

ajs6f:
@escowles Are you talking about actually modularizing "all-repo" stats so that people could plug in what they want?

escowles:
I was thinking we could literally plug in those two lines of code as an optional module.  They work fine if you aren't using federation.  Of course,people could plug in alternate implementations if they wanted to.

Of course, we could do this in fcrepo4: if there is an appropriate bean injected, then add those properties, otherwise just skip them.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/RootRdfContext.java
```diff
@@ -77,40 +68,21 @@ public RootRdfContext(final FedoraResource resource,
...
             b.add(create(subject(), HAS_OBJECT_COUNT.asNode(), createLiteral(String
                     .valueOf(getRepositoryCount(repository)))));
             b.add(create(subject(), HAS_OBJECT_SIZE.asNode(), createLiteral(String
                     .valueOf(getRepositorySize(repository)))));
         */
 
-        // Get the cluster configuration, if available
-        // this ugly test checks to see whether this is an ordinary JCR
-        // repository or a ModeShape repo, which will possess the extra info
-        if (JcrRepository.class.isAssignableFrom(repository.getClass())) {
```
awoods: I believe this cluster information is still valuable... especially with the recent flurry of interest in clustering.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/functions/GetCacheManager.java
```diff
@@ -1,52 +0,0 @@
...
-import org.infinispan.manager.DefaultCacheManager;
-import org.modeshape.jcr.JcrRepository;
-import org.slf4j.Logger;
-
-/**
- * Get the internal Infinispan CacheManager from a Repository
- *
- * @author acoburn
- */
-class GetCacheManager implements Function<Repository, DefaultCacheManager> {
```
awoods: Why are we dropping this class?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/functions/GetClusterConfiguration.java
```diff
@@ -1,86 +0,0 @@
...
-import org.modeshape.jcr.value.binary.infinispan.InfinispanBinaryStore;
-import org.slf4j.Logger;
-
-/**
- * Extract the Infinispan cluster configuration and state
- * from a running Modeshape repository
- * @author Gregory Jansen
- * @since Apr 26, 2013
- */
-public class GetClusterConfiguration implements Function<Repository, Map<String, String>> {
```
awoods: Why are we dropping this class?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/RootRdfContext.java
ajs6f: The info may be valuable, bu that does not mean it should be published here. Perhaps in logging statements.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/functions/GetCacheManager.java
ajs6f: Sans clustering info we don't need it.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/functions/GetClusterConfiguration.java
ajs6f: See above.

ajs6f:
@escowles I would prefer an entirely separate module. I wouldn't mind shoving the Metrics-produced stats in there too.

awoods:
Normally I do not like dead code, but in this case, it continues to serve as a good reminder. I would suggest leaving this commented-out block around until we actually split the logic into a new, separate module.

acoburn:
@awoods ok, then I will leave this PR as it currently stands; it is ready for review.

escowles:
@awoods: I've created a ticket for an external module for repository object/byte counts: https://jira.duraspace.org/browse/FCREPO-1820

ajs6f:
I think this is ready to go. Any objections?

PR #948, opened 2015-11-12
trying to read build properties.

nianma:
This is only for sharing, not for merging.

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
@@ -83,6 +83,12 @@ public final class RdfLexicon {
             "http://www.w3.org/ns/sparql-service-description#";
 
     /**
+     * DOAP namespace.
+     */
+    public static final String DOAP_NAMESPACE =
+            "http://usefulinc.com/ns/doap#";
+
+    /**
      * The namespaces that the repository manages internally.
      */
     public static final Set<String> managedNamespaces = of(
@@ -316,6 +322,10 @@ public final class RdfLexicon {
 
     private static Predicate<Property> hasFedoraNamespace =
         p -> !p.isAnon() && p.getNameSpace().startsWith(REPOSITORY_NAMESPACE);
+
+    // DOAP release decribes fedora build information
+    public static final Property DOAP_RELEASE =
+            createProperty(DOAP_NAMESPACE + "release");
 
     /**
      * Detects whether an RDF property is managed by the repository.
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/BuildPropertiesLoader.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/BuildPropertiesLoader.java
@@ -42,18 +42,21 @@ public class BuildPropertiesLoader {
     private String buildRevision;
 
     /**
-     * This method loads build System Properties
+     * This method loads build Properties
      */
 
     public BuildPropertiesLoader() {
 
         final Properties buildProperties = new Properties();
-        try (final InputStream pStream = getClass().getResource("/build-info.properties").openStream()) {
+        try (final InputStream pStream = getClass().getResourceAsStream("/build-info.properties")) {
             buildProperties.load(pStream);
 
         } catch (final IOException e) {
-           LOGGER.info("IOException for property file");
-           e.printStackTrace();
+           LOGGER.info("IOException for build-info.properties file");
+        //   e.printStackTrace();
+        buildName = buildProperties.getProperty(BUILD_NAME_KEY, "unknown");
+        buildDate = buildProperties.getProperty(BUILD_DATE_KEY, "unknown");
+        buildRevision = buildProperties.getProperty(BUILD_REVISION_KEY, "unknown");
         }
 
         buildName = buildProperties.getProperty(BUILD_NAME_KEY, "unknown");
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/BuildPropertiesLoader.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/BuildPropertiesLoader.java
@@ -1,0 +1,67 @@
+/**
+ * Copyright 2015 DuraSpace, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.fcrepo.kernel.modeshape.rdf.impl;
+
+/**
+ *
+ * @author Nianli Ma
+ *         Date: 11/11/15
+ */
+
+import org.slf4j.Logger;
+import static org.slf4j.LoggerFactory.getLogger;
+import java.util.Properties;
+import java.io.InputStream;
+import java.io.IOException;
+import java.io.FileNotFoundException;
+
+public class BuildPropertiesLoader {
+
+    private static final Logger LOGGER = getLogger(BuildPropertiesLoader.class);
+
+    /**
+     * This method loads build System Properties
+     */
+
+    public void loadSystemProperties() {
+
+        final Properties buildProperties = new Properties();
+        try (final InputStream pStream = getClass().getResource("/build-info.properties").openStream()) {
+            buildProperties.load(pStream);
+            final String projectVersion = buildProperties.getProperty("version");
+            if (projectVersion != null ) {
+                LOGGER.debug("project version is {}", projectVersion);
+                setProperty("version", projectVersion);
+            } else {
+                setProperty("version", "Unknown");
+            }
+
+        } catch (final FileNotFoundException e) {
+           LOGGER.info("File not Found");
+           //set unknown value for all the build properties here
+           setProperty("version", "Unknown");
+        } catch (final IOException e) {
+           LOGGER.info("IOException for property file");
+           e.printStackTrace();
+        }
+    }
+
+    private static void setProperty(final String prop, final String propValue) {
+        System.setProperty(prop, propValue);
+    }
+
+}
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/BuildPropertiesLoader.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/BuildPropertiesLoader.java
@@ -48,15 +48,16 @@ public class BuildPropertiesLoader {
     public BuildPropertiesLoader() {
 
         final Properties buildProperties = new Properties();
-        try (final InputStream pStream = getClass().getResourceAsStream("/build-info.properties")) {
-            buildProperties.load(pStream);
 
+        try (final InputStream pStream = getClass().getResourceAsStream("/build-info.properties")) {
+          if (pStream != null) {
+            buildProperties.load(pStream);
+          }
         } catch (final IOException e) {
            LOGGER.info("IOException for build-info.properties file");
-        //   e.printStackTrace();
-        buildName = buildProperties.getProperty(BUILD_NAME_KEY, "unknown");
-        buildDate = buildProperties.getProperty(BUILD_DATE_KEY, "unknown");
-        buildRevision = buildProperties.getProperty(BUILD_REVISION_KEY, "unknown");
+           buildName = buildProperties.getProperty(BUILD_NAME_KEY, "unknown");
+           buildDate = buildProperties.getProperty(BUILD_DATE_KEY, "unknown");
+           buildRevision = buildProperties.getProperty(BUILD_REVISION_KEY, "unknown");
         }
 
         buildName = buildProperties.getProperty(BUILD_NAME_KEY, "unknown");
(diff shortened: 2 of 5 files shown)


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
```diff
@@ -317,6 +323,10 @@
     private static Predicate<Property> hasFedoraNamespace =
         p -> !p.isAnon() && p.getNameSpace().startsWith(REPOSITORY_NAMESPACE);
 
+    // DOAP release decribes fedora build information
```
awoods: May as well move this property up closer to the `DOAP namespace`... instead of down here at the bottom of the file.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/BuildPropertiesLoader.java
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * This method loads build Properties
+     */
+
+    public BuildPropertiesLoader() {
+
+        final Properties buildProperties = new Properties();
+
+        try (final InputStream pStream = getClass().getResourceAsStream("/build-info.properties")) {
```
awoods: Define `/build-info.properties` as a private static class field.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/RootRdfContext.java
```diff
@@ -129,7 +132,25 @@ private void concatRepositoryTriples() {
                     createTypedLiteral(counters.get(PREFIX + FIXITY_REPAIRED_COUNTER).getCount()).asNode()));
         }
 
+        //add project build description triples here
+        final BuildPropertiesLoader buildLoader = new BuildPropertiesLoader();
+
+        final com.hp.hpl.jena.graph.Node releaseSubject = createURI(REPOSITORY_NAMESPACE + "release");
+            b.add(create(subject(), DOAP_RELEASE.asNode(), releaseSubject));
+            b.add(create(releaseSubject, type.asNode(), createURI(DOAP_NAMESPACE + "Version")));
```
awoods: The subject of all /root triples should be the /root itself, not `REPOSITORY_NAMESPACE + "release"`

ajs6f: True, so maybe we need to make resources available to describe the releases, hosted maybe at fedora.info or somewhere. Where else would we make the assertion above, which, while it doesn't belong in the response at the repository root, is correct and useful?

awoods: Would a simpler approach be reasonable:
```
            b.add(create(subject(), createURI(DOAP_NAMESPACE + "name"),
                         createLiteral(buildLoader.getBuildName())));
            b.add(create(subject(), createURI(DOAP_NAMESPACE + "created"),
                         createLiteral(buildLoader.getBuildDate())));
            b.add(create(subject(), createURI(DOAP_NAMESPACE + "revision"),
                         createLiteral(buildLoader.getBuildRevision())));
```

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/BuildPropertiesLoader.java
```diff
@@ -0,0 +1,91 @@
...
+          if (pStream != null) {
+            buildProperties.load(pStream);
+          }
+        } catch (final IOException e) {
+           LOGGER.info("IOException for build-info.properties file");
```
awoods: Do not swallow this exception, but instead throw a `RepositoryRuntimeException`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/RootRdfContext.java
ajs6f: I'm not sure those triples are actually correct.I think you are conflating the repository software and this instance thereof. This repository instance wasn't created on `buildLoader.getBuildDate()`, the _software_ was. But triples with the repository root resource as their subject describe this repository _instance_, not the software.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/BuildPropertiesLoader.java
```diff
@@ -0,0 +1,91 @@
...
+           buildName = buildProperties.getProperty(BUILD_NAME_KEY, "unknown");
```
awoods: Remove these three property definitions.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/RootRdfContext.java
awoods: Agreed. Would a hash URI (`#release` or the like) make sense.

ajs6f: We could do that. I would really much rather we just publish a description for each release as part of the release workflow (just some N-Triples hosted somewhere) and just link to it here. That seems much cleaner to me.

awoods:
@nianma, I would suggest adding two integration test methods:
1. In `fcrepo-kernel-modeshape::FedoraResourceImplIT` that checks that the release triples are all set to `unknown`
2. In `fcrepo-webapp::SanityCheckIT` that checks that the release triples are all NOT set to `unknown`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/RootRdfContext.java
awoods: The question should be posed to @coblej whether an external resource with triples would serve the intent of this ticket:
https://jira.duraspace.org/browse/FCREPO-1774

awoods: On further thought, one of the primary benefits of the build-number (git commit number) and build-date (mvn install date) is for debugging snapshot and branch releases, as well as ensuring that a repository instance is on the expected commit.
I suspect we will not be automating the creation of external RDF documents for each github commit.

ajs6f: I didn't realize that you were getting that granular. Actually, I think you very well could automate that, as long as you are willing to point to Github resources, but I don't care that much.

acoburn: I would think that a `doap.rdf` file for the Fedora project would be very useful here. Would adding one be part of a separate ticket or part of this ticket?

awoods: @acoburn, what exactly do you have in mind with "doap.rdf file"?

acoburn: I mean that many open source projects contain a `doap.rdf` file (Description Of A Project). E.g. https://github.com/apache/storm/blob/master/doap_Storm.rdf, https://github.com/apache/kafka/blob/trunk/doap_Kafka.rdf. https://github.com/ruby-rdf/rdf/blob/develop/etc/doap.ttl

Since that file typically contains these sorts of values (major/minor version, etc), and since that file is always an RDF resource, it could be a convenient resource to link to from the repository.

awoods: Yes, that seems like a separate issue from the topic of this ticket... unless such a doap.rdf would be created for each github commit.

awoods: To be clear, currently, every time the command `mvn install` is run three pieces of information are dynamically created and added to the HTML splash page:
* `project.version` e.g. 4.4.1-SNAPSHOT
* `build.timestamp` e.g. 2015-11-29
* `buildNumber` e.g. 80571244

This ticket is to provide that same information in a machine-readable fashion... hence the initial suggestion of triples on the repository root resource.

ajs6f: Is that actually what the ticket is asking for? In my reading it's just asking for major-minor-bugfix version numbers. Not individual build numbers nor timestamps.

awoods: @coblej?

coblej: For our purposes, the major-minor-bugfix version numbers should be sufficient.  No particular objection to more granular information but something like '4.4.1' should suffice for our use case.

awoods: Thanks, @coblej. That said, could a usable solution be to link off to a remote doap.rdf source? or does the version number need to be directly accessible from a repository GET request?

fcrepo-webapp/src/main/resources/build-info.properties
```diff
@@ -0,0 +1,3 @@
+build.name=${project.version}
+build.date=${build.timestamp}
+build.revision=${buildNumber}
```
acoburn: I really don't think we should use a `properties` file for this. The version information is already included in the `MANIFEST.MF` file and it can be retrieved like so:

    this.getClass().getPackage().getImplementationVersion();

The build date and build revision are not typically part of the manifest file, and if they are not needed, I would omit them. IMO, we would also define and use the `SpecificationVersion`, which would correspond to the version of `fcrepo-kernel-api` (the `ImplementationVersion` would be the version of `fcrepo-kernel-modeshape`) -- that would involve updating `./fcrepo-parent/pom.xml` to include this line in the bundle plugin:

    <Specification-Version>${project.version}</Specification-Version>

Once the kernel-api is completely separated from the implementation, the version of fcrepo-kernel-api could be different than `${project.version}`, but for now, they are the same.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/RootRdfContext.java
coblej: @awoods I was certainly anticipating it would be directly accessible from the repository software using a GET request (as it was in Fedora 3).  Linking off to a remote doap.rdf source would probably not work for us since one of our repository applications runs on a protected network where linking out to a remote site is problematic.
cc: @dchandekstark

awoods: Circling back on this ticket, it seems that the solution that is being sought is a simple predicate on the /root of the repository that holds the property of the Fedora release version number. Given the apparent lack of suitable existing terms ([doap](http://usefulinc.com/ns/doap) properties seems to have a range of doap:Project or doap:Version which would require new Fedora resources) would it instead make sense to create a new fedora:hasReleaseNumber property to http://fedora.info/definitions/v4/repository# ?

escowles: @awoods a third option would be to use a hash URI resource:

```
</rest> doap:release </rest#version> .
</rest#version> a doap:Version; doap:revision "4.4.3" .
```

awoods: @escowles, we are circling a bit in this conversation:
https://github.com/fcrepo4/fcrepo4/pull/948/files#r46093452
with response from @ajs6f of:
https://github.com/fcrepo4/fcrepo4/pull/948/files#r46093474

escowles: @awoods Right.  @ajs6f's suggestion of an external document isn't what the user is after, so I'm in favor of going back to a hash URI.

awoods: @escowles, to be clear, you would prefer a hash URI over a simple, new fedora namespaced property?

escowles: @awoods yes, using an existing ontology for this seems much better than minting our own properties to me.

awoods: The following suggested solution sounds reasonable to me:
https://github.com/fcrepo4/fcrepo4/pull/948/files#r48074015

@ajs6f ? @coblej ? @acoburn ?

coblej: @awoods If it provides the desired functionality, then I am fine with it.  I don't really feel knowledgable enough to comment directly at the implementation level.

awoods: Thanks, @coblej. Thoughts, @ajs6f? @acoburn?

acoburn: :+1: to @escowles' suggestion to use an existing ontology (`doap`) and hash URIs

Surfrdan:
It seems this was never meant to be merged and is marked wont fix. Closing

PR #953, opened 2015-11-13
fcr:backup error response

yinlinchen:
-Resolved: https://jira.duraspace.org/browse/FCREPO-1788


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NamespaceTools.java
```diff
@@ -75,7 +75,13 @@ public static void validatePath(final Session session, final String path) {
                     try {
                         namespaceRegistry.getURI(prefix);
                     } catch (final NamespaceException e) {
-                        throw new FedoraInvalidNamespaceException("Prefix " + prefix + " has not been registered", e);
+                        if (segment.indexOf("fcr:backup") >= 0) {
```
awoods: Please change to `segment.equals()` instead of `segment.indexOf()`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NamespaceTools.java
```diff
@@ -75,7 +75,13 @@ public static void validatePath(final Session session, final String path) {
...
+                            throw new FedoraInvalidNamespaceException(
```
awoods: Please create a new exception type, `BackupException`, and throw that instead of a `FedoraInvalidNamespaceException`.

awoods:
Please add an integration test that fails before this update and passes after your PR.

PR #958, opened 2015-11-28
Add 'Limit' header to LDP GET

awoods:
Limit HTML UI to 99 children
Add 'numberOfChildren' triple to containers

Resolves: https://jira.duraspace.org/browse/FCREPO-1808

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -152,6 +152,12 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
 
     protected Response getContent(final String rangeValue,
                                   final RdfStream rdfStream) throws IOException {
+        return getContent(rangeValue, -1, rdfStream);
+    }
+
+    protected Response getContent(final String rangeValue,
+                                  final int limit,
+                                  final RdfStream rdfStream) throws IOException {
         if (resource() instanceof FedoraBinary) {
 
             final String contentTypeString = ((FedoraBinary) resource()).getMimeType();
@@ -185,7 +191,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
             }
 
         } else {
-            rdfStream.concat(getResourceTriples());
+            rdfStream.concat(getResourceTriples(limit));
             if (prefer != null) {
                 prefer.getReturn().addResponseHeaders(servletResponse);
             }
@@ -196,6 +202,10 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
     }
 
     protected RdfStream getResourceTriples() {
+        return getResourceTriples(-1);
+    }
+
+    protected RdfStream getResourceTriples(final int limit) {
         // use the thing described, not the description, for the subject of descriptive triples
         if (resource() instanceof NonRdfSourceDescription) {
             resource = ((NonRdfSourceDescription) resource()).getDescribedResource();
@@ -237,7 +247,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
 
             // containment triples about this resource
             if (ldpPreferences.prefersContainment()) {
-                rdfStream.concat(getTriples(ChildrenRdfContext.class));
+                rdfStream.concat(getTriples(ChildrenRdfContext.class).limit(limit));
             }
 
             // LDP container membership triples for this resource
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -155,6 +155,15 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
         return getContent(rangeValue, -1, rdfStream);
     }
 
+    /**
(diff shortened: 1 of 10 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -152,6 +152,12 @@
 
     protected Response getContent(final String rangeValue,
                                   final RdfStream rdfStream) throws IOException {
+        return getContent(rangeValue, -1, rdfStream);
+    }
+
+    protected Response getContent(final String rangeValue,
+                                  final int limit,
```
ajs6f: Javadocs. What is the unit for `limit`?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
```diff
@@ -247,6 +248,25 @@ public String getObjectsAsString(final Graph graph,
...
 
     /**
+     *
+     * @param graph
+     * @param subject
+     * @return
+     */
+    public int getNumChildren(final Graph graph, final Node subject) {
+        LOGGER.trace("Getting number of children: s:{}, g:{}", subject, graph);
+        final Iterator<Node> iterator = listObjects(graph, subject, HAS_CHILD_COUNT.asNode());
```
ajs6f: Maybe just use `GraphUtils::getStringValue`.

awoods: The pattern between the javascript and ViewHelpers is to pass in a `graph` vs. a `Resource`, which is required by `GraphUtils::getStringValue`. If you have a pull-request against this branch that uses `GraphUtils.getStringValue`, I would be happy to merge it into this work. Otherwise, I would suggest moving forward with this impl as is.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
```diff
@@ -247,6 +248,25 @@ public String getObjectsAsString(final Graph graph,
     }
 
     /**
+     *
+     * @param graph
```
awoods: Javadoc is incomplete

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
ajs6f: Then better something like
```
Iterator<String> count = listObjects(graph, subject, HAS_CHILD_COUNT.asNode()).filterKeep(Node::isLiteral).mapWith(Node::getLiteralLexicalForm);
return count.hasNext() ? Integer.parseInt(count.next()) : 0;
```

instead of several times the length of `if-then`, but even better to switch the pattern. Also, that property should be internally typed, and there should be no need to cast into `Integer`, but that's another story.

awoods: This looks like a good suggestion; however, it would appear that `Node` needs to be a functional interface to for `.filterKeep(Node::isLiteral)` to compile. Suggestions?
Otherwise, I would suggest moving forward with this impl as is.

ajs6f: No, I have no problem compiling that, and it's perfectly legal Java. Something may be wrong with your IDE.

awoods: Hmm. I wonder what I have wrong?
https://gist.github.com/awoods/784a14da703c0b1c6c87
Thoughts?

ajs6f: Probably Fedora's failure to get to Jena 3. That seems to be what causes all problems now.

awoods: Does this compile for you, @ajs6f?

ajs6f: Does what compile?

awoods: The question is, does the linked "diff.txt" patch compile for you, @ajs6f? Also in that link is the output I see when running `mvn clean install -pl fcrepo-http-commons`.
https://gist.github.com/awoods/784a14da703c0b1c6c87

ajs6f: I already told you it did, but I'm using Jena 3. I wouldn't even bother working tickets like this one until that upgrade is in.

awoods:
@acoburn, I believe this ticket is ready to go. @ajs6f, you are not suggesting holding this PR on the non-functional, syntactic polish of Java8/Jena3, are you?
p.s. These two commits should be squashed.

acoburn:
merged via 03adc0b85166c2e16c280fd183b2b97252210a7c

PR #960, opened 2015-12-01
Update REST API for specifying checksums

Rarian:
Initially clients could specify checksums for uploaded binary content
via a custom `checksum` query parameter, instead of this non-standard
approach, we want to refactor to follow RFC-3230 4.3.2.

* Adds `Warning` header to announce deprecation of `checksum`
* Adds support for `Digest` header as per RFC-3230
* Adds utility function for parsing `Digest` headers
* Adds integration tests for checksum match and mismatch
* Adds documentation for deprecation of `checksum`
* Adds `TODO` hints for removal of deprecated code

Resolves: https://jira.duraspace.org/browse/FCREPO-1468


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -235,16 +240,42 @@ public Response deleteObject() {
...
+            throws InvalidChecksumException, MalformedRdfException {
+        return createOrReplaceObjectRdf(requestContentType, requestBodyStream,
+            checksum, contentDisposition, ifMatch, link, null);
+    }
+
+    /**
+     * Create a resource at a specified path, or replace triples with provided RDF.
+     * @param requestContentType the request content type
+     * @param requestBodyStream the request body stream
+     * @param checksumDeprecated the deprecated checksum value
```
ajs6f: This needs a better comment. Why deprecated? How to avoid it? What to prefer?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -399,15 +437,44 @@ public Response updateSparql(@ContentLocation final InputStream requestBodyStrea
...
+                                 @HeaderParam("Slug") final String slug,
+                                 @ContentLocation final InputStream requestBodyStream,
+                                 @HeaderParam("Link") final String link)
+            throws InvalidChecksumException, IOException, MalformedRdfException, AccessDeniedException {
+        return createObject(checksum, contentDisposition, requestContentType, slug, requestBodyStream, link, null);
+    }
+    /**
+     * Creates a new object.
+     *
+     * application/octet-stream;qs=1001 is a workaround for JERSEY-2636, to ensure
```
ajs6f: Please leave a `TODO` in the code for this, and open a ticket.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -645,4 +725,25 @@ private void checkLinkForLdpResourceCreation(final String link) {
...
+     * Parse the RFC-3230 Digest response header value.  Look for a
+     * sha1 checksum and return it as a urn, if missing an empty string is
+     * returned.
+     * @param digest The Digest header value
+     * @return the sha1 checksum value
+     */
+    private String parseDigestHeader(final String digest) {
+        if (digest != null) {
+            final Map<String,String> digestPairs = RFC3230_SPLITTER.split(digest);
+            for (Map.Entry<String, String> digestPair : digestPairs.entrySet()) {
```
ajs6f: Prefer `Map::forEach` for readability.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -645,4 +725,25 @@ private void checkLinkForLdpResourceCreation(final String link) {
...
+                switch(digestPair.getKey().toLowerCase()) {
```
ajs6f: Is this with some expectation that there will be other choices in the future?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -829,6 +830,47 @@ public void testIngestWithBinary() throws IOException {
...
+     */
+    @Test
+    public void testIngestWithBinaryAndChecksum() throws IOException {
+        final HttpPost method = postObjMethod();
+        final File img = new File("src/test/resources/test-objects/img.png");
+        method.addHeader("Content-Type", "application/octet-stream");
+        method.addHeader("Digest", "SHA1=f0b632679fab4f22e031010bd81a3b0544294730");
+        method.setEntity(new FileEntity(img));
+
+        try (final CloseableHttpResponse response = execute(method)) {
```
ajs6f: Prefer [the simpler method](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java#L221).

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -829,6 +830,47 @@ public void testIngestWithBinary() throws IOException {
...
+     */
+    @Test
+    public void testIngestWithBinaryAndChecksumMismatch() throws IOException {
+        final HttpPost method = postObjMethod();
+        final File img = new File("src/test/resources/test-objects/img.png");
+        method.addHeader("Content-Type", "application/octet-stream");
+        method.addHeader("Digest", "SHA1=fedoraicon");
+        method.setEntity(new FileEntity(img));
+
+        try (final CloseableHttpResponse response = execute(method)) {
```
ajs6f: See above.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
Rarian: The RFC specification allows for sending multiple digests.  `Digest: md5=6668675a91f39ca1afe46c084e8406ba,sha1=f0b632679fab4f22e031010bd81a3b0544294730` is a valid header we would want to support, even if we're only looking for the sha1.

ajs6f: Sure, the question is why this is a `switch` and not a simple `if`, which would be more readable, or even just a `digestPairs.stream()::filter` invocation.

Rarian: Fair enough, I was thinking we may have wanted to keep track of submitted hashes with an eye to extendability, but this could definitely be simplified if we aren't thinking about tracking anything but sha1.

ajs6f: Oh, I'm not sure of that myself.It was an honest question. I find bit fixity confusing and depressing, like every other desperate attempt to stab the beating heart of time and lay it cold and still, so I try to stay away from it. If we have any plans to add acceptable hashes, it's okay by me, but we should have a comment here to that effect. Otherwise, yeah, maybe just the simpler conditional.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
Rarian: Right, forgot we could use features from Java 8.

ajs6f: Wrap a pretty lambda about your shoulders, then flail it wildly through the air like Isadora Duncan with a scarf.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
escowles: I think it would be great to support other checksums.  We could calculate other checksums and check them too.  And we could also add the additional checksums as properties on the binary.  So I'm fine with leaving some of the framework in place for handling additional checksum algorithms, and adding a new ticket for adding that support.

awoods: Here is the ticket from the community: https://jira.duraspace.org/browse/FCREPO-1842

ruebot: :+1: on supporting additional checksums. Thinking here of the use case of migrating fcrepo3 to fcrepo4, and having a variety of checksums algorithms implemented.

ajs6f: Works for me, just as long as we document it, and a ticket is a great way. That ticket could feature a pointer to this place in the code, to help whoever works it.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
Rarian: Let me know if the new commit addresses this in an appropriate fashion.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
Rarian: Alrighty, I'll back out my .stream()::filter lambda in favour of forEach, with a TODO about supporting additional checksums and a reference to 1842.

Rarian: Actually, now that I think about it the code as currently written won't handle multiple checksums cleanly as currently envisioned.  
Should parseDigestHeader take a second parameter that specifies the hash to return?
Should parseDigestHeader return a Map of Digests => Hashes?
Should I leave the code as is, as of 3880db5?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -645,4 +736,20 @@ private void checkLinkForLdpResourceCreation(final String link) {
...
+    /**
+     * Parse the RFC-3230 Digest response header value.  Look for a
+     * sha1 checksum and return it as a urn, if missing or malformed
+     * an empty string is returned.
+     * @param digest The Digest header value
+     * @return the sha1 checksum value
+     */
+    private String parseDigestHeader(final String digest) {
+        final Map<String,String> digestPairs = RFC3230_SPLITTER.split(nullToEmpty(digest));
+        final String hash = digestPairs.entrySet().stream()
```
ajs6f: Just use `findFirst` here, to return an `Optional<String>`. Then below, use `Optional::orElse` to create your defaulting behavior. There's no need to create then check for the blank string.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -645,4 +736,20 @@ private void checkLinkForLdpResourceCreation(final String link) {
...
+                    .filter(s -> s.getKey().toLowerCase().equals("sha1"))
+                    .map(Map.Entry::getValue)
+                    .findFirst()
+                    .orElse("");
+        return StringUtils.isBlank(hash) ? "" : "urn:sha1:" + hash;
```
ajs6f: Like `return hash.map("urn:sha1:"::concat).orElse("")`.

Rarian: That is definitely more elegant.  I'll update that as soon as there is a [consensus as to the direction to go with this function](https://github.com/fcrepo4/fcrepo4/pull/960#discussion_r46338824).

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
awoods: Before deciding on how or if the `parseDigestHeader` signature should change, what is the expected behavior of this PR (noting that there is another [ticket](https://jira.duraspace.org/browse/FCREPO-1842) to allow for alternate checksums)?
To be more specific, what do we expect the response to be for the following client requests:
1. POST -H"Digest: md5=some-md5" /resource
2. POST -H"Digest: md5=some-md5,SHA1=some-sha1" /resource
3. POST -H"Digest: sha1=some-sha1" /resource

Rarian: My thoughts in originally writing this were that this original basic implementation would respond to those headers in the following fashion:
# No hash retained
# SHA1 retained
# SHA1 retained

And that when work is done on [FCREPO-1842](https://jira.duraspace.org/browse/FCREPO-1842) this code could be refactored in such a way as to best suit the particulars of the chosen implementation of multiple digest support.

awoods: Therefore, it seems like the answer to your questions: https://github.com/fcrepo4/fcrepo4/pull/960#discussion_r46338824

...is "leave the code generally as is, as of 3880db5 (made slightly more elegant)?"

Rarian: Excellent, elegance hopefully upgraded in new commit.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
Rarian: Thanks for the pointers, I don't routinely write Java so haven't had much of an opportunity to play with features from 8.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -645,4 +737,19 @@ private void checkLinkForLdpResourceCreation(final String link) {
...
+     * @param digest The Digest header value
+     * @return the sha1 checksum value
+     */
+    private String parseDigestHeader(final String digest) {
+        final Map<String,String> digestPairs = RFC3230_SPLITTER.split(nullToEmpty(digest));
+        final Optional<String> hash = digestPairs.entrySet().stream()
+                    .filter(s -> s.getKey().toLowerCase().equals("sha1"))
+                    .map(Map.Entry::getValue)
+                    .findFirst();
+        return hash.map("urn:sha1:"::concat).orElse("");
```
acoburn: You can even go further and simply `.map(...).orElse("")` off of the `.findFirst()` line:

    .findFirst()
    .map("urn:sha1:"::concat)
    .orElse("");

Then you don't even need to define the `hash` value or import `Optional`.

Rarian: I... should have seen that.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/4aefe3e8f9a9541a3b8ca7ba222b8ebd9853f634

PR #962, opened 2015-12-04
Replace Iterator-based RdfStream with Java 8 Streams

acoburn:
This is an implementation of a context-bearing RdfStream class using the Java8 stream API. The intention here is that this will replace the existing `org.fcrepo.kernel.api.utils.iterators.RdfStream` class.

Relates to: https://jira.duraspace.org/browse/FCREPO-1312

The design of this implementation addresses two related issues. First, to make `RdfStream` behave like a Java 8 `Stream` rather than an `Iterator`.

And secondly, to make the classes that implement these RDF streams (`org.fcrepo.kernel.modeshape.rdf.impl.*`) isolated to the `fcrepo-kernel-modeshape` module. The current architecture places a hard dependency on `fcrepo-kernel-modeshape` from the `fcrepo-http-*` modules, which would preclude any alternate implementation. This PR goes most of the way toward severing that direct dependency. It does this by declaring an RdfContext enum in the `fcrepo-kernel-api` module. The values defined there can be passed into the `FedoraResource::getTriples` method, which are handled by the implementing (e.g. modeshape) module.


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfStreamContext.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /**
+     * Create an RdfStreamContext from an existing Model.
+     *
+     * @param model the model
+     * @return the corresponding RdfStreamContext
+     */
+    public static RdfStreamContext fromModel(final Model model) {
+        return new RdfStreamContext(StreamSupport.stream(spliteratorUnknownSize(model.listStatements(), NONNULL), false)
```
ajs6f: Just `Iter.asStream`.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfStreamContext.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /**
+     * A namespace-aware {@link Collector}
+     * @return a Collector for use with aggregating the Triple stream.
+     */
+    public Collector<Triple, ?, Model> toModel() {
+        final Model model = createDefaultModel();
+        model.setNsPrefixes(namespaces());
+        return Collector.of(() -> model,
```
ajs6f: This isn't bad, but it might make more sense to just roll up a `Collection` of `Statements` and add it to a `Model`

ajs6f:
Eh. I guess this works. I don't like the fact that the graph is an internal member. I don't think we have a good understanding of the nature of identity or equality for this type. But at least it gets us moving forward.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfStreamContext.java
acoburn: I don't think that's present in jena 2.13.x. I know it's in 3.x.

ajs6f: I thought we went to 3 a while ago?!

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfStreamContext.java
acoburn: You mean drop this method and have users of the class simply use `Collectors.toList()` (and manage the `Model::add` themselves)? That would work, too, but seemed to me that this method would lead to less code overall.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfStreamContext.java
acoburn: Nope: https://jira.duraspace.org/browse/FCREPO-1672

I'm working on that, but I'm stuck on how string literals get mapped to/from the JCR layer w/r/t their Datatype. It must be a simple thing that I'm overlooking, but if you'd like to take over that ticket, you'd be welcome to.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfStreamContext.java
ajs6f: No, I mean have something like what is there now, which returns the `Model` directly. What you have here is flexible to a fault. I could mistakenly do:
```
     Stream<Triple> triples1 = ... ;
     Stream<Triple> triples2 = ... ;
     triples1.stream().collect(triples2.toModel());
```
Frankly, now might be a good time to just drop namespace-prefix management from this layer of the code entirely.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfStreamContext.java
ajs6f: No, I'm on "vacation" with my new second child. If you can point me to a specific test, I could try to make it pass, but I'm not up for an epic, saga, book-length sequence of cantos, or other long-form piece of poetry.

acoburn:
I also don't like that the stream is an internal member, but java8 streams don't carry state. And if we need things like `topic()` and `namespaces()`, then the class will need to carry state with it and push the stream into a private member.

In my opinion, the entire `RdfStream` type would be as simple as:

    interface RdfStream extends Stream<Triple> { }

As for identity, what exactly _is_ identity in the context of a `Stream`? I certainly don't think there is any reasonable way to express that.

ajs6f:
This is why I want a subtype of `Stream<Triple>`, which _could_ carry state. `::namespaces` I would rather see gone and `::session` is a impl detail which should go, but `::topic` is _not_ incidental. It is crucial.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfStreamContext.java
acoburn: I would be strongly in favor of dropping namespace management from this. A better impl of `toModel` would be a static function with a usage pattern like this:

     Stream<Triple> triples1 = ... ;
     Stream<Triple> triples2 = ... ;
     triples1.stream().collect(RdfStreamContext.toModel(namespaces));

ajs6f: If you want to drop them, and I want to drop them, then why not drop them?

I agree that what you show immediately above is better, but I'm still not sure why you want to produce a `Collector` instead of a `Model`-valued instance method.

acoburn: Very well, they will be dropped. As for why a `Collector`, that's because `Stream::collect` requires one, which fits the java8 `Stream` model better than creating our own thing.

ajs6f: No one is arguing for "our own thing". My suggestion is an instance method on this class that directly returns a `Model`.

acoburn: I could compromise on a static method. An instance method that consumes an internal `Stream` is problematic if that stream has already been operated on:

    RdfStreamContext rdf = new RdfStreamContext(...);
    rdf.stream().skip(3);
    Model model = rdf.toModel(); // IllegalStateException: stream has already been operated upon or closed

Alternately, this would be fine (as a static method):

    RdfStreamContext rdf = new RdfStreamContext(...);
    Model model = rdf.toModel(rdf.stream().skip(3));

ajs6f: These are other good examples of why this internal stream thing doesn't work for me.

acoburn:
Updated based on suggestions. This still uses an internal `Stream`, and the `Node` topic is now a necessary part of the constructor.

ajs6f:
I guess I like this more, now, but I still don't think it captures the notion. I would much rather see a subtype of `Stream<Triple>`. There is some grossness here:
```
RdfStreamContext triples = ...;
out.println("The first triple is: " + triples.stream().findFirst().orElse("not there!"));
Stream<Triple> stillThere = triples.stream();
assert stillThere.count() == 0;
// explode!
```
The `findFirst` is terminal on the internal stream, but leaves behind a "shell" `RdfStreamContext`. What exactly _is_ that shell? For all practical purposes, it's the same thing as a "post-terminal" `Stream`. I don't like that. It's correct behavior given the Streams API, but it couldn't happen if `RdfStreamContext` < `Stream<Triple>`.

acoburn:
@ajs6f  `RdfStream` is now defined as an interface with a context-bearing member. I think this captures what we both want. (of course, there is no implementation yet, but it's a start)

ajs6f:
:+1:

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/RdfContext.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+package org.fcrepo.kernel.api.rdf;
+
+/**
+ * A collection of RDF contexts that can be used to extract triples from FedoraResources
+ *
+ * @author acoburn
+ * @since Dec 4, 2015
+ */
+public enum RdfContext {
```
ajs6f: Is this a bad coupling in the type system? If I add a new algorithm for generating triples, do I have to add a class and a member here? Seems like we only need the class…

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/utils/GraphDifferencer.java
```diff
@@ -0,0 +1,104 @@
...
+import static com.hp.hpl.jena.graph.Node.ANY;
+import static com.hp.hpl.jena.sparql.graph.GraphFactory.createDefaultGraph;
+
+/**
+ * A wrapping {@link Stream} that calculates two differences between a
+ * {@link Graph} A and a source Stream B. The differences are (A - (A ∩ B)) and
+ * (B - (A ∩ B)). The ordinary output of this stream is (B - (A ∩ B)), and
+ * after exhaustion, sets containing (A - (A ∩ B)) and (A ∩ B) are available.
+ *
+ * @author ajs6f
```
ajs6f: Add `@author acoburn`

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/DefaultRdfStream.java
```diff
@@ -0,0 +1,321 @@
...
+import com.hp.hpl.jena.rdf.model.Model;
+import com.hp.hpl.jena.rdf.model.Statement;
+
+/**
+ * Implementation of a context-bearing RDF stream
+ *
+ * @author acoburn
+ * @since Dec 6, 2015
+ */
+public class DefaultRdfStream implements RdfStream {
```
ajs6f: Can we factor out a `WrappingStream<T>` here?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/RdfStream.java
```diff
@@ -13,12 +13,23 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+package org.fcrepo.kernel.api.rdf;
```
ajs6f: Why `api.rdf`? Why not just `org.fcrepo.kernel.api`? It's one of the most fundamental types we use.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/StreamUtils.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Convert an Iterator to a Stream
+     *
+     * @param <T> the type of the Stream
+     * @param iterator the iterator
+     * @param parallel whether to parallelize the stream
+     * @return the stream
+     */
+    @SuppressWarnings("unchecked")
```
ajs6f: Why is this a type safety issue?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/StreamUtils.java
```diff
@@ -0,0 +1,59 @@
...
+public class StreamUtils {
+
+    /**
+     * Convert an Iterator to a Stream
+     *
+     * @param iterator the iterator
+     * @param <T> the type of the Stream
+     * @return the stream
+     */
+    @SuppressWarnings("unchecked")
```
ajs6f: Why is this a type safety issue?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/iterators/PersistingRdfStreamConsumer.java
```diff
@@ -118,25 +118,30 @@ public PersistingRdfStreamConsumer(final IdentifierConverter<Resource, FedoraRes
...
     }
 
     @Override
     public void consume() throws MalformedRdfException {
-        while (stream.hasNext()) {
-            final Statement t = m.asStatement(stream.next());
-            LOGGER.debug("Operating triple {}.", t);
+        stream.forEach(t -> {
+            final Statement s = m.asStatement(t);
+            LOGGER.debug("Operating triple {}.", s);
```
ajs6f: `Operating on triple`

ajs6f:
I don't understand the advantage of enumerating the algorithms for generating triples.

acoburn:
The entire reason for creating this enum is so that the fcrepo-http-* modules don't reference fcrepo-modeshape classes directly. Otherwise, (and this is the current situation) it would not be possible to reuse the fcrepo-http-* modules with a non modeshape impl.


On Feb 15, 2016, at 8:52 AM, A. Soroka <notifications@github.com<mailto:notifications@github.com>> wrote:


I don't understand the advantage of enumerating the algorithms for generating triples.

ajs6f:
Can't we just get rid of the ability to select different triple generation algorithms? Just have a simple `getTriples()` without parameters?

acoburn:
And how would the prefer headers work? Especially the inbound refs and/or child resources? Fetch all triples all the time and then filter? Wouldn't there be a significant performance penalty for that?


On Feb 15, 2016, at 9:07 AM, A. Soroka <notifications@github.com<mailto:notifications@github.com>> wrote:


Can't we just get rid of the ability to select different triple generation algorithms? Just have a simple getTriples() without parameters?

ajs6f:
If the concern is the `Prefer` header, why doesn't the new `enum` contain direct matches to the available values for that header?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/VersionsRdfContext.java
```diff
@@ -72,37 +73,35 @@
...
     @SuppressWarnings("unchecked")
-    private Iterator<Triple> versionTriples() throws RepositoryException {
-        final Iterator<Version> allVersions = versionHistory.getAllVersions();
-        return Iterators.concat(Iterators.transform(allVersions, version2triples::apply));
+    private Stream<Triple> versionTriples() throws RepositoryException {
+        return iteratorToStream(versionHistory.getAllVersions()).flatMap(version2triples);
     }
 
-    private final Function<Version, Iterator<Triple>> version2triples = new Function<Version, Iterator<Triple>> () {
+    private final Function<Version, Stream<Triple>> version2triples = new Function<Version, Stream<Triple>> () {
```
ajs6f: Why not use lambda notation here?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/DefaultRdfStream.java
```diff
@@ -0,0 +1,321 @@
...
+        this.stream = stream;
+    }
+
+    /**
+     * Create an RdfStream from an existing Model.
+     * @param model An input Model
+     * @return a new RdfStream object
+     */
+    public static RdfStream fromModel(final Model model) {
+        return fromModel((Node)null, model);
```
awoods: `(Node)null` casting is unnecessary.

ajs6f: Do we even want to have a method that creates a stream with no topic?With hindsight, I would much rather we made the topic `final

awoods: +1 to "all streams must have a non-null topic".

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpTripleUtil.java
```diff
@@ -59,17 +59,19 @@ public void setApplicationContext(final ApplicationContext applicationContext) {
...
             final FedoraResource resource, final UriInfo uriInfo,
             final IdentifierConverter<Resource,FedoraResource> idTranslator) {
 
         LOGGER.debug("Adding additional HTTP context triples to stream");
-        getUriAwareTripleFactories().forEach((bean, factory) -> {
-            LOGGER.debug("Adding response information using: {}", bean);
-            final Model m = factory.createModelForResource(resource, uriInfo, idTranslator);
-            rdfStream.concat(fromModel(m));
-        });
+        final Map<String, UriAwareResourceModelFactory> foo = getUriAwareTripleFactories();
```
awoods: `foo` is never used... can probably remove this line.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/DefaultRdfStream.java
ajs6f: No more billion dollar mistakes!

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/DefaultRdfStream.java
```diff
@@ -0,0 +1,321 @@
...
+    public DefaultRdfStream(final Node node) {
+        this(node, empty());
+    }
+
+    /**
+     * Create an RdfStream
+     *
+     * @param stream a stream of triples
+     */
+    public DefaultRdfStream(final Stream<Triple> stream) {
```
ajs6f: Maybe get rid of this constructor?

awoods: +1, it appears to only be used once, and topic is known.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/DefaultRdfStream.java
```diff
@@ -0,0 +1,321 @@
...
+        this.node = null;
+        this.stream = stream;
+    }
+
+    /**
+     * Create an RdfStream
+     * @param node the topic of the stream
+     * @param stream the incoming stream
+     */
+    public DefaultRdfStream(final Node node, final Stream<Triple> stream) {
```
ajs6f: Let's put `Objects::requireNonNull` here!

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/DefaultRdfStream.java
```diff
@@ -0,0 +1,321 @@
...
+ */
+public class DefaultRdfStream implements RdfStream {
+
+    protected Stream<Triple> stream;
+    protected final Node node;
+
+    /**
+     * Create an empty RdfStream
+     */
+    public DefaultRdfStream() {
```
ajs6f: I say kill this-- I will bet even money that anywhere it is used `Stream::empty` could be used instead.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/DefaultRdfStream.java
acoburn: I'm all for that, too

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/RdfStream.java
acoburn: Using the `org.fcrepo.kernel.api` package for `RdfStream` makes sense. What about the other Classes/Interfaces in the `api.rdf` package? Would you recommend putting all of those into the `org.fcrepo.kernel.api` package? (That would include `DefaultRdfStream`, `RdfContext` and `RdfCollectors`.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/DefaultRdfStream.java
acoburn: Yes, a `WrappingStream<T>` makes a lot of sense. Should that go into the `org.fcrepo.kernel.api.utils` package?

ajs6f: Yeah, prolly. I'm starting to think we need to step back and take a quick look at the whole package structure here.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/RdfStream.java
ajs6f: Erm, mebbe. See my note [here](https://github.com/fcrepo4/fcrepo4/pull/962/files#diff-d21cc8996ca7ecac4305348e5792e256R55). I think `RdfContext` and `RdfCollectors` are more like API, but not `DefaultRdfStream`. Sort of.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/DefaultRdfStream.java
acoburn: +1 on evaluating the package structure. Currently there are some embedded loops that should be removed (api.exception references api.models and vice-versa), but that is beyond the scope of this ticket.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/RdfStream.java
acoburn: OK, I will move RdfStream, RdfContext and RdfCollectors into the org.fcrepo.kernel.api package, but leave DefaultRdfStream in api.rdf for now. We should have a larger conversation about package naming and organization.

acoburn:
Initial code review comments addressed

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfContext.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+package org.fcrepo.kernel.api;
+
+/**
+ * A collection of RDF contexts that can be used to extract triples from FedoraResources
+ *
+ * @author acoburn
+ * @since Dec 4, 2015
+ */
+public enum RdfContext {
```
ajs6f: This is more what I would be hoping for. @awoods , @escowles , @barmintor can I get you to glance at this? Is this in line with our prospective moves towards developing versioning, fixity, etc. as extensions to LDP? I would think that anything that develops from a Prefer header value or as an extension based on LDP type would show up here in some reasonably explicit way.

I'm still not happy about having to expose this stuff using an `enum` instead of polymorphism, but I don't want to drag us around and around with that right now. The question I want to be a little more sure about is whether we are capturing here the "flavors" of RDF that any Fedora impl would have to be able to supply, to the best of our current ability.

awoods: I suspect we may need to iterate on the completeness of this `enum` based on feedback from the spec-writing exercises and associated alt-F4-implementations, but that should not hold-up this solid initial cut.

I hope to finish my review of this entire ticket today.

barmintor: I don't have a lot of context, but 1) If this is supposed to be a complete list, shouldn't the minimal container Prefer value be here? 2) Doesn't using an enum here fairly sharply limit the reusability of the code? 3) Strictly speaking the Prefer header's 'include' and 'omit' params are multivalued lists, so I hope this enum is used in a way that supports that?

awoods: @barmintor, what do you mean by #2:
"Doesn't using an enum here fairly sharply limit the reusability of the code?"

ajs6f: Yes, this is all why I would rather not use the `enum` here, although I appreciate that @acoburn is just trying to get information passed around in a reasonable way. The context is: @acoburn is redoing RDF generation mechanics to use the new Java 8 Stream API, which is the general-purpose and well-supported version of what @cbeer and I did when we first built this stuff. In order to do that, he wants (quite reasonably) to provide some means by which clients can indicate what kind of RDF they want. There is no point to generating RDF and then filtering it, and it could be very expensive for some implementations. That's where this comes in. We _were_ passing around actual values for classes (`Foo.class`) but that couples the client directly to impls (boo!). The ideal,  to my mind, would be to develop a good set of interface types that actually model the RDF generation flavors we expect to support, and then pass those around as types. No coupling, extensible, no problem. But that woul admittedly slow this down and be more work for @acoburn.

barmintor: @awoods building a code API around enums requires adherence to that list of enums by other implementations, which is the kind of API I suspect we have all been frustrated by in the past. Though like I said, limited context.

ajs6f: Suppose you write a new impl and you want to experiment with providing a new kind of RDF around resources (something we want to _encourage_). No way to do that here and still reuse the HTTP layers. You can't add members to an `enum`, you have to use polymorphism. That's the natural way in a language like Java.

barmintor: @ajs6f I am inclined to mumble something about generics and interfaces here, but if I am to do that I need to read the whole thing more carefully.

ajs6f: Yeah, there isn't any question in my mind that there are better options than an `enum`. There is a question as to whether we just eat the `enum` for now with the good intention of revisiting it (the kernel Java API has never been very public and I have no problem grinding on it, because we personally know everyone who would care and we can just buy them a beer in Dublin for the inconvenience) or whether we stop and catch our breath for a bit and doo it really right, which is annoying and obstacular to @acoburn , who has put in a lot of work on this and deserves to see his commit numbers go up.

ajs6f: Maybe we hold off for another day and talk this over tomorrow?

acoburn: I agree that an enum is not ideal, but it seems the simplest way to decouple the HTTP layers from the modeshape implementation, which is an absolute blocker for any alternate implementation.

This PR is already on the agenda for tomorrow's call.

escowles: I agree that using an enum will prevent the kind of future extensibility we want to encourage, but I think it's fine to address that in another ticket and not hold this up for that.

awoods: Agreed. My preference is for incremental steps. @acoburn has done a great job of initial decoupling from fcrepo-kernel-modeshape with this PR. I would be inclined to move forward with the `enum` approach for this increment.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/TypeRdfContext.java
```diff
@@ -51,15 +55,15 @@ public TypeRdfContext(final FedoraResource resource,
...
+        final Stream.Builder<Triple> other = Stream.builder();
+        if (resource instanceof NonRdfSource) {
             // gather versionability info from the parent
-            if (resource().getNode().getParent().isNodeType("mix:versionable")) {
-                concat(create(subject(), type.asNode(), createURI(MIX_NAMESPACE + "versionable")));
+            if (resource.getNode().getParent().isNodeType("mix:versionable")) {
+                other.accept(create(subject(), type.asNode(), createURI(MIX_NAMESPACE + "versionable")));
             }
         }
+        this.stream = Stream.concat(
```
awoods: You are changing the `*RdfContext` classes to set `this.stream` directly, vs. `concat()`ing to whatever used to be in `this.stream`. This may result in the unexpected loss of triples in `*RdfContext` classes in the inheritance tree above or below any of your updated `*RdfContext` classes.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfContext.java
barmintor: Can I ask the question another way? I don't grok how the enum part has
anything to do with the stream/iterator part; that seems like it should
have been another ticket, and another PR. I trust @acoburn's assessment
about whether that's possible or not, but this seems like a PR that is
trying to to do a little too much. Maybe that's a consequence of mvn
juggling?

acoburn: @barmintor in one sense these are two separate tickets: 1) change RdfStream to be Stream-based instead of Iterator-based and 2) decouple the modeshape-based RdfStream implementations from the HTTP layers.

Approaching these separately, however, would mean completely rewriting the RdfStream implementations (all 18 of them), and then rewriting them again. IMO, this is like removing a band-aid: it is better to do it all at once.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -233,69 +237,60 @@ protected RdfStream getResourceTriples(final int limit) {
...
             // LDP container membership triples for this resource
             if (ldpPreferences.prefersMembership()) {
-                rdfStream.concat(getTriples(LdpContainerRdfContext.class));
-                rdfStream.concat(getTriples(LdpIsMemberOfRdfContext.class));
+                streams.add(getTriples(LDP_MEMBERSHIP));
             }
 
-            // Embed all hash and blank nodes
-            // using IS_MANAGED_TRIPLE directly to avoid Prefer header logic (we never want them for hash fragments)
-            rdfStream.concat(filter(getTriples(HashRdfContext.class), IS_MANAGED_TRIPLE.negate()::test));
```
awoods: It appears that you are changing the logic to include `HashRdfContext` and `SkolemNodeRdfContext` in both `Prefer=minimal` and `!Prefer=minimal`, whereas they previously were only included in `!Prefer=minimal`.

Note: Hash and Skolem are being added here: https://github.com/fcrepo4/fcrepo4/pull/962/files#diff-d372035f89c8d0846920e20bd13e0145R245

acoburn: That is true. Should Hash and Skolem contexts be excluded from `representation=minimal`?

awoods: That is the way we have been interpreting `minimal` to-date. So I would say, yes.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -233,69 +237,60 @@ protected RdfStream getResourceTriples(final int limit) {
...
-        rdfStream.concat(filter(getTriples(TypeRdfContext.class), tripleFilter::test));
+        final List<Stream<Triple>> streams = new ArrayList<>();
 
-        rdfStream.concat(filter(getTriples(PropertiesRdfContext.class), tripleFilter::test));
+        streams.add(getTriples(PROPERTIES).filter(tripleFilter));
 
-        if (!returnPreference.getValue().equals("minimal")) {
+        if (returnPreference.getValue().equals("minimal")) {
+            if (ldpPreferences.prefersServerManaged()) {
+                streams.add(getTriples(SERVER_MANAGED)
```
awoods: You are adding much more to `minimal` than before in the case where `Prefer=ServerManaged`. Previously, only the `LdpRdfContext` triples were added. You have changed that to add:
`LdpRdfContext` and `AclRdfContext`, `RootRdfContext`, `ContentRdfContext`, and `ParentRdfContext`.

Also, I am not clear why you are surgically filtering out triples that include "?s rdf:type ldp:*".

As a general note, the refactoring in this PR would be smoother if you were to minimize changes to existing logic.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/TypeRdfContext.java
acoburn: `this.stream` will only ever be empty at that point, but if you'd prefer that I use `Stream.concat` instead of direct assignment, that's fine.

awoods: My concern is that `NodeRdfContext` or `DefaultRdfStream` or `WrappingStream` may add something to `this.stream` (in future updates by someone who does not realize that subclasses are overwriting that data member) that becomes lost with the construction of any of the sibling classes to `TypeRdfContext`.

acoburn: @awoods you are completely right. I was thinking of the current subclasses of NodeRdfContext, not _any_ subclass of NodeRdfContext.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
```diff
@@ -187,35 +186,26 @@ void init() throws IOException {
...
 
-        try {
-            final RdfStream nsRdfStream = new NamespaceRdfContext(rdfStream.session());
-
-            rdfStream.namespaces(nsRdfStream.namespaces());
-
-            final Node subject = VIEW_HELPERS.getContentNode(rdfStream.topic());
-
-            final Model model = rdfStream.asModel();
+        final Node subject = VIEW_HELPERS.getContentNode(nsStream.stream.topic());
```
awoods: It seems like an anit-pattern for clients of `RdfNamespacedStream` to need to be aware of the internals of `RdfNamespacedStream.stream` in order to get `RdfNamespacedStream.stream.topic`. 
If however, `RdfNamespacedStream` implemented `RdfStream`, clients could instead simply call `RdfNamespacedStream.topic()`.

acoburn: RdfNamespacedStream is simply a Tuple with two members. I would really prefer that it is no more complex than that.

This is a consequence of removing the namespaces() method from RdfStream (previously, it had been carrying around a jcr.Session). This structure (RdfNamespacedStream) sets the stage for injecting a NamespaceFactory, which is not included here because it would require changing the Spring configuration in the webapp and therefore a major version bump.

awoods: Does implementing the `RdfStream` interface change any of what you are saying?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
acoburn: No. your logic is backwards. Only LdpRdfContext triples are permitted through that filter (though I admit that this line of code is not pretty).

The issue is that SERVER_MANAGED triples come in two forms: those that are part of a `return=minimal` response and those that are part of a `return=representation` response.

One option would be to have two versions in the `enum`: SERVER_MANAGED_ALL and SERVER_MANAGED_MINIMAL, but I really did not like that, so I opted to have a single SERVER_MANAGED enum value and filter out any non LdpRdfContext triples in the case or `return=minimal`.

awoods: I have two concerns here:
1. The logic is changing from the original implementation - minor concern, although it may have unexpected impacts
2. Your use of the `RdfContext` enum has been consistent except for this block that does some magic assumption-mapping that the new filtering will map to only the desired `LdpRdfContext` triples. For maintenance sake, a more transparent approach would be helpful.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
acoburn: Implementing `RdfStream` would require considerably more code. You would need to have `RdfNamespacedStream extend RdfStream` and then have an implementing class along the lines of `DefaultRdfStream`. I would really prefer to stick with my method -- it's just a Tuple.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/RdfNamespacedStream.java
```diff
@@ -0,0 +1,44 @@
...
+    public final Map<String, String> namespaces;
+
+    /**
+     * Creates an object to hold an RdfStream and an associated namespace mapping.
+     *
+     * @param stream the RdfStream
+     * @param namespaces the namespace mapping
+     */
+    public RdfNamespacedStream(final RdfStream stream, final Map<String, String> namespaces) {
+        this.stream = stream;
```
awoods: Since we are giving direct access to these data members, we should assert that they are non-null in this constructor.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
awoods: You are correct. I was misled by the simple signature on the `RdfStream` interface:
https://github.com/fcrepo4/fcrepo4/pull/962/files#diff-890f544f149c4ed8a3017db0cc81846bR29

..overlooking the fact that it extends another, larger interface.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
acoburn: 1) The logic should not be changing from the original implementation. If it is, that is an oversight on my part and a failure of our testing mechanism not to catch that.

2) The logic appears inconsistent because I tried to align the `RdfContext enum` with the set of expected Prefer headers. The same issue applies to `PROPERTIES` and whether Hash and Skolem nodes should be included. The better approach would be what @ajs6f has been suggesting and use something other than an `enum`. OTOH, this is already an extremely large and complex PR, and I would prefer not to make it considerably bigger.

awoods: I certainly do not want to make this PR any larger either. However, the filter logic here is subverting the `RdfContext` logic and hence ripe for misunderstanding and bugs.
I am suggesting a more transparent approach, which might mean adding more values to the `RdfContext` enum since the current values appear to be insufficient.

acoburn: Previously, the `RdfContext` enum consisted of names for all the implementing `*RdfContext` values (ACL, REFERENCES, ROOT, SKOLEM, etc). I was asked to align the enum to the expected Prefer contexts so I reimplemented it. Which will it be?

awoods: I must have missed the request "to align the enum to the expected Prefer contexts". I would suggest that the current approach does not exactly hit the mark because we see here that filtering and tweaking is still required to get the desired triples in the various Prefer contexts.
If we can come up with a clean and transparent approach along the lines of what you have started here, that is great. If not, I have no problem with mapping the `RdfContext` enum to the `*RdfContext` values.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -352,7 +352,7 @@ public Response createOrReplaceObjectRdf(
         final RdfStream resourceTriples;
 
         if (resource.isNew()) {
-            resourceTriples = new RdfStream();
+            resourceTriples = new DefaultRdfStream(translator().reverse().convert(resource()).asNode());
```
awoods: We do not necessarily need to address it here, but we have `translator().reverse().convert(resource())` all over the place. Would it make sense to have a helper method on `IdentifierConverter` to simplify this?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
ajs6f: The underlying fact here is that the `enum` is not only in-extensible, but doesn't work all that well out of the gate. I think it's time to talk about alternatives (well, really it _will_ be time tomorrow, on the call).

awoods: Also, as @barmintor mentions [here](https://github.com/fcrepo4/fcrepo4/pull/962/files#r53178615), having an enum that tries to support the variety of `Prefer` contexts defined by the combinations of `include` and `omit` statements will be an ongoing challenge.

ajs6f: Just so's I don't forget, one of the ways that the `enum` improves on the current design is that it packages up the current impl classes into sensible bundles like "makes LDP membership triples" or "makes LDP containment triples" etc. That's a good characteristic we want in any design we choose. (I would think we could attach those semantics to types and not singletons.)

acoburn: @awoods: that particular concern is a non-issue here. The `FedoraResource::getTriples` method accepts an `EnumSet`, so if a particular class of triples is to be `include`d or `omit`ed, the implementing client either includes or omits that particular `RdfContext` value: `obj.getTriples(translator(), of(PROPERTIES, SERVER_MANAGED))` or `obj.getTriples(translator(), of(LDP_MEMBERSHIP))`.

The bigger issue is what @ajs6f mentions: namely that the `enum` can't be extended by any sort of hierarchical mechanism that way a class or interface can. 

What I would like to resolve is whether that change (from an `enum` to an `interface`) needs to happen now, as part of this PR, or later, after this gets merged. My opinion is that it should happen later, but only because rebasing this PR is truly hellish, and if there are too many more commits that get merged in the meantime, it will become unbearable.

ajs6f: Why can't we step this PR back to accepting the `Class` of impl classes, as in the current codebase, merge all the great stuff you did using the Streams API, then take the second step after more discussion?

ajs6f: Also, as a total side note, if we do this PR as it stands, you should add `getTriples(translator, RdfContext context)`, because `obj.getTriples(translator(), of(LDP_MEMBERSHIP))` is a little small bit icky. Just being picky.

acoburn: What is wrong with `EnumSet<RdfContext>.of(...)`?

The signature of `FedoraResource::getTriples` accepts either a single `RdfContext` or an `EnumSet` of those contexts.

acoburn: @ajs6f we could certainly pull apart this PR and do it more incrementally, but in all likelihood, I wouldn't have time to do that until April, which probably means completely starting over (which means probably not doing it at all). Unless someone else wants to take over the shepherding of this PR.

acoburn: To @ajs6f's point about bundling logical classes together, I completely agree. Having initially implemented this `enum` as a long list with all the implementing classes (ACL, ROOT, HASH, SKOLEM, etc), I think the current design is considerably better. It is not ideal, as shown with the `return=minimal` combinations, but from a design perspective, I find it much better to work with.

awoods: There is no doubt that almost everything here is great. Can we do anything to avoid the magic filtering that is found [below](https://github.com/fcrepo4/fcrepo4/pull/962/files#diff-d372035f89c8d0846920e20bd13e0145R250)?

The need for this mysterious filtering highlights a misalignment between Prefer contexts and the current `RdfContext` enum.

acoburn: The simple way to handle that is to add a MINIMAL enum value that would allow for those permutations. It's not the most elegant design, but it would work and could be replaced with a stronger `interface`-based design later.

awoods: I would be interested in seeing what an additional MINIMAL enum value could offer.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/VersionsRdfContext.java
```diff
@@ -72,64 +73,41 @@
     public VersionsRdfContext(final FedoraResource resource,
                               final IdentifierConverter<Resource, FedoraResource> idTranslator)
         throws RepositoryException {
-        super();
+        super(idTranslator.reverse().convert(resource).asNode());
         this.idTranslator = idTranslator;
         this.subject = idTranslator.reverse().convert(resource).asNode();
```
awoods: It is unsettling that `VersionsRdfContext` has its own `private` data member `Node subject` that appears to be eclipsing the data member `Node super.node`.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
ajs6f: `EnumSet<RdfContext>.of(...)` is only icky when its a singleton and we could have another method instead, because it's an unneeded short-lived object. It's just me being picky-- we have other much more important things about which to worry, so don't worry about it.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/iterators/PersistingRdfStreamConsumer.java
```diff
@@ -118,25 +118,30 @@ public PersistingRdfStreamConsumer(final IdentifierConverter<Resource, FedoraRes
             return true;
         };
         // we fail on non-Fedora RDF
-        this.stream = stream.withThisContext(stream.filter(isFedoraSubjectTriple::test));
+//new RdfAdder(idTranslator, getSession(), replacementStream
```
awoods: Looks like some dead code needs removal here.

awoods:
Done with review.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
barmintor: I just want to add that, while I would rather not release before moving
from enum to interface, I am personally swayed by the PR growing too large
to rebase if we need to merge and spin new tickets off.

ajs6f: What @barmintor said, although I am surprised that we can't just back away from the `enum` without a lot of work. That seems odd to me, like opening a sardine can and having a gush of thick purple sweet-smelling smoke shoot out. Not wrong, per se. Just… unexpected.

acoburn:
Updates applied.

awoods:
Thanks for your patience, @acoburn, with all of the back-and-forth. I think you landed this in an excellent place.
Now the foundation is set for the `enum` debates.
@ajs6f, I will wait on your feedback before squashing and merging.

ajs6f:
After discussion today, it seems like merging is the right thing to do.

awoods:
I'm on it...

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/523953db47ff09c0311de1dd51e98ae384cc0836

PR #966, opened 2015-12-22
GET w/ bad Accept header now returns proper error

bseeger:
- When an invalid or unsupported format type was requested the wrong
HTTP error code was returned.  Now it returns the correct one.

Resolves: https://jira.duraspace.org/browse/FCREPO-1840

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -103,7 +103,6 @@ import org.springframework.context.annotation.Scope;
 import com.codahale.metrics.annotation.Timed;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Splitter;
-import com.google.common.collect.ImmutableList;
 import static com.google.common.base.Strings.nullToEmpty;
 
 /**
@@ -204,8 +203,7 @@ public class FedoraLdp extends ContentExposingResource {
 
         LOGGER.info("GET resource '{}'", externalPath);
 
-        final ImmutableList<MediaType> acceptableMediaTypes =
-                new ImmutableList.Builder<MediaType>().addAll(headers.getAcceptableMediaTypes()).build();
+        final List<MediaType> acceptableMediaTypes = headers.getAcceptableMediaTypes();
 
         if (acceptableMediaTypes.size() > 0) {
 
@@ -223,8 +221,8 @@ public class FedoraLdp extends ContentExposingResource {
                     x.isWildcardType() || possibleVariants.stream().anyMatch(t -> t.getMediaType().isCompatible(x)));
 
             if (!match) {
-                 LOGGER.info("Unable to produce content in the requested mime-type(s): " );
-                 acceptableMediaTypes.stream().forEach(x -> LOGGER.info(x.toString()));
+                LOGGER.info("Unable to produce content in the requested mime-type(s): " );
+                acceptableMediaTypes.stream().forEach(x -> LOGGER.info("MediaType: ", x));
 
                 return notAcceptable(possibleVariants).build();
             }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -198,7 +198,7 @@ public class FedoraLdp extends ContentExposingResource {
     @GET
     @Produces({TURTLE + ";qs=10", JSON_LD + ";qs=8",
             N3, N3_ALT2, RDF_XML, NTRIPLES, APPLICATION_XML, TEXT_PLAIN, TURTLE_X,
-            TEXT_HTML, APPLICATION_XHTML_XML, "*/*"})
+            TEXT_HTML, APPLICATION_XHTML_XML})
     public Response describe(@HeaderParam("Range") final String rangeValue) throws IOException {
         checkCacheControlHeaders(request, servletResponse, resource(), session);
 
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -21,14 +21,15 @@ import static javax.ws.rs.core.MediaType.APPLICATION_XHTML_XML;
 import static javax.ws.rs.core.MediaType.APPLICATION_XML;
 import static javax.ws.rs.core.MediaType.TEXT_HTML;
 import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
-import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
-import static javax.ws.rs.core.Response.Status.NOT_IMPLEMENTED;
 import static javax.ws.rs.core.Response.created;
 import static javax.ws.rs.core.Response.noContent;
 import static javax.ws.rs.core.Response.ok;
+import static javax.ws.rs.core.Response.notAcceptable;
+import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
 import static javax.ws.rs.core.Response.Status.CONFLICT;
-import static javax.ws.rs.core.Response.Status.UNSUPPORTED_MEDIA_TYPE;
 import static javax.ws.rs.core.Response.Status.FORBIDDEN;
+import static javax.ws.rs.core.Response.Status.NOT_IMPLEMENTED;
+import static javax.ws.rs.core.Response.Status.UNSUPPORTED_MEDIA_TYPE;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.apache.http.HttpStatus.SC_BAD_REQUEST;
 import static org.apache.jena.riot.WebContent.contentTypeSPARQLUpdate;
@@ -81,6 +82,7 @@ import javax.ws.rs.core.Link;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriBuilderException;
+import javax.ws.rs.core.Variant;
 
 import org.fcrepo.http.commons.domain.ContentLocation;
 import org.fcrepo.http.commons.domain.PATCH;
@@ -101,6 +103,7 @@ import org.springframework.context.annotation.Scope;
 
 import com.codahale.metrics.annotation.Timed;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
 import com.google.common.base.Splitter;
 import static com.google.common.base.Strings.nullToEmpty;
 
@@ -202,13 +205,26 @@ public class FedoraLdp extends ContentExposingResource {
         checkCacheControlHeaders(request, servletResponse, resource(), session);
 
         LOGGER.info("GET resource '{}'", externalPath);
-        addResourceHttpHeaders(resource());
 
         final RdfStream rdfStream = new RdfStream().session(session)
                     .topic(translator().reverse().convert(resource()).asNode());
 
-        return getContent(rangeValue, getChildrenLimit(), rdfStream);
+        // If requesting a binary, check the mime-type if "Accept:" header is present.
+        // (This needs to be done before setting up response headers, as getContent
+        // returns a response - so changing headers after that won't work so nicely.)
+        final ImmutableList<MediaType> acceptableMediaTypes =
+                ImmutableList.copyOf(headers.getAcceptableMediaTypes());
 
+        if (resource() instanceof FedoraBinary && acceptableMediaTypes.size() > 0) {
+            final MediaType mediaType = MediaType.valueOf(((FedoraBinary) resource()).getMimeType());
+            if (!acceptableMediaTypes.stream().anyMatch(t -> t.isCompatible(mediaType))) {
+                return notAcceptable(
+                        Variant.VariantListBuilder.newInstance().mediaTypes(mediaType).build()).build();
+            }
+        }
+
+        addResourceHttpHeaders(resource());
+        return getContent(rangeValue, getChildrenLimit(), rdfStream);
     }
 
     private int getChildrenLimit() {
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -191,14 +191,14 @@ public class FedoraLdp extends ContentExposingResource {
      * Retrieve the node profile
      *
      * @param rangeValue the range value
-     * @return triples for the specified node
+     * @return a binary or the triples for the specified node
      * @throws IOException if IO exception occurred
      */
     @GET
     @Produces({TURTLE + ";qs=10", JSON_LD + ";qs=8",
             N3, N3_ALT2, RDF_XML, NTRIPLES, APPLICATION_XML, TEXT_PLAIN, TURTLE_X,
             TEXT_HTML, APPLICATION_XHTML_XML})
-    public Response describe(@HeaderParam("Range") final String rangeValue) throws IOException {
+    public Response getResource(@HeaderParam("Range") final String rangeValue) throws IOException {
         checkCacheControlHeaders(request, servletResponse, resource(), session);
 
         LOGGER.info("GET resource '{}'", externalPath);
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -18,25 +18,27 @@ package org.fcrepo.http.api;
 
 import static com.hp.hpl.jena.rdf.model.ResourceFactory.createResource;
 import static javax.ws.rs.core.MediaType.APPLICATION_XHTML_XML;
+import static javax.ws.rs.core.MediaType.APPLICATION_XML;
 import static javax.ws.rs.core.MediaType.TEXT_HTML;
 import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
-import static javax.ws.rs.core.Response.created;
-import static javax.ws.rs.core.Response.notAcceptable;
-import static javax.ws.rs.core.Response.noContent;
-import static javax.ws.rs.core.Response.ok;
 import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
 import static javax.ws.rs.core.Response.Status.NOT_IMPLEMENTED;
+import static javax.ws.rs.core.Response.created;
+import static javax.ws.rs.core.Response.noContent;
+import static javax.ws.rs.core.Response.ok;
 import static javax.ws.rs.core.Response.Status.CONFLICT;
 import static javax.ws.rs.core.Response.Status.UNSUPPORTED_MEDIA_TYPE;
 import static javax.ws.rs.core.Response.Status.FORBIDDEN;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.apache.http.HttpStatus.SC_BAD_REQUEST;
 import static org.apache.jena.riot.WebContent.contentTypeSPARQLUpdate;
+import static org.fcrepo.http.commons.domain.RDFMediaType.JSON_LD;
 import static org.fcrepo.http.commons.domain.RDFMediaType.N3;
 import static org.fcrepo.http.commons.domain.RDFMediaType.N3_ALT2;
 import static org.fcrepo.http.commons.domain.RDFMediaType.NTRIPLES;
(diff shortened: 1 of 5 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -403,6 +404,24 @@ public void testGetNonRDFSourceDescription() throws IOException {
...
+     */
+    @Test
+    public void testGetRDFSourceWrongAccept() throws IOException {
+        final String id = getRandomUniqueId();
+        createObjectAndClose(id);
+
+        final HttpGet get = new HttpGet(serverAddress + id);
+        get.addHeader("Accept", "application/turtle");
+
+        try (final CloseableHttpResponse response = execute(get)) {
```
ajs6f: I think there's a `getStatus` method in the superclass that does the closing for you.

acoburn: Here's an example of what @ajs6f is suggesting: http://git.io/vEcjc

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -192,21 +194,45 @@ public Response options() {
...
-            N3, N3_ALT2, RDF_XML, NTRIPLES, APPLICATION_XML, TEXT_PLAIN, TURTLE_X,
-            TEXT_HTML, APPLICATION_XHTML_XML, "*/*"})
+    @Produces({TURTLE + ";qs=10", "*/*"})
     public Response describe(@HeaderParam("Range") final String rangeValue) throws IOException {
         checkCacheControlHeaders(request, servletResponse, resource(), session);
 
         LOGGER.info("GET resource '{}'", externalPath);
+
+        final ImmutableList<MediaType> acceptableMediaTypes =
+                new ImmutableList.Builder<MediaType>().addAll(headers.getAcceptableMediaTypes()).build();
```
ajs6f: Use the static factory method `ImmutableList.builder()`, and `acceptableMediaTypes` can just be typed `List<MediaType>`. Actually, do you really need to copy the value of `getAcceptableMediaTypes()` over at all?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -192,21 +194,45 @@ public Response options() {
...
+            } else {
+                possibleVariants.addAll(RDFMediaType.POSSIBLE_RDF_VARIANTS);
+            }
+
+            final boolean match = acceptableMediaTypes.stream().anyMatch(x ->
+                    x.isWildcardType() || possibleVariants.stream().anyMatch(t -> t.getMediaType().isCompatible(x)));
+
+            if (!match) {
+                 LOGGER.info("Unable to produce content in the requested mime-type(s): " );
+                 acceptableMediaTypes.stream().forEach(x -> LOGGER.info(x.toString()));
```
ajs6f: You can use a method reference here: `acceptableMediaTypes.stream().forEach(LOGGER::info);`

bseeger: Cool! I didn't know about that. I'll change it.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
bseeger: No, I don't think so, but I was trying to prevent a few function calls to access that list. I could just call `headers.getAcceptableMediaTypes()` whenever I need the list, or just use a 
```java
final List<MediaType> acceptableMediaTypes = headers.getAcceptableMediaTypes();
```
Which way is preferred in Java?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
bseeger: Actually, it doesn't appear as if I can do this here, since I'd be using an instanceMethod that doesn't take a MediaType as it's argument.  
`acceptableMediaTypes.stream().forEach(LOGGER::info)` equals, I think, 
`acceptableMediaTypes.stream().forEach(x -> LOGGER::info(x))`  <-- x is a MediaType; won't compile
Seems that I really need the `x.toString()` as the argument.  Thoughts?

ajs6f: No, you are right. I mistakenly thought that there was a `Logger::info` that accepted an object and called `toString()` on it. I must have been thinking of a different logging API. But you could let `Logger` do the `toString()` with `x -> LOGGER.info("MediaType: ", x)`. It's pretty minor either way. Just for the sake of completeness, you could also do
```
acceptableMediaTypes.stream().map(MediaType::toString).forEach(LOGGER::info)
```

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
ajs6f: Given that the result of `getAcceptableMediaTypes()` isn't likely to change in the course of this method, I think `final List<MediaType> acceptableMediaTypes = headers.getAcceptableMediaTypes();` is fine.

ajs6f: Just for the record, function calls in Java are incredibly cheap (JIT). Creating short-lived objects is what you really want to avoid.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/RDFMediaType.java
```diff
@@ -67,10 +67,11 @@
 
     public static final List<Variant> POSSIBLE_RDF_VARIANTS = mediaTypes(
             RDF_XML_TYPE, TURTLE_TYPE, N3_TYPE, N3_ALT2_TYPE, NTRIPLES_TYPE, APPLICATION_XML_TYPE,
-            TEXT_PLAIN_TYPE, TURTLE_X_TYPE, JSON_LD_TYPE).add().build();
+            TEXT_PLAIN_TYPE, TURTLE_X_TYPE, JSON_LD_TYPE, TEXT_HTML_TYPE, APPLICATION_XHTML_XML_TYPE).add().build();
```
awoods: I have serious concerns about introducing arguably incorrect RDF variants: html and xhtml.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/RDFMediaType.java
```diff
@@ -67,10 +67,11 @@
...
 
     public static final String POSSIBLE_RDF_RESPONSE_VARIANTS_STRING[] = {
-        TURTLE, N3, N3_ALT2, RDF_XML, NTRIPLES, TEXT_PLAIN, APPLICATION_XML, TURTLE_X, JSON_LD };
+            TURTLE, N3, N3_ALT2, RDF_XML, NTRIPLES, TEXT_PLAIN, APPLICATION_XML, TURTLE_X, JSON_LD, TEXT_HTML,
```
awoods: I have serious concerns about introducing arguably incorrect RDF variants: html and xhtml.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/RDFMediaType.java
ajs6f: By what means could Fedora possibly be producing HTML RDF output? Are you thinking about the HTML "admin" presentation?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -192,21 +193,44 @@ public Response options() {
...
      *
      * @param rangeValue the range value
-     * @return triples for the specified node
+     * @return the binary or the triples for the specified node
      * @throws IOException if IO exception occurred
      */
     @GET
-    @Produces({TURTLE + ";qs=10", JSON_LD + ";qs=8",
-            N3, N3_ALT2, RDF_XML, NTRIPLES, APPLICATION_XML, TEXT_PLAIN, TURTLE_X,
-            TEXT_HTML, APPLICATION_XHTML_XML, "*/*"})
```
awoods: Hmm, reflecting on @acoburn's [comment](https://jira.duraspace.org/browse/FCREPO-1840?focusedCommentId=47464&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-47464), I would like to see us move more towards increasing the use of the jaxrs machinery instead of further away.

ajs6f: Of course we want to. But this whole method borks JAX-RS (look at the return type). That's the underlying problem. If we can factor it into two methods with proper return types, we can unshoulder the work onto JAX-RS.

awoods: The return type of `Response` is perfectly reasonable jax-rs. What are you suggesting?

ajs6f: The return type is _not_ "perfectly reasonable jax-rs". Correct use of JAX-RS implies narrowly typed return values with corresponding mappers and serializers, not generically and opaquely typed return values. Correct usage here wold be one method with a return type of `RdfStream` for RDF and another with a return type of `StreamingOutput` for binaries, but we can't easily do that because of our funky URI structure.

acoburn: I believe the issue is that this endpoint can produce _anything_. Hence the `@Produces({"*/*"})`. A better use of JAX-RS would be to have one endpoint that produces content-negotiated RDF (LDP-RS with explicit `Produces` values) and another that produces `Binary` (LDP-NR) resources where any content type is permitted. The difficulty with that approach is (AFAIK) that Fedora uses the same endpoint to produce both of these resource types.

ajs6f: What @acoburn said is what I'm trying to say. Endpoints should _not_ be able to produce _anything_. The fact that this one can is exactly why it has a `Response` return type, which is wrong _because_ it represents the fact that this endpoint can produce _anything_.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/RDFMediaType.java
bseeger: My bad -- I was a little confused about those two, though I do understand it more now and see that they return the entire webpage. I'll take them out of the RDF Variants list.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
awoods: The first question is, 
1. "Is it conceivable for our endpoints/URIs would take on a different structure to differentiate between rdf and non-rdf sources"?
2. If not, is there a more useful return type here?

ajs6f: 1. Sure. It's always conceivable that we will change. I don't know how useful a claim that is. The real question is "how much change are you as project lead willing to countenance bringing in front of the community"?

ajs6f: 2. No. The whole point is that the return type is (correctly) indicating a deeper problem with the method (that @acoburn described). The return type is "correct" in the sense that it is accurate. It is "wrong" in the sense that we would prefer to be something else, but cannot do that without factoring apart the method.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/RDFMediaType.java
ajs6f: I don't whether you are really wrong here, @bseeger , since some endpoints using this constant can produce HTML Maybe we can break this apart into "flavors of RDF that many endpoints can produce" and "HTML flavors". Some endpoints get just the first, and some get both together.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
bseeger: I'm not sure where to go with this now, other then maybe going back a commit or two - to what @acoburn suggested before ([comment](https://jira.duraspace.org/browse/FCREPO-1840?focusedCommentId=47464&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-47464)) - my original commit on this pull request.  It's clear that there's a bigger design issue here that shouldn't probably be solved via this bug.  One thought is to change the method name 'describe' to something like 'getResource' so it's clear (to developers) that it's fetching NonRdfResources as well as RdfResources.  I did put it in the method's description, so maybe that's enough. 

I don't know how to leverage more JaxRS here.  I tried to see if I could break up the method into two (sub) methods, each with their own @Produces [here](https://gist.github.com/bseeger/d2c48ba1ac0a2e37ab1d). 

I predicted this wouldn't work, and it hasn't (at least not yet, I'm still testing it out).  I couldn't find anything online specifically saying this wouldn't work, though it does seem unlikely.

awoods: Thanks for the update, @bseeger. For this round, I agree that the deeper architectural issue is out of scope. 
* Let's rename the method along the lines of `getResource`
* Roll back to a simpler commit
* Assess things from there.

Let's focus on the issue described in the ticket:
> When making a GET request with an incorrect mimeType, e.g. "Accept: application/turtle", fedora responds with a 500 status code. It should, instead, respond with a 400 code.

bseeger:
Reverted back to simpler mime-type checking. Changed the method name from describe(...) to getResource(...).  Let me know what you think.

awoods:
Thanks, @bseeger. 
First question, does the PR resolve the issue described in the ticket? 
Second question, does the PR introduce unexpected side effects?

bseeger:
First question: yes!  it now returns a 406 if someone requests an invalid mime-type. 
Legit request: `$>curl -XGET -i -H"Accept: text/turtle" http://localhost:8080/fcrepo/rest/pizza/
HTTP/1.1 200 OK`
Bad Request: `$>curl -XGET -i -H"Accept: application/turtle" http://localhost:8080/fcrepo/rest/pizza/
HTTP/1.1 406 Not Acceptable

Second question: yes!  (ugh) you can't specify the mime-type for a binary... you'll get a 406 every time you do - even if you pick the correct mime-type.  And... and... I'm not sure what to do with this.   I can put the `*/*` back in the @Produces line and do mime-type checking only on binary resources, but then the original bug exists again, unless I do mime-type checking on everything (skipping JAX-RS) and then we go in circles. ;)

awoods:
Thanks again, @bseeger. Your summary is helpful for expectation-setting.
We don't do conneg on binaries... so it is not completely insane to respond with a `406` when it is attempted.
I have put your ticket back into "Review".

ajs6f:
No, it's not insane and it may be as good as we can do. But it is weird. Are we saying that if I stick in a binary, then request it with the _correct mimetype_, I will still get an error?

bseeger:
Yes, you will get a 406.  It is weird, though unless we list all the relevant mime-types for binaries in the @Produces line, we need to do our own mime-type checking further down (or fix the larger issue of both Rdf and Non-Rdf resources being fetched via the same method).

ajs6f:
Well, we definitely can't list all the relevant mimetypes, because we don't know them! :)

awoods:
After testing this updated PR, I will add one point of clarification to the comments above:
* If you GET a binary resource with an `Accept` header from this [list](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java#L199-L201), the binary is downloaded correctly, with the real content-type, not the `Accept` content-type. 
I think this is better than the current behavior of returning a `500`, but not ideal.

I would suggest adding to the current PR a simple check in "ContentExposingResource.getContent()" before [line:194](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java#L195) for a match of the `Accept` header (if it exists) and either the real [contentTypeString](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java#L168) or [mediaType](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java#L184), which ever works better. If the `Accept` header exists and matches, great; if not, then `406`.
Reasonable?

bseeger:
With the PR the way it currently is, a GET request for a Binary with an Accept header would never get that far - because the `*/*` has been taken out of the `@Produces` line. If I put that back in, then the original bug is unresolved unless I go back to mime-type checking everything outside of JaxRS.  

I realize this is outside the scope of the bug, but how crazy would it be to reverse the semantics for getting a binary and it's associated metadata.   Ie, take out the `/fcr:metadata/` and always return metadata unless they tack on a `/fcr:binary/` on the end. 

Meaning, if `/blahbinary/` is a binary, then: 
`curl -XGET -v -i  -H"Accept: text/plain" http://localhost:8080/fcrepo/rest/blahbinary/`
Returns the metadata for that binary. 
And: 
`curl -XGET -v -i  -H"Accept: text/plain" http://localhost:8080/fcrepo/rest/blahbinary/fcr:binary`
Actually gets the binary.  Then we could split the GET into two functions (one for metadata, one for binaries) and clean up the api. 

... I have no clue how what I'm suggesting would impact current users.  Might be an absolutely terrible idea.

ajs6f:
That's what we used to do, except it was `fcr:content`, not `fcr:binary`. I can't remember why we changed. @barmintor, wasn't that your idea?

awoods:
@bseeger, When a `GET` request for a binary resource is made with and `Accept` header of something like `text/turtle`, it goes through successfully. With that in mind, please re-read: https://github.com/fcrepo4/fcrepo4/pull/966#issuecomment-183979826

bseeger:
Thanks for the clarification, Andrew.  I see the distinction now. Yes, I'm all for adding that mime-type checking before line:194 as described above.   I can add that today.

barmintor:
@ajs6f @bseeger it wasn't my idea, but I can point you to a couple of relevant posts:

https://github.com/fcrepo4/fcrepo4/commit/a2b9358842d2d46e07314533927ff14809f31880#commitcomment-7650435

https://groups.google.com/forum/#!searchin/fedora-tech/beta$20release$20notes/fedora-tech/mADFnf_1G30/i4XzmLIiq4kJ

IIRC: Our experience back in the early beta was that this issue set you chasing your own tail about various bad implementation experiences as a consequence of JAXRS & JCR, but that this approach minimized the weirdness required of the client to know about where & what to POST or DELETE to manage LDP-NRs.

cbeer:
`POST`, `DELETE`, and `PUT` are all pretty weird for the client with `fcr:content`:

- `PUT` for an LDP-NR ought to create the resource at the location given in the request. Requiring `fcr:content` either requires client knowledge about interacting fcrepo4, or some awkward hand-waving about when you're updating content or metadata.
- `DELETE` for an LDP-NR cascades to the LDP-RS describing it, and having a delete request for a resource also delete its "parent" is not intuitive

ajs6f:
Now I wish we had taken the high road and refused to use any magic URLs. When people create bitstreams, we could have returned the URI of the description in a header. We could have done the work of decoupling hierarchy and lifecycle. We would have been much better off in the long run.

barmintor:
But we do this! When you create a bitstream, the URI of the description is
in a Link@rel="describedBy" header! This is a LDP-ism.

awoods:
May I suggest creating a GitHub issue for the `fcr:metadata` discussion. It is out of scope for the work that @bseeger is doing here.

ajs6f:
Then why do we need `fcr:metadata` at all?

bseeger:
The latest commit fixes the loophole where asking for a binary with Accept header containing a mime-type listed in the @Produces line, but not the binary's actual mime-type, returned the resource.  Now, if the mime-type of the binary does not match a mime-type requested in the Accept field, a 406 will be returned instead. 

I had to introduce this change in the FedoraLdp.java file before the getContent() call because it seems you have to have the header info complete before the response builder (in getContent()) calls build().  I'm open to moving things around, this just seemed the cleanest of all the options I thought about and tried.

awoods:
@bseeger, please put the JIRA ticket back into "Review" if you are ready.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/825d26fb10396b4f3a5aad5f6a680ad6b6ef49ea

PR #968, opened 2015-12-23
FCREPO-1868: remove dependency on javax.jcr.Session in kernel-api

ajs6f:


ajs6f:
Removing `Session` from the API means that it is now appearing explicitly with much greater frequency in the HTTP layers, for now. I think that's unavoidable for the moment. Removing it from the HTTP layers is a whole separate task because of the mistake we made in tightly coupling `Session` with HTTP requests.

acoburn:
Parameterizing the `fcrepo-kernel-api` interfaces makes a lot of sense to me, but pushing the `Session` type into the HTTP layers just shifts the problem. I would expect that the better solution would have `Session` types isolated in the `fcrepo-kernel-modeshape` module.

ajs6f:
It definitely would, and I don't much care. The object here for me is to get this stuff out of the kernel APi so that reimplementation is no longer blocked. The other work of cleaning the HTTP layer can go on later. (And by someone else.)

Eventually, that type parameter must be replaced by an actual Fedora construction. Probably `Transaction`.

acoburn:
Is there anything preventing one from simply using `org.fcrepo.kernel.api.Transaction` as a concrete parameterized type in those cases? In this PR, `Transaction<T>` can already parameterized with a `Session` type.

ajs6f:
No, `Transaction` isn't ready for that. It doesn't actually offer the ability to do anything useful with Fedora resources yet. What we're talking about doing is de-generifying `Transaction` as we move the actual functions that alter repository contents from JCR-specific forms in `Session` to Fedora-specific forms in `Transaction`. If that _is_ what we end up doing.

ajs6f:
@acoburn , is this in abeyance because you are on vacation or because you have continuing or fresh objections?

acoburn:
@ajs6f I am still concerned about having `Session` pushed into the http layers, but I recognize your point about this blocking other development. Unless anyone else has objections, I will plan to merge this later today.

ajs6f:
I'm not trying to push you forward. I'm not going to do anything about this in the next few days anyway (Jena stuff) so if you want to step back and make a more concerted plan to eject JCR from the layers above the kernel, we can do that. Put an item on Thursday's meeting.

acoburn:
That sounds good -- I'll add it to the tech meeting agenda

acoburn:
@ajs6f since the effort that is part of #1107 supersedes this PR, shall we close this one?

ajs6f:
Yeah, I'm fine with that.

PR #971, opened 2016-01-21
Added package-info files where missing for all non-kernel modules

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1823

### DIFF
--- a/fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/package-info.java
+++ b/fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/package-info.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright 2015 DuraSpace, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/package-info.java
```diff
@@ -0,0 +1,23 @@
+/**
```
awoods: License header format should be `/*` instead of `/**`
https://travis-ci.org/fcrepo4/fcrepo4/builds/103756794#L5428

acoburn: Oh that's right. I'll fix that

acoburn:
@awoods this is ready for your review

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/b73f27084fa164ae25cde0c4e6e75dcd787a4ef9

PR #974, opened 2016-02-04
remove /fcr:nodetypes endpoint

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1890


awoods:
Although I completely agree with this PR, we have received strong feedback not to make backwards incompatible changes without prior warning. I suggest we save this PR and add a deprecation header now.

acoburn:
Are you thinking that a 299 response code with a Warning header would be appropriate?

awoods:
Is 299 used in the deprecated case? In any case, at least adding the "Warning" header, ala:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java#L676

acoburn:
Adding a Warning header will be tracked with https://jira.duraspace.org/browse/FCREPO-1892

299 is unassigned, and I don't think it's "standard" in any sense. I will leave the response code as 200.

ajs6f:
+1 to the plan, just wanted to add that we should emit public messages explaining that this is going away, and _**explaining why**_.

PR #977, opened 2016-02-05
remove jcr.Property from fcrepo-kernel-api

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1894


ajs6f:
So the idea here is to stamp out `Property` in favor of `Node`, which we can further stamp out at some future point?

acoburn:
Yes, the idea is that this further consolidates the use of jcr to the Node type which will be addressed separately.

ajs6f:
I'll wait a day and merge.

PR #978, opened 2016-02-09
Add configuration for a MySQL backend, instead of LevelDB

awoods:
As of this update, the default backend is still LevelDB.
To use the MySQL object store, you need to:
1. Install an instance of MySQL
2. Run Fedora with the following JAVA_OPTS:
```
JAVA_OPTS="${JAVA_OPTS} -Dfcrepo.modeshape.configuration=classpath:/config/jdbc-mysql/repository.json"
JAVA_OPTS="${JAVA_OPTS} -Dfcrepo.ispn.mysql.username=<username>"
JAVA_OPTS="${JAVA_OPTS} -Dfcrepo.ispn.mysql.password=<password>"
JAVA_OPTS="${JAVA_OPTS} -Dfcrepo.ispn.mysql.host=<default=localhost>"
JAVA_OPTS="${JAVA_OPTS} -Dfcrepo.ispn.mysql.port=<default=3306>"
```
Note, the database and table used by ISPN are auto-created.

Resolves: https://jira.duraspace.org/browse/FCREPO-1833

### DIFF
--- a/fcrepo-configs/src/main/resources/config/infinispan/jdbc-mysql/infinispan.xml
+++ b/fcrepo-configs/src/main/resources/config/infinispan/jdbc-mysql/infinispan.xml
@@ -38,7 +38,7 @@
           <jdbc:string-keyed-table prefix="ispn_entry"
                                    create-on-start="true"
                                    drop-on-exit="false">
-            <jdbc:id-column name="id" type="VARCHAR(255) CHARACTER SET latin1 COLLATE latin1_general_cs"/>
+            <jdbc:id-column name="id" type="VARCHAR(255)"/>
             <jdbc:data-column name="datum" type="LONGBLOB"/>
             <jdbc:timestamp-column name="version" type="BIGINT"/>
           </jdbc:string-keyed-table>
--- a/fcrepo-configs/src/main/resources/config/infinispan/jdbc-mysql/infinispan.xml
+++ b/fcrepo-configs/src/main/resources/config/infinispan/jdbc-mysql/infinispan.xml
@@ -1,0 +1,52 @@
+<infinispan xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xmlns="urn:infinispan:config:7.2"
+            xmlns:jdbc="urn:infinispan:config:store:jdbc:7.2"
+            xsi:schemaLocation="urn:infinispan:config:7.2 http://www.infinispan.org/schemas/infinispan-config-7.2.xsd urn:infinispan:config:store:jdbc:7.2 http://infinispan.org/schemas/infinispan-cachestore-jdbc-config-7.2.xsd">
+
+  <cache-container default-cache="FedoraRepository">
+
+    <jmx duplicate-domains="true"/>
+
+    <local-cache name="FedoraRepository" statistics="true">
+
+      <transaction transaction-manager-lookup="org.infinispan.transaction.lookup.GenericTransactionManagerLookup"
+                   mode="NON_XA"
+                   locking="PESSIMISTIC"/>
+
+      <eviction max-entries="500" strategy="LIRS" thread-policy="DEFAULT"/>
+
+      <!--
+          Define the cache loaders (i.e., cache stores). Passivation is false because we want *all*
+          data to be persisted, not just what doesn't fit into memory. Shared is false because there
+          are no other caches sharing this jdbc store. We set preload to false for lazy loading;
+          may be improved by preloading and configuring eviction.
+
+          We can have multiple cache loaders, which get chained. But we'll define just one.
+
+          See: https://docs.jboss.org/infinispan/7.2/configdocs/infinispan-cachestore-jdbc-config-7.2.html
+       -->
+      <persistence passivation="false">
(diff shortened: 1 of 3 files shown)


ajs6f:
MySQL stuff is under GPL. Is that okay for us?

escowles:
I don't have any concerns about MySQL or its licensing -- we're not distributing the code, and users can decide whether they are OK with GPL or not.

Though I would be interested in what's involved in supporting other databases, since people might reasonably want to use Postgres, Oracle, etc., etc.

DiegoPino:
@escowles, the cache store documentation (we are at 7.x i think, docs are for 8.x) 
http://infinispan.org/cache-store-implementations/
If those Relational DBs can use JDBC then it's pretty easy. We are just using this for strings right?

escowles:
@DiegoPino There are a couple of options, but the current config uses the one that stores values as serialized java objects in a blob column.  I assume it will work with any JDBC database.  But it would be nice to see a list of supported or tested databases to be sure.

I'm mostly interested in what the config looks like.  Right now, we have the MySQL dependency configured in pom.xml -- would we have to configure every database we wanted to support like this?  Could we make that parameterized too?

ajs6f:
@escowles We're going to have to pull anything we want to use for automated tests, at least at `test` scope.

escowles:
@ajs6f Yes, but the GPL is really only different from other OSS licenses in what it demands of people who distribute changes (i.e., that they're distributed under the GPL).  So I don't think we're going to have to do anything different, or put any burden on our users, by pulling GPL-licensed code from Maven.

ajs6f:
@escowles, sorry, my remark was in response to your remark "Right now, we have the MySQL dependency configured in pom.xml -- would we have to configure every database we wanted to support like this? Could we make that parameterized too?".

escowles:
@ajs6f Oh, my mistake.  Yes, we'd need to configure anything we wanted to test.  But my understanding is that we'd test against one database (embedded Derby/Cloud/HSQLDB) but want to let users swap in their JDBC database of choice.  It would be nice if they could do that by supplying system properties specifying the Maven package, JDBC driver class and connection properties.

ajs6f:
@escowles Fine with me. But it would mean a more onerous release process. Not saying we can't do that, but we should account for it.

escowles:
@ajs6f And maybe it's not worth doing if we expect users to use precompiled WAR files, where they would copy their JDBC driver and configure the connection in their servlet container.

ajs6f:
@escowles  Honestly, that's what I would choose. That's the J2EE way. Ops people will likely be more comfortable with it, too.

fcrepo-configs/src/main/resources/config/infinispan/jdbc-mysql/infinispan.xml
```diff
@@ -0,0 +1,52 @@
...
+                                 read-only="false"
+                                 purge="false">
+          <jdbc:connection-pool connection-url="jdbc:mysql://${fcrepo.ispn.mysql.host:localhost}:${fcrepo.ispn.mysql.port:3306}/ispn?createDatabaseIfNotExist=true"
+                                driver="com.mysql.jdbc.Driver"
+                                username="${fcrepo.ispn.mysql.username}"
+                                password="${fcrepo.ispn.mysql.password}"/>
+          <jdbc:string-keyed-table prefix="ispn_entry"
+                                   create-on-start="true"
+                                   drop-on-exit="false">
+            <jdbc:id-column name="id" type="VARCHAR(255) CHARACTER SET latin1 COLLATE latin1_general_cs"/>
```
escowles: Is there a reason why the character set is forced to latin1?  On a fresh install of MySQL 5.7.9 on MacOSX, this required changing my server settings to use latin1 instead of utf8.  Removing the character set specification here allowed it to run on a vanilla MySQL.

awoods: @escowles, I would be delighted to remove the `latin1` references. Should we remove the `COLLATE` clause and replace the `CHARACTER SET` with `utf8`?

escowles: It works for me with:

```
<jdbc:id-column name="id" type="VARCHAR(255)"/>
```

Does that work for you?

awoods:
@ajs6f / @escowles : re: GPL, I think MySQL's exception is made for us:
http://www.mysql.com/about/legal/licensing/foss-exception/

ajs6f:
Okay by me. We are FOSS. (I think that was the name of a really obscure Krautrock band in the early Eighties.)

fcrepo-configs/src/main/resources/config/infinispan/jdbc-mysql/infinispan.xml
awoods: Worked for me as well. See: https://github.com/awoods/fcrepo4/commit/d0368234704f83b05fee176ff3bed569d52f01f6

escowles:
Resolved with https://github.com/fcrepo4/fcrepo4/commit/f805f67e859be90d9c4fecaeff7cfcd0bb70fdc3

PR #982, opened 2016-02-16
Remove empty hash URI fragments after updates.

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-1764

### DIFF
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -581,8 +581,8 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
             if (node.hasNode("#")) {
                 for (final Iterator<Node> hashNodes = node.getNode("#").getNodes(); hashNodes.hasNext(); ) {
                     final Node n = hashNodes.next();
-                    final Iterator userProps = Iterators.filter(n.getProperties(),
-                            p -> !isManagedPredicate.test(propertyConverter.convert((Property)p)));
+                    final Iterator<Property> userProps = Iterators.filter((Iterator<Property>)n.getProperties(),
+                            p -> !isManagedPredicate.test(propertyConverter.convert(p)));
                     if ( !userProps.hasNext() ) {
                         LOGGER.debug("Removing empty hash URI node: {}", n.getName());
                         n.remove();
@@ -590,7 +590,7 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
                 }
             }
         } catch (RepositoryException ex) {
-            LOGGER.warn("Error removing empty fragments: {}", ex.toString());
+            throw new RepositoryRuntimeException("Error removing empty fragments", ex);
         }
     }
 
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -27,6 +27,7 @@ import static java.util.Arrays.asList;
 import static java.util.stream.Collectors.joining;
 import static org.apache.commons.codec.digest.DigestUtils.shaHex;
 import static org.fcrepo.kernel.api.RdfLexicon.REPOSITORY_NAMESPACE;
+import static org.fcrepo.kernel.api.RdfLexicon.isManagedPredicate;
 import static org.fcrepo.kernel.modeshape.FedoraJcrConstants.JCR_CREATED;
 import static org.fcrepo.kernel.modeshape.FedoraJcrConstants.JCR_LASTMODIFIED;
 import static org.fcrepo.kernel.modeshape.FedoraJcrConstants.FROZEN_MIXIN_TYPES;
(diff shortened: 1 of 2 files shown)


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -563,11 +569,31 @@ public void replaceProperties(final IdentifierConverter<Resource, FedoraResource
...
             throw new MalformedRdfException(exceptions.toString());
         }
     }
 
+    private void removeEmptyFragments() {
+        try {
+            if (node.hasNode("#")) {
+                for (final Iterator<Node> hashNodes = node.getNode("#").getNodes(); hashNodes.hasNext(); ) {
+                    final Node n = hashNodes.next();
+                    final Iterator userProps = Iterators.filter(n.getProperties(),
```
ajs6f: Can we not parameterize `Iterator` here?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -563,11 +569,31 @@ public void replaceProperties(final IdentifierConverter<Resource, FedoraResource
...
+                            p -> !isManagedPredicate.test(propertyConverter.convert((Property)p)));
+                    if ( !userProps.hasNext() ) {
+                        LOGGER.debug("Removing empty hash URI node: {}", n.getName());
+                        n.remove();
+                    }
+                }
+            }
+        } catch (RepositoryException ex) {
+            LOGGER.warn("Error removing empty fragments: {}", ex.toString());
```
ajs6f: Isn't this more like an `error`? Maybe this should even throw the exception?

ajs6f: In any event, no need to call `toString` in that position. Maybe you meant `getMessage`?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -563,11 +569,31 @@ public void replaceProperties(final IdentifierConverter<Resource, FedoraResource
...
+        removeEmptyFragments();
+
         if (exceptions.length() > 0) {
             throw new MalformedRdfException(exceptions.toString());
         }
     }
 
+    private void removeEmptyFragments() {
+        try {
+            if (node.hasNode("#")) {
```
ajs6f: Is there any need for this test when the loop below won't execute if there are no # nodes?

awoods:
The essence of this PR looks good, pending response to @ajs6f's comments.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
escowles: We can throw the exception, which would cause the update being processed to be abandoned.  As long as there aren't any errors in our node navigation, then I would expect the errors to be serious enough that abandoning the update probably makes sense.

ajs6f: That's my vote!

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
escowles: If there isn't a "#" node, then this will fail.  I tried removing it, and there are about a dozen IT failures, so I think we need to keep it.

ajs6f: Oh, that's the use of `getNode`, which throws `PathNotFoundException` when there is no node. That's a crap API. You could do `getNodes("#")` but that's not really any better. Oh, well.

escowles:
@ajs6f I've pushed a new commit that wraps any exceptions in a RepositoryRuntimeException, and casts the JCR PropertyIterator to an Iterator<Property>.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -111,6 +113,8 @@
 
     private static final Logger LOGGER = getLogger(FedoraResourceImpl.class);
 
+    private static final PropertyConverter propertyConverter = new PropertyConverter();
```
ajs6f: Just a note to all: there is no state in a `PropertyConverter`. We could probably make it a singleton and cut out some short-lived objects.

awoods: @ajs6f, would it make sense to make a ticket for Singleton-izing `PropertyConverter`? If so, please do.

ajs6f: I dunno. Would it?

awoods: Since you are the one doing the suggesting, and you don't know, then probably not.

ajs6f: I just don't want to add any more work to the Mode impl bag.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/a71298f4817887538112dccc1979ad40956c8ff3

PR #984, opened 2016-02-16
Translating invalid path requests to InvalidResourceIdentifierException/400 Bad Request

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-1837


ajs6f:
What are we doing now? Some kind of 500?

escowles:
Yes, it falls through to WildcardExceptionMapper and results in a 500.

ajs6f:
Cool, cool.

PR #985, opened 2016-02-17
Improving Prefer/Accept handling when creating resources

escowles:
* Fixes https://jira.duraspace.org/browse/FCREPO-1878
* Returning RDF of the newly-created resource when requested via Prefer header.

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -592,7 +592,19 @@ public class FedoraLdp extends ContentExposingResource {
         return createUpdateResponse(resource, true);
     }
 
-    private Response createUpdateResponse(final FedoraResource resource, final boolean created) throws IOException {
+    /**
+     * Create the appropriate response after a create or update request is processed.  When a resource is created,
+     * examine the Prefer and Accept headers to determine whether to include a representation.  By default, the
+     * URI for the created resource is return as plain text.  If a minimal response is requested, then no body is
+     * returned.  If a non-minimal return is requested, return the RDF for the created resource in the appropriate
+     * RDF serialization.
+     *
+     * @param resource The created or updated Fedora resource.
+     * @param created True for a newly-created resource, false for an updated resource.
+     * @return 204 No Content (for updated resources), 201 Created (for created resources) including the resource
+     *    URI or content depending on Prefer headers.
+     */
+    private Response createUpdateResponse(final FedoraResource resource, final boolean created) {
         addCacheControlHeaders(servletResponse, resource, session);
         addResourceLinkHeaders(resource, created);
         if (!created) {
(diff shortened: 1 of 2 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -586,15 +589,52 @@ public Response createObject(@QueryParam("checksum") final String checksumDeprec
...
+                prefer.getReturn().addResponseHeaders(servletResponse);
+            }
+            final RdfNamespacedStream rdfStream = new RdfNamespacedStream(
+                new DefaultRdfStream(translator().reverse().convert(resource()).asNode(), getResourceTriples()),
+                getNamespaces(session()));
+            return builder.entity(rdfStream).build();
+        }
+    }
 
+    private boolean acceptsPlainText() {
```
awoods: Please add a comment documenting what constitutes "acceptable" scenarios for plain/text.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -586,15 +589,52 @@ public Response createObject(@QueryParam("checksum") final String checksumDeprec
...
+        final List<String> acceptHeaders = headers.getRequestHeader(HttpHeaders.ACCEPT);
```
awoods: It may be cleaner to use: `headers.getAcceptableMediaTypes()`

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -586,15 +589,52 @@ public Response createObject(@QueryParam("checksum") final String checksumDeprec
         }
 
         LOGGER.debug("Finished creating resource with path: {}", newObjectPath);
+        return createUpdateResponse(resource, true);
+    }
 
-        addCacheControlHeaders(servletResponse, result, session);
-
-        final URI location = getUri(result);
+    private Response createUpdateResponse(final FedoraResource resource, final boolean created) throws IOException {
```
awoods: There is a fair amount of logic in this method. Please add a method comment giving an overview of what is expected in the returned `Response`.
...Or if you prefer, add inline comments throughout the logic of this method.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
escowles: Yes, and I think it'll be better to rework this method a bit to use the MediaType objects isWildcardType()  and isCompatible() methods, which will be more robust than the current impl.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/34b50e13f249a575d2b0fa23788e11d120ebd45e

PR #986, opened 2016-02-18
Adding configuration for PostgreSQL and generic JDBC databases

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-1916


awoods:
The postgres config files appear to work. I would suggest we remove the two `generic` configs and move this PR through.

PR #989, opened 2016-02-19
First widening of contract for FedoraResource::getTriples

ajs6f:


ajs6f:
DO NOT MERGE YET, MORE COMMITS TO COME

awoods:
This WIP PR is associated with: https://jira.duraspace.org/browse/FCREPO-1921 , no?

ajs6f:
Sure.

ajs6f:
@barmintor Take a look when you get a chance. This is really simple, much simpler than I thought we would have to be. Yay!

ajs6f:
Do not merge until squashing.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -192,16 +193,16 @@
...
-                new SimpleEntry<>(RdfContext.SERVER_MANAGED, getServerManagedTriples),
-                new SimpleEntry<>(RdfContext.LDP_MEMBERSHIP, getLdpMembershipTriples),
-                new SimpleEntry<>(RdfContext.LDP_CONTAINMENT, getLdpContainsTriples))
+                new SimpleEntry<>(RequiredRdfContext.PROPERTIES, getDefaultTriples),
+                new SimpleEntry<>(RequiredRdfContext.VERSIONS, getVersioningTriples),
+                new SimpleEntry<>(RequiredRdfContext.EMBED_RESOURCES, getEmbeddedResourceTriples),
+                new SimpleEntry<>(RequiredRdfContext.INBOUND_REFERENCES, getInboundTriples),
+                new SimpleEntry<>(RequiredRdfContext.SERVER_MANAGED, getServerManagedTriples),
+                new SimpleEntry<>(RequiredRdfContext.LDP_MEMBERSHIP, getLdpMembershipTriples),
+                new SimpleEntry<>(RequiredRdfContext.LDP_CONTAINMENT, getLdpContainsTriples))
```
acoburn: `import static` for these values?

acoburn:
I am :+1: on this design.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
ajs6f: Couldn't have said it better myself. I'll do that in a further commit.

ajs6f:
Worked @acoburn 's comments.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -135,8 +140,11 @@
 
     private static final PropertyConverter propertyConverter = new PropertyConverter();
 
-    private static Function<FedoraResource, Function<IdentifierConverter<Resource, FedoraResource>, Function<Boolean,
-            Stream<Triple>>>> getDefaultTriples = resource -> translator -> uncheck(minimal -> {
+    // A curried type accepting resource, translator, and "minimality", returning triples.
+    private static interface RdfGenerator extends Function<FedoraResource,
```
acoburn: In C++, this sort of "`typedef`" decl is done all the time, but I thought it was an anti-pattern in Java. Am I wrong?

ajs6f: Does it clarify the code or not? I feel like it does. In any event, if you aren't happy with it, I can definitely break it out into its own compilation unit and we can comment it up thoroughly. It might even provide part of a useful extension point that might get packed into the `kernel-api`.

acoburn: I think it definitely clarifies the code. I would be inclined to keep it as it is, especially since `RdfGenerator` is a private type.

ajs6f: Okely-dokely. Let's wait a bit to see if anyone else objects?

escowles: Well, it certainly reduces the repetitive cumbersome Function<<<>>> declarations, so it seems good to me.  If this pattern is useful in other implementations, it could be nice to make it part of the kernel-api.  But if it's not, then keeping it here and private makes sense to me.

ajs6f: I dunno if it would be useful outside the current context. It's not hard to take a pure type up into the API.  But it would be completely disconnected from any other type in the API, because fundamentally, this type is about impl. Hm.

acoburn:
Merged via 864f85e7b49989ee6a53b0df305139dfd745ce95

PR #990, opened 2016-02-23
FedoraResource::getChildren() now returns a Stream

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1924

Note: the object in the `hasChildCount` triple now uses the datatype `xsd:long` instead of an `xsd:int`. This is because `Stream::count` returns a `long`, but if we want to keep the triple's datatype as an `xsd:int` I could simply cast the `long` into an `int`. I don't expect any overflows, though doing so would also introduce that limitation.

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
@@ -54,7 +54,7 @@ public interface FedoraResource {
 
     /**
      * Get the children of this resource
-     * @return iterator
+     * @return a stream of Fedora resources
      */
     Stream<FedoraResource> getChildren();
 
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
@@ -18,8 +18,8 @@ package org.fcrepo.kernel.api.models;
 import java.net.URI;
 import java.util.Date;
 import java.util.EnumSet;
-import java.util.Iterator;
 import java.util.List;
+import java.util.stream.Stream;
 
 import javax.jcr.AccessDeniedException;
 import javax.jcr.Node;
@@ -56,7 +56,7 @@ public interface FedoraResource {
      * Get the children of this resource
      * @return iterator
      */
-    Iterator<FedoraResource> getChildren();
+    Stream<FedoraResource> getChildren();
 
     /**
      * Get the container of this resource
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -15,9 +15,6 @@
  */
 package org.fcrepo.kernel.modeshape;
 
-import static com.google.common.collect.Iterators.concat;
-import static com.google.common.collect.Iterators.filter;
-import static com.google.common.collect.Iterators.singletonIterator;
 import static com.google.common.collect.Iterators.transform;
 import static com.google.common.collect.Lists.newArrayList;
 import static com.hp.hpl.jena.update.UpdateAction.execute;
@@ -154,8 +151,7 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
 
     private static Function<FedoraResource, Function<IdentifierConverter<Resource, FedoraResource>, Function<Boolean,
             Stream<Triple>>>> getEmbeddedResourceTriples = resource -> translator -> uncheck(_minimal -> {
-        return iteratorToStream(resource.getChildren())
-                .flatMap(child -> child.getTriples(translator, RdfContext.PROPERTIES));
+        return resource.getChildren().flatMap(child -> child.getTriples(translator, RdfContext.PROPERTIES));
     });
 
     private static Function<FedoraResource, Function<IdentifierConverter<Resource, FedoraResource>, Function<Boolean,
@@ -242,9 +238,9 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
      * @see org.fcrepo.kernel.api.models.FedoraResource#getChildren()
      */
     @Override
-    public Iterator<FedoraResource> getChildren() {
+    public Stream<FedoraResource> getChildren() {
         try {
-            return concat(nodeToGoodChildren(node));
+            return nodeToGoodChildren(node);
         } catch (final RepositoryException e) {
             throw new RepositoryRuntimeException(e);
         }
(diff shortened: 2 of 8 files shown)


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
```diff
@@ -56,7 +56,7 @@
      * Get the children of this resource
      * @return iterator
```
ajs6f: Change this Javadoc?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -256,24 +252,15 @@ public String getPath() {
...
-                    if (input.isNodeType(FEDORA_PAIRTREE)) {
-                        return concat(nodeToGoodChildren(input));
-                    }
-                    return singletonIterator(nodeToObjectBinaryConverter.convert(input));
-                } catch (final RepositoryException e) {
-                    throw new RepositoryRuntimeException(e);
+    @SuppressWarnings("unchecked")
+    private Stream<FedoraResource> nodeToGoodChildren(final Node input) throws RepositoryException {
+        return iteratorToStream(input.getNodes()).filter(nastyChildren.negate())
+            .flatMap(uncheck((final Node x) -> {
```
ajs6f: Do we need the `()` around the param? Or the `{}` around the body? Actually. there's a general question: do we have style guides for lambdas? I don't remember.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -256,24 +252,15 @@ public String getPath() {
...
-            public Iterator<FedoraResource> apply(final Node input) {
-                try {
-                    if (input.isNodeType(FEDORA_PAIRTREE)) {
-                        return concat(nodeToGoodChildren(input));
-                    }
-                    return singletonIterator(nodeToObjectBinaryConverter.convert(input));
-                } catch (final RepositoryException e) {
-                    throw new RepositoryRuntimeException(e);
+    @SuppressWarnings("unchecked")
+    private Stream<FedoraResource> nodeToGoodChildren(final Node input) throws RepositoryException {
```
ajs6f: How can this throw a `RepositoryException`? Am I missing something? It's all one big `return`...

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/ChildrenRdfContext.java
```diff
@@ -62,29 +57,20 @@ public ChildrenRdfContext(final FedoraResource resource,
             LOGGER.trace("Found children of this resource: {}", resource.getPath());
 
             // Count the number of children
-            final Iterator<FedoraResource> childrenCounter = resource().getChildren();
-            concat(of(createNumChildrenTriple(Iterators.size(childrenCounter))));
-            concat(iteratorToStream(resource().getChildren()).map(child2triple));
-
+            concat(of(createNumChildrenTriple(resource().getChildren().count())));
+            concat(resource().getChildren().peek(x -> LOGGER.info("Creating triple for child node: {}", x))
+                    .map(x -> create(subject(), CONTAINS.asNode(), x instanceof NonRdfSourceDescription ?
```
ajs6f: Maybe use `child` instead of `x`? I hate the variable name `x`, even though I something use it.

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/FedoraResourceImplTest.java
```diff
@@ -388,8 +389,8 @@ public void testGetChildrenExcludesModeSystem() throws RepositoryException {
         when(mockNode.getNodes()).thenReturn(nodeIterator(mockChild));
         when(mockChild.isNodeType("mode:system")).thenReturn(true);
         when(mockChild.getName()).thenReturn("x");
-        final Iterator<FedoraResource> children = testObj.getChildren();
-        assertFalse("Expected an empty iterator", children.hasNext());
+        final Stream<FedoraResource> children = testObj.getChildren();
+        assertFalse("Expected an empty iterator", children.iterator().hasNext());
```
ajs6f: Why not `findFirst().isPresent()` as above?

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/FedoraResourceImplTest.java
```diff
@@ -397,16 +398,16 @@ public void testGetChildrenExcludesTombstones() throws RepositoryException {
         when(mockNode.getNodes()).thenReturn(nodeIterator(mockChild));
         when(mockChild.isNodeType(FEDORA_TOMBSTONE)).thenReturn(true);
         when(mockChild.getName()).thenReturn("x");
-        final Iterator<FedoraResource> children = testObj.getChildren();
-        assertFalse("Expected an empty iterator", children.hasNext());
+        final Stream<FedoraResource> children = testObj.getChildren();
+        assertFalse("Expected an empty iterator", children.iterator().hasNext());
```
ajs6f: Why not `findFirst().isPresent()` as above?

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/FedoraResourceImplTest.java
```diff
@@ -397,16 +398,16 @@ public void testGetChildrenExcludesTombstones() throws RepositoryException {
...
     }
 
     @Test
     public void testGetChildrenExcludesJcrContent() throws RepositoryException {
         when(mockNode.getNodes()).thenReturn(nodeIterator(mockChild));
         when(mockChild.getName()).thenReturn(JCR_CONTENT);
-        final Iterator<FedoraResource> children = testObj.getChildren();
-        assertFalse("Expected an empty iterator", children.hasNext());
+        final Stream<FedoraResource> children = testObj.getChildren();
+        assertFalse("Expected an empty iterator", children.iterator().hasNext());
```
ajs6f: Why not `findFirst().isPresent()` as above?

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/rdf/impl/ChildrenRdfContextTest.java
```diff
@@ -101,8 +99,16 @@ public void testNoChildren() throws RepositoryException {
 
     @Test
     public void testChildren() throws RepositoryException {
+        final FedoraResource mockRes1 = mock(FedoraResource.class);
```
ajs6f: Maybe use `@Mock FedoraResource mockRes1, mockRes2, mockRes3`?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
acoburn: You're right -- this function cannot throw an `Exception`

acoburn: I should have looked more closely. The `input.getNodes()` expression will throw a `RepositoryException`, so the `throws` bit needs to be left in there.

ajs6f: Ah, stupid JCR API.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
acoburn: AFAIK, we do not have a style guide for lambdas. And yes, it was necessary to put the `()` around the param, though I can remove the `{}` block

ajs6f: Hm, we should make up a style guide. Mine is simple: don't use any punctuation you don't have to. Just makes things harder to read.

acoburn:
Code review comments have all been addressed.

ajs6f:
Indeed they are. :+1:

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/85ad606706245761c1ba1c148f979230debea1bb

PR #992, opened 2016-02-23
change property2values function to return a Stream

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1926

### DIFF
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -15,13 +15,12 @@
  */
 package org.fcrepo.kernel.modeshape;
 
-import static com.google.common.collect.Iterators.transform;
-import static com.google.common.collect.Lists.newArrayList;
 import static com.hp.hpl.jena.update.UpdateAction.execute;
 import static com.hp.hpl.jena.update.UpdateFactory.create;
 import static java.util.Arrays.asList;
 import static java.util.EnumSet.of;
 import static java.util.stream.Collectors.joining;
+import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toMap;
 import static java.util.stream.Stream.empty;
 import static org.apache.commons.codec.digest.DigestUtils.shaHex;
@@ -343,15 +342,10 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
 
             while (inboundProperties.hasNext()) {
                 final Property prop = inboundProperties.next();
-                final List<Value> newVals = new ArrayList<>();
-                final Iterator<Value> propIt = property2values.apply(prop);
-                while (propIt.hasNext()) {
-                    final Value v = propIt.next();
-                    if (!node.equals(getSession().getNodeByIdentifier(v.getString()))) {
-                        newVals.add(v);
-                        LOGGER.trace("Keeping multivalue reference property when deleting node");
-                    }
-                }
+                final List<Value> newVals = property2values.apply(prop).filter(
+                        UncheckedPredicate.uncheck(x -> !node.equals(getSession().getNodeByIdentifier(x.getString()))))
+                    .collect(toList());
+
                 if (newVals.size() == 0) {
                     prop.remove();
                 } else {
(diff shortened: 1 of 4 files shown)


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/functions/AnyTypesPredicate.java
```diff
@@ -55,9 +52,8 @@ public AnyTypesPredicate(final String...types) {
     public boolean testThrows(final Node input) throws RepositoryException {
         requireNonNull(input, "null node has no types!");
         if (isFrozen.test(input) && input.hasProperty(FROZEN_MIXIN_TYPES)) {
-            final Property frozenTypesProperty = input.getProperty(FROZEN_MIXIN_TYPES);
-            final List<Value> values = copyOf(property2values.apply(frozenTypesProperty));
-            if (values.stream().map(uncheck(Value::getString)).anyMatch(nodeTypes::contains)) {
+            if (property2values.apply(input.getProperty(FROZEN_MIXIN_TYPES)).map(uncheck(Value::getString))
```
ajs6f: So this a style thing, but I would do something like:
```
final Stream<Value> values = property2values.apply(input.getProperty(FROZEN_MIXIN_TYPES));
return values.map(uncheck(Value::getString)).anyMatch(nodeTypes::contains))
|| nodeTypes.stream().anyMatch(uncheck(input::isNodeType));
```
but maybe that's less readable.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -360,15 +360,10 @@ public void delete() {
...
-                final Iterator<Value> propIt = property2values.apply(prop);
-                while (propIt.hasNext()) {
-                    final Value v = propIt.next();
-                    if (!node.equals(getSession().getNodeByIdentifier(v.getString()))) {
-                        newVals.add(v);
-                        LOGGER.trace("Keeping multivalue reference property when deleting node");
-                    }
-                }
+                final List<Value> newVals = property2values.apply(prop).filter(
+                        UncheckedPredicate.uncheck(x -> !node.equals(getSession().getNodeByIdentifier(x.getString()))))
```
awoods: Per @ajs6f's preference, shall we go with a variable name ("val"?) instead of `x`?

acoburn: I find `val` confusing because it makes me think of Scala's `val`. I could change it to `v`, though. Personally, I find using just `x` everywhere makes the lambdas _easier_ to read and make sense of.

awoods: I am fine any which way... was just looking out for @ajs6f.

awoods:
This PR looks good from my end. Please move it forward, @ajs6f, when you have a chance.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
ajs6f: I'm not actually fine with this (thanks for catching it). I'm at a loss as to how you can think that `x` is a better variable name than a meaningful one, @acoburn . Can you please use a proper name here, e.g. `value`?

awoods: I think having an informal policy of "meaningful variable names" makes good sense... and that could certainly apply here. 
I also think we should be careful to respect individual style.

ajs6f: Yeah, I'm conscious of the need for balance. I'm honestly very surprised, almost astonished, by @acoburn 's opinion that `x` is a more clear idiom...

acoburn:
I can't very well claim that `value` is _less_ clear than `x`, so `value` it is.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/f1b898f9c54b6a7f503c0a8cb668972ece9e3e17

PR #993, opened 2016-02-24
Resolve Jersey session injection error

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1785

This changes the SessionFactory in `fcrepo-http-commons` so that it doesn't throw a `TransactionMissingException` during Jersey injection (which causes lots of Jersey errors and is arguably a bad implementation) but rather returns a `null` value for a `Session` in the case of a missing transaction ID. This also led to a refactoring of `FedoraBaseResource` so that calls to `session()` all involve an appropriate `null` check.

The `TransactionMissingException` class was renamed to `SessionMissingException` to better capture what this exception really means with this implementation.

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/session/SessionFactory.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/session/SessionFactory.java
@@ -16,7 +16,7 @@
 package org.fcrepo.http.commons.session;
 
 import static java.util.Objects.requireNonNull;
-import static javax.ws.rs.core.Response.Status.GONE;
+//import static javax.ws.rs.core.Response.Status.GONE;
 import static org.slf4j.LoggerFactory.getLogger;
 
 import java.security.Principal;
@@ -28,10 +28,10 @@ import javax.jcr.RepositoryException;
 import javax.jcr.Session;
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.BadRequestException;
-import javax.ws.rs.ClientErrorException;
+//import javax.ws.rs.ClientErrorException;
 
 import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
-import org.fcrepo.kernel.api.exception.TransactionMissingException;
+import org.fcrepo.kernel.api.exception.SessionMissingException;
 import org.fcrepo.kernel.api.Transaction;
 import org.fcrepo.kernel.api.services.TransactionService;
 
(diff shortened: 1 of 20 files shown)


fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/session/SessionFactory.java
```diff
@@ -16,7 +16,7 @@
 package org.fcrepo.http.commons.session;
 
 import static java.util.Objects.requireNonNull;
-import static javax.ws.rs.core.Response.Status.GONE;
+//import static javax.ws.rs.core.Response.Status.GONE;
```
awoods: remove

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/session/SessionFactory.java
```diff
@@ -28,10 +28,10 @@
 import javax.jcr.Session;
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.BadRequestException;
-import javax.ws.rs.ClientErrorException;
+//import javax.ws.rs.ClientErrorException;
```
awoods: remove

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/session/SessionFactory.java
```diff
@@ -127,8 +127,10 @@ public Session getSession(final HttpServletRequest servletRequest) {
             } else {
                 session = getSessionFromTransaction(servletRequest, txId);
             }
-        } catch (final TransactionMissingException e) {
-            throw new ClientErrorException(GONE, e);
+        } catch (final SessionMissingException e) {
+            LOGGER.warn("Transaction missing: {}", e.getMessage());
+            return null;
+            //throw new ClientErrorException(GONE, e);
```
awoods: remove

awoods:
Did you skip these on purpose?
* https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/repository/FedoraRepositoryTransactions.java#L37
* https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/repository/FedoraRepositoryExport.java#L43
* https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/repository/FedoraRepositoryImport.java#L42

acoburn:
@awoods I removed the unnecessary `Session` injection. Thanks!

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/2f24542c07a0a9becbb1b99d4215081aac91bf04

PR #995, opened 2016-02-29
Refactor translator().reverse().convert(resource).asNode()

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-1932


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
```diff
@@ -56,6 +57,10 @@
         return idTranslator;
     }
 
+    protected Node asNode(final FedoraResource resource) {
```
ajs6f: Obviously, this is a good shortcut. Maybe a quick JavaDoc here just for those unfamiliar with the codebase?

awoods: Javadoc added...
Please do not commit this PR without first squashing these two commits. I would be happy to do the squash if you prefer.

ajs6f:
You squash, I'll merge. Mental note: we should probably revisit the Guava `Converter` abstraction on which we built the id translation classes. Post Java 8, there might be better ways to do this stuff.

awoods:
squashed

PR #996, opened 2016-03-01
rework HTML template rendering

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1934

This introduces an RDFS Class: `fedora:RepositoryRoot`. Once we have the green light on this impl, that class will need to be added to the core ontology.

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
@@ -15,18 +15,18 @@
  */
 package org.fcrepo.http.api.responses;
 
+import static java.util.stream.Stream.of;
 import static javax.ws.rs.core.MediaType.APPLICATION_XHTML_XML;
 import static javax.ws.rs.core.MediaType.APPLICATION_XHTML_XML_TYPE;
 import static javax.ws.rs.core.MediaType.TEXT_HTML;
 import static javax.ws.rs.core.MediaType.TEXT_HTML_TYPE;
 import static com.google.common.collect.ImmutableMap.builder;
 import static com.hp.hpl.jena.graph.Node.ANY;
-import static com.hp.hpl.jena.rdf.model.ResourceFactory.createProperty;
-import static com.hp.hpl.jena.rdf.model.ResourceFactory.createResource;
-import static org.apache.commons.lang3.StringUtils.isBlank;
-import static org.fcrepo.kernel.api.RdfLexicon.JCR_NAMESPACE;
+import static com.hp.hpl.jena.sparql.util.graph.GraphUtils.multiValueURI;
+import static com.hp.hpl.jena.vocabulary.RDF.type;
+import static org.fcrepo.kernel.api.RdfLexicon.LDP_NAMESPACE;
+import static org.fcrepo.kernel.api.RdfLexicon.REPOSITORY_NAMESPACE;
 import static org.fcrepo.kernel.api.RdfCollectors.toModel;
-import static org.fcrepo.kernel.modeshape.rdf.JcrRdfTools.getRDFNamespaceForJcrNamespace;
 import static org.slf4j.LoggerFactory.getLogger;
 
 import java.io.IOException;
@@ -37,18 +37,12 @@ import java.io.Writer;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
 import java.net.URL;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Properties;
 
 import javax.annotation.PostConstruct;
-import javax.jcr.RepositoryException;
-import javax.jcr.Session;
-import javax.jcr.nodetype.NodeType;
-import javax.jcr.nodetype.NodeTypeIterator;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.MultivaluedMap;
@@ -56,7 +50,6 @@ import javax.ws.rs.core.UriInfo;
 import javax.ws.rs.ext.MessageBodyWriter;
 import javax.ws.rs.ext.Provider;
 
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.hp.hpl.jena.graph.Node;
 import com.hp.hpl.jena.rdf.model.Model;
@@ -69,11 +62,8 @@ import org.apache.velocity.tools.generic.FieldTool;
 import org.fcrepo.http.commons.responses.HtmlTemplate;
 import org.fcrepo.http.commons.responses.RdfNamespacedStream;
 import org.fcrepo.http.commons.responses.ViewHelpers;
-import org.fcrepo.http.commons.session.SessionFactory;
 import org.fcrepo.kernel.api.RdfLexicon;
-import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
 import org.slf4j.Logger;
-import org.springframework.beans.factory.annotation.Autowired;
 
 /**
  * Simple HTML provider for RdfNamespacedStreams
@@ -85,9 +75,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 @Produces({TEXT_HTML, APPLICATION_XHTML_XML})
 public class StreamingBaseHtmlProvider implements MessageBodyWriter<RdfNamespacedStream> {
 
-
-    @Autowired
-    SessionFactory sessionFactory;
 
     @javax.ws.rs.core.Context
     UriInfo uriInfo;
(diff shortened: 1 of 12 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
```diff
@@ -236,32 +207,18 @@ private Template getTemplate(final Model rdf, final Node subject,
...
 
         if (!template.isPresent()) {
-            LOGGER.trace("Attempting to discover mixin types of node for resource in question: {}", subject);
-            template = rdf.listObjectsOfProperty(createResource(subject.getURI()),
-                                                 createProperty(getRDFNamespaceForJcrNamespace(JCR_NAMESPACE) +
-                                                     "mixinTypes"))
-                          .toList().stream()
-                          .map(x -> x.asLiteral().getLexicalForm())
+            LOGGER.trace("Attempting to discover RDF types of node for resource in question: {}", subject);
+            template = rdf.listObjectsOfProperty(subjectURI, type).toList().stream()
```
ajs6f: I don't think you need `toList().stream()` here. You can take the `NodeIterator` result of `listObjectsOfProperty` and use `mapWith` and `filterKeep` to get the same results.

acoburn: That's mostly true, but then I have to do some gymnastics to convert the iterator value (if present) into an `Optional`

ajs6f: Guava `Iterators::getlast`.

acoburn: The other issue is that since we're still on Jena 2.13, the jena.util.iterator.Filter doesn't implement a functional interface. We could use the Guava `filter` and `transform` functions, but then we'd have a lot more nested code, which IMO is much less readable than this.

ajs6f: That's more annoying. I wouldn't use Guava. I would use Jena's fluent `Iter`, e.g. `static import Iter.iter` and then
```
iter(listObjectsOfProperty(subjectURI, type))
    .map(Node::.getURI)
    .filter(typeMap::containsKey)
    .map(typeMap::get)
```
etc.

acoburn:
@ajs6f the PR now uses Jena's `Iter.iter` function.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
```diff
@@ -143,44 +137,23 @@ void init() throws IOException {
...
+        templatesMap = templatesMapBuilder.build();
+
+        final ImmutableMap.Builder<String, String> typeMapBuilder = builder();
+        typeMapBuilder.put(REPOSITORY_NAMESPACE + "RepositoryRoot", "root");
+        typeMapBuilder.put(REPOSITORY_NAMESPACE + "Binary", "binary");
+        typeMapBuilder.put(REPOSITORY_NAMESPACE + "Version", "resource");
+        typeMapBuilder.put(REPOSITORY_NAMESPACE + "Pairtree", "resource");
+        typeMapBuilder.put(REPOSITORY_NAMESPACE + "Container", "resource");
+        typeMapBuilder.put(LDP_NAMESPACE + "NonRdfSource", "binary");
+        typeMapBuilder.put(LDP_NAMESPACE + "RdfSource", "resource");
```
ajs6f: Style-wise, I tend to prefer:
```
typeMap = ImmutableMap.Builder<String, String>builder()
    .put(REPOSITORY_NAMESPACE + "RepositoryRoot", "root")
    .put(REPOSITORY_NAMESPACE + "Binary", "binary")
    etc.
   .build()
```

ajs6f: Did you try `static import ImmutableMap.builder`? Sometimes the type inferencing works.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
```diff
@@ -236,32 +210,17 @@ private Template getTemplate(final Model rdf, final Node subject,
...
 
         if (!template.isPresent()) {
-            LOGGER.trace("Attempting to discover mixin types of node for resource in question: {}", subject);
-            template = rdf.listObjectsOfProperty(createResource(subject.getURI()),
-                                                 createProperty(getRDFNamespaceForJcrNamespace(JCR_NAMESPACE) +
-                                                     "mixinTypes"))
-                          .toList().stream()
-                          .map(x -> x.asLiteral().getLexicalForm())
+            LOGGER.trace("Attempting to discover RDF types of node for resource in question: {}", subject);
+            template = ofNullable(getFirst(iter(rdf.listObjectsOfProperty(subjectURI, type))
```
ajs6f: This seems better to me, although I honestly don't see much value to `Optional` here. A possibly `null`-valued variable actually seems shorter and cleaner. If it's not actually the return type of a method, I tend to avoid `Optional`.

acoburn: We're using an `Optional` because `findFirst` returns an `Optional` in the block above (line 209)

ajs6f: I know, I'm saying I would resolve that monad immediately. But it's not a big thing. If you think that would be less clear, fine.

acoburn:
@ajs6f more code updates -- the `Optional` monad is now immediately resolved.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
```diff
@@ -228,40 +196,17 @@ protected Context getContext(final Model model, final Node subject) {
...
-                          .filter(templatesMap::containsKey)
-                          .map(templatesMap::get)
-                          .findFirst()
-                          .orElse(templatesMap.get("node"));
-        }
+        return annotations.stream().filter(x -> x instanceof HtmlTemplate).map(x -> ((HtmlTemplate) x).value())
+            .filter(templatesMap::containsKey).map(templatesMap::get).findFirst().orElseGet(() -> {
+                LOGGER.trace("Attempting to discover RDF types of node for resource in question: {}", subject);
+                final Template tpl = getFirst(iter(rdf.listObjectsOfProperty(createResource(subject.getURI()), type))
+                              .map(x -> x.asResource().getURI()).filter(x -> typeMap.containsKey(x))
```
ajs6f: How come `x -> x.asResource().getURI()`? Isn't `Node::getURI` fine here? And why move from `typeMap::containsKey` to `x -> typeMap.containsKey(x)`. Otherwise, I like this better.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
```diff
@@ -228,40 +196,17 @@ protected Context getContext(final Model model, final Node subject) {
...
+                              .map(typeMap::get).filter(templatesMap::containsKey).map(templatesMap::get), null);
+                if (tpl != null) {
```
ajs6f: Could be just `return tpl != null ? tpl : templatesMap.get("default")`. But that's just my taste.

ajs6f: Actually, there's `MoreObjects::firstNonNull`, too.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
```diff
@@ -143,44 +134,21 @@ void init() throws IOException {
...
-            throw new RepositoryRuntimeException(e);
-        }
+
+        of("jcr:nodetypes", "fcr:versions", "fcr:fixity", "binary", "resource", "root", "default")
+            .forEach(key -> templatesMapBuilder.put(key, velocity.getTemplate(getTemplateLocation(key))));
+
+        templatesMap = templatesMapBuilder.build();
+
+        final ImmutableMap.Builder<String, String> typeMapBuilder = builder();
+        typeMap = typeMapBuilder.put(REPOSITORY_NAMESPACE + "RepositoryRoot", "root")
```
acoburn: I agree with you there -- the issue I was trying to avoid was having the same `Template` object as the value of multiple `Map<String, Template>` entries. For example, fedora:Version, fedora:Pairtree and fedora:Container all need to result in the same `Template`. If that's not a concern, then I can certainly remove that `typeMap`.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
ajs6f: Or maybe, assuming we drop the type map stuff as suggested above, we just do:
```
final Resource resource = createResource(subject.getURI());
final String templateName = getFirst(iter(rdf.listObjectsOfProperty(resource, type)).map(Node::getURI), "default");
return templatesMap.get(templateName);
```
or you could add the type map stuff back in, although I wouldn't. The point is to move the "get first" forward to the stream of names.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
ajs6f: I have no problem with that at all, and I don't see how it would result in any later problems. References are cheap, but distinguishing types is expensive, so let's have lots of types all pointing at the same thing.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
acoburn: Oh, that's nice.

ajs6f: If there's only one value for a default, we should only have to make that choice once. I think. Or maybe I'm looking at the wrong project. Is this Jena or Fedora?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
ajs6f: Ah, right, right. Okay, you're going to freaking shoot me, but what about `com.hp.hpl.jena.sparql.util.graph.GraphUtils.multiValueURI(Resource, Property)`? I.e. like:
```
final Resource resource = rdf.getResource(subject.getURI());
final String templateName = getFirst(multiValueURI(resource, type),"default"));
return templatesMap.get(templateName);
```

acoburn:
@ajs6f more updates applied.

ajs6f:
I'm down with it. I'll wait a day in case anyone else wants to opine, otherwise I'll merge it tomorrow.

awoods:
`private StreamingBaseHtmlProvider.addTemplate()` can now be removed... dead code.
https://github.com/fcrepo4/fcrepo4/pull/996/files#diff-479f0eb6453708a90e8c7cc92f0657a6R220

acoburn:
@awoods thanks for spotting that -- I removed the dead code.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/RootRdfContext.java
```diff
@@ -71,6 +74,8 @@ public RootRdfContext(final FedoraResource resource,
 
         final Stream.Builder<Triple> b = Stream.builder();
 
+        b.accept(create(subject(), type.asNode(), createURI(REPOSITORY_NAMESPACE + "RepositoryRoot")));
```
awoods: We should make a note (new JIRA ticket?) to include `fedora:RepositoryRoot` in the Fedora ontology.

acoburn: @awoods see https://github.com/fcrepo4/ontology/pull/35 as a follow-on to the existing JIRA ticket

awoods: :+1:

awoods:
..an additional PR to clean up the `StreamingBaseHtmlProviderTest`:
https://github.com/acoburn/fcrepo4/pull/4

awoods:
This looks good to go. We can wait for tomorrow or squash and merge now.

ajs6f:
@acoburn Can you squash?

acoburn:
@ajs6f squashed

ajs6f:
Merging.

PR #997, opened 2016-03-02
add fedora RDF types to Root resource

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1939


awoods:
:+1:

ajs6f:
Just a note to self: we have stuff like `REPOSITORY_NAMESPACE + "RepositoryRoot"` all over. At some point we should pull out full constants for those URIs. Otherwise there will be pain when we have to update the name of an ontology type, which sooner or later we will.

PR #998, opened 2016-03-02
Clean-up related to: https://jira.duraspace.org/browse/FCREPO-1878

awoods:


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -291,7 +291,7 @@ public Response createOrReplaceObjectRdf(
             @HeaderParam("Content-Disposition") final ContentDisposition contentDisposition,
             @HeaderParam("If-Match") final String ifMatch,
             @HeaderParam("Link") final String link)
-            throws IOException, InvalidChecksumException, MalformedRdfException {
+            throws InvalidChecksumException, MalformedRdfException {
```
acoburn: `MalformedRdfException` is a `RuntimeException`, so is it necessary to declare it?

awoods: Probably not necessary.

acoburn: OK, perhaps that can be cleaned up as part of another ticket -- there are plenty of method signatures in the codebase with declared `RuntimeException`s.

awoods: Agreed... lots of opportunities for future cleanup.

PR #999, opened 2016-03-02
fix template selection in cases where the RDF types are not disjoint

acoburn:
See: https://jira.duraspace.org/browse/FCREPO-1939

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
@@ -196,9 +196,8 @@ public class StreamingBaseHtmlProvider implements MessageBodyWriter<RdfNamespace
                 final List<String> types = multiValueURI(rdf.getResource(subject.getURI()), type);
                 if (types.contains(REPOSITORY_NAMESPACE + "RepositoryRoot")) {
                     return REPOSITORY_NAMESPACE + "RepositoryRoot";
-                } else {
-                    return types.stream().filter(templatesMap::containsKey).findFirst().orElse("default");
                 }
+                return types.stream().filter(templatesMap::containsKey).findFirst().orElse("default");
             });
         LOGGER.debug("Using template: {}", tplName);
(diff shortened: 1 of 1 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
```diff
@@ -192,8 +192,14 @@ private Template getTemplate(final Model rdf, final Node subject,
 
         final String tplName = annotations.stream().filter(x -> x instanceof HtmlTemplate)
             .map(x -> ((HtmlTemplate) x).value()).filter(templatesMap::containsKey).findFirst()
-            .orElseGet(() -> multiValueURI(rdf.getResource(subject.getURI()), type).stream()
-                    .filter(templatesMap::containsKey).findFirst().orElse("default"));
+            .orElseGet(() -> {
+                final List<String> types = multiValueURI(rdf.getResource(subject.getURI()), type);
+                if (types.contains(REPOSITORY_NAMESPACE + "RepositoryRoot")) {
+                    return REPOSITORY_NAMESPACE + "RepositoryRoot";
+                } else {
```
awoods: Channeling @ajs6f: Since the `if` block has a `return`, you should be able to simply omit `else` and follow the `if` block with: 
"return types.stream().filter(templatesMap::containsKey).findFirst().orElse("default");"

ajs6f: Yes, the `else` isn't needed here. If you are using Eclipse, there is definitely a setting to notify you about this kind of thing. I imagine other IDEs also can do that.

acoburn: I removed the `else`. It's just habit from other languages where `if` - `else` represent expressions rather than control flow.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/cae2ac8d8cc152643172b5b60820f5bdca5dcc90

PR #1001, opened 2016-03-03
Confine all internal mode/jcr properties and type to fcrepo-kernel-modeshape

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1935

Confines all internal modeshape/jcr types and properties to the fcrepo-kernel-modeshape module.

Previously, these values were escaping to the fcrepo-http-api layer and then filtered out (because of how template rendering worked).

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -73,6 +73,7 @@ import static org.fcrepo.kernel.api.RdfLexicon.HAS_PRIMARY_IDENTIFIER;
 import static org.fcrepo.kernel.api.RdfLexicon.HAS_PRIMARY_TYPE;
 import static org.fcrepo.kernel.api.RdfLexicon.INBOUND_REFERENCES;
 import static org.fcrepo.kernel.api.RdfLexicon.INDIRECT_CONTAINER;
+import static org.fcrepo.kernel.api.RdfLexicon.JCR_NAMESPACE;
 import static org.fcrepo.kernel.api.RdfLexicon.JCR_NT_NAMESPACE;
 import static org.fcrepo.kernel.api.RdfLexicon.LDP_MEMBER;
 import static org.fcrepo.kernel.api.RdfLexicon.LDP_NAMESPACE;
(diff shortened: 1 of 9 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -73,6 +73,7 @@
 import static org.fcrepo.kernel.api.RdfLexicon.HAS_PRIMARY_TYPE;
 import static org.fcrepo.kernel.api.RdfLexicon.INBOUND_REFERENCES;
 import static org.fcrepo.kernel.api.RdfLexicon.INDIRECT_CONTAINER;
+import static org.fcrepo.kernel.api.RdfLexicon.JCR_NAMESPACE;
```
ajs6f: I'm not saying you're wrong here, but if we are serous about people eventually picking up the `kernel-api`, impling it, and then using the HTTP modules with that impl, we will eventually want to remove this kind of JCR-related constant from the API entirely (even from our own namespaces) so this kind of testing is a little weird.

acoburn: I agree that the testing is a little weird. An alternative is to just trust that the testing at the fcrepo-kernel-modeshape level is complete, thereby removing theses sorts of tests from the fcrepo-http-api module. For now, though, I'd prefer to keep these tests in place to guarantee that those JCR types aren't escaping.

ajs6f: I'm cool with it. I was more just reminding us about the long term. We can do Maven tricks to prevent JCR stuff from encrusting alt impls.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersionsIT.java
```diff
@@ -453,9 +455,11 @@ public void testDatastreamAutoMixinAndRevert() throws IOException {
         // datastream should not be versionable
         try (final CloseableGraphStore originalObjectProperties =
                 getContent(serverAddress + pid + "/" + dsid + "/fcr:metadata")) {
-            assertFalse("Node must not have versionable mixin.",
-                    originalObjectProperties.contains(ANY, createURI(serverAddress + pid + "/" + dsid),
-                            type.asNode(), createURI(MIX_NAMESPACE + "versionable")));
+            final Node subject = createURI(serverAddress + pid + "/" + dsid);
+            assertFalse("Node must not have versionable mixin.", originalObjectProperties.contains(ANY,
+                    subject, type.asNode(), createURI(MIX_NAMESPACE + "versionable")));
+            assertFalse("Node is expected to contain hasVersions triple.", originalObjectProperties.contains(ANY,
```
awoods: Should this text say: `Node is <not> expected...` instead?

acoburn: yes, good catch.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/02f98e1ac64cb9537320cab43ab520d761f744fd

PR #1003, opened 2016-03-06
Emit complete events for MOVE operations

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1937

This PR also refactors much of the `Iterator`-based event machinery into `Stream`-based data flows.

The implementation emits a NODE_MOVED event both for the moved resource and for all child resources. A NODE_REMOVED event is emitted for the _old_ location of the moved resource and the former locations of all children.


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/DefaultFilter.java
```diff
@@ -59,7 +60,8 @@
     @Override
     public boolean test(final Event event) {
         try {
-            return getMixinTypes(event).anyMatch(fedoraMixins::contains);
+            // NODE_MOVED events do not contain a mixin type
+            return getMixinTypes(event).anyMatch(fedoraMixins::contains) || event.getType() == NODE_MOVED;
```
ajs6f: I this going to catch nonFedora stuff moving around in the JCR for MODE system purposes?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/FedoraEventImpl.java
```diff
@@ -199,6 +199,20 @@ public String getEventID() {
         return eventID;
     }
 
+    /**
+     * Extract a Map with any additional information from the underlying JCR Event.
+     * @return a Map of additional information.
+     */
+    @Override
+    @SuppressWarnings("unchecked")
```
ajs6f: I like to use a var inside the method to hold suppressions so that they aren't visible at the method (which, to me, is an inexact semantic), but not everyone does. Do we have a policy on this?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/JcrEvent.java
```diff
@@ -0,0 +1,97 @@
...
+import static java.util.UUID.randomUUID;
+
+import java.util.Map;
+
+import javax.jcr.observation.Event;
+
+/**
+ * @author acoburn
+ */
+public class JcrEvent implements Event {
```
ajs6f: Do you want to just cash out `FedoraEventImpl` in exchange for this class and a factory? Or putting a `static` `from(Event e)` factory method on this class that just copies the data?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/SimpleObserver.java
```diff
@@ -68,6 +81,53 @@
...
+        evt.getProperties().stream()
+            .filter(property -> validJcrProperties.contains(property) || !property.startsWith("jcr:"))
+            .forEach(property -> {
+                final String[] parts = property.split(":", 2);
+                if (parts.length == 2) {
+                    try {
+                        event.addProperty(
+                            getRDFNamespaceForJcrNamespace(namespaceRegistry.getURI(parts[0])) + parts[1]);
+                    } catch (final RepositoryException ex) {
+                        LOGGER.debug("Prefix could not be dereferenced using the namespace registry: {}", property);
```
ajs6f: Is this just a log-and-move-on situation? If the prefix can't be resolved, what exactly are we publishing?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/SimpleObserver.java
```diff
@@ -68,6 +81,53 @@
...
+                } else {
+                    event.addProperty(property);
+                }
+            });
+        evt.getTypes().forEach(event::addType);
+        return event;
+    };
+
+    private static Function<Session, Function<FedoraEvent, Stream<FedoraEvent>>> handleMoveEvents = session -> evt -> {
+        if (evt.getTypes().contains(org.fcrepo.kernel.api.observer.EventType.NODE_MOVED)) {
```
ajs6f: Static import `NODE_MOVED`?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/eventmappings/AllNodeEventsOneEvent.java
```diff
@@ -66,67 +67,30 @@
...
-                }
-                if (PROPERTY_EVENT_TYPES.contains(ev.getType())) {
-                    final String eventPath = ev.getPath();
-                    fedoraEvent.addProperty(eventPath.substring(eventPath.lastIndexOf('/') + 1));
-                } else {
-                    LOGGER.trace("Not adding non-event property: {}, {}", fedoraEvent, ev);
-                }
-            } catch (final RepositoryException e) {
-                throw new RepositoryRuntimeException(e);
+    public Stream<FedoraEvent> apply(final Stream<Event> events) {
```
ajs6f: I believe that this new method is correct, but how about some in-line comments for the poor bastard who has to alter it two years from now? Could be you or me.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/FedoraTypesUtils.java
```diff
@@ -78,7 +78,7 @@
             FROZEN_MIXIN_TYPES,
             FROZEN_PRIMARY_TYPE);
 
-    private static Set<String> validJcrProperties = of(
+    public static Set<String> validJcrProperties = of(
```
ajs6f: If this becomes `public` it should be equipped with a doc-comment of some kind.

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/integration/kernel/modeshape/observer/MoveObserverIT.java
```diff
@@ -0,0 +1,156 @@
...
+    public void TestEventBusPublishing() throws RepositoryException {
+
+        final Session se = repository.login();
+        se.getRootNode().addNode("/object1").addMixin(FEDORA_CONTAINER);
+        se.getRootNode().addNode("/object2").addMixin(FEDORA_CONTAINER);
+        se.save();
+        se.logout();
+
+        try {
+            sleep(500);
```
ajs6f: This is not a very strong test. How about using Awaitility to build something a little less brittle?

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/integration/kernel/modeshape/observer/MoveObserverIT.java
```diff
@@ -0,0 +1,156 @@
...
+        n.addMixin(FEDORA_CONTAINER);
+        n.addNode("/child1").addMixin(FEDORA_CONTAINER);
+        n.addNode("/child2").addMixin(FEDORA_CONTAINER);
+        se.save();
+        ns.moveObject(se, "/object4", "/object5");
+        se.save();
+        se.logout();
+
+        try {
+            sleep(500);
```
ajs6f: See above re: Awaitility.

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/observer/eventmappings/AllNodeEventsOneEventTest.java
```diff
@@ -78,14 +78,11 @@
     @Mock
     private Event mockEvent5;
 
-    @Mock
-    private Iterator<Event> mockIterator;
+    private Stream<Event> mockStream;
```
ajs6f: :+1:  to using actual `Stream`s for this testing. Maybe use actual `Event`s, too.

acoburn:
@ajs6f I have added an incremental change to address your comments. This last commit removes the `JcrEvent` class as well as the wrapped `javax.jcr.observation.Event` object from within the `FedoraEventImpl` class. This way we don't need the additional implementation of a jcr Event and can simply use `FedoraEventImpl::from(Event)` to translate from an `Event` to a `FedoraEvent`. Is this along the lines of what you were thinking?

acoburn:
@ajs6f this is now ready for your review

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/SimpleObserver.java
```diff
@@ -68,6 +82,60 @@
     static final Integer EVENT_TYPES = NODE_ADDED + NODE_REMOVED + NODE_MOVED + PROPERTY_ADDED + PROPERTY_CHANGED
             + PROPERTY_REMOVED;
 
+    private static Stream<FedoraResource> getAllChildren(final FedoraResource resource) {
```
ajs6f: Should this be a method on `FedoraResource`? Different impls may have important optimizations for it.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/SimpleObserver.java
```diff
@@ -68,6 +82,60 @@
...
+                            property);
+                    }
+                } else {
+                    LOGGER.warn("Prefix could not be determined, skipping: {}", property);
+                }
+            });
+        return event;
+    };
+
+    private static Function<Session, Function<FedoraEvent, Stream<FedoraEvent>>> handleMoveEvents = session -> evt -> {
```
ajs6f: Can we have a method-level doc explaining the semantics of this function? I understand what it does-- I want to doc what it is _supposed_ to do.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/FedoraTypesUtils.java
```diff
@@ -128,6 +128,12 @@
             type.startsWith("mix:");
 
     /**
+     * Predicate for determining whether a JCR property should be converted to the fedora namespace.
+     */
+    public static Predicate<String> isPublicJcrProperty = property ->
```
ajs6f: Is this just `validJcrProperties::contains`?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/SimpleObserver.java
acoburn: Possibly? how about a specialization of the existing function like so: `FedoraResource::getChildren(final boolean recursive)`? Then, with no params, you get the direct descendants, and with a `recursive=true` you get all children, recursively.

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/integration/kernel/modeshape/observer/SimpleObserverIT.java
```diff
@@ -116,30 +125,94 @@ public void contentEventCollapsing() throws RepositoryException, InvalidChecksum
...
+        awaitEvent("/object6/object7", NODE_REMOVED);
+        awaitEvent("/object6/object7/child1", NODE_REMOVED);
+        awaitEvent("/object6/object7/child2", NODE_REMOVED);
+        // should produce two of these
+        awaitEvent("/object6", PROPERTY_CHANGED);
+
+        assertEquals("Move operation didn't generate additional events", (Integer) 12, eventBusMessageCount);
+    }
+
+    private void awaitEvent(final String id, final EventType eventType, final String property) {
```
ajs6f: Much better than just `sleep`ing, but is there a timeout on this method?

acoburn: There is a timeout on all the methods that call `awaitEvent`. Do you think it is necessary to also have a timeout on this method?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/SimpleObserver.java
ajs6f: Yeah, that allow a simple migration with a `default` impl for the current method. Are there other cases? I.e. There's recursive/non-recursive, are there other characteristics we need to encode?

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/integration/kernel/modeshape/observer/SimpleObserverIT.java
ajs6f: Not additionally, more that you could get rid of the multiple timeouts, and if someone wants to add a test later, they wouldn't have to remember to add a timeout themselves. Not a blocker-- your call.

acoburn: Like the way there is no `timeout` annotation on the `TestEventBusPublishing` method? https://github.com/fcrepo4/fcrepo4/pull/1003/files#diff-168576a30696166dad8294a45745cbe2L75 

Sounds like a good idea.

ajs6f: Weird that that test method name is capitalized. Oh, well.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
```diff
@@ -60,6 +60,13 @@
     Stream<FedoraResource> getChildren();
```
ajs6f: `default Stream<FedoraResource> getChildren() { return getChildren(false); }`?

acoburn: Oh right, give this latest commit a look.

ajs6f:
Okay, I'll let it hang out until tomorrow in case someone else has comments, then I'll merge. Thanks, @acoburn !

PR #1004, opened 2016-03-11
Fix malformed path in JMS events for URI fragments

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1528

### DIFF
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/FedoraEventImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/FedoraEventImpl.java
@@ -253,6 +253,11 @@ public class FedoraEventImpl implements FedoraEvent {
         }
     }
 
+    /**
+     * The JCR-based Event::getPath contains some Modeshape artifacts that must be removed or modified in
+     * order to correspond to the public resource path. For example, JCR Events will contain a trailing
+     * /jcr:content for Binaries, a trailing /propName for properties, and /#/ notation for URI fragments.
+     */
     private static String cleanPath(final Event event) throws RepositoryException {
         // remove any trailing data for property changes
         final String path = PROPERTY_TYPES.contains(event.getType()) ?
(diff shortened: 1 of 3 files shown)


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/FedoraEventImpl.java
```diff
@@ -255,5 +253,15 @@ public static FedoraEvent from(final Event event) {
         }
     }
 
+    private static String cleanPath(final Event event) throws RepositoryException {
+        // remove any trailing data for property changes
+        final String path = PROPERTY_TYPES.contains(event.getType()) ?
+            event.getPath().substring(0, event.getPath().lastIndexOf("/")) : event.getPath();
+
+        // reformat any hash URIs and remove any trailing /jcr:content
+        final HashConverter converter = new HashConverter();
```
ajs6f: Is `HashConverter` stateful? If not, can this be a `static` field, instead of building a new object over and over? The compiler/runtime might be smart enough to factor this out, but we should help it as much as possible.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/FedoraEventImpl.java
```diff
@@ -255,5 +253,15 @@ public static FedoraEvent from(final Event event) {
         }
     }
 
+    private static String cleanPath(final Event event) throws RepositoryException {
```
ajs6f: Maybe a real brief comment as to what "cleaning" means?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/FedoraEventImpl.java
acoburn: Sadly, no, the `Converter` classes are all stateful. This is another reason why it may be worthwhile to refactor our use of these away from Guava's `Converter` classes, but that is WAAAAY beyond the scope of this ticket. (They could easily be refactored into pairs of curried functions, but it would have wide-ranging implications for the entire codebase.)

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/FedoraEventImpl.java
acoburn: I added some method-level comments about the purpose of this `cleanPath` function.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/161d42666e186d20eee6d28cf6d7a45018d7d306

PR #1005, opened 2016-03-13
Ensure versioned child resources do not result in 404

awoods:
Resolves:
- https://jira.duraspace.org/browse/FCREPO-1942
- https://jira.duraspace.org/browse/FCREPO-1944

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
@@ -242,4 +242,11 @@ public interface FedoraResource {
      * @return the node for this object at the version provided
      */
     Node getNodeVersion(String label);
+
+    /**
+     * This method returns the version label of this frozen resource.
+     * If this resource is not frozen, null is returned.
+     * @return version label
+     */
+    String getVersionLabelOfFrozenResource();
 }
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -757,9 +757,8 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
                 // ..Or it is a child-version-history on a frozen path
             } else if (hasProperty(JCR_CHILD_VERSION_HISTORY)) {
                 final Node childVersionHistory = getNodeByProperty(getProperty(JCR_CHILD_VERSION_HISTORY));
-                final Node childNode;
                 try {
-                    childNode = getNodeByProperty(childVersionHistory.getProperty(JCR_VERSIONABLE_UUID));
+                    final Node childNode = getNodeByProperty(childVersionHistory.getProperty(JCR_VERSIONABLE_UUID));
                     return new FedoraResourceImpl(childNode);
                 } catch (ItemNotFoundException e) {
                     // The unfrozen resource has been deleted, return the tombstone.
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -747,13 +747,24 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
         try {
             // Either this resource is frozen
             if (hasProperty(JCR_FROZEN_UUID)) {
-                return new FedoraResourceImpl(getNodeByProperty(getProperty(JCR_FROZEN_UUID)));
+                try {
+                    return new FedoraResourceImpl(getNodeByProperty(getProperty(JCR_FROZEN_UUID)));
+                } catch (ItemNotFoundException e) {
+                    // The unfrozen resource has been deleted, return the tombstone.
+                    return new TombstoneImpl(getNode());
+                }
 
                 // ..Or it is a child-version-history on a frozen path
             } else if (hasProperty(JCR_CHILD_VERSION_HISTORY)) {
                 final Node childVersionHistory = getNodeByProperty(getProperty(JCR_CHILD_VERSION_HISTORY));
-                final Node childNode = getNodeByProperty(childVersionHistory.getProperty(JCR_VERSIONABLE_UUID));
-                return new FedoraResourceImpl(childNode);
+                final Node childNode;
+                try {
+                    childNode = getNodeByProperty(childVersionHistory.getProperty(JCR_VERSIONABLE_UUID));
+                    return new FedoraResourceImpl(childNode);
+                } catch (ItemNotFoundException e) {
+                    // The unfrozen resource has been deleted, return the tombstone.
+                    return new TombstoneImpl(childVersionHistory);
+                }
 
             } else {
                 throw new RepositoryRuntimeException("Resource must be frozen or a child-history!");
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -78,6 +78,8 @@ import javax.jcr.nodetype.NodeType;
 import javax.jcr.version.Version;
 import javax.jcr.version.VersionHistory;
 import javax.jcr.NamespaceRegistry;
+import javax.jcr.version.VersionIterator;
+import javax.jcr.version.VersionManager;
 
 import com.google.common.base.Converter;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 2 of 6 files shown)


whikloj:
Okay I tested this against the two tickets and it does seem to solve them both. 
So kudos to you @awoods. 

I don't fully understand what you are doing in it though. The tombstone stuff seems to mean you can access the path even thought the resource is gone? It appears you are creating a tombstone on the fly for use in the request...is that correct?

Either way I can merge this if you don't mind that I don't fully understand it. Do you want it squashed as well?

awoods:
I think the three commits can stand alone, squash-free.
Yes, the tombstones are created on the fly in the situations where the "real" resource has been removed.

awoods:
p.s. If you do not mind, this PR also addresses:
- https://jira.duraspace.org/browse/FCREPO-1817 and
- https://jira.duraspace.org/browse/FCREPO-1817

ajs6f:
Hang on, what does it mean to be "creating a tombstone on the fly"? Does that mean that a tombstone is accessible via the HTTP API?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
```diff
@@ -242,4 +242,11 @@ void replaceProperties(final IdentifierConverter<Resource, FedoraResource> idTra
      * @return the node for this object at the version provided
      */
     Node getNodeVersion(String label);
+
+    /**
+     * This method returns the version label of this frozen resource.
+     * If this resource is not frozen, null is returned.
+     * @return version label
+     */
+    String getVersionLabelOfFrozenResource();
```
ajs6f: This seems like a really JCR-specific method to expose on this interface in the API.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -739,7 +745,30 @@ public FedoraResource getUnfrozenResource() {
...
+                    return new FedoraResourceImpl(getNodeByProperty(getProperty(JCR_FROZEN_UUID)));
+                } catch (ItemNotFoundException e) {
+                    // The unfrozen resource has been deleted, return the tombstone.
+                    return new TombstoneImpl(getNode());
+                }
+
+                // ..Or it is a child-version-history on a frozen path
+            } else if (hasProperty(JCR_CHILD_VERSION_HISTORY)) {
+                final Node childVersionHistory = getNodeByProperty(getProperty(JCR_CHILD_VERSION_HISTORY));
+                final Node childNode;
```
ajs6f: Why not just declare the variable as it is set? There's no clarity gained by declaring it outside the `try`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -774,6 +800,62 @@ public Node getNodeVersion(final String label) {
...
+                final Node childNode = getNodeByProperty(childVersionHistory.getProperty(JCR_VERSIONABLE_UUID));
+                final Version childVersion = getVersionManager().getBaseVersion(childNode.getPath());
+                frozenResource = childVersion.getFrozenNode();
+
+            } else {
+                frozenResource = getNode();
+            }
+
+            // Loop versions
+            final VersionIterator versions = versionHistory.getAllVersions();
```
ajs6f: Use `Iterator<Version>` here at the cost of a known-good typecheck and you can use `forEachRemaining` which is shorter.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
awoods: The concept of "FrozenResource" does leave a taste of JCR. The point of the method is to get the label of a version. I was keeping the terminology in line with the other methods on this interface:
- `isFrozenResource`: https://github.com/awoods/fcrepo4/blob/fcrepo-1942/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java#L225
- `getUnfrozenResource`: https://github.com/awoods/fcrepo4/blob/fcrepo-1942/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java#L237

..but would be happy to change the name if you have a suggestion.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
awoods: Good point.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
awoods: This is a marginal difference in style with no apparent benefit in functionality. I would prefer to leave it as is... but would also be happy to merge in a patch to this PR if you would like to provide it.

ajs6f: The benefit is leaving the traversal to the things traversed, which is always the correct approach. See @acoburn 's PR for an even better form.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -774,6 +799,55 @@ public Node getNodeVersion(final String label) {
...
+            }
+
+            // Loop versions
+            @SuppressWarnings("unchecked")
+            final Stream<Version> versions = iteratorToStream(versionHistory.getAllVersions());
+            return versions
+                .filter(UncheckedPredicate.uncheck(version -> version.getFrozenNode().equals(frozenResource)))
+                .map(uncheck(versionHistory::getVersionLabels))
+                .filter(ArrayUtils::isNotEmpty)
+                .map(labels -> labels[0])
```
acoburn: Actually, you can replace `.filter(ArrayUtils::isNotEmpty).map(labels -> labels[0])` with `.flatMap(Arrays::stream)`, which is a simpler way to express this.

escowles:
:+1: 

Looks good, and I just verified that this resolves 1715, 1817, 1942 and 1944.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
ajs6f: Is that actually the same semantic? Isn't `labels[0]` picking out an element?

awoods:
Thanks, @escowles. It would be great to move this through for the 4.5.1 code freeze (this week?).
Let's wait for @acoburn and @ajs6f to work out their streaming [polish](https://github.com/fcrepo4/fcrepo4/pull/1005#discussion_r55998269)... then onward.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -774,6 +799,55 @@ public Node getNodeVersion(final String label) {
...
+                final Node childVersionHistory = getNodeByProperty(getProperty(JCR_CHILD_VERSION_HISTORY));
+                final Node childNode = getNodeByProperty(childVersionHistory.getProperty(JCR_VERSIONABLE_UUID));
+                final Version childVersion = getVersionManager().getBaseVersion(childNode.getPath());
+                frozenResource = childVersion.getFrozenNode();
+
+            } else {
+                frozenResource = getNode();
+            }
+
+            // Loop versions
```
ajs6f: I'm not sure this comment quite makes sense any more...

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
ajs6f: It's not the name (although that is a little weird) it's that we're passing around notions like "frozen" and "unfrozen" which are themselves JCR-specific. Given that this solves a couple of outstanding tickets, I'm not going to worry about it now, but for the long term, we need to look hard at the `kernel-api` module and get rid of this stuff, replacing it with types and notions that actual reflect our modeling commitments.

awoods: :+1: to giving the `kernel-api` a thorough review from the perspective of modeling commitments... as a follow-on task.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
acoburn: yes, `labels[0]` is picking an element, but so is `findFirst()` (below). Using `Arrays::stream` simply defers that selection until `findFirst`.

awoods: @acoburn, If you would like to create a patch for this tweak for `.flatMap(Arrays::stream)`, great. Otherwise, is there anything else that this PR is waiting on?

acoburn:
:+1: looks good to me

awoods:
@whikloj, are you in a position to do a full squash/merge of this?

whikloj:
Sure thing, doing that now.

whikloj:
Resolved with f0db4f21a1dcf72eeb2ffb565462ab7fc4600a8c

PR #1007, opened 2016-03-15
Allow references to the Root resource to omit a trailing slash

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1884

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
@@ -150,7 +150,9 @@ public class HttpResourceConverter extends IdentifierConverter<Resource,FedoraRe
     @Override
     public boolean inDomain(final Resource resource) {
         final Map<String, String> values = new HashMap<>();
-        return uriTemplate.match(resource.getURI(), values) && values.containsKey("path");
+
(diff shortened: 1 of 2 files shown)


fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
```diff
@@ -225,6 +227,8 @@ private String asString(final Resource resource, final Map<String, String> value
                 throw new InvalidResourceIdentifierException("Path contains empty element! " + path);
             }
             return path;
+        } else if (isRootWithoutTrailingSlash(resource)) {
```
ajs6f: Is `else` needed here?

acoburn: Yes, because it is an `else if`. Anything that falls through returns `null`.

ajs6f: Yeah, I get the `if` part of `else if`. I am wondering about the `else` part. I prefer just a series of `if () return x` which makes the case structure a little more clear. Maybe that's less clear to others?

acoburn: Oh, I see. Sure, the `else` can be removed. I don't have a particular preference on that style.

PR #1008, opened 2016-03-18
FCREPO-1945: Node Properties should not be included in JMS events

ajs6f:
https://jira.duraspace.org/browse/FCREPO-1945

### DIFF
--- a/fcrepo-jms/src/main/java/org/fcrepo/jms/headers/DefaultMessageFactory.java
+++ b/fcrepo-jms/src/main/java/org/fcrepo/jms/headers/DefaultMessageFactory.java
@@ -16,6 +16,7 @@
 package org.fcrepo.jms.headers;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
+import static java.lang.String.join;
 import static java.util.stream.Collectors.joining;
 import static org.fcrepo.kernel.api.RdfLexicon.REPOSITORY_NAMESPACE;
 import static org.slf4j.LoggerFactory.getLogger;
@@ -58,12 +59,10 @@ public class DefaultMessageFactory implements JMSEventMessageFactory {
     public static final String BASE_URL_HEADER_NAME = JMS_NAMESPACE
             + "baseURL";
 
-    public static final String PROPERTIES_HEADER_NAME = JMS_NAMESPACE
-            + "properties";
-
     public static final String USER_HEADER_NAME = JMS_NAMESPACE + "user";
     public static final String USER_AGENT_HEADER_NAME = JMS_NAMESPACE + "userAgent";
     public static final String EVENT_ID_HEADER_NAME = JMS_NAMESPACE + "eventID";
+    public static final String RESOURCE_TYPE_HEADER_NAME = JMS_NAMESPACE + "resourceTypes";
 
     private String baseURL;
     private String userAgent;
@@ -104,6 +103,7 @@ public class DefaultMessageFactory implements JMSEventMessageFactory {
         message.setStringProperty(USER_HEADER_NAME, event.getUserID());
         message.setStringProperty(USER_AGENT_HEADER_NAME, userAgent);
         message.setStringProperty(EVENT_ID_HEADER_NAME, event.getEventID());
+        message.setStringProperty(RESOURCE_TYPE_HEADER_NAME, join(",", event.getResourceTypes()));
 
         LOGGER.trace("getMessage() returning: {}", message);
         return message;
--- a/fcrepo-jms/src/main/java/org/fcrepo/jms/headers/DefaultMessageFactory.java
+++ b/fcrepo-jms/src/main/java/org/fcrepo/jms/headers/DefaultMessageFactory.java
@@ -103,7 +103,6 @@ public class DefaultMessageFactory implements JMSEventMessageFactory {
         message.setStringProperty(BASE_URL_HEADER_NAME, baseURL);
         message.setStringProperty(USER_HEADER_NAME, event.getUserID());
         message.setStringProperty(USER_AGENT_HEADER_NAME, userAgent);
-        message.setStringProperty(PROPERTIES_HEADER_NAME, String.join(",", event.getProperties()));
         message.setStringProperty(EVENT_ID_HEADER_NAME, event.getEventID());
 
         LOGGER.trace("getMessage() returning: {}", message);
--- a/fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/HeadersJMSIT.java
+++ b/fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/HeadersJMSIT.java
@@ -18,6 +18,7 @@ package org.fcrepo.integration.jms.observer;
 import static com.jayway.awaitility.Awaitility.await;
 import static com.jayway.awaitility.Duration.ONE_SECOND;
 import static javax.jms.Session.AUTO_ACKNOWLEDGE;
+import static org.fcrepo.integration.jms.observer.HeadersJMSIT.DangerousSupplier.guard;
 import static org.fcrepo.jms.headers.DefaultMessageFactory.BASE_URL_HEADER_NAME;
 import static org.fcrepo.jms.headers.DefaultMessageFactory.EVENT_TYPE_HEADER_NAME;
 import static org.fcrepo.jms.headers.DefaultMessageFactory.IDENTIFIER_HEADER_NAME;
@@ -33,6 +34,8 @@ import java.io.ByteArrayInputStream;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.function.Predicate;
+import java.util.function.Supplier;
+
 import javax.inject.Inject;
 import javax.jcr.Repository;
 import javax.jcr.RepositoryException;
@@ -257,15 +260,22 @@ public class HeadersJMSIT implements MessageListener {
         return msg.getStringProperty(BASE_URL_HEADER_NAME);
     }
 
-    private static <T> T guard(final DangerousSupplier<T> get) {
-        try {
-            return get.get();
-        } catch (final JMSException e) {
-            throw new AssertionError();
-        }
-    }
+    @FunctionalInterface
+    static interface DangerousSupplier<T> extends Supplier<T> {
 
-    private static interface DangerousSupplier<T> {
-        T get() throws JMSException;
+        static <T> T guard(final DangerousSupplier<T> get) {
+            return get.get();
+        }
+
+        T dangerousGet() throws JMSException;
+
+        @Override
+        default T get() {
+            try {
+                return dangerousGet();
+            } catch (final JMSException e) {
+                throw new AssertionError();
+            }
+        }
     }
 }
--- a/fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/HeadersJMSIT.java
+++ b/fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/HeadersJMSIT.java
@@ -22,9 +22,7 @@ import static javax.jms.Session.AUTO_ACKNOWLEDGE;
 import static org.fcrepo.jms.headers.DefaultMessageFactory.BASE_URL_HEADER_NAME;
 import static org.fcrepo.jms.headers.DefaultMessageFactory.EVENT_TYPE_HEADER_NAME;
 import static org.fcrepo.jms.headers.DefaultMessageFactory.IDENTIFIER_HEADER_NAME;
-import static org.fcrepo.jms.headers.DefaultMessageFactory.PROPERTIES_HEADER_NAME;
 import static org.fcrepo.jms.headers.DefaultMessageFactory.TIMESTAMP_HEADER_NAME;
-import static org.fcrepo.kernel.api.RdfLexicon.HAS_SIZE;
 import static org.fcrepo.kernel.api.RdfLexicon.REPOSITORY_NAMESPACE;
 import static org.fcrepo.kernel.api.RequiredRdfContext.PROPERTIES;
 import static org.jgroups.util.UUID.randomUUID;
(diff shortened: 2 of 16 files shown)


fcrepo-jms/src/main/java/org/fcrepo/jms/headers/DefaultMessageFactory.java
```diff
@@ -103,7 +103,6 @@ public Message getMessage(final FedoraEvent event,
         message.setStringProperty(BASE_URL_HEADER_NAME, baseURL);
         message.setStringProperty(USER_HEADER_NAME, event.getUserID());
         message.setStringProperty(USER_AGENT_HEADER_NAME, userAgent);
-        message.setStringProperty(PROPERTIES_HEADER_NAME, String.join(",", event.getProperties()));
```
acoburn: Should we also remove the declaration of `PROPERTIES_HEADER_NAME` above? I suspect it is no longer used.

ajs6f: Good call. I'll do that.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/FedoraEventImpl.java
```diff
@@ -227,46 +226,68 @@ public String toString() {
...
+     * @param event
+     * @return the types recorded on the resource associated to this event
+     */
+    public static Stream<String> getResourceTypes(final Event event) {
+        if (event instanceof org.modeshape.jcr.api.observation.Event) {
+            try {
+                final org.modeshape.jcr.api.observation.Event modeEvent =
+                        (org.modeshape.jcr.api.observation.Event) event;
+                final Stream.Builder<NodeType> types = Stream.builder();
+                for (final NodeType type : modeEvent.getMixinNodeTypes()) {
```
acoburn: `Arrays.stream(modeEvent.getMixinNodeTypes()).forEach(types::add);` ?

ajs6f: No, I thought about that, and if you _really_ want it, fine, but it seems like the creation of several short-lived objects for very little gain. I'm getting more sensitive to that as evidence mounts that stuff coming out of GC eden is a contributor to lower throughput for a lot of code.

I would rather change our code standards to allow the elision of braces for a one-line `for` like this, which would provide as much concision at a lower cost.

acoburn: That's a good point. Your construction is fine, let's leave it as-is.

acoburn:
I'm not seeing where the `resourceType` values are stored in the newly created JMS message. I would expect a header such as `RESOURCE_TYPES_HEADER_NAME` and an equivalent event.getResourceTypes() joined into a `String`, along with some tests to verify that the headers are now present.

fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/HeadersJMSIT.java
```diff
@@ -215,10 +212,9 @@ public void testRemoval() throws RepositoryException {
     @Override
     public void onMessage(final Message message) {
         try {
-            LOGGER.debug(
-                    "Received JMS message: {} with path: {}, timestamp: {}, event type: {}, properties: {},"
-                            + " and baseURL: {}", message.getJMSMessageID(), getPath(message), getTimestamp(message),
-                            getEventTypes(message), getProperties(message), getBaseURL(message));
+            LOGGER.debug("Received JMS message: {} with path: {}, timestamp: {}, event type: {}, and baseURL: {}",
```
acoburn: possibly list the values from `FedoraEvent::getResourceTypes` here?

ajs6f:
I think I may have just plain forgot to stick the types into JMS messages. Let me go look at that...

fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/HeadersJMSIT.java
ajs6f: Sure, done.

fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/HeadersJMSIT.java
```diff
@@ -15,24 +15,27 @@
...
 import static org.fcrepo.kernel.api.RequiredRdfContext.PROPERTIES;
 import static org.jgroups.util.UUID.randomUUID;
+import static org.junit.Assert.assertTrue;
 import static org.slf4j.LoggerFactory.getLogger;
 
 import java.io.ByteArrayInputStream;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
```
acoburn: I also get a compile error with this -- it has to do with the static import of the internal Interface before `java.util.function.Supplier` is imported. You can get around it by putting the static imports after the other import statements.

ajs6f: How weird! I see no such error. Oh well, I'll switch them.

ajs6f: Done.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/FedoraEventImpl.java
```diff
@@ -227,46 +224,68 @@ public String toString() {
...
         requireNonNull(event);
         try {
-            @SuppressWarnings("unchecked")
             final Map<String, String> info = event.getInfo();
-
-            return new FedoraEventImpl(valueOf(event.getType()), cleanPath(event),
+            final FedoraEventImpl fedoraEvent = new FedoraEventImpl(valueOf(event.getType()), cleanPath(event),
                     event.getUserID(), event.getUserData(), event.getDate(), info);
-
+            getResourceTypes(event).forEach(fedoraEvent::addResourceType);
```
acoburn: The namespaces for these resource types need to be dereferenced in a manner similar to how they were handled with the (removed) namespaceResolver function. The problem is that types take this form internally `fedora:Resource`, but when published on the eventBus, they need to be `http://fedora.info/definitions/v4/repository#Resource`

fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/HeadersJMSIT.java
```diff
@@ -131,7 +134,8 @@ public void testIngestion() throws RepositoryException {
         try {
             containerService.findOrCreate(session, testIngested);
             session.save();
-            awaitMessageOrFail(testIngested, NODE_ADDED_EVENT_TYPE, null);
+            final Message ingestMsg = awaitMessageOrFail(testIngested, NODE_ADDED_EVENT_TYPE);
+            assertTrue(getResourceTypes(ingestMsg).contains("fedora:Resource"));
```
acoburn: Should be checking for `contains("http://fedora.info/definitions/v4/repository#Resource")`

ajs6f: Nope. If it did, it wouldn't pass. The URIs are coming out as QNames. I can put a NS-mapping step in there, but it will be kind of messy. Do you want that?

acoburn:
I found some strange behavior with `Binaries`. When I `PUT/POST` a `fedora:Binary`, I get an event for `fedora:NonRdfSourceDescription`. And when I update the properties on that `fedora:NonRdfSourceDescription`, I see an event for `fedora:Binary`, which seems exactly opposite to what I would expect. Containers all seem to behave as expected.

fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/HeadersJMSIT.java
acoburn: yes, because what happens if someone doesn't register a NS and the event includes a type for `ns001:Object`. That would be completely ambiguous, which we definitely don't want.

acoburn: ...and the NS mapping step should be in the `org.fcrepo.kernel.modeshape.observer.SimpleObserver` class

ajs6f: Actually, I don't think that's a problem with the SPI but a problem with a stupid client, but I'm fine to put a mapping step in.

ajs6f: No, I'm going to put it in the JMS module. `SimpleObserver` is an internal pickoff point. There's no need to expend time mapping namespaces there.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/FedoraEventImpl.java
ajs6f: See our conversation elsewhere. The event bus is not the right place to do that, if you insist on it being done. Externally-facing machinery like the JMS publishing gear is where that should be done, if it is to be done.

fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/HeadersJMSIT.java
acoburn: I disagree. You'll need access to a JCR Session to dereference the namespaces, and that should be kept out of the JMS module.

ajs6f:
No, that's what you should expect. Remember, when you create a binary, you also create a description. And properties that you push to the description of a binary actually get stored on that binary. If you don't like that, we can revisit that set of decisions, but the events coming out now are actually a good "trace" of what you are doing to the repository.

fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/HeadersJMSIT.java
ajs6f: The namespace-resolution is something that is a useless expense at the event bus.

ajs6f: The fact that you currently need access to JCR to resolve namespaces is a long-standing architectural flaw in the code. There are two possible solutions: manage namespaces independently of JCR, or do not manage namespaces. I would vaguely be in favor of the latter, but we should go to the users to understand the needs.

acoburn: My two priorities here are:

  1. make any resource types coming out of the JMS broker to be unambiguous. To me, that means they are fully qualified or a namespace mapping is provided _with every message_. I am strongly in favor of the former approach.
  2. keep the JMS module independent of JCR and the modeshape implementation; that way it could be re-used by an alternate implementation.

Both of these can very easily be addressed by simply making those qnames (`fedora:Resource`) into fully qualified URIs (`http://fedora.info/definitions/v4/repository#Resource`) inside the modeshape implementation code.

ajs6f: They could, but that would be wrong for the other reason I gave, which is why I'm not going to do that. I'd much rather put this on hold and deal properly with the namespace question, rather than gunk up the internals of the reference kernel implementation for short-term gain.
As for the two needs in specific, as I said, I don't recognize the first at all, but I'm willing to treat with it at your insistence, and the second seems far less important to me than I think it may seem to you. Neither of them is much of a priority to me, which is why I would rather wait.

ajs6f: Incidentally, unless/until we are actually publishing some kind of shared-meaningful serialization in the messages (RDF being the obvious candidate) there's not much point to arguing about this. Maybe we should pack this whole issue in until we get that resolved.

ajs6f: https://jira.duraspace.org/browse/FCREPO-1960

acoburn: ajs6f++

IMO the JMS headers that we're arguing about shouldn't even be part of the SPI specification but rather a "handy extra feature of the JMS implementation of that spec" -- if they are even part of that implementation at all. The key thing is to have an RDF serialization so the type of ambiguity I'm trying to avoid is resolved properly.

ajs6f: No, the headers about which we are arguing are in fact the (future) _content_ of that spec. What could possibly be more important to put in an RDF expression of the event than the type of the resource concerned? Unless you are saying that the ref impl should _also_, _additionally_ stick 'em in headers for JMS. I'm not against that, but I don't quite see the utility?

acoburn: I'm saying the same thing you are -- those headers will become the content of the message body, serialized as RDF. The SPI spec will be silent on such things as JMS headers, but if an impl wants to use such things, the spec shouldn't prevent it. Though the utility would be considerably less.

ajs6f: Okay, that's cool. But in that case, I want to turn this whole thing (as far as goes the scheduling of work) around and work on the RDF expression _first_. That will give us the grounds on which to do this ticket right from the beginning. But that also means it won't get into this upcoming point release. WDYT?

acoburn: I completely agree.

ajs6f: Okay, so we put this on hold. I'm not going to close this PR right away, because it could be useful in the next few days for discussion.

acoburn:
Superseded by #1053

PR #1015, opened 2016-04-05
Add link headers to versioned resources

awoods:
Resolves both:
https://jira.duraspace.org/browse/FCREPO-1975 and
https://jira.duraspace.org/browse/FCREPO-1974

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersionsIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersionsIT.java
@@ -20,6 +20,7 @@ import static com.hp.hpl.jena.graph.NodeFactory.createLiteral;
 import static com.hp.hpl.jena.graph.NodeFactory.createURI;
 import static com.hp.hpl.jena.rdf.model.ResourceFactory.createResource;
 import static com.hp.hpl.jena.vocabulary.RDF.type;
+import static javax.ws.rs.core.Link.fromUri;
 import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
 import static javax.ws.rs.core.Response.Status.CONFLICT;
 import static javax.ws.rs.core.Response.Status.NO_CONTENT;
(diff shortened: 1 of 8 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersionsIT.java
```diff
@@ -183,6 +195,55 @@ public void testVersioningANodeWithAVersionableChild() throws IOException {
...
+                DESCRIBED_BY.getLocalName()).build();
+
+        // Look for expected Link headers
+        final HttpHead headObjMethod = headObjMethod(versionId);
+        try (final CloseableHttpResponse response = execute(headObjMethod)) {
+
+            final Set<Link> resultSet = new HashSet<>();
+            final Collection<String> linkHeaders = getLinkHeaders(response);
+
+            linkHeaders.stream().map(Link::valueOf).forEach(link -> {
```
acoburn: rather than a `forEach` adding to a `Set<Link>` here, a `map` and `collect(toSet())` would be cleaner.

awoods: Can you elaborate with a PR?

acoburn:
It seems that most of this involves adding headers using `javax.servlet`. When the response is generated, though, we are using `javax.ws.rs`.
This certainly works, but I wonder why we wouldn't just use JAX-RS for everything? That would mean using the JAX-RS API such as `ok(...).header(...).header(...).build()` rather than relying on changing the state of the internal `servletResponse` field.

awoods:
That is a good point, however, this PR is actually a very minor change to `FedoraVersions` with some refactoring of the location of existing `private` and `protected` methods. I definitely think there would be value in re-examining/refactoring our use of JAX-RS.

acoburn:
I'm OK with using the servlet-api as is currently the case. I'm more noting the fact that we should evaluate our use of JAX-RS along the lines of this much neglected ticket that has my name on it: https://jira.duraspace.org/browse/FCREPO-1664

acoburn:
Merged via 15a0b2687156498a1c4722abe182fcdca9dde154

PR #1020, opened 2016-04-13
FCREPO-1989: Minor cleanup of compiler warnings

ajs6f:
https://jira.duraspace.org/browse/FCREPO-1989

### DIFF
--- a/fcrepo-http-api/pom.xml
+++ b/fcrepo-http-api/pom.xml
@@ -81,6 +81,7 @@
       <groupId>org.fcrepo</groupId>
       <artifactId>fcrepo-connector-file</artifactId>
       <version>${project.version}</version>
+      <scope>test</scope>
     </dependency>
 
     <dependency>
--- a/fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
@@ -61,7 +61,6 @@ import java.text.ParseException;
 import java.util.EnumSet;
 import java.util.List;
 
-import javax.jcr.AccessDeniedException;
 import javax.jcr.NamespaceRegistry;
 import javax.jcr.Node;
 import javax.jcr.RepositoryException;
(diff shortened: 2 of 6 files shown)


fcrepo-http-api/pom.xml
```diff
@@ -77,11 +77,6 @@
       <version>${project.version}</version>
     </dependency>
 
-    <dependency>
-      <groupId>org.fcrepo</groupId>
-      <artifactId>fcrepo-connector-file</artifactId>
-      <version>${project.version}</version>
```
acoburn: This can't be entirely removed, but you can scope it to `test`.

ajs6f: I can do that, but I want to ask now-- why do we need this for a test? This seems like a bad and confusing dependency.

acoburn: It seems like moving the projection tests into `fcrepo-webapp` might make the most sense.

ajs6f: It would make _more_ sense. I guess it depends on what view we take of the projection functionality. @awoods, thoughts?

awoods:
It seems most appropriate to move projection outside of core, into fcrepo-webapp-plus... either now or soon.

BTW, this PR should be targeted to the `DEV` branch instead of `master`.

acoburn:
@awoods If I understand you correctly, you're suggesting that the entire `fcrepo-connector-file` module be moved into the `fcrepo4-exts` organization. If that's the case, I'm _definitely_ in favor of that.

awoods:
That is indeed what I am saying... but that probably deserves its own ticket.

acoburn:
Yes, that's certainly beyond the scope of "Minor cleanup of compiler warnings"

ajs6f:
:+1:  to moving the connector jazz out of the core. Meanwhile, I'm just going to shift the scope for the dep here.

ajs6f:
https://jira.duraspace.org/browse/FCREPO-1992

awoods:
Note, this PR needs to be re-targeted to the `DEV` branch instead of `master`.

ajs6f:
Closed in favor of https://github.com/fcrepo4/fcrepo4/pull/1023

PR #1021, opened 2016-04-15
Read PATCH requests as UTF-8 encoded input streams

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1993

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -17,6 +17,7 @@ package org.fcrepo.http.api;
 
 
 import static com.hp.hpl.jena.rdf.model.ResourceFactory.createResource;
+import static java.nio.charset.StandardCharsets.UTF_8;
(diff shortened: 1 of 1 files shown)


whikloj:
So I am guessing you need to be on a system with a default encoding scheme that is **NOT** UTF-8 to be able to test this properly, yes?

ajs6f:
Theoretically, [no](https://stackoverflow.com/questions/361975/setting-the-default-java-character-encoding), but it could make for some messiness in Maven Surefire and Failsafe. Still, we might want to do it.

awoods:
This looks good to me, especially given the integration test and error handling provided by the related 
tickets:
https://jira.duraspace.org/browse/FCREPO-1973
https://jira.duraspace.org/browse/FCREPO-1979

whikloj:
So we are leaving this as being tested manually by Windows users? I don't have a particular problem with that, just wondering.

acoburn:
Anyone can test this by starting the JVM with the file.encoding property set to something other than utf-8. Using a Mac, without the file.encoding value set, this fixes PATCH requests containing utf-8 characters.

awoods:
Given the other testing context around this PR, I am happy to see this move through as-is.

PR #1022, opened 2016-04-16
Tie binary description ETags/lastModified dates to the binaries they describe

acoburn:

The current behavior advertises ETags/lastModified dates as if they are the same for binaries and binary descriptions. This PR makes this the _behavior_ of Fedora by relying on the same lastModified property for both binaries and their descriptions.

If, on the other hand, binaries and their descriptions are to contain _different_ ETag and lastModified dates, then a different implementation should be pursued. At the very least, though, this fixes the currently broken behavior.

With this impl, the ETag value for both a binary and binary description will be the same. Likewise, the last-modified date will be the same. The NonRdfSourceDescription simply proxies those calls to the FedoraBinary. If a property of the binary description is changed, then the ETag and last-modified date will change for both. If it is problematic to reuse the ETag value for both resources, there is an easy way around that.

### DIFF
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/NonRdfSourceDescriptionImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/NonRdfSourceDescriptionImpl.java
@@ -19,6 +19,8 @@ import static org.fcrepo.kernel.modeshape.utils.FedoraTypesUtils.isNonRdfSourceD
 import static org.modeshape.jcr.api.JcrConstants.JCR_CONTENT;
 import static org.slf4j.LoggerFactory.getLogger;
 
+import java.util.Date;
+
 import javax.jcr.Node;
 import javax.jcr.RepositoryException;
 
@@ -49,6 +51,21 @@ public class NonRdfSourceDescriptionImpl extends FedoraResourceImpl implements N
     @Override
     public NonRdfSource getDescribedResource() {
         return new FedoraBinaryImpl(getContentNode());
+    }
+
+    @Override
+    public String getEtagValue() {
+        return getDescribedResource().getEtagValue();
+    }
+
+    @Override
+    public Date getCreatedDate() {
+        return getDescribedResource().getCreatedDate();
+    }
+
+    @Override
+    public Date getLastModifiedDate() {
+        return getDescribedResource().getLastModifiedDate();
     }
 
     private Node getContentNode() {
--- a/fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/NonRdfSourceDescriptionImplTest.java
+++ b/fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/NonRdfSourceDescriptionImplTest.java
@@ -43,6 +43,7 @@ import static org.junit.Assert.fail;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 import static org.mockito.MockitoAnnotations.initMocks;
+import static org.modeshape.jcr.api.JcrConstants.JCR_CONTENT;
 
 /**
  * <p>{@link NonRdfSourceDescriptionImplTest} class.</p>
@@ -66,6 +67,9 @@ public class NonRdfSourceDescriptionImplTest implements FedoraTypes {
     private Node mockDsNode;
 
     @Mock
+    private Node mockContentNode;
+
+    @Mock
     private InputStream mockStream;
 
     @Mock
@@ -83,6 +87,7 @@ public class NonRdfSourceDescriptionImplTest implements FedoraTypes {
             when(mockDsNode.getMixinNodeTypes()).thenReturn(nodeTypes);
             when(mockDsNode.getName()).thenReturn(testDsId);
             when(mockDsNode.getSession()).thenReturn(mockSession);
+            when(mockDsNode.getNode(JCR_CONTENT)).thenReturn(mockContentNode);
             final NodeType mockNodeType = mock(NodeType.class);
             when(mockNodeType.getName()).thenReturn("nt:file");
             when(mockDsNode.getPrimaryNodeType()).thenReturn(mockNodeType);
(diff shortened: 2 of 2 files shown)


ajs6f:
I _think_ that one ETag to rule them all makes sense here, because we have only one repository resource. It is what I would call a datastream, what others are calling a binary. It is a bitstream with properties and altering either the bits or the properties counts as a mutation. So one last-mod for both makes sense to me.

escowles:
This change definitely improves the current behavior where property updates don't change the description's lastModified or ETag.  And it seems like a logical outcome of providing triples about only a single subject in the binary description.

Though I am worried this means that changing a property on a binary's description could force needlessly retrieving a (potentially very large) binary because its ETag changed even though the content did not.  It also seems like we're missing the opportunity to use the binary's SHA1 as a strong ETag.  l have opened a JIRA ticket to consider using a binary's SHA1 as its ETag, and whether that is workable, what other changes would be required, etc: https://jira.duraspace.org/browse/FCREPO-1996

awoods:
I am thinking about this in the context of the Fedora specification. This current PR would suggest that we specify that binary resources and their descriptions share the same etag/last-mod. Do we want to require this of all Fedora implementations?

Although this addresses the UI issue of https://jira.duraspace.org/browse/FCREPO-1983, it creates a rift with the LDP spec which calls for two distinct resources:
When a binary resource is created, per the LDP spec, a `Link: <some-description-url>; rel:describedby` header is returned that points to a separate resource that describes the binary.
Ref: https://www.w3.org/TR/ldp/#ldpc-post-createbinlinkmetahdr

ajs6f:
I don't see the incompatibility. Nothing in that section of the spec calls for independent ETags, that I can see.

ajs6f:
To the other point, yes, I think this is a spec requirement. @escowles has a good point about the bitstream being retrieved needlessly, but I think we want to resolve that with something more subtle than ETag.

acoburn:
I am of two minds on this. On the one hand, the current behavior is clearly broken, and this PR fixes it. And it does so while preserving the non-broken portion of the current behavior as much as possible.

On the other, I see binary resources and their descriptions as two distinct resources (resources, though, whose lifecycles are inextricably bound together). I would expect the bitstream (fedora:Binary) portion to have a strong ETag derived from the content of that binary (as @escowles suggests: the SHA1). For the NonRdfSourceDescription, I would expect weak ETags that are based on the last-mod date AND the content-type. That way, a caching proxy could hold different representations, e.g. for JSON+LD and n-triples; otherwise, requests for the different types would return 304s regardless of the Accept header. 

I would note that at present, all ETags produced by Fedora are strong ETags. We should discuss this further on a tech call.

ajs6f:
Do you contemplate the possibility of multiple descriptions for a single bitstream, or multiple bistreams answering to a single description?

acoburn:
I had been operating under the assumption that there is a strict one-to-one correspondence between `fedora:Binary` and `fedora:NonRdfSourceDescription` resources.

ajs6f:
If that is the case, Occam's razor suggests we consider such pairs to be one resource.

acoburn:
That's fine to consider them one resource, but the currency of ETags is _representations_ (we've had this conversation before). And by presenting a single ETag for _all_ representations of a resource and the various content-types of its description, we're doing a major disservice to caching proxies.

escowles:
I agree with @acoburn about the resource/representation issue, and having a weak ETag for the description based on last modified, and a strong ETag for the content based on the SHA1 makes sense to me.  One question I have is whether changing the last-modified of the content without changing the ETag will result in it being invalidated in caches.  If it is, then we would want to have two separate last-modified dates.

awoods:
Regarding whether we consider the NonRdfSource and its Description as one or two resources, the wording of the [LDP spec](https://www.w3.org/TR/ldp/#ldpc-post-createbinlinkmetahdr), and the related fact that the NonRdfSource and its Description have distinct URI is a strong indication of two separate resources.

As a note, this Weak ETag ticket seems to have gone stagnant.

ajs6f:
I'm fine with separate ETags-- as @acoburn says, ETags go to representations. What I do not want are separate las-mod dates. That would be inaccurate, in my view. To @awoods' point: I don't much care about LDP in this context. I'm not using the work resource to mean anything connected with that, but to mean something in the Fedora model. Put more shortly: a datastream.

cbeer:
> I am thinking about this in the context of the Fedora specification. This current PR would suggest that we specify that binary resources and their descriptions share the same etag/last-mod. Do we want to require this of all Fedora implementations?

Absolutely not; they happen to be the same in the Fedora implementation, but we shouldn't encourage clients to expect that or require other implementations to do the same. I suggest its a quirk of our backend, not a feature.

ajs6f:
I think we're having two different conversations. One is about ETags, and the other about last-mod. @cbeer's point seems to me to be about ETag. @escowles and I had a conversation at some point (can't dig it up now) during which we agred that last-mod != ETag. last-mod is _one_ of the ingredients for ETag. ETag summarizes more information than just last-mod.

acoburn:
@ajs6f https://wiki.duraspace.org/display/FF/Design+-+Last-Modified+Property

awoods:
@cbeer, 
> Absolutely not; they happen to be the same in the Fedora implementation, but we shouldn't encourage...
Actually, this is not the case with the current Fedora implementation. Consolidating the ETag and Last-modified date is the essence of this PR. I would suggest that the reference Fedora implementation follow whatever expectation we document in the specification.

acoburn:
No, the essence of this PR is fixing something that is currently broken. It does this without changing the _design_ of how ETags are represented and acted on. 

On the other hand, the essence of this discussion seems to be about changing the _design_ of ETags (which I am fully supportive of).

I see this PR as a stop-gap measure: fix the bug now, fix the design later. Alternately, we can make the change (bug + design) in one fell swoop. I was trying to separate those into smaller pieces.

As a note: the _bug_ portion of this is blocking progress on https://jira.duraspace.org/browse/FCREPO-1498 and https://jira.duraspace.org/browse/FCREPO-1742. Hence this PR.

acoburn:
See https://github.com/fcrepo4/fcrepo4/pull/1024 for an alternate resolution of this (https://github.com/fcrepo4/fcrepo4/pull/1024 is, in my opinion, a better fix).

acoburn:
closed in favor of #1024

PR #991, opened 2016-02-23
eTags should be Weak eTags

Rarian:
This is a new pull request to address outstanding comments on #920.

PR #1024, opened 2016-04-18
Fix ETag handling for Binaries and Binary descriptions

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1754

This is an alternate resolution of this issue. See #1022 for a different approach.

The advantage of this PR (from my perspective) is that ETag and last-mod values on the fedora:Binary and associated description are decoupled.

Note: this PR also addresses the Strong/Weak ETag issue: Binaries produce strong ETags, everything else produces a weak ETag. So it also supersedes #991  

See: http://irclogs.fcrepo.org/2016-04-18.html for further discussion of this.

Based on discussion, this approach is to be pursued instead of #1022

### DIFF
--- a/fcrepo-http-api/pom.xml
+++ b/fcrepo-http-api/pom.xml
@@ -81,6 +81,7 @@
       <groupId>org.fcrepo</groupId>
       <artifactId>fcrepo-connector-file</artifactId>
       <version>${project.version}</version>
+      <scope>test</scope>
     </dependency>
 
     <dependency>
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -497,13 +497,16 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
         // must change.
         if (resource instanceof NonRdfSourceDescription) {
             final NonRdfSourceDescription description = (NonRdfSourceDescription)resource;
+            // Use a weak ETag for the LDP-RS
             etag = new EntityTag(description.getDescribedResource().getEtagValue(), true);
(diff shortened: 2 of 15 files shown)


acoburn:
Note: this PR also addresses the Strong/Weak ETag issue: Binaries produce strong ETags, everything else produces a weak ETag.

awoods:
To be clear, this current PR bases the:
- etag and last-mod for NonRdfSources on the Description, and 
- etag and last-mod for Descriptions on the NonRdfSources.

A brief rationale would be helpful here.

acoburn:
@awoods I agree a rationale would be helpful here. Would you like that as an inline comment or written here in the github comments? In brief, though, it follows from the principle of the binary being the subject of the triples in the description.

acoburn:
@awoods as a side note, this is exactly why we need a specification. It is impossible to fix incorrect behavior if there is no clear document describing what _correct_ behavior is.

awoods:
Maybe an inline comment would help folks who look at the code in the future understand what is going on.

ajs6f:
Six kinds of amen.

ajs6f:
To be clear, my comment was in support of @acoburn's remark about specification.

sprater:
@acoburn:  I pulled your fix into my fork, compiled, and tested in the web UI again the case I documented in FCREPO-1983.  That problem appears to be fixed.

acoburn:
@sprater thanks for checking! I would like to add some additional integration tests before this is actually merged. If that happens next week (before the 4.5.1 release), I'll close this PR and re-target it at the `DEV` branch; otherwise, it will stay where it is. In any case, this ticket is certainly moving forward.

sprater:
@acoburn sounds good.  I'll wait for the merge when it occurs, then test again, just to make sure.

acoburn:
Superseded by #1028

PR #1027, opened 2016-04-22
Update fcrepo-parent to use the latest version of fcrepo-build-tools

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-2004


ajs6f:
Do we need to wait for the new build tools to get through Maven Central distribution, or are we already good on that?

acoburn:
I think we're good on that: http://search.maven.org/#artifactdetails%7Corg.fcrepo%7Cfcrepo-build-tools%7C4.4.1%7Cbundle

ajs6f:
Cool, merging.

PR #1030, opened 2016-04-28
Only create one-click if Maven profile is used: -Pone-click

awoods:
* This enables a default ModeShape config for the one-click

Related to: https://jira.duraspace.org/browse/FCREPO-1994


whikloj:
@awoods so the goal here is to only create the one click war file if you use the _one-click_ profile?

whikloj:
I answered my own question, yes it does.

PR #1032, opened 2016-05-02
remove FedoraResource::getNode from fcrepo-kernel-api

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1869

This will clearly have significant implications on certain downstream projects. If this approach seems sensible, then I will create related PRs for those projects.

PR #1033, opened 2016-05-03
Use a userspace lastmodified property instead of jcr:lastModified

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1742
Resolves: https://jira.duraspace.org/browse/FCREPO-1498

Supersedes #917

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -62,6 +62,7 @@ import static org.fcrepo.kernel.api.FedoraTypes.FCR_METADATA;
 import static org.fcrepo.kernel.api.RdfLexicon.BASIC_CONTAINER;
 import static org.fcrepo.kernel.api.RdfLexicon.CONSTRAINED_BY;
 import static org.fcrepo.kernel.api.RdfLexicon.CONTAINS;
+import static org.fcrepo.kernel.api.RdfLexicon.CREATED_DATE;
 import static org.fcrepo.kernel.api.RdfLexicon.DC_TITLE;
 import static org.fcrepo.kernel.api.RdfLexicon.DIRECT_CONTAINER;
 import static org.fcrepo.kernel.api.RdfLexicon.HAS_CHILD;
@@ -76,6 +77,7 @@ import static org.fcrepo.kernel.api.RdfLexicon.INBOUND_REFERENCES;
 import static org.fcrepo.kernel.api.RdfLexicon.INDIRECT_CONTAINER;
 import static org.fcrepo.kernel.api.RdfLexicon.JCR_NAMESPACE;
 import static org.fcrepo.kernel.api.RdfLexicon.JCR_NT_NAMESPACE;
+import static org.fcrepo.kernel.api.RdfLexicon.LAST_MODIFIED_DATE;
 import static org.fcrepo.kernel.api.RdfLexicon.LDP_MEMBER;
 import static org.fcrepo.kernel.api.RdfLexicon.LDP_NAMESPACE;
 import static org.fcrepo.kernel.api.RdfLexicon.MEMBERSHIP_RESOURCE;
@@ -2171,7 +2173,38 @@ public class FedoraLdpIT extends AbstractResourceIT {
     }
 
     @Test
-    public void testCreationResponseDefault() {
+    public void testInboundLinksDoNotUpdateEtag() throws IOException {
+        final String id1 = getRandomUniqueId();
+        final HttpPut httpPut = putObjMethod(id1);
+        final String oldETag;
+        final String oldMod;
+        try (final CloseableHttpResponse response = execute(httpPut)) {
+            assertEquals(CREATED.getStatusCode(), getStatus(response));
+            oldETag = response.getFirstHeader("ETag").getValue();
+            oldMod = response.getFirstHeader("Last-Modified").getValue();
+        }
+
+        final String id2 = getRandomUniqueId();
+        createObject(id2).close();
+
+        final HttpPatch patch = patchObjMethod(id2);
+        patch.addHeader("Content-Type", "application/sparql-update");
+        patch.setEntity(new StringEntity(
+                "INSERT { <> <http://purl.org/dc/elements/1.1/relation> <" + serverAddress + id1 + "> } WHERE {}"));
+        try (final CloseableHttpResponse response = execute(patch)) {
+            assertEquals(NO_CONTENT.getStatusCode(), getStatus(response));
+        }
+
+        try (final CloseableHttpResponse response = execute(getObjMethod(id1))) {
+            final String etag = response.getFirstHeader("ETag").getValue();
+            final String lastmod = response.getFirstHeader("Last-Modified").getValue();
+            assertEquals(oldMod, lastmod);
+            assertEquals(oldETag, etag);
+        }
+    }
+
+    @Test
+    public void testCreationResponseDefault() throws Exception {
         testCreationResponse(null, null, CREATED, "text/plain");
         testCreationResponse(null, "application/ld+json", NOT_ACCEPTABLE, "text/html");
     }
@@ -2220,5 +2253,113 @@ public class FedoraLdpIT extends AbstractResourceIT {
         assertEquals(BAD_REQUEST.getStatusCode(), getStatus(patchObjMethod("%5bfoo")));
         assertEquals(BAD_REQUEST.getStatusCode(), getStatus(postObjMethod("%5bfoo")));
         assertEquals(BAD_REQUEST.getStatusCode(), getStatus(putObjMethod("%5bfoo")));
+    }
+
+    @Test
+    public void testBinaryLastModified() throws Exception {
+        final String objid = getRandomUniqueId();
+        final String objURI = serverAddress + objid;
+        final String binURI = objURI + "/binary1";
+
+        final long lastmod1;
+        try (final CloseableHttpResponse response = execute(putDSMethod(objid, "binary1", "some test content"))) {
+            assertEquals(CREATED.getStatusCode(), getStatus(response));
+            lastmod1 = headerFormat.parse(response.getFirstHeader("Last-Modified").getValue()).getTime();
+        }
+
+        sleep(1000); // wait a second to make sure last-modified value will be different
+
+        try (final CloseableGraphStore graph = getGraphStore(new HttpGet(binURI + "/fcr:metadata"))) {
+            verifyModifiedMatchesCreated(graph);
+        }
+
+        final HttpPatch patchBinary = new HttpPatch(binURI + "/fcr:metadata");
+        patchBinary.addHeader("Content-Type", "application/sparql-update");
+        patchBinary.setEntity(new StringEntity("INSERT { <" + binURI + "> " +
+                "<http://www.w3.org/TR/rdf-schema/label> \"this is a label\" } WHERE {}"));
+        final long lastmod2;
+        try (final CloseableHttpResponse response = execute(patchBinary)) {
+            assertEquals(NO_CONTENT.getStatusCode(), getStatus(response));
+            lastmod2 = headerFormat.parse(response.getFirstHeader("Last-Modified").getValue()).getTime();
+            assertTrue(lastmod2 > lastmod1);
+        }
+
+        sleep(1000); // wait a second to make sure last-modified value will be different
+
+        final long lastmod3;
+        try (final CloseableHttpResponse response = execute(putDSMethod(objid, "binary1", "new test content"))) {
+            assertEquals(NO_CONTENT.getStatusCode(), getStatus(response));
+            lastmod3 = headerFormat.parse(response.getFirstHeader("Last-Modified").getValue()).getTime();
+            assertTrue(lastmod3 > lastmod2);
+        }
+    }
+
+    @Test
+    public void testContainerLastModified() throws Exception {
+        final String objid = getRandomUniqueId();
+        final String objURI = serverAddress + objid;
+
+        // create an object
+        final long lastmod1;
+        try (final CloseableHttpResponse response = execute(putObjMethod(objid))) {
+            assertEquals(CREATED.getStatusCode(), getStatus(response));
+            lastmod1 = headerFormat.parse(response.getFirstHeader("Last-Modified").getValue()).getTime();
+        }
+
+        sleep(1000); // wait a second to make sure last-modified value will be different
+
+        // initial created and last-modified properties should match
+        try (final CloseableGraphStore graph = getGraphStore(getObjMethod(objid))) {
+            verifyModifiedMatchesCreated(graph);
+        }
+
+        // update the object properties (last-modified should be updated)
+        final HttpPatch patchObject = new HttpPatch(objURI);
+        patchObject.addHeader("Content-Type", "application/sparql-update");
+        patchObject.setEntity(new StringEntity("INSERT { <> " +
+                "<http://www.w3.org/TR/rdf-schema/label> \"this is a label\" } WHERE {}"));
+        final long lastmod2;
+        try (final CloseableHttpResponse response = execute(patchObject)) {
+            assertEquals(NO_CONTENT.getStatusCode(), getStatus(response));
+            lastmod2 = headerFormat.parse(response.getFirstHeader("Last-Modified").getValue()).getTime();
+            assertTrue(lastmod2 > lastmod1);
+        }
+
+        sleep(1000); // wait a second to make sure last-modified value will be different
+
+        // create a direct container (last-modified should be updated)
+        final long lastmod3;
+        final HttpPut createContainer = new HttpPut(objURI + "/members");
+        createContainer.addHeader("Content-Type", "text/turtle");
+        final String membersRDF = "<> a <http://www.w3.org/ns/ldp#DirectContainer>; "
+            + "<http://www.w3.org/ns/ldp#hasMemberRelation> <http://pcdm.org/models#hasMember>; "
+            + "<http://www.w3.org/ns/ldp#membershipResource> <" + objURI + "> . ";
+        createContainer.setEntity(new StringEntity(membersRDF));
+        try (final CloseableHttpResponse response = execute(createContainer)) {
+            assertEquals(CREATED.getStatusCode(), getStatus(response));
+            lastmod3 = headerFormat.parse(response.getFirstHeader("Last-Modified").getValue()).getTime();
+            assertTrue(lastmod3 > lastmod2);
+        }
+
+        sleep(1000); // wait a second to make sure last-modified value will be different
+
+        // create child in the container
+        final long lastmod4;
+        assertEquals(CREATED.getStatusCode(), getStatus(new HttpPut(objURI + "/members/member1")));
+
+        // last-modified should be updated
+        try (final CloseableHttpResponse response = execute(headObjMethod(objid))) {
+            assertEquals(OK.getStatusCode(), getStatus(response));
+            lastmod4 = headerFormat.parse(response.getFirstHeader("Last-Modified").getValue()).getTime();
+            assertTrue(lastmod4 > lastmod3);
+        }
+    }
+
+    private static void verifyModifiedMatchesCreated(final GraphStore graph) {
+        final Node cre = graph.find(ANY, ANY, CREATED_DATE.asNode(), ANY).next().getObject();
+        final Node mod = graph.find(ANY, ANY, LAST_MODIFIED_DATE.asNode(), ANY).next().getObject();
+        System.out.println("cre: " + cre.getLiteralValue());
+        System.out.println("mod: " + mod.getLiteralValue());
+        assertEquals(cre.getLiteralValue(), mod.getLiteralValue());
     }
 }
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -2191,9 +2191,7 @@ public class FedoraLdpIT extends AbstractResourceIT {
         patch.addHeader("Content-Type", "application/sparql-update");
         patch.setEntity(new StringEntity(
                 "INSERT { <> <http://purl.org/dc/elements/1.1/relation> <" + serverAddress + id1 + "> } WHERE {}"));
-        try (final CloseableHttpResponse response = execute(patch)) {
-            assertEquals(NO_CONTENT.getStatusCode(), getStatus(response));
-        }
+        assertEquals(NO_CONTENT.getStatusCode(), getStatus(execute(patch)));
 
         try (final CloseableHttpResponse response = execute(getObjMethod(id1))) {
             final String etag = response.getFirstHeader("ETag").getValue();
@@ -2358,8 +2356,6 @@ public class FedoraLdpIT extends AbstractResourceIT {
     private static void verifyModifiedMatchesCreated(final GraphStore graph) {
         final Node cre = graph.find(ANY, ANY, CREATED_DATE.asNode(), ANY).next().getObject();
         final Node mod = graph.find(ANY, ANY, LAST_MODIFIED_DATE.asNode(), ANY).next().getObject();
-        System.out.println("cre: " + cre.getLiteralValue());
-        System.out.println("mod: " + mod.getLiteralValue());
         assertEquals(cre.getLiteralValue(), mod.getLiteralValue());
     }
 }
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -68,6 +68,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -389,7 +390,7 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
 
             // This is resolved immediately b/c we delete the node before updating an indirect container's target
             final boolean shouldUpdateIndirectResource = ldpInsertedContentProperty(node)
-                .flatMap(resourceToProperty.apply(getSession())).filter(this::hasProperty).isPresent();
+                .flatMap(resourceToProperty(getSession())).filter(this::hasProperty).isPresent();
 
             node.remove();
 
@@ -581,11 +582,11 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
 
         model.register(listener);
 
-        final PropertyChangedListener insertedContentListener = new PropertyChangedListener();
+        final AtomicBoolean propertyChanged = new AtomicBoolean();
 
         ldpInsertedContentProperty(getNode()).ifPresent(resource -> {
-            insertedContentListener.addProperty(resource);
-            model.register(insertedContentListener);
+            final PropertyChangedListener propertyListener = new PropertyChangedListener(resource, propertyChanged);
+            model.register(propertyListener);
         });
 
         model.setNsPrefixes(request.getPrefixMapping());
@@ -597,7 +598,7 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
 
         try {
             touch();
-            if (insertedContentListener.propertyChanged()) {
+            if (propertyChanged.get()) {
                 touchLdpMembershipResource(getNode());
             }
         } catch (final RepositoryException ex) {
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -24,13 +24,9 @@ import static java.util.stream.Collectors.toList;
 import static java.util.stream.Stream.concat;
 import static java.util.stream.Stream.empty;
 import static java.util.stream.Stream.of;
-import static java.util.TimeZone.getTimeZone;
 import static org.apache.commons.codec.digest.DigestUtils.shaHex;
 import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_LASTMODIFIED;
 import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_REPOSITORY_ROOT;
-import static org.fcrepo.kernel.api.FedoraTypes.LDP_DIRECT_CONTAINER;
-import static org.fcrepo.kernel.api.FedoraTypes.LDP_INDIRECT_CONTAINER;
-import static org.fcrepo.kernel.api.FedoraTypes.LDP_MEMBER_RESOURCE;
 import static org.fcrepo.kernel.api.RdfLexicon.REPOSITORY_NAMESPACE;
 import static org.fcrepo.kernel.api.RdfLexicon.isManagedNamespace;
 import static org.fcrepo.kernel.api.RdfLexicon.isManagedPredicate;
@@ -54,6 +50,7 @@ import static org.fcrepo.kernel.modeshape.services.functions.JcrPropertyFunction
 import static org.fcrepo.kernel.modeshape.utils.FedoraTypesUtils.hasInternalNamespace;
 import static org.fcrepo.kernel.modeshape.utils.FedoraTypesUtils.isFrozenNode;
 import static org.fcrepo.kernel.modeshape.utils.FedoraTypesUtils.isInternalNode;
+import static org.fcrepo.kernel.modeshape.utils.FedoraTypesUtils.touchLdpMembershipResource;
 import static org.fcrepo.kernel.modeshape.utils.NamespaceTools.getNamespaceRegistry;
 import static org.fcrepo.kernel.modeshape.utils.StreamUtils.iteratorToStream;
 import static org.fcrepo.kernel.modeshape.utils.UncheckedFunction.uncheck;
@@ -63,7 +60,6 @@ import static org.slf4j.LoggerFactory.getLogger;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Calendar;
 import java.util.Collection;
 import java.util.Date;
 import java.util.Iterator;
@@ -124,6 +120,7 @@ import org.fcrepo.kernel.modeshape.rdf.impl.ReferencesRdfContext;
 import org.fcrepo.kernel.modeshape.rdf.impl.RootRdfContext;
 import org.fcrepo.kernel.modeshape.rdf.impl.SkolemNodeRdfContext;
 import org.fcrepo.kernel.modeshape.rdf.impl.VersionsRdfContext;
+import org.fcrepo.kernel.modeshape.utils.FedoraTypesUtils;
 import org.fcrepo.kernel.modeshape.utils.JcrPropertyStatementListener;
 import org.fcrepo.kernel.modeshape.utils.UncheckedPredicate;
 import org.fcrepo.kernel.modeshape.utils.iterators.RdfAdder;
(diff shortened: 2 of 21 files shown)


whikloj:
I don't understand all of this, but what I do understand, I like 👍

acoburn:
Most of the heavy lifting was from @escowles 👏

whikloj:
well congrats to both of you

whikloj:
This is just an observation and is probably a question of timing and PR order, but I noticed that you guys have a couple `getNode()`'s in here and PR #1032 aims to remove that (I think).

I'm going to try this with my pcdm:Collections and see if the membershipResource gets reindexed when a new proxy is added.

Again, nice job. @escowles++ & @acoburn++

acoburn:
@whikloj related to timing, yes, some rebasing will be required between this PR and #1032 
I've been waiting for the 4.5.1 release for these two PRs. I have one more on deck, but that one depends more explicitly on this PR.

whikloj:
Soooo, maybe I'm missing something but was this supposed to resolve the issue of not emitting events when `ldp:membershipResource`s are updated by an indirectContainer? I think I made a mistake and this is not what I thought it was.

acoburn:
@whikloj This PR addresses a number of related issues. The first is that of ETags. The current Fedora release uses the JCR-based lastModified property, which changes every time something changes with a resource (good) and every time any other resource links to that resource (bad). As a byproduct of using a user-space last-modified property, we have more control over when the last-modified property changes -- including when a resource changes due to a change of LDP-based membership triples (i.e. direct and indirect container). And with this PR, when LDP membership triples change, the user-space last-modified property changes, causing an event to be emitted.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -2171,7 +2173,38 @@ public void testDeleteLargeLiteralStoredAsBinary() throws IOException {
...
+        }
+
+        final String id2 = getRandomUniqueId();
+        createObject(id2).close();
+
+        final HttpPatch patch = patchObjMethod(id2);
+        patch.addHeader("Content-Type", "application/sparql-update");
+        patch.setEntity(new StringEntity(
+                "INSERT { <> <http://purl.org/dc/elements/1.1/relation> <" + serverAddress + id1 + "> } WHERE {}"));
+        try (final CloseableHttpResponse response = execute(patch)) {
```
ajs6f: Isn't this just [this](https://github.com/acoburn/fcrepo4/blob/b456ebc624ce73dd358abc2acd7d1c783c71e16d/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java#L225)?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -2221,4 +2254,112 @@ public void testPathsWithBrackets() {
...
+            assertEquals(OK.getStatusCode(), getStatus(response));
+            lastmod4 = headerFormat.parse(response.getFirstHeader("Last-Modified").getValue()).getTime();
+            assertTrue(lastmod4 > lastmod3);
+        }
+    }
+
+    private static void verifyModifiedMatchesCreated(final GraphStore graph) {
+        final Node cre = graph.find(ANY, ANY, CREATED_DATE.asNode(), ANY).next().getObject();
+        final Node mod = graph.find(ANY, ANY, LAST_MODIFIED_DATE.asNode(), ANY).next().getObject();
+        System.out.println("cre: " + cre.getLiteralValue());
```
ajs6f: `System.out.println`? Maybe use a proper `Logger`?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
acoburn: yes, thanks for catching that

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
acoburn: where on earth did _that_ come from? Of course, I'll remove it.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -2221,4 +2254,112 @@ public void testPathsWithBrackets() {
...
+        final String objURI = serverAddress + objid;
+
+        // create an object
+        final long lastmod1;
+        try (final CloseableHttpResponse response = execute(putObjMethod(objid))) {
+            assertEquals(CREATED.getStatusCode(), getStatus(response));
+            lastmod1 = headerFormat.parse(response.getFirstHeader("Last-Modified").getValue()).getTime();
+        }
+
+        sleep(1000); // wait a second to make sure last-modified value will be different
```
ajs6f: Urb. This is kind of icky. Can we use the internal event bus here and wait on it to know when to proceed?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -432,15 +448,31 @@ public Date getLastModifiedDate() {
...
+            return created;
+        }
+        return timestamp;
+    }
 
+    /**
+     * Set the last-modified date to the current date.
+     * @throws RepositoryException
+     */
+    public void touch() throws RepositoryException {
```
ajs6f: If no one is going to do anything with this `RepositoryException` but rethrow it, maybe just catch and rethrow it here?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
escowles: Sorry, that almost certainly came from my WIP branch...

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BinaryServiceImpl.java
```diff
@@ -58,9 +62,26 @@ public FedoraBinary findOrCreate(final Session session, final String path) {
 
             if (dsNode.isNew()) {
                 initializeNewDatastreamProperties(dsNode);
+
+                final Node parent = dsNode.getDepth() > 0 ? dsNode.getParent() : null;
```
ajs6f: I'm not sure I understand the logic in this section-- it's pretty dense. Maybe some inline comments?

whikloj:
@acoburn ok, that's what I thought but in creating an indirect container and adding an object to it, the membershipResource resource in Fedora gets the updated property but it does not seem to be "re-indexed" by the camel-toolbox. Perhaps it is my setup, I'll try and log the actual JMS messages coming out.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/ContainerServiceImpl.java
```diff
@@ -61,10 +64,29 @@ public Container findOrCreate(final Session session, final String path) {
 
             if (node.isNew()) {
                 initializeNewObjectProperties(node);
+
+                final Node parent = node.getDepth() > 0 ? node.getParent() : null;
```
ajs6f: Am I nuts or is this very much like what's new in `BinaryServiceImpl` above?

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/integration/kernel/modeshape/FedoraResourceImplIT.java
```diff
@@ -185,15 +186,40 @@ public void testRandomNodeGraph() {
 
     @Test
     public void testLastModified() throws RepositoryException {
-        final String pid = UUID.randomUUID().toString();
+        final String pid = getRandomPid();
+        containerService.findOrCreate(session, "/" + pid);
+
+        session.save();
```
ajs6f: It's usual to use `try-finally` with `Session`s.

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/integration/kernel/modeshape/FedoraResourceImplIT.java
```diff
@@ -185,15 +186,40 @@ public void testRandomNodeGraph() {
...
+
+        final Container obj2 = containerService.findOrCreate(session, "/" + pid);
+        final Date created = roundDate(obj2.getCreatedDate());
+        final Date modified = roundDate(obj2.getLastModifiedDate());
+        assertFalse(modified + " should not be before " + created, modified.before(created));
+
+        final Graph graph = obj2.getTriples(subjects, PROPERTIES).collect(toModel()).getGraph();
+        final Node s = createGraphSubjectNode(obj2);
+        final ExtendedIterator<Triple> iter = graph.find(s, LAST_MODIFIED_DATE.asNode(), ANY);
+        assertTrue("Should have one lastModified triple", iter.hasNext());
```
ajs6f: Maybe just `assertEquals` and `org.apache.jena.atlas.iterator.Iter.count(Iterator<T>)`?

ajs6f:
Aside from sniping comments about code style and the like, this looks basically good to me. I mean, it's definitely the right direction. Just as a note to myself, the integration test for FedoraLdp is getting crazy long. I think we may need to break it up at some point.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
acoburn: I'm not actually sure I understand this suggestion. The `RepositoryException` is an important part of this method because there may be an access denied exception, which is handled further up the stack.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
acoburn: I'm going to defer this one to a separate effort: https://jira.duraspace.org/browse/FCREPO-2023

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/ContainerServiceImpl.java
acoburn: I put this and several other instances of similar code into a single utility function.

acoburn:
I believe I have addressed all of the review comments related to this PR

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
acoburn: The FCREPO-2023 ticket is still relevant, but in this case, we'll still need to use `Thread.sleep`. I say that because this test is specifically about verifying that the reported last modified date occurs in a particular order corresponding to the order of operation. Because the reported last modified date is rounded to the nearest second, it is entirely possible that two operations in quick succession would produce last-modified dates that are equal, even if the event bus is used.

acoburn:
@whikloj you are right -- the eventing machinery is missing some changes related to indirect containers. Stay tuned for another commit related to that.

whikloj:
@acoburn cool thanks.

ajs6f:
Is this ready?

acoburn:
yes, please review/test

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/FedoraTypesUtils.java
```diff
@@ -329,22 +337,67 @@ public static Node getJcrNode(final FedoraResource resource) {
...
+                    return Optional.of(createResource(parent.getProperty(LDP_INSERTED_CONTENT_RELATION).getString()));
+                }
+            }
+        } catch (final RepositoryException ex) {
+            throw new RepositoryRuntimeException(ex);
+        }
+        return empty();
+    }
+
+    public static Function<Session, Function<Resource, Optional<String>>> resourceToProperty = session -> resource -> {
```
ajs6f: This might be clearer as a method:
```
public static Function<Resource, Optional<String>> resourceToProperty(Session session) {
```
1. The only time you use it that I see, you explicitly call `apply`. That's the very semantic of a plain method, which is more concise.
2. The `Session` is actually part of the state of the `Resource -> Optional<String>` function.

acoburn: Yeah, I previously changed it to a plain method and then changed it back for some reason (I don't recall exactly why). I'll see about changing to a plain method.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/PropertyChangedListener.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+
+    /**
+     *  Add a property to watch
+     *
+     *  @param property the property
+     *  @return whether the property was added
+     */
+    public boolean addProperty(final Resource property) {
```
ajs6f: This might be easier to reason about in the future if `properties` was immutable, built by defensive copy at construction.

acoburn: I initially wrote this class as you describe, but the use I had for this class didn't fit well with that approach. By adding properties separately from the constructor, I could write the implementation like so:

    final PropertyChangedListener insertedContentListener = new PropertyChangedListener();
 
     ldpInsertedContentProperty(getNode()).ifPresent(resource -> {
          insertedContentListener.addProperty(resource);
          model.register(insertedContentListener);
    });

    if (insertedContentListener.propertyChanged()) {
        ...
    }

An alternative would be to use a builder pattern.

ajs6f: I saw that. I don't see why you couldn't calculate the presence and value and use them:
```
ldpInsertedContentProperty(getNode()).ifPresent(resource -> {
      final PropertyChangedListener insertedContentListener = new PropertyChangedListener(resource);
      model.register(insertedContentListener);
});
```

acoburn: well yes, but you need to define the `PropertyChangedListener` outside of the lambda expression and then later you have to do an additional null check before checking for `PropertyChangedListener::propertyChanged`, which _must_ happen outside of that lambda. Doing so makes the code kinda ugly.

acoburn: I suppose an alternative is to just use a lot of `if (...)` blocks. I was just trying to avoid a lot of verbosity.

ajs6f: I think you're looking at this backwards. You don't need to define the listener outside of the lambda. You need to define whatever you are going to check outside of the lambda. You made that a field in the listener but you don't have to. You could just as well pass in the property and an `AtomicBoolean` into the constructor. That's how I tend to write this kind of code for Jena.

ajs6f: To be clear, I don't think this matters that much. Another way to keep things simpler would be to scrap the `Set` and just have a `Property`-valued field. You have at most one property to check, right?

acoburn: The `AtomicBoolean` approach seems like a good one to me.

ajs6f: Well, you have to catch the result in something outside the lambda, so whatever seems most pleasant. The whole listener apparatus feels really old and creaky these days. It would be nice to redesign it for a Java 8 world.

acoburn:
@ajs6f I have added another commit to address your most recent comments, along with a unit test for the `PropertyChangedListener` class.

ajs6f:
Looks good to me. I'll wait a bit for other's comments.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/FedoraTypesUtils.java
```diff
@@ -153,6 +168,12 @@
     });
 
     /**
+     * Check if a property is intentionally suppressed.
+     */
+    final static List<String> suppressed = Arrays.asList("jcr:mimeType", "jcr:lastModified");
+    private static Predicate<Property> isSuppressedProperty = uncheck(p -> suppressed.contains(p.getName()));
```
awoods: This Predicate does not appear to be used. If so, please remove.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/FedoraTypesUtils.java
```diff
@@ -153,6 +168,12 @@
     });
 
     /**
+     * Check if a property is intentionally suppressed.
+     */
+    final static List<String> suppressed = Arrays.asList("jcr:mimeType", "jcr:lastModified");
```
awoods: May as well use the constant defined for `jcr:lastModified`: FedoraJcrConstants.JCR_LASTMODIFIED.

You are welcome to add a new FedoraJcrConstants.JCR_MIMETYPE if you like a consistent list of elements.

acoburn: This also seems not to be used, and so I would be inclined to remove this along with the `Predicate` on the next line.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/FedoraTypesUtils.java
acoburn: @awoods this Predicate has been removed.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/FedoraTypesUtils.java
```diff
@@ -314,4 +335,79 @@ public static Node getJcrNode(final FedoraResource resource) {
...
+    /**
+     * Update the fedora:lastModified date of the parent's ldp:membershipResource if that node is a direct
+     * or indirect container, provided the LDP constraints are valid.
+     *
+     * @param node The JCR node
+     * @throws RepositoryException if a repository exception occurred
+     */
+    public static void touchLdpMembershipResource(final Node node) throws RepositoryException {
+        if (node.getDepth() > 0) {
+            final Node parent = node.getParent();
```
awoods: This logic only works as expected if the `parent` is the Direct/Indirect container... which is true if the `node` was created with a `slug`. However, in the common case where no `slug` is used and the UUIDPathMinter creates the hierarchy, the true Direct/Indirect container does not get updated. 
This can be demonstrated in an integration test.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BinaryServiceImpl.java
awoods: If the `dsNode` was created without a `slug` and therefore the `parent` is a PairTree, do we want to be touching the PairTree or the logical parent container? I suspect the latter.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/ContainerServiceImpl.java
awoods: If the `dsNode` was created without a `slug` and therefore the `parent` is a PairTree, do we want to be touching the PairTree or the logical parent container? I suspect the latter.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/FedoraTypesUtils.java
acoburn: @awoods that is an excellent point. I wonder if reusing some of the code in FedoraResource::getContainer would be appropriate.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/NodeServiceImpl.java
```diff
@@ -112,11 +116,27 @@ public void moveObject(final Session session, final String source, final String
...
+
         } catch (final RepositoryException e) {
             throw new RepositoryRuntimeException(e);
         }
     }
 
+    private void touchParentLdpMemberResource(final Session session, final String path) throws RepositoryException {
+        final FedoraResource resource = find(session, path);
+        final Node node = getJcrNode(resource);
+        final Node parent = node.getDepth() > 0 ? node.getParent() : null;
```
awoods: See: https://github.com/fcrepo4/fcrepo4/pull/1033/files#r63611100

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/FedoraTypesUtils.java
```diff
@@ -314,4 +328,79 @@ public static Node getJcrNode(final FedoraResource resource) {
...
+     * Given a JCR Node, fetch the parent's ldp:insertedContentRelation value, if
+     * one exists.
+     *
+     * @param node the JCR Node
+     * @return the ldp:insertedContentRelation Resource, if one exists.
+     */
+    public static Optional<Resource> ldpInsertedContentProperty(final Node node) {
+        try {
+            if (node.getDepth() > 0) {
+                final Node parent = node.getParent();
```
awoods: See other comments related to logical and hierarchical parents.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/FedoraTypesUtils.java
awoods: FedoraResource::getContainer should resolve the issue.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/ContainerServiceImpl.java
```diff
@@ -61,10 +63,22 @@ public Container findOrCreate(final Session session, final String path) {
...
+                }
+            }
+
+            if (node.isNew()) {
+                touch(node);
             }
 
             return new ContainerImpl(node);
         } catch (final RepositoryException e) {
+            LOGGER.info("Error: {}", e);
```
awoods: Please either change this to printing the `e.getMessage()` or removing this logging statement, as opposed to dumping exception stacks to the log.

acoburn:
@awoods I have addressed your comments in a new commit to this PR.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/ContainerServiceImpl.java
ajs6f: This is an actual error, so it deserves `ERROR` level logging. Otherwise, I disagree with @awoods , I think it is helpful to have the stack trace, but it's not something I would care very much about. As long as whatever is logged gives enough info to take definite and immediate action so that the request can be reissued successfully.

escowles: The general principle here is that stacktraces are not useful for the vast majority of users, and they dramatically increase logfile sizes.

I agree with @ajs6f that I like to see the stacktraces, and find them useful for debugging.  This issue comes up fairly frequently in code review.  Maybe we should have a debug mode that enables dumping stacktraces?

acoburn: I removed that line in the last commit. Doing so also makes the code consistent with the existing BinaryService::findOrCreate method. Plus, because it rethrows the exception as a RepositoryRuntimeException, it seems like the additional logging is superfluous. Or do you think otherwise? If it goes back in, are you suggesting it be at DEBUG level, containing the whole stacktrace? and should the corresponding BinaryService::findOrCreate function have a similar logging? I'm also beginning to wonder if this is out of scope for this ticket.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -436,15 +452,31 @@ public Date getLastModifiedDate() {
...
 
-        final Date createdDate = getCreatedDate();
         if (createdDate != null) {
             LOGGER.trace("Using created date for last modified date for node {}", node);
             return createdDate;
         }
 
         return null;
     }
+    private long getTimestamp(final String property, final long created) throws RepositoryException {
```
awoods: Minor: add a newline before this `private` method.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -408,7 +420,7 @@ private void createTombstone(final Node parent, final String path) throws Reposi
     public Date getCreatedDate() {
         try {
             if (hasProperty(JCR_CREATED)) {
-                return new Date(getProperty(JCR_CREATED).getDate().getTimeInMillis());
+                return new Date(getTimestamp(JCR_CREATED, 0L));
```
awoods: The `0L` in `getTimestamp(JCR_CREATED, 0L)` seems a little opaque. Would a boolean work?

acoburn: I'll need to defer to @escowles on that one.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/ContainerServiceImpl.java
escowles: @acoburn In this particular case, I think it's fine to just rethrow the exception and it'll be handled elsewhere.  But I'm responding to the general issue that's come up several times in code review: developers would like to be able to see stacktraces, but we frequently ask them to remove them in code review for the reasons I mentioned.

What I'm thinking of specifically is being able to add a System property (`-Dfcrepo.debug=true` or something like that) to output stacktraces any time we normally log exceptions.  This would probably involve some kind of logging helper with Spring config to enable/disable the stacktrace logging.

ajs6f: I think that "debug mode" actually _is_ `DEBUG` level logging. Here's my proposed rule: log informative messages that enable immediate concrete action to correct the specific problem, intended for integrators, administrators, and programmers, at `ERROR` level. Log stacktraces and other diagnostic information that would be primarily of interest to integrators and programmers, for the purpose of understanding how to correct the integration or code to avoid the problem in future, at `DEBUG` level. Log deep reporting of control flow and logic that would be primarily of interest to programmers at `TRACE` level.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
escowles: The `0L` here matches up with this line: https://github.com/fcrepo4/fcrepo4/pull/1033/files/9f68918ebc5007adc656a619ff7916320ab56be6#diff-7aee50d5030524028026aa94cd75f350R465

The intention is to be able to use the `getTimestamp` method for both created and modified dates.  For modified dates, it's important to make sure that the returned value isn't before the created value.  But for created dates, that doesn't apply.  We could just revert the change to this line if that's confusing.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/ContainerServiceImpl.java
ajs6f: -1 to using Spring-specific tools for AOP. If we're going to do deep black logging magic, at least we should use tools built for the purpose, like SLF4J `Marker` and Logback filtering.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -436,15 +452,31 @@ public Date getLastModifiedDate() {
...
+            return created;
+        }
+        return timestamp;
+    }
 
+    /**
+     * Set the last-modified date to the current date.
+     * @throws RepositoryException if a repository exception occurs
+     */
+    public void touch() throws RepositoryException {
```
awoods: It would clean up other parts of the code if we `try/catch` the `RepositoryException` here, and `throw` a `RepositoryRuntimeException`.

ajs6f: +1 to going unchecked ASAP, unless there is a really specific action the software can take to recover.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/ContainerServiceImpl.java
escowles: That also sounds fine to me.  The only issue I see with that approach is potentially wanting to have multiple logging statements to handle WARN/ERROR messages intended for integrators, and also DEBUG messages intended for programmers for the same exception, which seems cumbersome.  Does SLF4J or Logback provide some mechanism for configuring whether exceptions are logged as messages or stacktraces?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -550,12 +582,28 @@ public void updateProperties(final IdentifierConverter<Resource, FedoraResource>
...
         listener.assertNoExceptions();
+
+        try {
+            touch();
+            if (propertyChanged.get()) {
+                touchLdpMembershipResource(getNode());
+            }
+        } catch (final RepositoryException ex) {
+            throw new RepositoryRuntimeException(ex);
+        }
```
awoods: Just a note, this `updateProperties()` method is getting denser and denser with no inline comments. This PR does not cause that problem, but it does add to it.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/ContainerServiceImpl.java
ajs6f: [Positionally](http://www.slf4j.org/faq.html#paramException).

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -643,6 +691,12 @@ public void replaceProperties(final IdentifierConverter<Resource, FedoraResource
             }
 
             removeEmptyFragments();
+            try {
+                touch();
```
awoods: In a similar fashion to `updateProperties()`, it seems like we should be doing some sort of `touchLdpMembershipResource` here in `replaceProperties()` as well, no?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -677,6 +731,7 @@ private void removeEmptyFragments() {
                         throw new RepositoryRuntimeException("Error removing empty fragments", ex);
                     }
                 });
+                touch();
```
awoods: We probably should move this `touch()` deeper into the logic (i.e. up a few lines) so that it is only called if `n.remove()` is called.

fcrepo-kernel-modeshape/src/main/resources/fedora-node-types.cnd
```diff
@@ -73,7 +74,9 @@
   - premis:hasMessageDigest (URI) COPY
 
 [fedora:Skolem] > mix:referenceable mixin
+  - fedora:lastModified (DATE)
 
 [fedora:Pairtree] mixin
+  - fedora:lastModified (DATE)
```
awoods: I am surprised that we want `fedora.lastModified` on Pairtrees. Is this an artifact of the previous logic of using `node.getParent()` instead of `resource.getContainer()`?

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/integration/kernel/modeshape/observer/SimpleObserverIT.java
```diff
@@ -219,8 +222,136 @@ public void testHashUriEvent() throws RepositoryException {
         awaitEvent("/object9", NODE_ADDED);
         awaitEvent("/object9", PROPERTY_ADDED);
         awaitEvent("/object9#contributor", PROPERTY_ADDED);
+        // NOT CORRECT
```
awoods: What does `NOT CORRECT` mean here?

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/integration/kernel/modeshape/observer/SimpleObserverIT.java
```diff
@@ -219,8 +222,136 @@ public void testHashUriEvent() throws RepositoryException {
...
+        awaitEvent("", PROPERTY_ADDED);
+
+        // This should be 2 -- should change with FCREPO-1985
```
awoods: Please add the text `FIXME` to this comment so that IDEs can surface it.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -377,21 +367,37 @@ public void delete() {
                 }
             }
 
-            final Node parent;
+            final Node parent = getNode().getDepth() > 0 ? getNode().getParent() : null;
```
awoods: This is likely a latent bug: adding the tombstone to the parent node (potentially a Pairtree) instead of the parent container.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
awoods: For clarity, I would suggest making `0L` a named constant, like `NO_TIME` or something.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
acoburn: My understanding of tombstones suggests that the current code is correct.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
acoburn: I agree that removing the checked exception would be a good thing. @awoods concern about `javax.jcr.AccessDeniedException`s can be addressed by catching those at the point where we're writing data to the JCR nodes and rethrowing them as `org.fcrepo.kernel.api.exceptions.AccessDeniedException`s (rather than higher up in the stack)

awoods: :+1:

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
acoburn: Actually, if I'm not mistaken, this `touch()` can be removed. The `removeEmptyFragments` method is private and only called twice in FedoraResourceImpl. Each time it is called, the `touch` method is already being called. I will simply remove this.

awoods: I was thinking the same thing, but thought it may be useful to leave this `touch` in place to mitigate new calls to `removeEmptyFragments()`... but I am also fine to see it removed.

acoburn:
I have addressed @awoods comments in a new commit.

awoods:
:+1: Looks good. Any final words before we commit?

PR #1035, opened 2016-05-12
Adding spans around release info

escowles:
* Adding spans with IDs around release info in HTML UI, to make it easier to parse.

Potential low-impact solution to https://jira.duraspace.org/browse/FCREPO-1774

### DIFF
--- a/fcrepo-webapp/src/main/webapp/index.html
+++ b/fcrepo-webapp/src/main/webapp/index.html
@@ -41,7 +41,7 @@ here.</p>
             <li><a href="https://wiki.duraspace.org/display/FF">Wiki</a></li>
             </ul>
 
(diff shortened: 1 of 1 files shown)


whikloj:
You are a whirlwind of productivity 👍

ajs6f:
Would we want to do this with RDFa/DOAP? It's not really semantically correct, but it might be a little clearer down the road. Or maybe I'm being paranoid.

whikloj:
I'd say this is making an unsupported method of determining version number, easier. So we aren't explicitly saying you should do this, but if you want to at least we've made it easier for you. But perhaps I'm not paranoid enough.

ajs6f:
That works for me.

ajs6f:
No one is paranoid enough.

No one.

PR #1037, opened 2016-05-16
FCREPO-2025: Switch to a version of Jersey that does not leak memory

birkland:
2.8 does not have the memory leak reported in
https://java.net/jira/browse/HK2-216

A possible fix was applied in Jersey 2.17, but this and newer
Jersey versions aren't drop-in compatible with 2.13/2.8.

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/SinglePrefer.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/SinglePrefer.java
@@ -19,7 +19,6 @@ import static java.util.Arrays.stream;
 import static java.util.stream.Collectors.toSet;
 import static org.fcrepo.http.commons.domain.PreferTag.emptyTag;
 
-import java.text.ParseException;
 import java.util.Set;
 import java.util.TreeSet;
 
@@ -38,16 +37,11 @@ public class SinglePrefer {
      * Parse a Prefer: header
      *
      * @param header the header
-     * @throws ParseException if parse exception occurred
      */
-    public SinglePrefer(final String header) throws ParseException {
-        preferTags.addAll(stream(header.split(",")).map(h -> {
-            try {
-                return new PreferTag(h);
-            } catch (final ParseException e) {
-                throw new IllegalArgumentException("Could not parse 'Prefer' header", e);
-            }
-        }).collect(toSet()));
+    public SinglePrefer(final String header) {
+        preferTags.addAll(stream(header.split(","))
+                .map(PreferTag::new)
+                .collect(toSet()));
     }
 
     /**
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/SinglePrefer.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/SinglePrefer.java
@@ -15,13 +15,13 @@
  */
 package org.fcrepo.http.commons.domain;
 
+import static java.util.Arrays.stream;
+import static java.util.stream.Collectors.toSet;
 import static org.fcrepo.http.commons.domain.PreferTag.emptyTag;
 
 import java.text.ParseException;
-import java.util.Arrays;
 import java.util.Set;
 import java.util.TreeSet;
-import java.util.stream.Collectors;
 
 /**
  * JAX-RS HTTP parameter parser for the Prefer header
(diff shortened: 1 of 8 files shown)


awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-2025

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/SinglePrefer.java
```diff
@@ -41,7 +41,13 @@
      * @throws ParseException if parse exception occurred
      */
     public SinglePrefer(final String header) throws ParseException {
-        preferTags.addAll(HttpHeaderReader.readList(PreferTag::new, header));
+        preferTags.addAll(Arrays.asList(header.split(",")).stream().map(h -> {
```
acoburn: static import Arrays.asList?

acoburn: Actually, even better, use Arrays.stream(...)

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/SinglePrefer.java
```diff
@@ -41,7 +41,13 @@
...
+            try {
+                return new PreferTag(h);
+            } catch (ParseException e) {
+                throw new IllegalArgumentException("Could not parse 'Prefer' header", e);
+            }
+        }).collect(Collectors.toSet()));
```
acoburn: static import Collectors.toSet?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/SinglePrefer.java
```diff
@@ -41,7 +41,13 @@
      * @throws ParseException if parse exception occurred
      */
     public SinglePrefer(final String header) throws ParseException {
-        preferTags.addAll(HttpHeaderReader.readList(PreferTag::new, header));
+        preferTags.addAll(Arrays.asList(header.split(",")).stream().map(h -> {
+            try {
+                return new PreferTag(h);
+            } catch (ParseException e) {
```
acoburn: `final ParseException e`

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/SinglePrefer.java
birkland: makes sense to me, will do

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/SinglePrefer.java
```diff
@@ -41,7 +41,13 @@
...
      */
     public SinglePrefer(final String header) throws ParseException {
-        preferTags.addAll(HttpHeaderReader.readList(PreferTag::new, header));
+        preferTags.addAll(stream(header.split(",")).map(h -> {
+            try {
+                return new PreferTag(h);
+            } catch (final ParseException e) {
+                throw new IllegalArgumentException("Could not parse 'Prefer' header", e);
+            }
+        }).collect(toSet()));
```
acoburn: And if we're catching the `ParseException` the method probably no longer needs the `throws ParseException`

birkland: Hm, good eye.  Outside of testing, it looks like the only place SinglePrefers are instantiated is inside the jax-rs impl (by virtue of `MultiPrefer(final @HeaderParam("Prefer") Set<SinglePrefer> prefers)`.  It certainly doesn't care about exception types, so removing the `throws ParseException` and throwing a runtime exception should be perfectly fine in practice.  However, it also implies that the `throws ParseException` in MultiPrefer will now never occur.  Should I remove it there too?

awoods: If it is not thrown, no reason to claim otherwise.

ajs6f: Mos def. The house style has evolved to minimize checked exceptions and when they are used, to give them very specific semantics (e.g. not `RepositoryException` but `OutOfDomainSubjectException`).

birkland: OK, great.  In that light, I went ahead and killed it at its source in `PreferTag`, getting rid of that nasty checked exception inside the lambda

awoods:
:+1: 
Let's ship this after tomorrow's Tech Meeting, unless anyone has an objection.

acoburn:
👍

awoods:
Note: Before committing this PR, the impact PR must also be included that fixes the breakage caused in fcrepo-transform: https://github.com/fcrepo4-exts/fcrepo-transform/pull/22

PR #1039, opened 2016-05-20
Updated to code to gracefully handle failure to migrate property.

mikedurbin:
Amends https://jira.duraspace.org/browse/FCREPO-1742 to support advanced authorization scenarios.

### DIFF
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -443,6 +443,8 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
      * Any method that exposes the last modified date must maintain this illusion so
      * that that external callers are presented with a sensible and consistent
      * representation of this resource.
+     * @return the last modified Date (or the created date if it was after the last
+     *         modified date)
      */
     @Override
     public Date getLastModifiedDate() {
@@ -1043,6 +1045,9 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
      * that triple for the given resource.  The current implementation of this method is used by
      * {@link PropertiesRdfContext} to replace the reported {@link org.fcrepo.kernel.api.RdfLexicon#LAST_MODIFIED_DATE}
      * with the one produced by {@link #getLastModifiedDate}.
+     * @param r the Fedora resource
+     * @param translator a converter to get the external identifier from a jcr node
+     * @return a function to convert triples
      */
     public static Function<Triple, Triple> fixDatesIfNecessary(final FedoraResource r,
                                                       final Converter<Node, Resource> translator) {
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -15,6 +15,7 @@
  */
 package org.fcrepo.kernel.modeshape;
 
+import static com.hp.hpl.jena.rdf.model.ResourceFactory.createTypedLiteral;
 import static com.hp.hpl.jena.update.UpdateAction.execute;
 import static com.hp.hpl.jena.update.UpdateFactory.create;
 import static java.util.Arrays.asList;
@@ -27,6 +28,7 @@ import static java.util.stream.Stream.of;
 import static org.apache.commons.codec.digest.DigestUtils.shaHex;
 import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_LASTMODIFIED;
 import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_REPOSITORY_ROOT;
+import static org.fcrepo.kernel.api.RdfLexicon.LAST_MODIFIED_DATE;
 import static org.fcrepo.kernel.api.RdfLexicon.REPOSITORY_NAMESPACE;
 import static org.fcrepo.kernel.api.RdfLexicon.isManagedNamespace;
 import static org.fcrepo.kernel.api.RdfLexicon.isManagedPredicate;
@@ -48,6 +50,7 @@ import static org.fcrepo.kernel.modeshape.rdf.JcrRdfTools.getRDFNamespaceForJcrN
 import static org.fcrepo.kernel.modeshape.services.functions.JcrPropertyFunctions.isFrozen;
 import static org.fcrepo.kernel.modeshape.services.functions.JcrPropertyFunctions.property2values;
 import static org.fcrepo.kernel.modeshape.utils.FedoraTypesUtils.getContainingNode;
+import static org.fcrepo.kernel.modeshape.utils.FedoraTypesUtils.getJcrNode;
 import static org.fcrepo.kernel.modeshape.utils.FedoraTypesUtils.hasInternalNamespace;
 import static org.fcrepo.kernel.modeshape.utils.FedoraTypesUtils.isFrozenNode;
 import static org.fcrepo.kernel.modeshape.utils.FedoraTypesUtils.isInternalNode;
@@ -63,6 +66,7 @@ import static org.slf4j.LoggerFactory.getLogger;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Calendar;
 import java.util.Collection;
 import java.util.Date;
 import java.util.Iterator;
@@ -456,15 +460,8 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
     private long getTimestamp(final String property, final long created) throws RepositoryException {
         LOGGER.trace("Using {} date", property);
         final long timestamp = getProperty(property).getDate().getTimeInMillis();
-        if (timestamp < created && created > NO_TIME && !isFrozenResource()) {
-            try {
-                LOGGER.trace("Updating {} with later created date", property);
-                getNode().setProperty(property, created);
-            } catch (final javax.jcr.AccessDeniedException e) {
-                LOGGER.trace("Unable to update {} with the later created date", property);
-            } catch (final javax.jcr.RepositoryException e) {
-                LOGGER.warn("Unable to update " + property + " with the later created date", e);
-            }
+        if (timestamp < created && created > NO_TIME) {
+            LOGGER.trace("Returning the later created date ({} > {}) for {}", created, timestamp, property);
             return created;
         }
         return timestamp;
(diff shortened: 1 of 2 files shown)


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -457,8 +457,14 @@ private long getTimestamp(final String property, final long created) throws Repo
         LOGGER.trace("Using {} date", property);
         final long timestamp = getProperty(property).getDate().getTimeInMillis();
         if (timestamp < created && created > NO_TIME && !isFrozenResource()) {
-            LOGGER.trace("Updating {} with later created date", property);
-            getNode().setProperty(property, created);
+            try {
+                LOGGER.trace("Updating {} with later created date", property);
+                getNode().setProperty(property, created);
+            } catch (final javax.jcr.AccessDeniedException e) {
+                LOGGER.trace("Unable to update {} with the later created date", property);
```
awoods: I think we will want to re-throw a `org.fcrepo.kernel.api.exception.AccessDeniedException` here.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -457,8 +457,14 @@ private long getTimestamp(final String property, final long created) throws Repo
...
+            } catch (final javax.jcr.RepositoryException e) {
+                LOGGER.warn("Unable to update " + property + " with the later created date", e);
```
awoods: I think we will want to re-throw a `org.fcrepo.kernel.api.exception.RepositoryRuntimeException` here.
Also, in the logging, please change to:
`LOGGER.warn("Unable to update {} with the later created date", propery, e.getMessage());`

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
mikedurbin: So that's the meat of the issue here... this method (getTimestamp) is supposed to GET a value.  It seems we also use it to correct data in the repository.  That would be fine if this method were only invoked by operations that are expected (and authorized) to update the resource in question (though, I'd hope the name changed to reflect that it actually updates the repository).  

The following execution path will likely illustrate the problem:

	at org.fcrepo.kernel.modeshape.FedoraResourceImpl.getTimestamp(FedoraResourceImpl.java:462)
	at org.fcrepo.kernel.modeshape.FedoraResourceImpl.getLastModifiedDate(FedoraResourceImpl.java:437)
	at org.fcrepo.kernel.modeshape.FedoraResourceImpl.getEtagValue(FedoraResourceImpl.java:758)
	at org.fcrepo.http.api.ContentExposingResource.addCacheControlHeaders(ContentExposingResource.java:502)
	at org.fcrepo.http.api.ContentExposingResource.checkCacheControlHeaders(ContentExposingResource.java:481)
	at org.fcrepo.http.api.ContentExposingResource.getBinaryContent(ContentExposingResource.java:317)
	at org.fcrepo.http.api.ContentExposingResource.getContent(ContentExposingResource.java:202)
	at org.fcrepo.http.api.FedoraLdp.getResource(FedoraLdp.java:228)

In cases where a user makes an HTTP GET, we end up invoking this code, and in some cases, with access controls enabled, this update is disallowed.  

Before the change proposed by this PR, that AccessDeniedException was thrown and later caught somewhere in the call stack and manifested as a 403.  That was extremely difficult to debug, since there are so many places where authorization checks occur.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
mikedurbin: I don't know how or whether a RepositoryException would be thrown here... until I added this line, I got a 409 from an integration test (though I couldnt' reproduce it).  Since I characterize this secret update operation within our method to get a timestamp as optional, I figured any exception thrown can be eaten and logged.

It's important to include the stack trace in the log in case this ever actually happens.  When these issues arise in production servers they are often hard to reproduce, but if we include the JCR exception as the root cause of the exception that actually gets logged, there's a chance we'll have the stack trace in our logs and we can see excactly what section of code to investigate as the cause of our unexpected error.

acoburn: I wonder if simply removing the setProperty part of this function would address this issue.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
escowles: The setProperty call is there to make sure that the created date is never later than the modified date (which often happens when creating nodes without this code, because we set the modified date, and then the created date is set a few milliseconds later when the node is saved).

Instead of updating the property, we could just return the later date, and accept that we'd be doing this every time.  What we're doing now (trying to update the property and burying the error if access is denied) effectively updates the property if the user has permissions to do so.

An alternative that I considered was doing a two-step process, where new nodes would be saved, and then we would manually set the created/modified times and re-save.  This has the advantage of making consistent dates possible from the start, at the cost of doing an extra save for each new node.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
awoods: It seems like the two primary options are:
1. Ensuring the dates are correct at the point of resource creation, which will require an extra `save` on create.
2. Do not update the property on subsequent `getTimestamp()` requests, but instead return the potentially adjusted value.

From a performance perspective, (2) is preferred. From a data integrity perspective, (1) is preferred. How do you all feel about going with option (2) with a `debug` statement that clearly indicates the two dates and that we are returning one that is adjusted?

Like @mikedurbin, I am not a fan of updating resources on HTTP GET requests.

escowles: I'm fine with option 2, so we could just do:

``` java
if (timestamp < created) {
    return created;
}
```

mikedurbin: Sounds like a consensus... I'll update the PR with code like that, @escowles , if all the tests still pass.

ajs6f:
Are we waiting on one more commit here?

awoods:
@ajs6f, I believe @mikedurbin is working it. See: http://irclogs.fcrepo.org/2016-05-20.html @10:25

mikedurbin:
I will have a little time to work on this this afternoon and then as much time as I need on Tuesday.  If that's too much of a holdup, I won't be offended if @ajs6f or someone more familiar with the FedoraResourceImpl wants to knock it out sooner.

mikedurbin:
Alright... I've pushed an approach that passes all the tests locally (for fcrepo4, fcrepo-module-auth-rbacl and fcrepo-module-auth-webac).

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -1022,4 +1025,21 @@ protected Property getProperty(final String relPath) {
         }
     }
 
+    /**
+     * A method that takes a Triple and returns a Triple that is the correct representation of
+     * that triple for the given resource.  The current implementation of this method is used by
+     * {@link PropertiesRdfContext} to replace the reported {@link org.fcrepo.kernel.api.RdfLexicon#LAST_MODIFIED_DATE}
+     * with the one produced by {@link #getLastModifiedDate}.
```
acoburn: This method needs some javadocs

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -1022,4 +1025,21 @@ protected Property getProperty(final String relPath) {
...
+     */
+    public static Triple fixTriples(final FedoraResource r, final Converter<Node, Resource> translator,
+                                    final Triple t) {
```
acoburn: Rather than returning a Triple, this function could return a `Function<Triple, Triple>`, making this function signature:

    public static Function<Triple, Triple> fixTriples(final FedoraResource r, final Converter<Node, Resource> translator)

The function body changes only slightly:

    return t -> {
        (what you currently have)
    }

Then, in `PropertiesRdfContext`, you can static import `fixTriples` and shorten the `map` to:

    .map(fixTriples(n, translator))

acoburn:
Other than some very minor suggestions this looks good

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -1022,4 +1025,21 @@ protected Property getProperty(final String relPath) {
         }
     }
 
+    /**
+     * A method that takes a Triple and returns a Triple that is the correct representation of
+     * that triple for the given resource.  The current implementation of this method is used by
+     * {@link PropertiesRdfContext} to replace the reported {@link org.fcrepo.kernel.api.RdfLexicon#LAST_MODIFIED_DATE}
+     * with the one produced by {@link #getLastModifiedDate}.
+     */
+    public static Triple fixTriples(final FedoraResource r, final Converter<Node, Resource> translator,
```
awoods: In addition to @acoburn's comments, I would suggest changing the name of this method to be more descriptive of what is happening... along the lines of `fixDatesIfNecessary()`... or the like.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
mikedurbin: I added some comments to that method and addressed the other suggestions.

acoburn: The comments are great! Maven, however, will complain unless you include:

    @param r the Fedora resource
    @param translator a tranlation function
    @return a function to convert triples ....

mikedurbin: We're trying to reduce those warnings?  Fair enough. 

I think it's better now.

awoods:
:+1: Looks good to me. Ship it when others are satisfied... and unblock the embargo on `master`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
ajs6f: To my recollection, those aren't just warnings. They _should_ fail the build.

PR #1040, opened 2016-05-20
Fix ETag handling for Binaries and Binary descriptions

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1754

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -483,6 +483,16 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
 
     /**
      * Add ETag and Last-Modified cache control headers to the response
+     * <p>
+     * Note: In this implementation, the HTTP headers for ETags and Last-Modified dates are swapped
+     * for fedora:Binary resources and their descriptions. Here, we are drawing a distinction between
+     * the HTTP resource and the LDP resource. As an HTTP resource, the last-modified header should
+     * reflect when the resource at the given URL was last changed. With fedora:Binary resources and
+     * their descriptions, this is a little complicated, for the descriptions have, as their subjects,
+     * the binary itself. And the fedora:lastModified property produced by that NonRdfSourceDescription
+     * refers to the last-modified date of the binary -- not the last-modified date of the
+     * NonRdfSourceDescription.
+     * </p>
      * @param servletResponse the servlet response
      * @param resource the fedora resource
      * @param session the session
@@ -500,22 +510,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
         final EntityTag etag;
         final Date date;
 
-        // The HTTP headers for ETags and Last-Modified dates are swapped for fedora:Binary
-        // resources and their descriptions. This may seem twisted.
-        //
-        // Here we are drawing a distinction between the HTTP resoruce and the LDP resource.
-        // As an HTTP resource, the last-modified header should reflect when the resource
-        // at the given URL was last changed. With fedora:Binary resources and their descriptions,
-        // this gets a little complicated, for the descriptions have, as their subjects, the
-        // binary itself. And the `fedora:lastModified` property produced by that NonRdfSourceDescription
-        // refers to the last-modified date of the binary -- not the last-modified date of the
-        // NonRdfSourceDescription. While that works for LDP, that falls on its face for HTTP.
-        //
-        // The reason being that a change to the NonRdfSourceDescription (e.g. adding or removing
-        // a property) does not change the fedora:lastModified date of the binary, yet, the
-        // representation of the triples in that NonRdfSourceDescription _did_ change. So, while
-        // the triple for fedora:lastModified will not change, the HTTP header for last-modified
-        // must change.
+        // See note about this code in the javadoc above.
         if (resource instanceof NonRdfSourceDescription) {
             final NonRdfSourceDescription description = (NonRdfSourceDescription)resource;
             // Use a weak ETag for the LDP-RS
@@ -574,7 +569,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
         final Date roundedDate = new Date();
 
         // See the related note about the next block of code in the
-        // ContentExposingResource::addCacheControlHeaders function
+        // ContentExposingResource::addCacheControlHeaders method
         if (resource instanceof NonRdfSourceDescription) {
             final NonRdfSourceDescription description = (NonRdfSourceDescription)resource;
             // Use a weak ETag for the LDP-RS
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -497,10 +497,40 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
             return;
         }
 
-        final FedoraResource mutableResource = resource instanceof NonRdfSourceDescription
-                ? ((NonRdfSourceDescription) resource).getDescribedResource() : resource;
-        final EntityTag etag = new EntityTag(mutableResource.getEtagValue());
-        final Date date = mutableResource.getLastModifiedDate();
+        final EntityTag etag;
+        final Date date;
+
+        // The HTTP headers for ETags and Last-Modified dates are swapped for fedora:Binary
+        // resources and their descriptions. This may seem twisted.
+        //
+        // Here we are drawing a distinction between the HTTP resoruce and the LDP resource.
+        // As an HTTP resource, the last-modified header should reflect when the resource
+        // at the given URL was last changed. With fedora:Binary resources and their descriptions,
+        // this gets a little complicated, for the descriptions have, as their subjects, the
+        // binary itself. And the `fedora:lastModified` property produced by that NonRdfSourceDescription
+        // refers to the last-modified date of the binary -- not the last-modified date of the
+        // NonRdfSourceDescription. While that works for LDP, that falls on its face for HTTP.
+        //
+        // The reason being that a change to the NonRdfSourceDescription (e.g. adding or removing
+        // a property) does not change the fedora:lastModified date of the binary, yet, the
+        // representation of the triples in that NonRdfSourceDescription _did_ change. So, while
+        // the triple for fedora:lastModified will not change, the HTTP header for last-modified
+        // must change.
+        if (resource instanceof NonRdfSourceDescription) {
+            final NonRdfSourceDescription description = (NonRdfSourceDescription)resource;
+            // Use a weak ETag for the LDP-RS
+            etag = new EntityTag(description.getDescribedResource().getEtagValue(), true);
+            date = description.getDescribedResource().getLastModifiedDate();
+        } else if (resource instanceof FedoraBinary) {
+            final NonRdfSource binary = (NonRdfSource)resource;
+            // Use a strong ETag for the LDP-NR
+            etag = new EntityTag(binary.getDescription().getEtagValue());
+            date = binary.getDescription().getLastModifiedDate();
+        } else {
+            // Use a weak ETag for the LDP-RS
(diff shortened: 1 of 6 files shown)


acoburn:
This supersedes #991 and #920

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -497,10 +497,40 @@ protected static void addCacheControlHeaders(final HttpServletResponse servletRe
...
         }
 
-        final FedoraResource mutableResource = resource instanceof NonRdfSourceDescription
-                ? ((NonRdfSourceDescription) resource).getDescribedResource() : resource;
-        final EntityTag etag = new EntityTag(mutableResource.getEtagValue());
-        final Date date = mutableResource.getLastModifiedDate();
+        final EntityTag etag;
+        final Date date;
+
+        // The HTTP headers for ETags and Last-Modified dates are swapped for fedora:Binary
```
ajs6f: This is all good info, but there's so much of it that maybe it should extracted to a Javadoc-level comment about the kinds of resources at play and the general considerations for each (e.g. LDP resources get triples for description, HTTP resources get headers, etc.).

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -539,10 +569,28 @@ private static void evaluateRequestPreconditions(final Request request,
...
 
-        final EntityTag etag = new EntityTag(resource.getEtagValue());
-        final Date date = resource.getLastModifiedDate();
+        final EntityTag etag;
+        final Date date;
         final Date roundedDate = new Date();
 
+        // See the related note about the next block of code in the
+        // ContentExposingResource::addCacheControlHeaders function
+        if (resource instanceof NonRdfSourceDescription) {
```
ajs6f: Note to self: Java has no real pattern matching, so we might want to install `isRdf`/`isNotRdf` methods on `FedoraResource`.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -434,7 +434,12 @@ public Response updateSparql(@ContentLocation final InputStream requestBodyStrea
             }
             session.save();
 
-            addCacheControlHeaders(servletResponse, resource(), session);
+            if (resource() instanceof FedoraBinary) {
+                final NonRdfSourceDescription description = ((FedoraBinary)resource()).getDescription();
```
ajs6f: Do stanzas like this become easier to understand if we move `getDescription()` to `FedoraResource` and let an RDF resource describe itself?

acoburn: Yes, that change would certainly make this type of thing easier.

acoburn: @ajs6f are you suggesting that we move the `getDescription` method to `FedoraResource` as part of this PR or would that be something separate?

acoburn:
@ajs6f: I restructured the large inline comment, placing it in the javadoc

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
ajs6f: As per discussion in RC, definitely a separate story.

ajs6f:
👍

awoods:
Let's hold of on merging more into `master` until the following issue has been resolved:
https://github.com/fcrepo4/fcrepo4/pull/1039

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -539,10 +564,28 @@ private static void evaluateRequestPreconditions(final Request request,
...
+        final EntityTag etag;
+        final Date date;
         final Date roundedDate = new Date();
 
+        // See the related note about the next block of code in the
+        // ContentExposingResource::addCacheControlHeaders method
+        if (resource instanceof NonRdfSourceDescription) {
+            final NonRdfSourceDescription description = (NonRdfSourceDescription)resource;
+            // Use a weak ETag for the LDP-RS
+            etag = new EntityTag(description.getDescribedResource().getEtagValue(), true);
```
awoods: This is the Etag of the binary... which should be strong, no?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -539,10 +564,28 @@ private static void evaluateRequestPreconditions(final Request request,
...
+            date = description.getDescribedResource().getLastModifiedDate();
+        } else if (resource instanceof NonRdfSource) {
+            final NonRdfSource binary = (NonRdfSource)resource;
+            // Use a strong ETag for the LDP-NR
+            etag = new EntityTag(binary.getDescription().getEtagValue());
```
awoods: Likewise, this is the Etag of the description, which should be weak, no?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
acoburn: Nope, it's the binary description: `resource instanceof NonRdfSourceDescription`, not the Binary

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
acoburn: No, you have it backwards. This is the binary: `resource instanceof NonRdfSource`. The confusing logic here is described in the javadoc comment.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -497,10 +507,25 @@ protected static void addCacheControlHeaders(final HttpServletResponse servletRe
...
+        // See note about this code in the javadoc above.
+        if (resource instanceof NonRdfSourceDescription) {
+            final NonRdfSourceDescription description = (NonRdfSourceDescription)resource;
+            // Use a weak ETag for the LDP-RS
+            etag = new EntityTag(description.getDescribedResource().getEtagValue(), true);
+            date = description.getDescribedResource().getLastModifiedDate();
+        } else if (resource instanceof FedoraBinary) {
+            final NonRdfSource binary = (NonRdfSource)resource;
+            // Use a strong ETag for the LDP-NR
+            etag = new EntityTag(binary.getDescription().getEtagValue());
```
awoods: Despite the text in the javadoc above, swapping the etags/lastmod is not clear to me. For example, if I perform a `GET` request on the binary, I would expect the Etag to reflect the state of the binary. If the binary's description changes, I do not want to have to re-download the binary on a subsequent `GET` because my Etag is now different.

acoburn: Are you actually seeing weak ETags with binaries? The code here will produce strong ETags with binaries and weak ETags for everything else.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -434,7 +434,12 @@ public Response updateSparql(@ContentLocation final InputStream requestBodyStrea
             }
             session.save();
 
-            addCacheControlHeaders(servletResponse, resource(), session);
+            if (resource() instanceof FedoraBinary) {
```
awoods: Is there a reason this same logic is not also included here:
https://github.com/acoburn/fcrepo4/blob/89f14e961631ba32b2525aeb36f58c648013683d/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java#L597

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -497,10 +507,25 @@ protected static void addCacheControlHeaders(final HttpServletResponse servletRe
...
+        final EntityTag etag;
+        final Date date;
+
+        // See note about this code in the javadoc above.
+        if (resource instanceof NonRdfSourceDescription) {
+            final NonRdfSourceDescription description = (NonRdfSourceDescription)resource;
+            // Use a weak ETag for the LDP-RS
+            etag = new EntityTag(description.getDescribedResource().getEtagValue(), true);
+            date = description.getDescribedResource().getLastModifiedDate();
+        } else if (resource instanceof FedoraBinary) {
```
awoods: Is this logic ever invoked given: https://github.com/fcrepo4/fcrepo4/pull/1040/files#diff-4a14fb4dcbde6f7d32274c43ca723c16R438

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
acoburn: Yes, because that logic is specific to handling PATCH operations on NonRdfSourceDescriptions -- for PUT and POST, NonRdfSourceDescriptions don't apply.

N.B. I realize that this code is checking for an `instanceof` a FedoraBinary. I actually do not understand why that works, but it is actually quite necessary. You will notice in the code above (in the same function), there is an earlier check for FedoraBinary, which returns false. Somehow, after calling `session.save()` the FedoraResource discovers that it is a FedoraBinary and not a NonRdfSourceDescription.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
acoburn: Yes the logic is invoked. Otherwise, when requesting a binary with an existing ETag:

    curl -i localhost:8080/fcrepo/rest/binary -H"If-None-Match: \"SOME-ETAG\""

you'd get a weak ETag in the response. My testing of this indicated that you receive a strong ETag. The same invocation on a container will result in a weak ETag.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -708,6 +708,128 @@ public void testPutBinary() throws IOException {
...
+        try (final CloseableHttpResponse response = execute(head1)) {
+            binaryEtag2 = response.getFirstHeader("ETag").getValue();
+            binaryLastModed2 = response.getFirstHeader("Last-Modified").getValue();
+        }
+
+        assertEquals("ETags should be the same", binaryEtag1, binaryEtag2);
+        assertEquals("Last-Modified should be the same", binaryLastModed1, binaryLastModed2);
+
+        final HttpGet get6 = new HttpGet(location);
+        get6.addHeader("If-None-Match", binaryEtag1.toString());
```
awoods: - Are `get6` and `get7` redundant with the `head1` and associated `assertEquals` checks?
- Also, `get6` uses `binaryEtag1` while `get7` uses `binaryLastModed2`... which seems slight asymmetric.
- Minor: The `toString()` in "get6.addHeader("If-None-Match", binaryEtag1.toString());" is unnecessary.

acoburn: In principle, `get6` and `get7` are redundant. In practice, this is checking for the error that currently exists in the main codebase -- that the reported ETag doesn't match the ETag used internally.

Yes, I can change `binaryLastModed2` to `binaryLastModed1`. There is an earlier check to make sure those two values are the same, but that change might make it easier to follow.

I will remove the toString() in a subsequent commit.

awoods:
Your tests in the PR are thorough, thanks! The logic is somewhat confusing in `ContentExposingResource`, but that is no fault of your own... and it works.

acoburn:
@awoods I added a commit to address your last comment.

PR #1041, opened 2016-05-23
Remove more modeshape dependencies from the HTTP layers

acoburn:
Partially resolves: https://jira.duraspace.org/browse/FCREPO-1694

In particular, this removes the dependency on the modeshape-based namespace service,
~~and extracts the TransactionServiceImpl from the HTTP layer.~~

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
@@ -26,7 +26,6 @@ import static org.fcrepo.kernel.modeshape.identifiers.NodeResourceConverter.node
 import static org.fcrepo.kernel.modeshape.services.TransactionServiceImpl.getCurrentTransactionId;
 import static org.fcrepo.kernel.modeshape.utils.FedoraTypesUtils.getClosestExistingAncestor;
 import static org.fcrepo.kernel.modeshape.utils.NamespaceTools.validatePath;
-import static org.modeshape.jcr.api.JcrConstants.JCR_CONTENT;
 import static org.slf4j.LoggerFactory.getLogger;
 import static org.springframework.web.context.ContextLoader.getCurrentWebApplicationContext;
 
@@ -369,16 +368,13 @@ public class HttpResourceConverter extends IdentifierConverter<Resource,FedoraRe
      * @return
      */
     private String doBackwardPathOnly(final FedoraResource resource) {
-        String path = reverse.convert(getPath(resource));
+        final String path = reverse.convert(getPath(resource));
         if (path != null) {
 
             if (resource instanceof NonRdfSourceDescription) {
-                path = path + "/" + FCR_METADATA;
+                return path + "/" + FCR_METADATA;
             }
 
-            if (path.endsWith(JCR_CONTENT)) {
-                path = replaceOnce(path, "/" + JCR_CONTENT, EMPTY);
-            }
             return path;
         }
         throw new RepositoryRuntimeException("Unable to process reverse chain for resource " + resource);
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/NamespaceService.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/NamespaceService.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2015 DuraSpace, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.kernel.api.services;
+
+import java.util.Map;
+
+import javax.jcr.Session;
+
+/**
+ * @author acoburn
+ * @since 5/20/2016
+ */
+public interface NamespaceService {
+    /**
+     * Retrieve a namespace map using a session.
+     *
+     * @param session the session
+     * @return a namespace mapping (prefix to URI)
+     */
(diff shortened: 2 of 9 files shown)


fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
```diff
@@ -389,7 +393,7 @@ protected void resetTranslationChain() {
         if (translationChain == null) {
             translationChain = getTranslationChain();
             final List<Converter<String, String>> newChain =
-                    new ArrayList<>(singleton(new TransactionIdentifierConverter(session)));
+                    new ArrayList<>(singleton(new TransactionIdentifierConverter(transactionService, session)));
```
ajs6f: Is this just `Collections::singletonList`?

ajs6f: Oh, this is weirder than that. Never mind. Do whatever here and I will send an orthogonal PR at some future point.

acoburn: Yes, it is. I'll change that

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/session/SessionFactory.java
```diff
@@ -136,6 +136,16 @@ public Session getSession(final HttpServletRequest servletRequest) {
     }
 
     /**
+     * Determine whether the given Session is part of a transaction
+     *
+     * @param session
+     * @return whether the session is part of a transaction
+     */
+    public boolean isInTransaction(final Session session) {
```
ajs6f: Are we trying to get JCR out of `-http-commons`? Is this is a temporary maneuver?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
ajs6f: No, the return of `singletonList` is immutable, don't worry about it. Like I said, I'll straighten this up later. Let's not tangle it in with the good stuff you are doing here.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/session/SessionFactory.java
acoburn: I would like to remove JCR from the HTTP layers completely, but this PR is specifically about chipping away at the Modeshape dependency in the `-http-commons` and `-http-api` modules. Removing JCR is a separate (though very much related) thing.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/NamespaceService.java
```diff
@@ -0,0 +1,34 @@
...
+ * @since 5/20/2016
+ */
+public interface NamespaceService {
+    /**
+     * Retrieve a namespace map using a session.
+     *
+     * @param session the session
+     * @return a namespace mapping (prefix to URI)
+     */
+    Map<String, String> getNamespaces(final Session session);
```
ajs6f: Just a thought: if we declare that all work goes on inside a transaction (a single isolated request is auto-wrapped into a short-lived transaction) this becomes `getNamespaces(final Transaction tx)` which is where we actually want to go, I think. That "autowrap" thing is working pretty well for Jena. We should strongly consider it in Fedora. It makes `Transaction` the natural and useful construct into which we dissolve `Session`.

acoburn: Yes, that makes a lot of sense to me. If I understand you correctly, the `org.fcrepo.kernel.api.Transaction` interface could eventually replace much of the existing instances where we rely on a `javax.jcr.Session` (at least outside of the `fcrepo-kernel-modeshape` module). Alternately, we could have a notion of a `FedoraSession` of which a `Transaction` is a specialization. Either way, removing `Session` from the kernel-api is the next big effort along this path to make room for alternate Fedora implementations.

ajs6f: Yeah, that's basically the idea. I'm not sure I understand your "alternate". What's the daylight between the hypothetical `FedoraSession` and `Transaction`?

acoburn: Well, the more I think about it, my proposed `FedoraSession` and `Transaction` seem pretty much the same.

ajs6f: Let's call it `FedTranSessAction` for now.

acoburn:
I added another commit to this PR, which scopes the `fcrepo-kernel-modeshape` dependency to `test` in the `fcrepo-http-api` module. There are still a few direct modeshape dependencies in the `fcrepo-http-common` module, but that will be part of a separate effort.

acoburn:
I updated this PR to be less aggressive, namely that it only moves the namespace handling into an external service -- it doesn't touch the TransactionService. Locally, I was able to establish that this does not adversely affect the auth modules or fcrepo-transform (and furthermore, requires no changes to those modules).

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
```diff
@@ -369,16 +368,13 @@ private static String getRelativePath(final FedoraResource child, final FedoraRe
...
-        String path = reverse.convert(getPath(resource));
+        final String path = reverse.convert(getPath(resource));
         if (path != null) {
 
             if (resource instanceof NonRdfSourceDescription) {
-                path = path + "/" + FCR_METADATA;
+                return path + "/" + FCR_METADATA;
             }
 
-            if (path.endsWith(JCR_CONTENT)) {
```
awoods: This PR looks good, but I wonder about the removal of this block. Is it that `JCR_CONTENT` no longer ever occurs at this point?

acoburn: `FedoraResource::getPath` will never end in `JCR_CONTENT`. See this: https://git.io/vrhaU (it was a somewhat recent change).

PR #1043, opened 2016-05-27
Remove jcr-based serialization component

acoburn:
Addresses: https://jira.duraspace.org/browse/FCREPO-1694 and https://jira.duraspace.org/browse/FCREPO-2045

~~This PR does _not_ remove the `fcrepo-serialization` module, nor does it remove the `fcr:export` or `fcr:import` endpoints. It _does_ remove the JCR-based serialization component from the core, so in practice, the core `fcrepo-webapp` does not contain links to `/fcr:export` or `/fcr:import`, and calling those endpoints will produce 404 errors. (404 may not be the correct response, but it seemed right as the serialization format is "not found").~~

~~If this appears to be the correct approach, the `JcrXmlSerialization` component will be put into its own extension module (e.g. `fcrepo-serialization-jcrxml`) that could live in `-exts` and be part of webapp-plus.~~

~~What this PR primarily achieves is the further removal of transitive dependencies on `fcrepo-kernel-modeshape` from the perspective of the HTTP-layers. At present `fcrepo-http-api` depends on `fcrepo-serialiation` which depends on `fcrepo-kernel-modeshape`. With this PR, `fcrepo-serialization` no longer depends on `fcrepo-kernel-modeshape`, but it still provides hooks for individual serializers to be injected into the module.~~

~~Alternately, if it is desired to completely remove the `fcr:export` / `fcr:import` endpoints, that would take more effort. Personally, I think that what is implemented here is a good structure.~~


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraExport.java
```diff
@@ -80,6 +79,10 @@ public Response exportObject(
         final FedoraObjectSerializer serializer =
             serializers.getSerializer(format);
 
+        if (serializer == null) {
+            return status(NOT_FOUND).entity("No " + format + " serialization available for " + resource.getPath())
```
ajs6f: Is this properly a `NOT FOUND` or is it more a 400 `BAD REQUEST`?

ajs6f: I know we're taking `format` as a query param now, but that's wrong. That should just be an `Accept` header.

ajs6f:
I think this goes in the right direction, but I would prefer to remove the endpoints entirely. Whether that's a good way for @acoburn to spend his time right now is a different question.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraExport.java
```diff
@@ -48,10 +48,9 @@
  */
 @Scope("request")
 @Path("/{path: .*}/fcr:export")
```
awoods: Is there any reason not to completely remove the `/fcr:export` endpoint?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraImport.java
```diff
@@ -51,10 +53,9 @@
  */
 @Scope("prototype")
 @Path("/{path: .*}/fcr:import")
-@Deprecated
 public class FedoraImport extends FedoraBaseResource {
```
awoods: Is there any reason to to completely remove the `/fcr:import` endpoint?

fcrepo-serialization/src/main/java/org/fcrepo/serialization/BaseFedoraObjectSerializer.java
```diff
@@ -19,35 +19,33 @@
...
-import org.springframework.beans.factory.annotation.Autowired;
 
 /**
  * Auto-wire some helpful services for the FedoraObjectSerializer
  *
  * @author cbeer
  */
-public abstract class BaseFedoraObjectSerializer implements
-        FedoraObjectSerializer {
+public abstract class BaseFedoraObjectSerializer implements FedoraObjectSerializer {
```
awoods: This class can probably go away as well.

acoburn: @awoods: all of this will go away. hang on for another commit: your review will be much easier.

awoods:
For clarity, we should put this PR under a new JIRA ticket, along the lines of: "Remove fcr:/import and fcr:/export Endpoints"

ajs6f:
https://jira.duraspace.org/browse/FCREPO-2045

acoburn:
@awoods @ajs6f the last commit completely removes the fcrepo-serialization module and the `/fcr:import` and `/fcr:export` endpoints.

PR #1048, opened 2016-06-03
Remove modeshape's ServletCredentials from the HTTP layers

acoburn:
Addresses: https://jira.duraspace.org/browse/FCREPO-1694

I am taking these one at a time now. Please bear with me. The `Repository::login` method requires a `javax.jcr.Credentials` object (which is a simple marker class: it doesn't do anything on its own), and the fcrepo-auth-* modules expect that the `Credentials` object is an `instanceof` modeshape's `ServletCredentials`.

Thus, in order to remove this particular modeshape dependency from `fcrepo-http-commons`, I added a `CredentialsService` as an injected "shim". There may be other ways to implement this, but I couldn't think of another way to do this without extensive changes. I would appreciate a review of this.

### DIFF
--- a/fcrepo-connector-file/pom.xml
+++ b/fcrepo-connector-file/pom.xml
@@ -49,6 +49,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>javax.servlet</groupId>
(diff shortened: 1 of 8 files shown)


awoods:
:+1:

PR #1053, opened 2016-06-15
Implement an RDF-based message serialization

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1945
Addresses: https://jira.duraspace.org/browse/FCREPO-1960

This is a fairly large PR. I don't expect it will be merged anytime soon, but I'd like to get it out there for review.

This retains the JMS headers as part of the message serialization only to maintain some bit of backwards compatibility with the existing camel-based messaging tooling, but I expect those will be removed once a messaging spec is formalized.

This also reorganizes the `fcrepo-jms` module -- the `headers` package no longer made sense and the overall package structure was simplified.


acoburn:
This PR supersedes #1008

fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/JmsIT.java
```diff
@@ -172,24 +167,24 @@ public void testMetadataEvents() throws RepositoryException {
...
     }
 
-    private void awaitMessageOrFail(final String id, final String eventType, final String property) {
-        await().pollInterval(ONE_SECOND).until(() -> messages.stream().anyMatch(msg -> {
+    private void awaitMessageOrFail(final String id, final String eventType, final String type) {
+        await().pollInterval(ONE_HUNDRED_MILLISECONDS).until(() -> messages.stream().anyMatch(msg -> {
             try {
                 return getPath(msg).equals(id) && getEventTypes(msg).contains(eventType)
-                        && (property == null || getProperties(msg).contains(property));
+                        && (type == null || getResourceTypes(msg).contains(type));
```
ajs6f: Might be a little more readable using [`nullToEmpty`](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/base/Strings.html#nullToEmpty(java.lang.String)).

fcrepo-kernel-api/pom.xml
```diff
@@ -69,6 +69,22 @@
       <groupId>javax.servlet</groupId>
       <artifactId>javax.servlet-api</artifactId>
     </dependency>
+    <dependency>
```
ajs6f: Whuh? Why is this JSON impl stuff coming into the kernal _API_? I don't like this at all.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/EventMessage.java
```diff
@@ -0,0 +1,171 @@
...
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.slf4j.Logger;
+
+/**
+ * A structure used for serializing a FedoraEvent into JSON
```
ajs6f: Why is serialization stuff for a single suddenly-privileged showing up in the API module?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/EventType.java
```diff
@@ -15,25 +15,27 @@
...
 public enum EventType {
-    NODE_ADDED("node added"),
-    NODE_REMOVED("node removed"),
-    PROPERTY_ADDED("property added"),
-    PROPERTY_REMOVED("property removed"),
-    PROPERTY_CHANGED("property changed"),
-    NODE_MOVED("node moved"),
-    PERSIST("persist");
+
+    RESOURCE_CREATION("resource creation", "ResourceCreation"),
```
ajs6f: How is `COPY` interpreted in this `enum`?

ajs6f:
A big 👎 to the massive privileging of JSON in this PR. That stuff does not belong in an API.

acoburn:
The reason I put the JSON code in the kernel-api was to make it possible for multiple messaging modules (from multiple impls) to reuse the serialization code. A pretty simple alternative would be to create another module, say fcrepo-message-serialization and put the JSON code there.

ajs6f:
Yeah, @escowles and I just had a hallway chat about this and that option came up as attractive. Might even be better to be more specific, `fcrepo-message-serialization-jsonld`. That "LD", by the way, is quite important. Without it my -1 would turn into a committer veto. We do RDF first and we can require a given serialization, but that's a dependency on the primary fact of RDF. Maybe LDP's [requirement](https://dvcs.w3.org/hg/ldpwg/raw-file/default/ldp.html#h-ldprs-http_get) of Turtle and JSON-LD is the right road here.

acoburn:
@ajs6f I pulled all of the serialization code out of `fcrepo-kernel-impl`. It now is spread across two new modules: `fcrepo-event-serialization-rdf` and `fcrepo-event-serialization-jsonld`. The `-serialization-rdf` module contains an interface that is implemented in `-jsonld`. (There is also a Turtle-based impl in -serialization-rdf, too). The fcrepo-jms code is hard wired to use the -jsonld module, but it would be possible to inject that dependency via spring, given this structure (I'm saving that for a separate effort).

acoburn:
In principle, the two new modules `fcrepo-event-serialization-rdf` and `fcrepo-event-serialization-jsonld` could be collapsed into a single module. I'm somewhat inclined to combine them, but there might also be some value in keeping them separate, especially since the dependency chains are rather different.

escowles:
👍 to moving the implementations out into separate modules to keep the API leaner, and I think it would be fine to combine them.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/observer/EventType.java
acoburn: COPY is interpreted the same as it always has been: as the creation of a new resource, i.e. RESOURCE_CREATION or what used to be NODE_ADDED.

fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/JmsIT.java
acoburn: Here we are distinguishing between an input parameter that is either `null` (in that case, just evaluate to `true`) or, in the case of a non-null value, check that it appears in the `Set` of resource types. With `nullToEmpty`, I suppose we could make the line into:

    && (nullToEmpty(type).isEmpty() || getResourceTypes(msg).contains(type));

but I don't think that's any more clear.

ajs6f: No, all strings contain the empty string.

acoburn: Then I clearly have no idea what you're suggesting. If you would like to make a PR against this PR, implementing what you propose, please do.

acoburn: @ajs6f nevermind, now I see what you're talking about -- your suggestion has been added in the latest commit.

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/TurtleSerializer.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Serialize a FedoraEvent as Turtle
+ * @author acoburn
+ * @since 6/16/16
+ */
+public class TurtleSerializer implements EventSerializer {
+
+    /**
+     * Serialize a FedoraEvent into a JSON String
```
awoods: This javadoc block should be changed to Turtle language from JSON.

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/JsonLDSerializer.java
```diff
@@ -0,0 +1,61 @@
...
+     * Serialize a FedoraEvent into a JSON String
+     * @param evt the Fedora event
+     * @return a JSON string
+     */
+    @Override
+    public String serialize(final FedoraEvent evt) {
+        try {
+            return MAPPER.writeValueAsString(from(evt));
+        } catch (final JsonProcessingException ex) {
+            LOGGER.warn("Error processing JSON: " + ex.getMessage());
```
awoods: As the message suggests, not processing a message seems like a `LOGGER.error()` vs. a `LOGGER.warn()`.

awoods: Additionally, please use the optimized template syntax offered by slf4j:
`"Error processing JSON: {}", ex.getMessage()`

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/EventSerializer.java
```diff
@@ -0,0 +1,94 @@
...
+ * @author acoburn
+ */
+public interface EventSerializer {
+
+    /**
+     * Convert an event to an Rdf Model
+     * @param evt the Fedora event
+     * @return an RDF model representing the event
+     */
+    public static Model toModel(final FedoraEvent evt) {
```
awoods: Since `toModel` is only used by `TurtleSerializer`, would it make sense to move this static method into `TurtleSerializer`?

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/EventMessage.java
```diff
@@ -0,0 +1,172 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.fcrepo.kernel.api.observer.FedoraEvent;
+import org.slf4j.Logger;
+
+/**
+ * A structure used for serializing a FedoraEvent into JSON
+ * @author acoburn
+ */
+class EventMessage {
```
awoods: I think this class would read more clearly if named `JsonLDEventMessage`.

fcrepo-event-serialization/src/test/java/org/fcrepo/event/serialization/FedoraEventTest.java
```diff
@@ -0,0 +1,141 @@
...
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+/**
+ * <p>FedoraEventTest class.</p>
+ *
+ * @author acoburn
+ */
+@RunWith(MockitoJUnitRunner.class)
+public class FedoraEventTest {
```
awoods: These class names may well be a result of splitting and combining maven modules, but this test leans towards the `TurtleSerializer`, particularly if `EventSerializer.toModel()` is moved to `TurtleSerializer`. 
Naming this class `FedoraEventTest` seems to give it greater emphasis... particularly since the default "Fedora Events" are serialized with the `JsonLDSerializer`.

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/EventSerializer.java
acoburn: I am inclined to keep this method where it is. The method converts a `FedoraEvent` into a generic RDF `Model`, independent of any concrete RDF syntax. For example, if someone wanted to serialize the messages as N-Triples, the implementation would need the `toModel` method. If we are certain that _only_ JSON-LD and Turtle are to be supported, then yes, it should move to the TurtleSerializer, but I don't think we want to make that assumption.

fcrepo-event-serialization/src/test/java/org/fcrepo/event/serialization/FedoraEventTest.java
acoburn: This may be a difference of opinion here, but I think fundamentally, what we are doing is converting a `FedoraEvent` to RDF, independent of any concrete RDF syntax. It is that conversion that I am verifying here. The `testTurtle` is thrown in more as an afterthought. Assuming that we keep `toModel` in the `EventSerializer`, are you suggesting that I create a separate test class for the Turtle serializer?

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/EventSerializer.java
awoods: Presumably the only difference required to support alternate RDF serializations (that are supported by Jena Models) would involve a parameterized version of your existing `TurtleSerializer`. Do you envision the possibility of creating separate classes for each Jena supported serialization in order to effectively change the string literal found here:
https://github.com/fcrepo4/fcrepo4/pull/1053/files/7bd36eddb3bae3e7a50dc672fdbb29a597fc6e27#diff-abb99f53759398650974df4fa035e481R42

acoburn: I was anticipating that (alternate) classes would be spring-injected. Clearly, one option would be to change the TurtleSerializer class to something like RdfSerializer and parameterize the class construction. I don't actually care that much about how this is organized, though. If you prefer to allow users to parameterize class construction like this, that's fine -- it just requires more knowledge of Jena models on the users' part. I like my approach better because it assumes the user/sysadmin knows that they want Turtle or NTriples or RDF/XML but don't necessarily need to care about how Jena creates those serializations (or what the String values are that correspond to those serializations).

acoburn:
@awoods I have addressed the code review comments (well, most of them -- the others I'm letting stand).

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/eventmappings/AllNodeEventsOneEvent.java
```diff
@@ -78,21 +77,8 @@
...
-                            final String eventPath = evt.getPath();
-                            fedoraEvent.addProperty(eventPath.substring(eventPath.lastIndexOf('/') + 1));
-                        } else {
-                            LOGGER.trace("Not adding non-event property: {}, {}", fedoraEvent, evt);
-                        }
-                    } catch (final RepositoryException ex) {
-                        throw new RepositoryRuntimeException(ex);
-                    }
+                    fedoraEvent.getTypes().add(FedoraEventImpl.valueOf(evt.getType()));
+                    fedoraEvent.getResourceTypes().addAll(getResourceTypes(evt).collect(toSet()));
```
awoods: These two lines appear to be redundant with:
https://github.com/fcrepo4/fcrepo4/pull/1053/files#diff-9801cf3467e0ba8fcc5862b6b0ca5c50R256 ...and below.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/DefaultFilter.java
```diff
@@ -56,31 +47,11 @@
 
     private static final Logger LOGGER = getLogger(DefaultFilter.class);
```
awoods: Looks like we can remove the `LOGGER` now.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/eventmappings/AllNodeEventsOneEvent.java
acoburn: No, actually those two lines are very important. The issue is this: say you have a lot of JCR events for a given node. These are collected into a single FedoraEvent in this class. But it turns out that those JCR events may have very different eventTypes and resourceTypes. These lines collect all of those into this FedoraEvent. You can try taking those lines out and seeing all the tests start to fail.

acoburn: You can see that the current impl does something similar, aggregating node properties here: https://git.io/vKet6

awoods: Got it. Thanks.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/FedoraEventImpl.java
```diff
@@ -58,56 +73,57 @@
...
         requireNonNull(path, "FedoraEvent requires a non-null path");
 
         this.eventTypes.addAll(types);
         this.path = path;
+        this.eventResourceTypes = resourceTypes;
         this.userID = userID;
-        this.userData = userData;
         this.date = date;
         this.info = isNull(info) ? emptyMap() : info;
-        this.eventID = pidMinter.get();
```
awoods: The reason a `pidMinter` is used here is because the `InternalAuditor` uses the `eventID` as the resource path at which the audit event is created:
https://github.com/fcrepo4-exts/fcrepo-audit/blob/master/src/main/java/org/fcrepo/audit/InternalAuditor.java#L203-L204

If a simple UUID is used, we will quickly run into the "many children" issue under the `/audit` container.

acoburn: That seems more like an issue with the `InternalAuditor` than with a `FedoraEvent::eventID`? For external applications (which don't generally have Modeshape's "many children" problem), I wanted to provide a UUID in a standard URI form: e.g. `urn:uuid:<some-uuid>`. See: https://en.wikipedia.org/wiki/Uniform_Resource_Name#Syntax

Couldn't the `InternalAuditor` be adjusted to convert this UUID into an appropriate pairtree structure?

escowles: Using a UUID here and creating the pairtree in the InternalAuditor seems reasonable to me.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/eventmappings/AllNodeEventsOneEvent.java
acoburn: To your point about redundancy, though, we could add `.skip(1)` to the `Stream<Event>`, since as you indicated, the values from the first JCR event are already part of the FedoraEvent

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/FedoraEventImpl.java
ajs6f: +1. Doing it here seems like a conflation of concerns.

awoods: Sounds good. We should update the `InternalAuditor` to create the "path-minter-like" hierarchy as a part of this effort... as well as making fcrepo-audit compile ;)

Additionally, similar logic will need to be implemented for the "external auditor" found in the fcrepo-camel-toolbox, since it uses the eventID as the subject of the audit triples stored in the external triplestore. It is important that the external triplestore audit subjects are the same URIs that are found in the repository:
https://github.com/fcrepo4-exts/fcrepo-camel-toolbox/blob/master/fcrepo-audit-triplestore/src/main/java/org/fcrepo/camel/audit/triplestore/AuditSparqlProcessor.java#L84

PR #1054, opened 2016-06-19
Allow FedoraResources to describe and be described by themselves

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-2059

This also removes the `org.fcrepo.kernel.api.models.NonRdfSource` interface, which, if it remained would no longer extend `FedoraResource` in any meaningful way, nor would it act as a marker in any way that `FedoraBinary` wouldn't. Now `FedoraBinary` extends `FedoraResource` directly rather than via `NonRdfSource`. The `NonRdfSourceDescription` interface also doesn't _extend_ `FedoraResource` in a particularly meaningful way, but it does provide a potentially useful marker on objects.


whikloj:
👏

fcrepo-http-api/src/main/java/org/fcrepo/http/api/url/HttpApiResources.java
```diff
@@ -64,7 +63,7 @@ public Model createModelForResource(final FedoraResource resource,
         }
 
         if (resource instanceof NonRdfSourceDescription) {
-            final NonRdfSource describedResource = ((NonRdfSourceDescription) resource).getDescribedResource();
+            final FedoraResource describedResource = resource.getDescribedResource();
```
awoods: Given this update, it seems like this block (lines 65 to 73) can be simplified.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -228,7 +227,7 @@ protected RdfStream getResourceTriples() {
     protected RdfStream getResourceTriples(final int limit) {
         // use the thing described, not the description, for the subject of descriptive triples
         if (resource() instanceof NonRdfSourceDescription) {
```
awoods: Is it still necessary to check if `resource()` is and `instanceof NonRdfSourceDescription`?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/url/HttpApiResources.java
acoburn: Yes, this block certainly can be simplified.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
acoburn: In principle, you are correct. In practice, I am unable to get the unit tests to pass. 

I am also concerned about this line in a more general sense: why is `resource` being _redefined_ in a method called `getResourceTriples`? That seems entirely wrong to me: any sort of accessor function should not be redefining internal, private variables. To me, this points to a need to refactor some of this code, and not just this particular `instanceof` check.

That is to say, I am leaving this as it stands for now.

PR #1055, opened 2016-06-20
Add interfaces for replacement of the existing IdentifierConverter class

acoburn:
Work-in-progress, addresses: https://jira.duraspace.org/browse/FCREPO-1957

This is a proof of concept interface that would eventually replace the existing IdentifierConverter abstract class.

If this seems like a good starting point, we can proceed from there.


awoods:
Although this is obviously very close to the existing [IdentifierConverter](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/identifiers/IdentifierConverter.java) and the Guava Converter, I can see how wrapping that functionality in something like you have here would be more intuitive.
Are you envisioning continuing to use "Guava-inside"? or a pure custom implementation?
@fcrepo4/committers, thoughts?

ajs6f:
What is the difference between `asString(a)` and `to().apply(a).toString()`?

ajs6f:
Are there invariants involving `equals` that we need here? What about ordering? Should impls of this type try to preserve the ordering of `Comparable`s?

acoburn:
@awoods I have two goals with this:
1. I'd like to see the interface for `IdentifierConverter` to be more understandable (I still find it confusing)
2. I'd like to see the `fcrepo-kernel-api` be _extremely_ light weight, and as much as I like using guava, I'd prefer to see it in implementation code and not in core APIs. This is why I keep bringing up the potential of rdf-commons on the tech calls (which keep being pushed to later weeks): to see if the jena deps can be pushed into impl code and out of core APIs.

ajs6f:
If the `Function`s being returned are stateful, but the state is in the converter, not the functions themselves, wouldn't it be more readable to have `B to(A a)` and `A from(B b)` and use method references where functions are needed?

acoburn:
@ajs6f I put the `asString` method in this interface only because it exists in the current implementation. I would be happy to remove it.

ajs6f:
👍 to killing it at this convenient juncture.

awoods:
I am happy to remove the `asString()` method, as long as we have a clear pattern for its replacement. This method on `IdentifierConverter` currently sees significant use.

ajs6f:
Multiplying out to `to().apply(a).toString()` should be automatic in an IDE. If `to().apply(a).toString()` does not instantly work as a drop-in replacement, we should know why and what the true semantic of `asString(A a)` is.

ajs6f:
More generally, I had understood this more like:
```
interface InvertibleFunction<A,B> extends Function<A,B> {

        A unapply(B b);

}
```

acoburn:
I think the `InvertibleFunction<A,B>` example above captures the intended semantics very well.

In addition to `apply` and `unapply`, should we have `inDomain` and `inRange` predicates? (the current implementation -- using Guava -- includes an `inDomain` predicate, which seems to be quite convenient, and an `inRange` predicate seems only fitting if this interface is trying to model a bijective function).

ajs6f:
That information _should_ be encoded in the type system, except Java. I would add, if anything, plain `boolean inDomain(A a)` and `boolean inRange(B b)` and the predicates are available as method references. Another way to think about this is more like Guava:
```
interface InvertibleFunction<A,B> extends Function<A,B> {

        boolean inDomain(A a);
        InvertibleFunction<B,A> inverse();
        default boolean inRange(B b) { return inverse().inDomain(b); }
}
```
I did something like that [here](https://github.com/apache/jena/commit/241060548a9fca777b7d40f1f216ae7ed930b20e#diff-dd8aae6715abe2f7535b2e858c78b925). In the end, that might be more flexible.

acoburn:
:+1: on the latest suggestion

ajs6f:
I would factor it apart a bit if you want to use it:
```
interface InvertibleFunction<A,B> extends Function<A,B> {
        InvertibleFunction<B,A> inverse();
}

interface PartialFunction<A,B> extends Function<A,B> {
        boolean inDomain(A a);
}

interface Converter<A,B> extends InvertibleFunction<A,B> ,  PartialFunction<A,B>{
        default boolean inRange(B b) { return inverse().inDomain(b); }
}
```
`PartialFunction` is a crummy name, but something like that.

acoburn:
How about `DomainRestrictedFunction` instead of `PartialFunction`?

acoburn:
@ajs6f the next iteration of this is ready for review.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/functions/InvertibleFunction.java
```diff
@@ -0,0 +1,34 @@
...
+ * @param <A> the type from which we are translating
+ * @param <B> the type to which we are translating
+ */
+public interface InvertibleFunction<A, B> extends Function<A, B> {
+    /**
+     * The inverse of the defined function
+     * @param <T> the type of the inverse of this function
+     * @return the inverse of the defined function.
+     */
+    <T extends InvertibleFunction<B, A>> T inverse();
```
ajs6f: This is definitely crisper and tastier than what I suggested; nice.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/functions/Converter.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public interface Converter<A, B> extends InvertibleFunction<A, B>, DomainRestrictedFunction<A, B> {
+
+    /**
+     * Whether the value is in the range of this function
+     * @param b a result of the converter function
+     * @return whether b is in the range of the converter
+     */
+    default boolean inRange(final B b) {
+        final Converter<B, A> converter = inverse();
```
ajs6f: Okay, we're getting into minutia here, but I'm honestly curious: why not just `return inverse().inDomain(b)`?

acoburn: I can't do just `inverse().inDomain(b)` -- I get a compiler error, since the returned type is `InvertibleFunction` which doesn't have an `inDomain` method. This, however, is possible: `((Converter)inverse()).inDomain(b)`, but I thought that looked ugly.

ajs6f: Oh, interesting. Maybe that's why I did all the gross `SelfType` and `InverseType` stuff in that Jena example. I can't remember. Okay, if you're happy, I'm happy. I think this expresses our intent about as clearly and concisely and completely as we can in the language in which we are working. At the moment. Let's see what kind of carnage this unleashes on the code base.

acoburn: OK, in terms of next steps, there is no way we're going to make the conversion from guava converters to these converters before the start of the 4.6.0 release. The question for @awoods is whether to merge these change now (independent of actually _using_ the interfaces) and then (after the 4.6.0 release) to transition away from the guava-based converters.

It would also be possible to do all of this in a single, massive PR, but I'd like to avoid that if possible.

Either way, I think it would be best to start work on this after the 4.6.0 release has been finalized; otherwise, the rebasing could be quite terrible.

ajs6f: I say leave 'em out for 4.6.0. They can do nothing but confuse and unsettle.

awoods:
Breaking this work into smaller pieces (with tickets for each) will make it more digestable... but I would prefer we save that for after 4.6.0.

acoburn:
@awoods @ajs6f is there any reason not to merge this into master now? Once #1075 (Jena 3.x) is merged, I will start working on the conversion to making use of this interface.

ajs6f:
Hm. One question now that I look at this afresh. Would you might want:
```
    default boolean inRange(final B b) {
        return inverse().inDomain(b);
    }

    @Override
    Converter<B,A> inverse();
```
in `Converter` and plain
```
   InvertibleFunction<B, A> inverse();
```
in `InvertibleFunction`? It has one advantage over what you have now: `Converter::inverse` is correctly constrained to be both `InvertibleFunction<B, A> & DomainRestrictedFunction<B,A>`. What you have there now just constrains to `InvertibleFunction<B, A>`. (That's why I can make `inRange` the one-liner where you couldn't.) But we lose the niceness of the generic in `InvertibleFunction::inverse`.

acoburn:
I like the simplicity of your suggestion. I will add a commit with that change. Once I get to _using_ these new interfaces, if it turns out we really need the generic of `InvertibleFunction::inverse` we can re-evaluate.

ajs6f:
I have a feeling that may have been the reason I did the self-type stuff in Jena, that I couldn't get a method-scoped type param to work for more arcane cases. But we may not even need that stuff. If we ever level up to the point that metamagic is interesting, we can take those feats then.

awoods:
@acoburn, you have landed on something nice here. :+1: to merging this PR once https://jira.duraspace.org/browse/FCREPO-1672 has completed review/merge.

barmintor:
Is there a reason to wait on Jena 3? I'm working on a branch to try and address the performance issue about reference properties, and one of the more immediate requirements is to decompose some of the IdentifierConverter impls. I'm forking on the MODE5 branch, but it would be nice to have something to cherrypick.

awoods:
Good point, @barmintor.

acoburn:
@awoods given @barmintor's work on the MODE5 branch, is there a reason _not_ to also merge that branch into master at this juncture, too?

awoods:
https://jira.duraspace.org/browse/FCREPO-1994 is ready to go.

PR #1057, opened 2016-06-24
Prevent SN-Sibling from creating that triggering the ConstraintViolationException when replacing binary contents. FCREPO-2055

lsitu:
@awoods, It's strange that we can't prevent SN-Sibling from creating at the first step when FedoraResource is created. So I think before calling the JCR session's save function should be the final chance to prevent any SN-Sibling nodes from created.


awoods:
Related to: https://jira.duraspace.org/browse/FCREPO-2055

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
```diff
@@ -357,6 +367,21 @@ protected void createDatastream(final String pid, final String dsid, final Strin
...
     }
 
+    protected Thread createSNSDatastreamThread(final String pid, final String dsid, final File content)
+            throws IOException {
+        return (new Thread() {
+            public void run() {
+
+            try {
+                logger.info("Attempting to create datastream for object: {} at datastream ID: {}", pid, dsid);
+                getStatus(putDSFileMethod(pid, dsid, content));
```
awoods: Like line 367 above, please `assertEquals` the status is a 201.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraHtmlIT.java
```diff
@@ -69,6 +73,57 @@ public void testGetDatastreamNode() throws IOException {
     }
 
     @Test
+    public void testConcurrentGetDatastreamNode() throws IOException, InterruptedException {
```
awoods: This test method should instead reside in `FedoraLdpIT.java`.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -389,7 +389,16 @@ public Response createOrReplaceObjectRdf(
         }
 
         try {
-            session.save();
+            // prevent SN-Sibling nodes from creating.
+            synchronized (LOGGER) {
```
awoods: Pushing this logic down to [FedoraLdp.createFedoraResource](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java#L717) will make it more reusable.

awoods:
Here is a PR against your dev branch that demonstrates that the SNS issue also exists when performing POST, as well as PUT:
https://github.com/lsitu/fcrepo4/pull/4

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
lsitu: Yes, it sounds good. I tried it again to push the logic down to FedoraLdp.createFedoraResource, but my test still get failed with the SN-sibling file get created (status code 201 instead of 409). So this should be the last chance to prevent the SN-sibling file from being created in this specific case.

awoods: That makes sense, since this is the last moment before the `save`.

awoods:
@lsitu, please incorporate useful elements of https://github.com/lsitu/fcrepo4/pull/4 in order to close this ticket.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
ajs6f: Why `synchronize` on `LOGGER`? `logback` loggers are threadsafe.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -195,6 +197,31 @@
...
+            TEST_FILE = File.createTempFile("content", ".txt");
+            TEST_FILE.deleteOnExit();
+            fw = new FileWriter(TEST_FILE);
+            for (int i = 0; i < 1000000; i++) {
+                fw.write("fooooooooooooooooooooooooooooooooooooooooooooooooooo");
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        } finally {
+            if (fw != null) {
```
ajs6f: `org.apache.commons.io.IOUtils.closeQuietly(Writer)` is neater than this.

lsitu:
@awoods I've merged your PR [lsitu#4|https://github.com/lsitu/fcrepo4/pull/4]. So it's ready for you now.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -2476,6 +2503,41 @@ public void testContainerLastModified() throws Exception {
...
+
+    @Test
+    public void testConcurrentPostDatastreamResource() throws IOException, InterruptedException {
+        final String pid = getRandomUniqueId();
+        createObject(pid);
+
+        final RequestThread t1 = new RequestThread(postDSFileMethod(pid, TEST_FILE));
+        final RequestThread t2 = new RequestThread(postDSFileMethod(pid, TEST_FILE));
+        t1.start();
+        Thread.sleep(100);
```
ajs6f: Why this `sleep`?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -195,6 +197,31 @@
         tripleFormat.setTimeZone(getTimeZone("GMT"));
     }
 
+    private static File TEST_FILE;
```
ajs6f: Why is `TEST_FILE` a `static` field? Why not just a local variable?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -2476,6 +2503,41 @@ public void testContainerLastModified() throws Exception {
...
 
+    @Test
+    public void testConcurrentPutDatastreamResource() throws IOException, InterruptedException {
+        final String pid = getRandomUniqueId();
+        createObject(pid);
+
+        final String ds = "ds2";
+        final RequestThread t1 = new RequestThread(putDSFileMethod(pid, ds, TEST_FILE));
+        final RequestThread t2 = new RequestThread(putDSFileMethod(pid, ds, TEST_FILE));
+        t1.start();
```
ajs6f: Again, why the `sleep`? If you want to sequence the threads, use a proper synchronization mechanism like `CountDownLatch`.

awoods:
@lsitu, the build is failing on a test I added that demonstrates that `POST` must also be considered in addition to `PUT`. Please take this PR forward. Thanks!

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
lsitu: @ajs6f I think this is just imitate the behavior of the HTML UI and make sure that first thread is done with status code 201 while the second thread failed with status code 409 or others. Thanks.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
```diff
@@ -357,6 +375,21 @@ protected void createDatastream(final String pid, final String dsid, final Strin
...
+            } catch (IOException e) {
+                e.printStackTrace();
```
ajs6f: `printStackTrace()` is bad form. Use a proper `Logger`, if logging a stacktrace is what is wanted.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
lsitu: @ajs6f Thanks. I think we can change it to use the FedoraLdp class or create a static object as the lock. What do you suggest?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
```diff
@@ -357,6 +375,21 @@ protected void createDatastream(final String pid, final String dsid, final Strin
         assertEquals(CREATED.getStatusCode(), getStatus(putDSMethod(pid, dsid, content)));
     }
 
+    protected Thread createSNSDatastreamThread(final String pid, final String dsid, final File content)
```
ajs6f: Where do you actually use this method?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
ajs6f: I do not understand why you are locking on anything.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
ajs6f: Where do you actually use this method?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
lsitu: @ajs6f Sorry maybe I need to explain that a user can use a program to submit the same ingest many times (see https://jira.duraspace.org/browse/FCREPO-2055), or click the HTML UI button many times to submit the same ingest with a malfunctioned browser?

ajs6f: Yes, they could. In no way does that prescribe locking on a static field in a JAX-RS resource that is not singleton-scoped. There is no possible diagnosis that leads to "lock on a static field" from that beginning. What is more, the fact that clients can misbehave in the way you describe, as they certainly can, is 1) not ameliorated by this locking and 2) absolutely _not_ something for which this code needs to account to begin with.

Can you actually demonstrate (with a test case or at least a careful bug report) the problem you are imagining here and demonstrate (with a test case) that your mechanism solves it?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
lsitu: Thanks. It becomes redundant with new test plan. I'll remove it then.

ajs6f: Code-prune like a weedwhacker!

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
```diff
@@ -423,4 +456,24 @@ protected static void assertDeleted(final String id) {
         assertThat("Expected object to be deleted", getStatus(new HttpHead(location)), is(GONE.getStatusCode()));
         assertThat("Expected object to be deleted", getStatus(new HttpGet(location)), is(GONE.getStatusCode()));
     }
+
+    class RequestThread extends Thread {
```
ajs6f: Rather than extending `Thread` and manually staging the executions, consider using calling `java.util.concurrent.Executors.newSingleThreadExecutor` to get an `Executor`, then calling `execute` sequentially on that with lambdas for each `Runnable`. That will guarantee the order of your two tasks being executed and should read _much_ more concisely.

lsitu:
@awoods I've added support to prevent SN-sibling from creating through POST and refactored the codes basing on the feedbacks. See commit https://github.com/fcrepo4/fcrepo4/pull/1057/commits/9b3c016e8311ad435ad7e9ee3157fadcbc62bec8. Thanks @ajs6f for your review and comments.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -763,6 +766,26 @@ private String mintNewPid(final String slug) {
...
+    /**
+     * perform save with checking to prevent SN-Sibling nodes from creating
+     * @param resource the FedoraResource need to save
+     * @throws javax.jcr.AccessDeniedException
+     * @throws VersionException
+     * @throws LockException
+     * @throws ConstraintViolationException
+     * @throws RepositoryException
+     */
+    private synchronized void save(final FedoraResource resource) throws javax.jcr.AccessDeniedException,
```
ajs6f: This is no better at all than what you had before. The keyword `synchronized` on an instance method just [locks on the monitor for the instance](https://docs.oracle.com/javase/tutorial/essential/concurrency/locksync.html). I do not know how much more clearly I can say this, but I will try once more: this class is a non-singleton JAX-RS resource. No two requests will be using the same instance. Synchronizing or locking on instance or static fields is _useless_ and _harmful_ here. Please remove this error.

lsitu: I think I can change it to a static method. But do you have other suggestions?

awoods: I am not sure making this a static method would be an improvement. My understanding is that without this `synchronized` block or method, the issue is not resolved. Is that true?
For example, do your integration tests fail if you remove the keyword `synchronized` here?

ajs6f: You cannot change it to a `static` method because you are clearly using instance fields in it (e.g. `session`, `resource`). Just remove `synchronized`, and _don't add anything_. That would be a good start.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -763,6 +766,26 @@ private String mintNewPid(final String slug) {
...
+            VersionException, LockException, ConstraintViolationException, RepositoryException {
```
ajs6f: Several of these exceptions are subtypes of `RepositoryException`. There is no point in throwing them specifically. Please remove them.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
lsitu: I just can't image if multiple threads are trying to call the save() method at the same time. Maybe we need another test with N threads to ingest the same file with PUT/POST concurrently?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraHtmlIT.java
```diff
@@ -68,6 +68,7 @@ public void testGetDatastreamNode() throws IOException {
         assertEquals(200, getStatus(method));
     }
 
+
```
awoods: Revert this space update.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
awoods: You are already doing this, no?
https://github.com/fcrepo4/fcrepo4/pull/1057/files/ba67f4a3027097bdaae93543c032f3d8a03a2aa0#diff-39785d72ffe23548cc619403f35d9028R2499
and
https://github.com/fcrepo4/fcrepo4/pull/1057/files/ba67f4a3027097bdaae93543c032f3d8a03a2aa0#diff-39785d72ffe23548cc619403f35d9028R2517

ajs6f: Do not imagine anything. Write a test, or recognize, as @awoods points out, that you already have.

lsitu: I think these two tests may not be sufficient. They just imitate the ingest of two large files that trigger the bug. Another test with several threads start at the same time with no assertion for the status code (status code is either 201 or 404) but check for existence of SN-sibling should be good.

lsitu:
For research purpose, I've updated the test cases to reflect the contest state with two threads to do the ingest at the same time. And I use the static variable LOGGER as the lock for synchronization purposely. All the tests are passed while it'll failed without synchronization, which is fairly interesting. But like @ajs6f said, we can use those two original test cases that imitate the bug and remove the synchronization keyword if we want, which will fix the bug reported in https://jira.duraspace.org/browse/FCREPO-2055. What do you think?

lsitu:
@awoods @ajs6f What's the next step would you like me to do, remove the synchronization keyword, change the save() method to static synchronized, or anything else?

ajs6f:
Remove the synchronization/locking and correct the tests to use a more understandable design that doesn't manually `start` and `join` threads. You can use my suggestions for that or a completely different design. If you really think that there exists some further issue to be resolved, please file an independent ticket detailing that new problem and we can discuss further in that context.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
awoods: I believe `TEST_FILE` is not a local variable because it creates a 50M file, which is used in two separate tests.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -763,6 +763,23 @@ private String mintNewPid(final String slug) {
         return pid;
     }
 
+    /**
+     * perform save with checking to prevent SN-Sibling nodes from creating
+     * @param resource the FedoraResource need to save
+     * @throws RepositoryException
+     */
+    private void save(final FedoraResource resource) throws RepositoryException {
+        synchronized (LOGGER) {
```
awoods: If I comment out this `synchronized`, sometimes the new tests pass and sometimes they fail. If `synchronized` stays, I have not seen any test failures.
Maybe more concurrent threads would demonstrate the failure case more definitively.

acoburn: I'm with @ajs6f -- the synchronization bit needs to go. Imagine what this would look like in a distributed context--it wouldn't even approach a workable solution. If you need some form of process coordination, you need something more than a static LOGGER variable.

acoburn:
I would add that, if this is fundamentally an issue with the HTML interface (i.e. clicking "Add" twice), why don't you just add a little javascript to "disable" the "Add" button upon clicking it. That is, does this issue _really_ appear when used with a regular HTTP client and not the browser?

acoburn:
One more point -- if there is _any_ locking or synchronization happening, PLEASE do not put that in the `fcrepo-http-*` modules. Doing so would potentially affect _all_ implementations of Fedora. If locking is absolutely necessary, isolate it in the modeshape layer.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
ajs6f: Is there any reason this needs to be a 50M file? Wouldn't a 5 byte file work just as well?

lsitu: I believe only big size of file can trigger the SN-sibling bug as reported in https://jira.duraspace.org/browse/FCREPO-2055, smaller file will trigger other errors that even FedoraLdp rest api fails to handle.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
lsitu: Yes, it look like something need to be done for concurrent support, which should be another story. The bug reported in [FCREPO-2055|https://jira.duraspace.org/browse/FCREPO-2055] is very specific case that relates to large file ingest only. I believe Modeshape doesn't handle well for nt:binary, which shouldn't allow SN-Sibling, so we are only trying to walk around it here ( see https://jira.duraspace.org/browse/FCREPO-2055?focusedCommentId=50506&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-50506 ).

lsitu:
I've updated the codes as suggested. Thanks everyone for your review, comments, and suggestions.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
```diff
@@ -451,23 +459,28 @@ protected static void assertDeleted(final String id) {
...
-        public int status() {
-            return status;
+        final ExecutorService executor = Executors.newFixedThreadPool(numberOfExecutions);
+        for (int i = 0 ; i < numberOfExecutions; i++) {
+            final Callable<Integer> task = () -> {
+                return getStatus(req);
+            };
+            final Future<Integer> future = executor.submit(task);
+            tasks.add(future);
+            Thread.sleep(100);
```
ajs6f: What is the purpose of this delay?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
```diff
@@ -451,23 +459,28 @@ protected static void assertDeleted(final String id) {
...
-
-        RequestThread(final HttpUriRequest request) {
-            this.request = request;
-        }
-
-        public int status() {
-            return status;
+        final ExecutorService executor = Executors.newFixedThreadPool(numberOfExecutions);
+        for (int i = 0 ; i < numberOfExecutions; i++) {
+            final Callable<Integer> task = () -> {
```
ajs6f: Just `final Callable<Integer> task = () -> getStatus(req);` is needed.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
ajs6f: If it is a question of triggering binary storage, you don't need anything _like_ a 50MB to do that. Refer to the [Modeshape configuration](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/test/resources/test_repository.json#L14) used for this test and you will see that the minimum size to trigger binary storage is only 40 bytes.

ajs6f:
The test code is definitely more readable, but still difficult to follow. I will send a direct PR with clarifications. That seems to be a more immediate course than going round and round with PR comments.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
awoods: I believe it is a question of having a file large enough such that a subsequent request comes in while the initial file upload is in-progress.

ajs6f: If that is the case than we need to actually write a mechanism to ensure that (using a streaming request) instead of gambling on a fixed size. A magic chosen fixed size will eventually fail, perhaps months or years from now, and it will be almost impossible to understand why.

ajs6f:
See https://github.com/lsitu/fcrepo4/pull/5.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
ajs6f: Why is this still here? It's not being used any more.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -196,6 +206,23 @@
         tripleFormat.setTimeZone(getTimeZone("GMT"));
     }
 
+    private static File TEST_FILE;
+
+    @BeforeClass
+    public static void beforeClass() {
```
ajs6f: Why is this still here? It's not being used any more.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
```diff
@@ -145,6 +146,31 @@ protected static HttpPut putDSMethod(final String pid, final String ds, final St
         return put;
     }
 
+    protected static HttpPut putDSFileMethod(final String pid, final String ds, final File content)
```
ajs6f: Why are any of these new methods in `AbstractResourceIT` still here? None of them are being used anymore.

ajs6f:
@lsitu , thanks for staying with this PR through a long process. My sense is that the changes are now minimal, clear, and efficacious. Given that this issue is not a blocker for 4.6, I'm inclined to let this PR sit until Monday in case anyone else has comments to add. If there are none by then, I will merge.

ajs6f:
@acoburn Do you think it's worth raising a ticket for the HTML/JS change you mention above?

awoods:
@acoburn's suggestion seemed like a good one: https://github.com/fcrepo4/fcrepo4/pull/1057#issuecomment-232834984

ajs6f:
@awoods See my comment above yours.

awoods:
@ajs6f, exactly. I was reinforcing your comment.

acoburn:
@awoods, @ajs6f: https://jira.duraspace.org/browse/FCREPO-2080

awoods:
Although this PR looks good, as a test, I reverted the following two lines from:
`save(resource);` to `session.save();`, i.e. basically reverting the functional code of this PR:
https://github.com/fcrepo4/fcrepo4/pull/1057/files#diff-4a14fb4dcbde6f7d32274c43ca723c16R394
and
https://github.com/fcrepo4/fcrepo4/pull/1057/files#diff-4a14fb4dcbde6f7d32274c43ca723c16R570

I then ran:
```
 for x in $(seq 0 9); 
    do echo $x; 
    mvn clean install -pl fcrepo-http-api/ | grep BUILD; 
done
```
And the build succeed all 10 times... no failures. Which would indicate the test is not testing the failure case, and potentially the functional update is making no difference.

lsitu:
Do you want to use the original tests to imitate the bug or not?

awoods:
That would be insightful... but I am signing off for the night. Please report back if you have time to do that.

awoods:
Another option here is to simply rely on @acoburn's upcoming resolution to:
https://jira.duraspace.org/browse/FCREPO-2080
That would resolve the HTML UI side of this issue... although not the REST API side.

lsitu:
I think the error is still there and it can be caught with my tests last night if I comment out the fixed codes:
FedoraLdpIT.testConcurrentPostDatastreamResource:2527->AbstractResourceIT.concurrentSNSDataStreamIngest:483 expected:<404> but was:<201>
  FedoraLdpIT.testConcurrentPutDatastreamResource:2517->AbstractResourceIT.concurrentSNSDataStreamIngest:483 expected:<404> but was:<201>

lsitu:
@ajs6f To replicated the bug reported in https://jira.duraspace.org/browse/FCREPO-2055, I think we can use curl as reported in the spec to replicate it, as long as the ingest file is big enough to ensoure both ingests can be overlapping for a while. If two ingests happen exactly at the same time, you may get an 500 error. Do you have any thoughts to ensure this without catching the 500 error? Thanks.

ajs6f:
Under no circumstances will we use the original code, for the reasons that @acoburn and I have already given several times. Using a shell command to operate inside a JUnit test is actually worse than the test code with which this PR began. We won't be doing that, out of pity for the maintainers who will follow us.

If the error is still present, as @lsitu reports above, then this PR is useless and should be closed. I will close it on Monday. We can wait for @acoburn 's ticket, but the fault in the REST API must still be corrected.

awoods:
@lsitu, I think you were correct in your original comment on this ticket, suggesting this is actually a ModeShape bug in that fedora:Container and fedora:Binary should not be allowed to have Same-Name-Siblings.
https://jira.duraspace.org/browse/FCREPO-2055?focusedCommentId=50506&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-50506

If you are interested in exploring the issue at the ModeShape level, fantastic. If not, I completely understand.
@ajs6f, feel free to close this ticket at your leisure.

lsitu:
@ajs6f I meant if I don't use the PR, I am still able to catch the bug. And you can always replicate the bug manually but not suggest to operate the curl command inside JUnit test. 
Actually the PR will fix the bug as a walk around solution, while the test in your last commit failed to test the bug (see @awoods' comment https://github.com/fcrepo4/fcrepo4/pull/1057#issuecomment-233104427).

ajs6f:
If this is a MODE bug (and I share @awoods 's opinion on that) then it would be nice to hand Matthew Hanlon a ModeShape ticket with which to track it if we close ours. Have we gotten that filed?

lsitu:
@ajs6f No. We haven't file it yet. And @awoods, I would like to look into it on the ModeShape level as well. I think at least we may need to find a way to prove the bug to Modeshape if we don't want to fix it and provide a patch to MODE with a test plan attached (What's the LOE?). As we've seen from the Modeshape 5 bug that we ran into previously, it could be months process to wait for a Modeshape release that include the bug fixing, even reporting the bug with a patch that fixed it.

With that said, if it just a test plan issue, maybe we can think about the option to update the test plan a little bit so that we can prove the fixing of the bug? We can rethink about the solution once the bug is fixed on the MODE side later. Just my two cents.

awoods:
@lsitu , @ajs6f , Although there is not yet a ModeShape ticket for this, I would be happy to file it... once we have a unit or integration test that demonstrates the issue. If we could provide a patch as well, that would be great; but including a test with the ticket is a minimum requirement. 
As a note, the test and ticket should be targeted against ModeShape5... assuming the problem exists in that release.

ajs6f:
@awoods You raise a good point that hasn't yet been discussed. If this gets fixed in MODE5, my choice would be do nothing other than what @acoburn has already done for <4.7.0. It's just not such a dangerous issue. Is it possible to use the info in the original ticket to check this?

awoods:
@ajs6f, the original ticket found this issue through the HTML UI. With @acoburn's fix, that is no longer possible. The only other way I have seen this issue surfaced is through @lsitu's original integration tests.
Those tests could be run against the [modeshape5](https://github.com/fcrepo4/fcrepo4/tree/modeshape5) branch.

lsitu:
@awoods, @ajs6f: The Same Name Sibling bug still exists in Modeshape 5. And I've created a PR to reproduce the Same Name Sibling bug against the modeshape5 branch with commit (https://github.com/fcrepo4/fcrepo4/pull/1068/commits/962bf4496944f64ff231ee6111d831b675343235) and added another commit (https://github.com/fcrepo4/fcrepo4/pull/1068/commits/c446cf2fc92d65bf8388aedbd0a97971479f1384) with the fixing codes that we agree to. So I think we can decide to move to the next step now.

lsitu:
Here is the PR I created against the modeshape5 branch: https://github.com/fcrepo4/fcrepo4/pull/1068

ajs6f:
See my comments at that PR. It is not acceptable without further work. Please file the appropriate ticket against MODE-- _that_ is the right step forward.

awoods:
For the ticket we submit to the Modeshape jira, we will need a test against the Modeshape code base. Have you looked into such a test?

ajs6f:
I disagree that we need a test, just a careful description. If you want to include a test, @lsitu 's earlier work should be refactored to run directly against MODE.

lsitu:
Okay, I am closing the PR and moving forward to file a MODE bug now.

ajs6f:
👍 Thanks, @lsitu !

PR #1059, opened 2016-06-27
FCREPO-2065 DelegateHeaderPrincipalProvider leaking on-behalf-of header when called with non-admin

mesemus:
ServletContainerAuthenticationProvider already handles DelegateHeaderPrincipalProvider differently
from other providers in authenticate() and getDelegatedPrincipal(). The same pattern was added into
collectPrincipals() to prevent instances of this class from filling FEDORA_ALL_PRINCIPALS.

Resolves: https://jira.duraspace.org/browse/FCREPO-2065

### DIFF
--- a/fcrepo-auth-common/src/test/java/org/fcrepo/auth/common/ServletContainerAuthenticationProviderTest.java
+++ b/fcrepo-auth-common/src/test/java/org/fcrepo/auth/common/ServletContainerAuthenticationProviderTest.java
@@ -20,6 +20,7 @@ import static org.fcrepo.auth.common.ServletContainerAuthenticationProvider.FEDO
 import static org.fcrepo.auth.common.ServletContainerAuthenticationProvider.FEDORA_USER_ROLE;
 import static org.fcrepo.auth.common.ServletContainerAuthenticationProvider.getInstance;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
@@ -176,6 +177,10 @@ public class ServletContainerAuthenticationProviderTest {
         when(delegateProvider.getDelegate(creds)).thenReturn(delegatePrincipal);
         when(delegatePrincipal.getName()).thenReturn("delegatedUserName");
 
+        // delegateProvider being HeaderProvider returns the header content in getPrincipals regardless of logged user
+        when(delegateProvider.getPrincipals(creds)).thenReturn(
+                new HashSet<>(Collections.singletonList(delegatePrincipal)));
+
         final ExecutionContext result =
                 provider.authenticate(creds, "repo", "workspace", context,
                         sessionAttributes);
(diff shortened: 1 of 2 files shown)


barmintor:
@mesemus thanks for this PR. I want us to look at whether this is a broader issue before we merge it: it ought to block the next release, but it seems like an encapsulation or responsibility problem in the providers.

ajs6f:
@barmintor Can you expand a bit on that? Is it a question of different providers doing different things but that not being reflected in some reasonably explicit way in the type hierarchy?

mesemus:
@barmintor, @ajs6f 
As I'm new to this, please let me know if the correct place for discussion is here on the pull request or in jira.

I've added a possible proposal of a change to the type hierarchy into the jira issue, which would allow  ServletContainerAuthenticationProvider deal only with interfaces, and an abstract class for header parsing AuthProviders.

ajs6f:
@barmintor et al., unless there are any live objections, I will be merging this on Monday and closing the associated ticket.

fcrepo-auth-common/src/test/java/org/fcrepo/auth/common/ServletContainerAuthenticationProviderTest.java
```diff
@@ -176,13 +177,30 @@ public void testNoDelegatedAuthenticationForUsers() {
         when(delegateProvider.getDelegate(creds)).thenReturn(delegatePrincipal);
         when(delegatePrincipal.getName()).thenReturn("delegatedUserName");
 
+        // delegateProvider being HeaderProvider returns the header content in getPrincipals regardless of logged user
+        when(delegateProvider.getPrincipals(creds)).thenReturn(
+                new HashSet<>(Collections.singletonList(delegatePrincipal)));
```
ajs6f: Before I merge this, we could clean it a bit. Instead of this complicated expression, how about using Guava's `Sets.newHashSet(E... elements)`?

fcrepo-auth-common/src/test/java/org/fcrepo/auth/common/ServletContainerAuthenticationProviderTest.java
```diff
@@ -176,13 +177,30 @@ public void testNoDelegatedAuthenticationForUsers() {
...
+                sessionAttributes.containsKey(FedoraAuthorizationDelegate.FEDORA_ALL_PRINCIPALS));
+
+        @SuppressWarnings("unchecked")
+        final Set<Principal> principals =
+                (Set<Principal>) sessionAttributes.get(FedoraAuthorizationDelegate.FEDORA_ALL_PRINCIPALS);
+
+        assertFalse(
+                "The sessionAttributes must not contain delegatedUserName " +
+                        "(delegated user must not leak to FEDORA_ALL_PRINCIPALS)",
+                principals.stream().anyMatch(x->"delegatedUserName".equals(x.getName())));
```
ajs6f: It would read a little better as
```
principals.stream().map(Principal::getName).anyMatch("delegatedUserName"::equals));
```

awoods:
Thanks for surfacing this issue @mesemus and for providing the patch. In order to both get this fix in the 4.6.0 release, and to not delay that release, I would be inclined to go forward with this PR (pending resolution of the code review comments from @ajs6f).
However, @mesemus, I would also be interested in seeing a follow-on patch that implements your [interfaces approach](https://jira.duraspace.org/secure/attachment/18030/18030_fedora+auth.pdf)... that seems to potentially be cleaner.
Thanks again!

ajs6f:
To be clear, this _is_ a blocker for 4.6.0, so rest assured that we will get a fix for this before releasing. This PR seems like the most expeditious way forward right now.

ajs6f:
@mesemus Will you be responding to the code review comments today? I would like to get this merged.

mesemus:
@ajs6f Sorry, different time zone, will do it just now.

mesemus:
@ajs6f, @awoods: changes committed, thanks for code review.

ajs6f:
@mesemus Looks great, merging!

PR #1060, opened 2016-06-29
FCREPO-2064 standardize exception logging in JAXRS ExceptionMapper impls

barmintor:


fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/ExceptionDebugLogging.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public interface ExceptionDebugLogging {
+    /**
+     * Log a Throwable at the DEBUG level, log the stacktrace at the TRACE level.
+     *
+     * @param context ExceptionDebugLogging the exception intercepting context.
+     * @param error Throwable the intercepted error.
+     * @param logger Logger the logger to use
+     */
+    public default void debugException(ExceptionDebugLogging context, Throwable error, Logger logger) {
```
ajs6f: This may be more a matter of style, but my preference would be to factor out `logger` into an abstract method `Logger logger()` and let it be implemented at greatest common factors in the mapper type-hierarchy.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/ExceptionDebugLogging.java
```diff
@@ -0,0 +1,43 @@
...
+        if (!logger.isDebugEnabled()) {
```
ajs6f: If you are only making one call of `logger.debug()` and the expression used therein is simple (and I would say that an accessor is generally simple) then there is no real value in `isDebugEnabled()` compared to the extra cost. `debug()` calls it anyway, so you're going to pay for it at least once. I know you know this, @barmintor, but for anyone who doesn't, the value to using `isXEnabled()` is to avoid calculating a complex expression within a `Logger::X` call, which, because of Java semantics, happens before the logging framework executes `isXEnabled()` internally to `X()` and therefore before the logging framework can short-circuit.

barmintor: I'm thinking that isDebugEnabled is cheaper than .getClass.getSimpleName, but your point is incontestable about the trace.

barmintor: (especially since in the most frequent case neither will do anything, so all the method overhead is wasted)

barmintor: also, thank you for having a look!

ajs6f: It may very well be. It's all an exception path anyway. I really wanted to mention it as a general point, because I am a pretentious pettifogging pedant.

barmintor: Would you like me to comment the 2 accessor vs 1 and early-exit
justification there?

ajs6f: I _NEVER_ say no to extra commenting.

barmintor: Done.

ajs6f:
@barmintor Are you done reacting to my comments? If so, I'll wait through this morning and if there are no other comments, I'll merge.

awoods:
@barmintor, as a note, I do not know a PR is ready for review until the associated [JIRA item](https://jira.duraspace.org/browse/FCREPO-2064) has been moved into the "Start Review" state.

barmintor:
Moved to "Start Review"

PR #1061, opened 2016-07-04
Remove fcrepo-kernel-api dependency on javax.servlet.http package

acoburn:
Addresses: https://jira.duraspace.org/browse/FCREPO-1664

`javax.jcr.Credentials` is just a marker class, and as such it there is no need for it to make `fcrepo-kernel-api` depend on `javax.servlet.http`. (Modeshape _does_ require a dependency on `javax.servlet.http`, but this PR isolates that dependency in the `fcrepo-kernel-modeshape` module).

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/CredentialsService.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/CredentialsService.java
@@ -16,7 +16,6 @@
 package org.fcrepo.kernel.api.services;
 
 import javax.jcr.Credentials;
-import javax.servlet.http.HttpServletRequest;
 
 /**
  * @author acoburn
@@ -25,11 +24,11 @@ import javax.servlet.http.HttpServletRequest;
 public interface CredentialsService {
 
     /**
-     * Get the credentials for the given servlet request
+     * Get the credentials for the given request
      *
-     * @param request the servlet request
+     * @param request the request
      * @return the JCR credentials for the given request
      */
-    Credentials getCredentials(final HttpServletRequest request);
+    Credentials getCredentials(final Object request);
 
 }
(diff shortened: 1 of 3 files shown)


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/CredentialsService.java
```diff
@@ -25,11 +24,11 @@
...
     /**
-     * Get the credentials for the given servlet request
+     * Get the credentials for the given request
      *
-     * @param request the servlet request
+     * @param request the request
      * @return the JCR credentials for the given request
      */
-    Credentials getCredentials(final HttpServletRequest request);
+    Credentials getCredentials(final Object request);
```
ajs6f: So our idea seems to be that anything could be used to generate JCR credentials, but secretly we know that we're going to factor out JCR credentials anyway, so who cares? Is that right?

acoburn: Basically, yes.

This `getCredentials` service was added in order to remove one more Modeshape dependency from the HTTP layer: https://github.com/fcrepo4/fcrepo4/commit/2d3506051df643e18da53a684fd1b81e77d2a7d8#diff-be95e9939194d194be91ae919de03353L148

As you know, I am keenly interested in removing JCR dependencies from the kernel-api and not adding to them, but a related step in that same direction involves removing MODE dependencies from the HTTP layers. In fact, that seems to be a necessary precondition (or at least a logical precondition).

I suspect that when we eventually replace the use of `jcr.Session` with a Fedora-specific one, this `jcr.Credentials` service will go away, as it should be pretty easy at that point to factor this method into a generic `FedoraSession` interface.

ajs6f: I have no objection-- just wanted to make sure I understood the move correctly. This is better than what we have and blocks no future moves to get even better, so 👍 .

PR #1063, opened 2016-07-07
Throw an exception if a non-SHA1 digest is passed

whikloj:
Addresses [FCREPO-2075](https://jira.duraspace.org/browse/FCREPO-2075)

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -791,8 +791,7 @@ public class FedoraLdp extends ContentExposingResource {
     private static String parseDigestHeader(final String digest) throws InvalidChecksumException {
         try {
             final Map<String,String> digestPairs = RFC3230_SPLITTER.split(nullToEmpty(digest));
-            final boolean checksumTypeIncludeSHA1 = digestPairs.keySet().stream().map(String::toLowerCase)
-                .anyMatch("sha1"::equals);
+            final boolean checksumTypeIncludeSHA1 = digestPairs.keySet().stream().anyMatch("sha1"::equalsIgnoreCase);
             // If you have one or more digests and one is sha1 or no digests.
             if (digestPairs.isEmpty() || checksumTypeIncludeSHA1) {
                 return digestPairs.entrySet().stream()
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -59,6 +59,8 @@ import java.net.URLDecoder;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.inject.Inject;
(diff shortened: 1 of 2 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -786,16 +788,24 @@ private static void checkLinkForLdpResourceCreation(final String link) {
...
         try {
             final Map<String,String> digestPairs = RFC3230_SPLITTER.split(nullToEmpty(digest));
-            return digestPairs.entrySet().stream()
-                .filter(s -> s.getKey().toLowerCase().equals("sha1"))
-                .map(Map.Entry::getValue)
-                .findFirst()
-                .map("urn:sha1:"::concat)
-                .orElse("");
+            final Set<String> checksumTypes =
+                    digestPairs.keySet().stream().map(s -> s.toLowerCase()).collect(Collectors.toSet());
```
ajs6f: `s -> s.toLowerCase()` is `String::toLowerCase`.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -786,16 +788,24 @@ private static void checkLinkForLdpResourceCreation(final String link) {
...
+            // If you have one or more digests and one is sha1 or no digests.
+            if (digestPairs.isEmpty() || checksumTypes.contains("sha1")) {
```
ajs6f: If you don't need to do anything else with `checksumTypes`, then you could do:
```
boolean checksumTypeIncludeSHA = digestPairs.keySet().stream().map(String::toLowerCase).anyMatch("sha1");
```

whikloj: Thank you, I knew there was something cleaner. Just couldn't pull it together.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -793,10 +791,10 @@ private static void checkLinkForLdpResourceCreation(final String link) {
     private static String parseDigestHeader(final String digest) throws InvalidChecksumException {
         try {
             final Map<String,String> digestPairs = RFC3230_SPLITTER.split(nullToEmpty(digest));
-            final Set<String> checksumTypes =
-                    digestPairs.keySet().stream().map(s -> s.toLowerCase()).collect(Collectors.toSet());
+            final boolean checksumTypeIncludeSHA1 = digestPairs.keySet().stream().map(String::toLowerCase)
+                .anyMatch("sha1"::equals);
```
ajs6f: Ah, I'm a dope-- I totally forgot `String::equalsIgnoreCase`, so just:
```
digestPairs.keySet().stream().anyMatch("sha1"::equalsIgnoreCase)
```

whikloj: Oh yeah, I kept looking for case insensitive matching in the `Set<String>` when I started this. Now that we are back to just using `String` I never thought about those useful methods. 👍

awoods:
:+1: Looks good to me. @ajs6f, feel free to squash/merge when you are content.

PR #1066, opened 2016-07-07
Remove fedora:numberOfChildren predicate

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-1880

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -66,7 +66,6 @@ import static org.fcrepo.kernel.api.RdfLexicon.CREATED_DATE;
 import static org.fcrepo.kernel.api.RdfLexicon.DC_TITLE;
 import static org.fcrepo.kernel.api.RdfLexicon.DIRECT_CONTAINER;
 import static org.fcrepo.kernel.api.RdfLexicon.HAS_CHILD;
-import static org.fcrepo.kernel.api.RdfLexicon.HAS_CHILD_COUNT;
 import static org.fcrepo.kernel.api.RdfLexicon.HAS_MEMBER_RELATION;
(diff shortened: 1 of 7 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -1251,20 +1250,9 @@ public void testGetObjectGraphWithChildren() throws IOException {
...
-
-                // Total child count should be provided
-                final Node child = children.next().getObject();
-                assertTrue("Object should be a literal!", child.isLiteral());
-                assertEquals(CHILDREN_TOTAL, Integer.parseInt(child.getLiteralValue().toString()));
-
                 final Iterator<Quad> contains = graph.find(ANY, createURI(location), CONTAINS.asNode(), ANY);
                 assertTrue("Should find contained child!", contains.hasNext());
-                assertEquals(CHILDREN_LIMIT - 1, Iterators.size(contains));
+                assertEquals(CHILDREN_LIMIT, Iterators.size(contains));
```
whikloj: I'm am good with merging this. I have pulled it and tested it, no issues building.

I am just wondering how come this test is changing with this PR? I must be missing something.

awoods: The "Limit" header determines the number of triples that are returned from the `ChildrenRdfContext` RDF generation... not specifically restricted to the `ldp:contains` triples.
Previously, the number of `ldp:contains` triples that were returned was `limit - 1` because there used to be an additional triple that was included in the response: `fedora:numberOfChildren`.

Now that the `fedora:numberOfChildren` has been removed, the only triples that `ChildrenRdfContext` generates are `ldp:contains` triples; therefore, the "Limit" header exactly corresponds to the number of `ldp:contains` triples returned to the caller.

whikloj: Ahh I see. Thanks for the clarification.

PR #1068, opened 2016-07-17
Reproduce the Same Name Sibling bug with walk around fixing solution.

lsitu:
Related to: https://jira.duraspace.org/browse/FCREPO-2055


ajs6f:
Does the bug actually appear in this branch?

ajs6f:
👎 to this PR-- we should be following this up with MODE instead, especially after @acoburn 's fix for FCREPO-2080.

lsitu:
Okay. @awoods Is it sufficient to report the SN-Sibling bug with test case above?

lsitu:
I would close the PR now since we all aware that the Same Name Sibling issue is still existing in Modeshape 5.

PR #1070, opened 2016-07-20
Remove jQuery dependencies on the javascript code in fcrepo-http-api

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-2081

PR #1071, opened 2016-07-21
Add a servlet filter for handling ETags for static resources

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-2083

### DIFF
--- a/fcrepo-webapp/src/main/webapp/WEB-INF/web.xml
+++ b/fcrepo-webapp/src/main/webapp/WEB-INF/web.xml
@@ -34,6 +34,16 @@
 
 	</servlet-mapping>
 
+  <filter>
+    <filter-name>ETagFilter</filter-name>
+    <filter-class>org.springframework.web.filter.ShallowEtagHeaderFilter</filter-class>
+  </filter>
+
+  <filter-mapping>
+    <filter-name>ETagFilter</filter-name>
+    <url-pattern>/static/*</url-pattern>
+  </filter-mapping>
+
   <!--Uncomment section below to enable Basic-Authentication-->
   <!--
   <security-constraint>
--- a/fcrepo-webapp/pom.xml
+++ b/fcrepo-webapp/pom.xml
@@ -149,7 +149,6 @@
       <artifactId>infinispan-cachestore-jdbc</artifactId>
     </dependency>
 
-    <!-- test gear -->
     <dependency>
       <groupId>org.springframework</groupId>
       <artifactId>spring-web</artifactId>
@@ -158,6 +157,8 @@
       <groupId>org.infinispan</groupId>
       <artifactId>infinispan-cachestore-leveldb</artifactId>
     </dependency>
+
+    <!-- test gear -->
     <dependency>
       <groupId>net.sourceforge.htmlunit</groupId>
       <artifactId>htmlunit</artifactId>


fcrepo-webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -34,6 +34,16 @@
...
 	</servlet-mapping>
 
+  <filter>
+    <filter-name>ETagFilter</filter-name>
+    <filter-class>org.springframework.web.filter.ShallowEtagHeaderFilter</filter-class>
+  </filter>
+
+  <filter-mapping>
+    <filter-name>ETagFilter</filter-name>
+    <url-pattern>/static/*</url-pattern>
```
ajs6f: Do we have this URL yet, or are you setting this up for the future?

acoburn: This already exists and is used, e.g. by the HTML UI and also for delivering static RDF related to constraint violations. This change simply adds ETags to those responses.

ajs6f: Oh, okay. I thought we had some kind of `fcr:assets` thing going on.

acoburn: Yes, `fcr:assets` also exists (and also doesn't respect ETags), but that's a completely separate issue. It's separate because `fcr:assets` is handled by `fcrepo-http-api` and not by `fcrepo-webapp` (which is what's going on here).

I would very much like to sort out the `fcr:assets` issue, but that's something of a larger deal (and relates to the LDTemplate refactoring discussed earlier).

ajs6f: Okay, this seems sensible enough.

fcrepo-webapp/src/main/webapp/WEB-INF/web.xml
```diff
@@ -34,6 +34,16 @@
 
 	</servlet-mapping>
 
+  <filter>
+    <filter-name>ETagFilter</filter-name>
+    <filter-class>org.springframework.web.filter.ShallowEtagHeaderFilter</filter-class>
```
awoods: As noted in the java docs, the `ShallowEtagHeaderFilter` saves bandwidth, but the server still needs to calculate the checksum. 
Since the static content is... static, we could save performance as well by overwriting the method [ShallowEtagHeaderFilter.generateEtagHeaderValue](http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/filter/ShallowEtagHeaderFilter.html#generateETagHeaderValue-java.io.InputStream-boolean-).

ajs6f: Well, I done merged this already, so you want to open another ticket?

acoburn: @awoods I saw that too, but as this really only affects the HTML UI, and as these resources are small, I didn't think it was worth the effort.

ajs6f: You would want to just send the filenames or something? Something that doesn't ever change and that we wouldn't need to keep up to date?

awoods: @acoburn, which is fine, but then why is it worth the effort of supporting ETags at all if "this really only affects the HTML UI, and as these resources are small"?

acoburn: @awoods because doing so is good practice. It is for much the same reason I refactored the JS in #1070 

And by "small" I meant that it would be presumably faster for the server to compute the checksum by reading the file locally than it would be to send all the bytes over the network.

ajs6f: So this was an improvement, and by making the ETag values static we could improve even further. How about a fresh ticket? I'm trying to get to code freeze, today.

PR #1073, opened 2016-07-21
Clean up intro text in index.html

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-2085

### DIFF
--- a/fcrepo-webapp/src/main/webapp/index.html
+++ b/fcrepo-webapp/src/main/webapp/index.html
@@ -15,34 +15,35 @@
             padding-top: 70px;
             padding-bottom: 70px;
         }
-
+        article p {
+          padding: 10px 0;
+        }
     </style>
 </head>
 <body>
-<div class="container">
-<div class="col-md-12">
-<div class="jumbotron">
-<a href="rest/" class="thumbnail" style=""><img src="static/images/fedora_logo_10in.png" /></a>
-<p>This is the exciting stub page for your Fedora ${project.version} repository that we have placed
-here.</p>
-
-<p>You probably want to visit something a little more interesting, such as:       </p>
-<a class="btn btn-primary btn-lg" href="rest/">the Fedora REST API endpoint</a>
-<p><br/></p>
-<p>Fedora is maintained by a community of volunteers. <a href="http://duraspace.org/register-repository" target="_duraspace">Register your repository</a> at <a href="http://duraspace.org/" target="_duraspace">DuraSpace</a> in order to facilitate information sharing and collaboration within the community.</p>
-</div>
-</div>
-</div>
-
-<nav class="navbar navbar-default navbar-fixed-bottom" role="navigation">
-    <div class="container">
-        <ul class="nav navbar-nav">
-            <li><a href="https://jira.duraspace.org/browse/FCREPO">Issue Tracker</a></li>
-            <li><a href="https://wiki.duraspace.org/display/FF">Wiki</a></li>
-            </ul>
-
-    <p class="navbar-text pull-right">Release: <span id="version">${project.version}</span> | Build <span id="build">#${buildNumber}</span> (<span id="timestamp">${build.timestamp}</span>)</p>
+<article class="container">
+  <div class="col-md-12">
+    <div class="jumbotron">
+      <a href="rest/" class="thumbnail" style=""><img src="static/images/fedora_logo_10in.png" /></a>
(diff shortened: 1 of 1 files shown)


fcrepo-webapp/src/main/webapp/index.html
```diff
@@ -15,30 +15,32 @@
...
+    </div>
+  </div>
+</article>
 
 <nav class="navbar navbar-default navbar-fixed-bottom" role="navigation">
     <div class="container">
         <ul class="nav navbar-nav">
             <li><a href="https://jira.duraspace.org/browse/FCREPO">Issue Tracker</a></li>
             <li><a href="https://wiki.duraspace.org/display/FF">Wiki</a></li>
+            <li><a href="https://wiki.duraspace.org/display/FEDORA4x/RESTful+HTTP+API">REST API Documentation</a></li>
```
escowles: Glad to see a link to the REST API docs here — but we should make sure to change the link to point to the version-specific docs at release time.

acoburn: I thought that ideally, we'd link to the REST docs associated with the particular version of Fedora, but it seemed like that would involve some Maven magic that didn't immediately occur to me. Or, maybe just a simple property pointing to the version of the docs?

escowles: Having master link to the FEDORA4x docs seems like the right thing to do, and then we could manually update the release branch to link to FEDORA45 or whatever the versioned URL that matches that release is.  In theory this could be automated somewhat, but for now maybe we can just add it to the release docs?

awoods: I suspect requiring the update of this link at release time is not going to be sustainable, i.e. there will be errors, times when it does not happen.
Maybe linking to this page would be more reasonable: https://wiki.duraspace.org/display/FF/Documentation

ajs6f: Please let me know whatever new work you are making for me, once you know what it is.

The obvious (although not obviously correct) way to move the value would be via [Maven Resource action](https://maven.apache.org/plugins/maven-resources-plugin/examples/filter.html).

acoburn:
Rebased and using @awoods' suggestion to link to https://wiki.duraspace.org/display/FF/Documentation

ajs6f:
Do we need a link to fedora.info or anything like that? Or is that in there already and I missed it?

acoburn:
@ajs6f, that's a good idea. I will add a link.

acoburn:
@ajs6f just a reminder about merging this prior to cutting a release candidate

ajs6f:
Too late. I thought it had been merged. I will try to add this in some way post-tag. Not sure what the least disruptive thing to do here is.

awoods:
@ajs6f, You can commit it against the release candidate branch.

ajs6f:
Okay, I'll do that a little later this morning. This is non-functional stuff, after all.

ajs6f:
@acoburn Want to squash first?

acoburn:
@ajs6f squashed

ajs6f:
thnx

awoods:
@acoburn, you probably need to re-target the PR to the RC branch.

acoburn:
No, that is precisely _not_ what I want to do. I will, however, make a second PR that targets the RC branch though.

ajs6f:
I just pushed this into RC-- you want to push it into DEV? I wouldn't mind you doing that just to check that DEV works correctly.

acoburn:
@ajs6f we need to rethink the DEV branch strategy. I do not like it.

ajs6f:
@acoburn To be clear, I manually pulled this branch into RC, via git pull on my local.

ajs6f:
Okay, I've got nothing in favor of DEV. I'm just trying to run through the process we now have to get to the end of it so I can go away and work on the specs, which is all I really care about anyway.

acoburn:
@ajs6f that's cool; I'm talking specifically about the strategy of freezing `master` for 3+ weeks and working on a `dev` branch during that time. It is needlessly complex and confusing. Instead, we should immediately create a `4.6.x-maintenance` branch. `master` would continue to be for the tip of development (and _not_ frozen). I'm trying to avoid this "target `DEV` nonsense".

acoburn:
@awoods this commit has already been merged with the `-RC` branch but not with `master`. Would you be able to review it for merging with `master`?

PR #1075, opened 2016-07-25
Migrate to Jena 3.1.0

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1672

This PR is divided into three commits as follows (mostly to make review easier):

1. Change the package names to `org.apache.jena` and increment version in pom.xml
2. Update implementation and test code to support the new version
3. Rework the ldp testsuite (which depends on Jena 2.x) so that we can avoid dependency convergences.


ajs6f:
Okay, obviously the vast majority of line changes are going to be `com.hp.hpl` => `org.apache`. Do you have any suggestions for substantive review? Are there places you remember making particular choices that should be reviewed? I don't think Jena's API actually changed all that much across the period in question, so nothing comes to my mind.

acoburn:
@ajs6f this is why I have three commits here. I'd look particularly at the second one https://github.com/fcrepo4/fcrepo4/pull/1075/commits/f99a317bc9263785c9eeb72c9318435f11bf44c3 -- that's where all of the impl code changes are.

acoburn:
@ajs6f also, in the first commit, there is this change: https://github.com/fcrepo4/fcrepo4/pull/1075/commits/890c31f86450e0b9965d0b58ac4f51f739661899#diff-f8c7031221097a6120fd8c87ded12841L20

GraphStoreWrapper -> GraphStoreBasic (GraphStoreWrapper doesn't seem to exist any longer)

Otherwise, the first commit is just about renaming the packages.

ajs6f:
`GraphStore` is now deprecated. I'm inclined to take this opportunity to replace it with either `Dataset` or `DatasetGraph` (preferably the former).

ajs6f:
So after the third commit, the LDP suite runs against the actual `fcrepo-webapp`?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraJcrConstants.java
```diff
@@ -46,6 +46,8 @@
 
     public static final String VERSIONABLE = "mix:versionable";
 
+    public static final String FIELD_DELIMITER = "\30^^\30";
```
ajs6f: Is this @cbeer 's unspeakable insignia?

acoburn:
@ajs6f yes, the LDP suite runs against `fcrepo-webapp`

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraJcrConstants.java
acoburn: Yes.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraBinaryImpl.java
```diff
@@ -246,7 +248,7 @@ public URI getContentDigest() {
     public String getMimeType() {
         try {
             if (hasProperty(HAS_MIME_TYPE)) {
-                return getProperty(HAS_MIME_TYPE).getString();
+                return getProperty(HAS_MIME_TYPE).getString().replace(FIELD_DELIMITER + XSDstring.getURI(), "");
```
awoods: Instead of hardcoding the clean-up of the `FIELD_DELIMITER` suffixes throughout the codebase, it seems cleaner to simply create a small utility in https://github.com/fcrepo4-exts/fcrepo4-upgrade-utils to be executed as part of the 4.6.0 to 4.7.0 upgrade process.

acoburn: So you are suggesting removing that construct altogether? What would you propose to replace that with?

To me, that seems like a separate issue.

acoburn: @awoods to be clear, with RDF 1.1, _all_ Literals have a datatype, so the string literals that previously had no datatype now will have a datatype, meaning that when they are persisted to Modeshape, they will always have the `FIELD_DELIMITER` + `xsd:string` suffix. Any pre Jena 3.x data will not have that suffix. For RDF output, that's not a problem, but for certain  methods (such as `getMimeType`) you clearly don't want that value (in fact, Jersey will blow up if the mimetype contains that suffix). An alternative would be to use the `ValueConverter` class here, which is probably a good idea. But again, I think that refactoring goes beyond the simple migration to Jena 3.x of this PR.

awoods: I appreciate the explanation, @acoburn. It first pass, it looked like you were just cleaning out the `FIELD_DELIMITER` cruft from the past.
I would like to give this PR a closer look, but have limited time today. If someone in addition to @ajs6f gives it the thumbs-up, feel free to move it forward. Otherwise, I will aim to give it a closer look tonight.

acoburn: @awoods I would like to address @ajs6f's comment about the use of `GraphStore` before this is merged. That will likely happen today.

ajs6f: Thank you for doing that, @acoburn . That `GraphStore` thing has been bugging me forever.

acoburn:
@ajs6f: I added another commit (e9292b0) replacing `GraphStore` with `Dataset`/`DatasetGraph`. I also replaced the (deprecated) use of `createAnon` with `createBlankNode`. I reordered the commits so that the LDP integration suite changes are still the final commit.

ajs6f:
Looks good to me.

acoburn:
@awoods ok, this PR is in your court now. I suspect that it will require additional PRs for the various downstream applications

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/integration/kernel/modeshape/FedoraResourceImplIT.java
```diff
@@ -752,6 +755,8 @@ public void testReplacePropertiesHashURIs() throws RepositoryException {
         final Model updatedModel3 = object.getTriples(subjects, PROPERTIES).collect(toModel());
 
         updatedModel3.remove(subject, dcCreator, hashResource);
+        updatedModel3.removeAll(hashResource, (Property)null, (RDFNode)null);
```
awoods: This line seemed peculiar, so I tested with it commented out.. with all tests passing. 
It appears that the update in this line can be removed/reverted.

awoods: I am happy to remove it as a part of this PR, unless there is some reason for this update that I am missing.

acoburn: Yes, go ahead and remove it

awoods:
Resolved with:
https://github.com/fcrepo4/fcrepo4/commit/46497f50f214d2ed8ebfa414f02f75138ea5a105
https://github.com/fcrepo4/fcrepo4/commit/a9dd8478bda6ccf54832d347d5e25e6568a23004
https://github.com/fcrepo4/fcrepo4/commit/336424ab6a010df725301a77501e288d8e3c1f5d
https://github.com/fcrepo4/fcrepo4/commit/4bfdd2594f6f31e12413c151323f9a89b2af6e63

PR #1076, opened 2016-07-25
Allow deletion of non-string literals via sparql-update

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-2090

Note: this should be merged _after_ https://jira.duraspace.org/browse/FCREPO-2091 but before #1075 (which can be simplified after this change)


acoburn:
@ajs6f: I added tests for this change. Incidentally, the test method I was building from didn't work in the first place: `createURI("<info:test#label>")` should have been `createURI("info:test#label")`

ajs6f:
I'm still really surprised no one noticed this.

PR #1080, opened 2016-07-25
Handle ldp:isMemberOfRelation on IndirectContainers

awoods:
This is a partial resolution related to: https://jira.duraspace.org/browse/FCREPO-2092

This commit:
- correctly handles _ref properties on proxy resources
- adds an integration test ensuring dynamically added property on indirect resource
  *\* the above integration test failing in this commit


awoods:
For Fedora 6 and beyond, I believe we should retain the integration test included in this PR.

bbpennel:
I believe this case is covered by a few tests in LdpContainerIT, so I am going to go ahead and close it

PR #1081, opened 2016-07-26
FCREPO-2093: Remove special case for RDF binary resources

birkland:
This pull request removes the special case from ContentExposingResource that parsed RDF from binary resources and combine with object properties.  It includes a test to verify that RDF persisted as binary is retrieved opaquely, without modification to its contents.


awoods:
Treating Binary resources as opaque bitstreams seems the cleanest approach. I am :+1: on removing the logic to parse and augment RDF Binaries, as implemented in this PR.

escowles:
I'm also 👍  to passing binaries as-is without modifying them if they are RDF.

I'm a little curious why this was even happening in the first place — we identify RDF in POST/PUT requests and create containers with the supplied RDF as the properties.  Maybe you could send RDF as text/plain, and then update the mime type afterwards?

acoburn:
a big 👍  from me, too

birkland:
@escowles I'm not sure why it's trying to parse binaries containing RDF in the first place - it first appeared in [a 2014 commit](https://github.com/fcrepo4/fcrepo4/commit/2611b7ad02c47f2f7ed8b3302513b0e6b218adf4) that doesn't seem to mention anything about binaries.

The key part to determining if an RDF in POST/PUT creates a binary vs container is apparently the presence of `Content-Disposition: attachment`.. which actually seems quite reasonable.  With the PR intentional binary creation works fine with any RDF mime type, so uploading as text/plain and changing is not necessary in practice

escowles:
Using `Content-Disposition: attachment` to trigger binary storage of RDF does sound very reasonable.  And since it's tied to how filenames are specified when uploading files, it seems like a very good pattern all the way around.  Can we update the REST API docs to include a note about this?

birkland:
@escowles The API docs do mention `Content-Disposition`, albeit only in the context of populating a property:

> The filename provided in the content disposition header will be stored in a ebucore:filename property

So maybe the docs can be updated to mention
- All requests to create a resource (PUT, POST) containing header `Content-Disposition: attachment` will result in the creation of a binary resource (LDP-NR) regardless of `Content-Type`
- All GET requests will contain a `Content-Disposition` header reflecting the filename present in `ebucore:filename`

Also, should this `Contentent-Disposition: attachment` -> LDP-NR behavior be mentioned in the [spec doc](https://docs.google.com/document/d/1ad3MhTQWGof0IJ03yY5jVofCK8uO9yPKMVXVYq7a44o/edit#) for discussion?  LDP doesn't really say all that much about the conditions under which an LDP-NR is created.

escowles:
Those doc updates look good to me!  I agree that triggering LDP-NR creation is an important thing for the spec to cover, so I think it's worth considering.  The LDP interaction model header may be a more standard approach, but it should be addressed one way or another.

awoods:
For reference: https://www.w3.org/TR/ldp/#ldpc-post-createrdf

ajs6f:
I like the functionality, but I would _really_ like to see this actuated by LDP headers. Why wouldn't we do that?

fcrepo-http-api/src/test/resources/test-objects/test.rdf
```diff
@@ -0,0 +1,123 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ajs6f: Minor point, but can we shorten this? There doesn't really seem to be any need for all this text for a test resource. Just a few lines of NTriples would do, which you could even just put in a string in the test, no? Unless I'm missing something about how the test works.

birkland: Good point - I simply included RDF in the test that I _knew_ always caused an exception.. but that aspect is not important for this test.  The point of the test is just to make sure that we get back exactly what we put in, for which a one liner is sufficient.

birkland:
@ajs6f So for now, should this fix be committed as-is (pending simplification of the test), with associated updating of the wiki docs to clarify what the current codebase does now?  Then the conversation about "how to create LDP-NRs" can be taken to the spec doc in order to find a way that broadly makes more sense?

ajs6f:
I won't stand in the way of this fix. It's an improvement over what we've got. But can you file a ticket for the next step?

birkland:
A ticket for the next step has been created:
https://jira.duraspace.org/browse/FCREPO-2095

Could the fix from this PR can make it into the next 4.6.0-RC? :)

acoburn:
@birkland it would be great if you could create a second PR targeting the RC branch.

birkland:
PR against the RC branch is here:
https://github.com/fcrepo4/fcrepo4/pull/1082

PR #1083, opened 2016-07-31
Properly support HTTP Patch with If-Match header checking

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-2098

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -543,8 +543,8 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
             etag = new EntityTag(resource.getDescription().getEtagValue());
             date = resource.getDescription().getLastModifiedDate();
         } else {
-            // Use a weak ETag for the LDP-RS
-            etag = new EntityTag(resource.getDescribedResource().getEtagValue(), true);
+            // Use a strong ETag for the LDP-RS when validating If-(None)-Match headers
+            etag = new EntityTag(resource.getDescribedResource().getEtagValue());
             date = resource.getDescribedResource().getLastModifiedDate();
         }
 
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -760,8 +760,9 @@ public class FedoraLdpIT extends AbstractResourceIT {
 
         // Next, update the description
         final HttpPatch httpPatch = patchObjMethod(id + "/binary/fcr:metadata");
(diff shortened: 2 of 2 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -761,6 +761,7 @@ public void testBinaryEtags() throws IOException, InterruptedException {
         // Next, update the description
         final HttpPatch httpPatch = patchObjMethod(id + "/binary/fcr:metadata");
         httpPatch.addHeader("Content-Type", "application/sparql-update");
+        httpPatch.addHeader("If-Match", descEtag1.replace("W/", ""));
```
ajs6f: This should only happen for "W/" at the start of the tag, right? I know it's picky, but can we be precise about that?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -543,8 +543,8 @@ private static void evaluateRequestPreconditions(final Request request,
             etag = new EntityTag(resource.getDescription().getEtagValue());
             date = resource.getDescription().getLastModifiedDate();
         } else {
-            // Use a weak ETag for the LDP-RS
-            etag = new EntityTag(resource.getDescribedResource().getEtagValue(), true);
+            // Use a strong ETag for the LDP-RS when validating If-(None)-Match headers
+            etag = new EntityTag(resource.getDescribedResource().getEtagValue());
```
ajs6f: Does using a strong ETag here imply anything about the API spec?

acoburn: No, it doesn't. And this brings the Fedora API in line with the [HTTP spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html#sec13.3.3). This section of code relates to evaluating (rather than generating) ETags. And for `GET` requests (according to the spec), ETags with the same opaque identifier are considered equal regardless of whether they are strong or weak. However, for requests that initiate changes to a server resource, the spec is clear that the ETags must be strong in order to evaluate to equal. In that case, we were comparing a strong ETag with a weak ETag, and it resulted in a non-match (unlike the _match_ in a read context):

> Strong validators are usable in any context. Weak validators are only usable in contexts that do not depend on exact equality of an entity. For example, either kind is usable for a conditional GET of a full entity. However, only a strong validator is usable for a sub-range retrieval, since otherwise the client might end up with an internally inconsistent entity.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
acoburn: Good point, I'll clarify the test code.

PR #1086, opened 2016-08-04
Fcrepo 2104 converters

barmintor:
Decompose the Identifier conversion from Jena Resource to Fedora Resource, so that a translator that doesn't require node resolution can be passed down into property value iterators.


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/functions/InjectiveFunction.java
```diff
@@ -0,0 +1,53 @@
...
+     * @return whether b is in the range of the converter
+     */
+    default boolean inRange(final B b) {
+        return true;
+    }
+
+    /**
+     * Reverse the function.
+     * @param b the candidate value for which a domain value should be calculated
+     * @return a domain value or null
```
ajs6f: Should this be `null` or throw an exception for out-of-bounds? The client should have already used other methods to check in-boundsness.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/functions/Converter.java
```diff
@@ -24,17 +24,29 @@
  * @param <A> the type from which we are translating
  * @param <B> the type to which we are translating
  */
-public interface Converter<A, B> extends InvertibleFunction<A, B>, DomainRestrictedFunction<A, B> {
+public interface Converter<A, B> extends InjectiveFunction<A, B>, DomainRestrictedFunction<A, B> {
+
+    @Override
+    public default Converter<B, A> inverse() {
```
ajs6f: If you want to put `toDomain` in, take this out, or have them both but in a subtype of `Converter`, same style as `InjectiveFunction` < `InvertibleFunction`.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/functions/InverseConverterWrapper.java
```diff
@@ -0,0 +1,67 @@
...
+        return original;
+    }
+
+    @Override
+    public <C> Converter<A, C> andThen(final Converter<B, C> after) {
+        return new CompositeConverter<>(this, after);
+    }
+
+    @Override
+    public <C> Converter<C, B> compose(final Converter<C, A> before) {
```
ajs6f: Isn't this just the default impl?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/functions/CompositeConverter.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * 
+ * @author barmintor
+ *
+ * @param <A> the domain of the first chained function
+ * @param <B> the range of the first and domain of the second functions
+ * @param <C> the range of the second function
+ */
+public class CompositeConverter<A, B, C> implements Converter<A, C> {
```
ajs6f: Does this need to be part of the public API? Wouldn't package-private be enough?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/functions/CompositeConverter.java
```diff
@@ -0,0 +1,78 @@
...
+    final Converter<C,B> secondReverse;
+
+    /**
+     * 
+     * @param first the first function
+     * @param second the function to be applied to outputs from the first
+     */
+    public CompositeConverter(final Converter<A,B> first, final Converter<B,C> second) {
+        this.first = first;
+        this.firstReverse = first.inverse();
```
ajs6f: No, I don't think this should be eager. `inverse()` might not be a simple reference-retrieval.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/functions/CompositeConverter.java
```diff
@@ -0,0 +1,78 @@
...
+        this.second = second;
+        this.secondReverse = second.inverse();
```
ajs6f: See above.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -810,18 +806,17 @@ public void disableVersioning() {
...
+            UncheckedPredicate.uncheck((final Node node) -> node.isNodeType("mix:versionable"));
+
     @Override
     public boolean isVersioned() {
-        try {
-            return node.isNodeType("mix:versionable");
-        } catch (final RepositoryException e) {
-            throw new RepositoryRuntimeException(e);
-        }
+        return isVersioned.test(node);
```
ajs6f: I would leave this as the logic-bearing part:
```
public static boolean isVersioned(Node node) {
    try {
        return node.isNodeType("mix:versionable");
    }
}

@Override
public boolean isVersioned() {
    return isVersioned(node);
}
```
and instead of introducing a new `Predicate<Node>`-valued member above, use a method reference `FedoraResourceImpl::isVersioned`. That would read much better to me.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -810,18 +806,17 @@ public void disableVersioning() {
 
     }
 
+    public static Predicate<Node> isVersioned =
+            UncheckedPredicate.uncheck((final Node node) -> node.isNodeType("mix:versionable"));
```
ajs6f: If you are going to do this, do we really need `(final Node node)`? Isn't just `node` enough for the compiler?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/identifiers/CompositeIdentifierConverter.java
```diff
@@ -0,0 +1,63 @@
...
+public class CompositeIdentifierConverter<A, B, C> extends IdentifierConverter<A, C> {
+
+    final IdentifierConverter<A,B> first;
+    final IdentifierConverter<B,A> firstReverse;
+    final IdentifierConverter<B,C> second;
+    final IdentifierConverter<C,B> secondReverse;
+
+    CompositeIdentifierConverter(final IdentifierConverter<A,B> first, final IdentifierConverter<B,C> second) {
+        this.first = first;
+        this.firstReverse = first.inverse();
```
ajs6f: See comments above. This shouldn't be eager.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/identifiers/HashConverter.java
```diff
@@ -26,22 +26,32 @@
 public class HashConverter extends InternalIdentifierConverter {
 
     @Override
-    protected String doForward(final String externalId) {
+    public String apply(final String externalId) {
+        if (!inDomain(externalId)) {
```
ajs6f: This is a bit confusing. I get the logic, but a comment would be good here, about how in(or out)-of-domain-ness, in this particular case, is actually about the existence of a `#`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/identifiers/IdentifierConverter.java
```diff
@@ -0,0 +1,112 @@
...
+    public boolean inDomain(final A resource) {
+        return apply(resource) != null;
+    }
+
+    /**
+     * Convert the given resource into a plain string representation of the conversion to the resource
+     * @param resource the given resource
+     * @return a plain string representation of the conversion to the resource
+     */
+    abstract public String asString(final A resource);
```
ajs6f: Can we not get rid of `asString`? Why on Earth not?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/identifiers/IdentifierConverter.java
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * An {@link IdentifierConverter} accepts and returns identifiers, translating
+ * them in some type-specific manner. The typical use of this
+ * contract is for translating between internal and external identifiers.
+ *
+ * @author ajs6f
+ * @since Mar 26, 2014
+ * @param <B> the type to and from which we are translating
+ */
+public abstract class IdentifierConverter<A, B> implements Converter<A, B> {
```
ajs6f: Can this whole class not go away in favor of one of the newly-introduced types? What's the new semantic that this is adding to `Converter`?

ajs6f:
I may be missing it (because this is a big PR) but has the meaning of "Internal Path" been documented somewhere here? My impression is that it's somewhere between the route into the REST API and the JCR path to a node, with bits from each.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/identifiers/NodeResourceConverter.java
```diff
@@ -33,7 +34,7 @@
  * @author cabeer
  * @since 10/15/14
  */
-public class NodeResourceConverter extends Converter<Node, FedoraResource> {
+public class NodeResourceConverter extends IdentifierConverter<Node, FedoraResource> {
```
ajs6f: This isn't actually an converter of identifiers, is it? It actually retrieves `Node`s. That matters to me, because part of the impetus for this PR was exactly the difference in cost between manipulating identifiers and retrieving nodes.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/identifiers/PathToNodeConverter.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.jcr.Session;
+
+import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
+
+/**
+ * 
+ * @author barmintor
+ *
+ */
+public class PathToNodeConverter extends IdentifierConverter<String, Node> {
```
ajs6f: See comment above. I may be misunderstanding something here, but this doesn't seem like something that is converting identifiers to me.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/HashRdfContext.java
```diff
@@ -57,12 +57,12 @@
...
      */
     public HashRdfContext(final FedoraResource resource,
-                          final IdentifierConverter<Resource, FedoraResource> idTranslator)
+                          final Converter<Resource, String> idTranslator)
             throws RepositoryException {
         super(resource, idTranslator);
 
         concat(getNodeStream(resource)
-                .flatMap(n -> nodeConverter.convert(n).getTriples(idTranslator, PROPERTIES))
+                .flatMap(n -> nodeConverter.apply(n).getTriples(idTranslator, PROPERTIES))
```
ajs6f: I'm not clear on what `nodeConverter` is here. There's no field named that here or in `NodeRdfContext`? (But there's a `nodeConverter()` method...?)

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/ExternalPathToInternalPathConverter.java
```diff
@@ -0,0 +1,95 @@
...
+        final ArrayList<Converter<String,String>> processors = new ArrayList<>();
+        processors.add(new HashConverter());
+        processors.add(new NamespaceConverter());
+        return processors;
+    }
+
+    @Override
+    public String toDomain(final String t) {
+        String result = t;
+        final ListIterator<Converter<String, String>> iter = processors.listIterator(processors.size());
```
ajs6f: Given the likely small size of this list, it might be more readable to use Guava's `Lists::reverse`.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/ExternalPathToInternalPathConverter.java
```diff
@@ -0,0 +1,95 @@
...
+        while (iter.hasPrevious()) {
+            result = iter.previous().inverse().apply(result);
+        }
+        return result;
+    }
+
+    @Override
+    public String apply(final String t) {
+        String result = t;
```
ajs6f: In the old impl, we roll up this chain as a composite in the state of the class and just execute that here. I think that's a little clearer. But this is nice too. If you want to put a smile on @acoburn 's face, let your `CompositeConverter` implement `java.util.stream.Collector<Converter, Converter, Converter>` and you can use all the Java 8 tasty sauce here.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/functions/InjectiveFunction.java
barmintor: You know, I'm not sure. Some one of these previous implementations tested that f(null) == null, so I went with it.

ajs6f: Hm. I'm trying to think of cases where you just throw a bunch of stuff at the converter and see what makes it through.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/functions/InverseConverterWrapper.java
```diff
@@ -0,0 +1,67 @@
...
+        return original.inRange(a);
+    }
+
+    @Override
+    public Converter<B, A> inverse() {
+        return original;
+    }
+
+    @Override
+    public <C> Converter<A, C> andThen(final Converter<B, C> after) {
```
barmintor: These need to stay, I think- but the composition methods are redundant.

ajs6f: These _are_ the composition methods, aren't they?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
```diff
@@ -257,4 +257,10 @@ void replaceProperties(final IdentifierConverter<Resource, FedoraResource> idTra
      * @return the resource being described
      */
     FedoraResource getDescribedResource();
+
+    /**
+     * Get a graph resource corresponding to this resource
+     * @return a graph resource
+     */
+    Resource graphResource(Converter<Resource, String> idTranslator);
```
ajs6f: I'll suggest going ahead and calling this a URI. It's going to be a URI, and that's what people want.

barmintor: I think I have addressed all of these that I can in the last commit; in the
next I will delete IdentifierConverter and hopefully we can get agreement
on a squashable set of changes.

ajs6f: A lot of what I was talking about is in there, but some is not and some is not addressed in the way I meant (not your fault, my fault for using unclear language). What do you say to me sending you a PR to this PR with my highly opinionated opinions? Then you can cherrypick from that what you like.

barmintor: Go for it. Just remember the end goal is an invertible Jena-resource to
internal path string function that doesn't load nodes in the common
execution path, passed down to the property resolution bits.

ajs6f: Yeah, that, _plus_ that Jena resource has to make sense in the HTTP context. Just for fun. :)

awoods:
Is there a JIRA ticket associated with this work?

barmintor:
I could've sworn it was 2104, but it's
https://jira.duraspace.org/browse/FCREPO-2105

awoods:
@barmintor, that ticket is in the `OPEN` state. When this PR is ready for review and testing, please update the ticket to the `START WORK`, then `START REVIEW` states.

barmintor:
Andrew, this is kind of a 2 part effort. Should I make up subtask tickets?

awoods:
@barmintor, sub-tasks are a little nasty in that a JIRA release can not be made (or sprint closed) unless the parent and sub-task(s) are all completed.
Instead, please create a new ticket and establish a linked relationship between the new and existing tickets.

barmintor:
https://jira.duraspace.org/browse/FCREPO-2111

PR #1088, opened 2016-08-09
Refuse updates to containers that include If-Match headers

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2115

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -255,12 +255,16 @@ public class FedoraLdp extends ContentExposingResource {
 
     /**
      * Deletes an object.
+     * @param ifMatch the if-match value
      *
      * @return response
      */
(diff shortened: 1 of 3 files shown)


dchandekstark:
I wonder what @ajs6f meant in https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java#L194-L195 ?

ajs6f:
I think it's fairly clear if you read the whole comment:
```
  * Construct an ETag value from the last modified date and path. JCR has a
     * mix:etag type, but it only takes into account binary properties. We
     * actually want whole-object etag data. TODO : construct and store an ETag
     * value on object modify
```

acoburn:
:-1: see https://github.com/fcrepo4/fcrepo4/pull/1089#issuecomment-238892821 for an explanation.

escowles:
Closed after discussion at https://wiki.duraspace.org/display/FF/2016-08-11+-+Fedora+Tech+Meeting

PR #1089, opened 2016-08-09
Refuse updates to containers that include If-Match headers

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2115

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -255,12 +255,16 @@ public class FedoraLdp extends ContentExposingResource {
 
     /**
      * Deletes an object.
+     * @param ifMatch the if-match value
      *
      * @return response
      */
     @DELETE
     @Timed
-    public Response deleteObject() {
+    public Response deleteObject(@HeaderParam("If-Match") final String ifMatch) {
+        if (resource() instanceof Container && ifMatch != null) {
+            throw new BadRequestException("An If-Match header is not allowed for Container updates");
+        }
         evaluateRequestPreconditions(request, servletResponse, resource(), session);
 
         LOGGER.info("Delete resource '{}'", externalPath);
@@ -359,7 +363,12 @@ public class FedoraLdp extends ContentExposingResource {
             throw new ClientErrorException("An If-Match header is required", 428);
         }
 
+        if (resource() instanceof Container && ifMatch != null) {
+            throw new BadRequestException("An If-Match header is not allowed for Container updates");
+        }
+
         evaluateRequestPreconditions(request, servletResponse, resource, session);
+
         final boolean created = resource.isNew();
 
         try (final RdfStream resourceTriples =
@@ -403,13 +412,15 @@ public class FedoraLdp extends ContentExposingResource {
      * Update an object using SPARQL-UPDATE
      *
      * @param requestBodyStream the request body stream
+     * @param ifMatch the if-match value
      * @return 201
      * @throws IOException if IO exception occurred
      */
     @PATCH
     @Consumes({contentTypeSPARQLUpdate})
     @Timed
-    public Response updateSparql(@ContentLocation final InputStream requestBodyStream)
+    public Response updateSparql(@ContentLocation final InputStream requestBodyStream,
+                                 @HeaderParam("If-Match") final String ifMatch)
             throws IOException {
 
         if (null == requestBodyStream) {
@@ -418,6 +429,10 @@ public class FedoraLdp extends ContentExposingResource {
 
         if (resource() instanceof FedoraBinary) {
             throw new BadRequestException(resource().getPath() + " is not a valid object to receive a PATCH");
+        }
+
+        if (resource() instanceof Container && ifMatch != null) {
+            throw new BadRequestException("An If-Match header is not allowed for Container updates");
         }
 
         try {
--- a/fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
@@ -580,7 +580,7 @@ public class FedoraLdpTest {
     @Test
     public void testDelete() throws Exception {
         final FedoraResource fedoraResource = setResource(FedoraResource.class);
-        final Response actual = testObj.deleteObject();
+        final Response actual = testObj.deleteObject(null);
         assertEquals(NO_CONTENT.getStatusCode(), actual.getStatus());
         verify(fedoraResource).delete();
     }
@@ -679,7 +679,7 @@ public class FedoraLdpTest {
 
         setResource(Container.class);
 
-        testObj.updateSparql(toInputStream("xyz"));
+        testObj.updateSparql(toInputStream("xyz"), null);
     }
 
 
@@ -699,24 +699,24 @@ public class FedoraLdpTest {
 
         doReturn(mockObject).when(testObj).resource();
 
-        testObj.updateSparql(toInputStream("xyz"));
+        testObj.updateSparql(toInputStream("xyz"), null);
     }
 
     @Test(expected = BadRequestException.class)
     public void testPatchWithoutContent() throws MalformedRdfException, IOException {
-        testObj.updateSparql(null);
+        testObj.updateSparql(null, null);
     }
 
     @Test(expected = BadRequestException.class)
     public void testPatchWithMissingContent() throws RepositoryException, MalformedRdfException, IOException {
         setResource(Container.class);
-        testObj.updateSparql(toInputStream(""));
+        testObj.updateSparql(toInputStream(""), null);
     }
 
     @Test(expected = BadRequestException.class)
     public void testPatchBinary() throws RepositoryException, MalformedRdfException, IOException {
         setResource(FedoraBinary.class);
-        testObj.updateSparql(toInputStream(""));
+        testObj.updateSparql(toInputStream(""), null);
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


acoburn:
No, no, no.

dchandekstark:
This does not seem consistent with https://www.w3.org/TR/ldp/#ldpr-resource.

escowles:
@dchandekstark @acoburn LDP says we SHOULD use If-Match, but RFC 7232 says we MUST use strong ETags with If-Match (https://tools.ietf.org/html/rfc7232#section-3.1):

> An origin server MUST use the strong comparison function when comparing entity-tags for If-Match (Section 2.3.2), since the client intends this precondition to prevent the method from being applied if there have been any changes to the representation data.

Since the vagaries of RDF and content negotiation prevent us from issuing strong ETags for containers, I think that means we can't use If-Match.

There is a draft that removes that restriction (https://tools.ietf.org/html/draft-ietf-httpbis-p4-conditional-21#section-3.1).  If we accept this draft and think we can use it, then we could use weak ETags.

dchandekstark:
Whatever the decision I am mainly concerned about client expectations and of course the Ruby ldp gem in particular.

escowles:
@dchandekstark I completely agree, and hope that using the `If-Unmodified-Since` header should be basically equivalent.

acoburn:
My objection to this runs as follows:

1) The LDP [spec](https://www.w3.org/TR/ldp/#ldpr-resource) section 4.2.4.5 says: 

>  LDP clients should use the HTTP If-Match header and HTTP ETags to ensure it isn’t modifying a resource that has changed since the client last retrieved its representation. LDP servers should require the HTTP If-Match header and HTTP ETags to detect collisions. 

By explicitly disallowing `If-Match` headers, we are running counter to the `SHOULD` specification for LDP clients.

2) We are using weak ETags for a reason. Different requests for LDP-RSs (for the same content-type) may appear without byte-to-byte correspondence, i.e. the order of triples varies from request to request. From [RFC 7232](https://tools.ietf.org/html/rfc7232#section-2.1):

> Likewise, a validator is weak if it is shared by two or more representations of a given resource at the same time, unless those representations have identical representation data.

and:

> If an origin server provides an entity-tag for a representation and the generation of that entity-tag does not satisfy all of the characteristics of a strong validator (Section 2.1), then the origin server MUST mark the entity-tag as weak by prefixing its opaque value with "W/" (case-sensitive).

and from [RFC 2616](https://www.ietf.org/rfc/rfc2616.txt):

> A "strong entity tag" MAY be shared by two entities of a resource only if they are equivalent by octet equality.

3) ETag headers are used for _representations_, not _resources_. From [RFC 2616](https://www.ietf.org/rfc/rfc2616.txt):

>  resource: A network data object or service that can be identified by a URI, as defined in section 3.2. Resources may be available in multiple representations (e.g. multiple languages, data formats, size, and resolutions) or vary in other ways.

> representation: An entity included with a response that is subject to content negotiation, as described in section 12. There may exist multiple representations associated with a particular response status.

4) There are two flavors of entity-tag comparison: weak and strong. They are outlined here: https://tools.ietf.org/html/rfc7232#section-2.3.2

```
+--------+--------+-------------------+-----------------+
| ETag 1 | ETag 2 | Strong Comparison | Weak Comparison |
+--------+--------+-------------------+-----------------+
| W/"1"  | W/"1"  | no match          | match           |
| W/"1"  | W/"2"  | no match          | no match        |
| W/"1"  | "1"    | no match          | match           |
| "1"    | "1"    | match             | match           |
+--------+--------+-------------------+-----------------+
```

5. The use of `If-Match` requires a strong comparison. From [RFC 7232](https://tools.ietf.org/html/rfc7232#section-3.1):

> An origin server MUST use the strong comparison function when comparing entity-tags for If-Match (Section 2.3.2), since the client intends this precondition to prevent the method from being applied if there have been any changes to the representation data.

Pulling all of these together, if a client is trying to modify a _resource_ (not a _representation_) the strong form of entity tag must be used. The fact that Fedora uses weak ETags for _representations_ of LDP-RSs is immaterial, because the client is not operating on the representation, it is operating on the resource.

I will change my mind if convinced otherwise, but until then, I am :-1: on this PR.

ajs6f:
I'm pretty confused, but that's normal. That's why I try to stay away from Fedora. But here's a question; @coblej 's example shows taking a weak ETag and extracting the opaque portion and using it as a strong ETag, which then works. That can't be right, can it?

dchandekstark:
@acoburn Makes a pretty strong case.

escowles:
The HTTP spec doesn't make a distinction between modifying resources and representations — it simply says that you have to use the strong comparison algorithm for any update.  Since only representations have ETags, they are all clients would ever have to use with the If-Match header.

I think this is going to be an issue for anyone trying to implement LDP unless they are using a serialized form of the RDF for storage (or can otherwise make guarantees re: triple order, etc.).  So I think this issue is worth bringing up in the broader LDP context, to revisit whether If-Match is really appropriate.  The skew in this area between LDP's focus on resources and HTTP's focus on representations makes it hard to satisfy both.

acoburn:
I also disagree with that. The spec is quite clear in its distinction b/t resources and representations (and entities): https://www.ietf.org/rfc/rfc2616.txt

    resource
      A network data object or service that can be identified by a URI,
      as defined in section 3.2. Resources may be available in multiple
      representations (e.g. multiple languages, data formats, size, and
      resolutions) or vary in other ways.

    entity
      The information transferred as the payload of a request or
      response. An entity consists of metainformation in the form of
      entity-header fields and content in the form of an entity-body, as
      described in section 7.

    representation
      An entity included with a response that is subject to content
      negotiation, as described in section 12. There may exist multiple
      representations associated with a particular response status.

Here, a _representation_ is part of a server response. How could a client _ever_ be operating on that (e.g. via PUT/PATCH/DELETE)? Furthermore: 

    9.6 PUT

    The PUT method requests that the enclosed entity be stored under the
    supplied Request-URI. If the Request-URI refers to an already
    existing resource, the enclosed entity SHOULD be considered as a
    modified version of the one residing on the origin server. If the
    Request-URI does not point to an existing resource, and that URI is
    capable of being defined as a new resource by the requesting user
    agent, the origin server can create the resource with that URI.

This section is just about `PUT`, but it refers specifically to _resources_ (and not representations).

I ask: where does it state that, if a server produces a weak ETag, a client MUST respond with that same weak ETag in its request? I have seen nothing indicating such a thing. In my reading of the spec, the strong/weak distinction refers to the strength of the `entity` (and not the resource). And that entity refers to different things: in the context of a server response, that entity is the bitstream that makes up the HTTP response; in the case of a client, that entity is the HTTP request.

Furthermore, in the case of a client, the weakness indicator is an indication of the strength of that attached entity: for GET the entity-body is empty; for PATCH the entity-body is a sparql-update payload; for DELETE the entity-body is also null; for PUT (on LDP-RSs) the entity-body is a RDF document. In all cases, it refers to the entity being sent by the client and _not_ the resource/entity/representation produced by the server. In a word, it refers to the semantics of the client request and the relative _strength_ of the attached entity. As stated above, these client operations can be split into two categories: READ and WRITE, and any WRITE operation using a weak entity-tag will fail (as per the various specs referred to above); therefore, if a client wishes to make an UPDATE to a resource using a conditional request (e.g. `If-Match`), then it must use a strong ETag and not simply echo back what the server produces.

Let me also be clear -- this PR would mean that _every_ java-based Fedora implementation that intends to use the `fcrepo-http-api` code would not support `If-Match` headers from LDP clients.

escowles:
@acoburn So the alternative that I'm hearing is normalizing the behavior that @coblej's gist shows:

https://gist.github.com/coblej/4fb90ec8f4900be2b01da9692d5eecb5#file-etag-mismatch-L19

A client receives a weak ETag, and somehow knows that it can use it as a strong ETag in subsequent requests?

acoburn:
@escowles a client should respond with whatever semantics it intends to use. If it is using conditional GETs (e.g. `If-None-Match`), then it can use a strong or weak ETag -- it doesn't really matter. If it is using conditional PUT/PATCH/DELETE (e.g. `If-Match`), then the client would need to use a strong ETag. Again, my understanding of the spec(s) is that the strength of an ETag refers to the strength of the HTTP entity (whether it's a server response or a client request).

awoods:
Here is a related ticket for why we are using weak ETags for LDP-RSs:

no-reply:
It seems to me that the problem is that Fedora is trying to use strong validators which fail to meet the criteria. Since the requirements aren't met, using strong validation is inappropriate.

---

Speaking to @acoburn's points in https://github.com/fcrepo4/fcrepo4/pull/1089#issuecomment-238892821.

*1*, *2*, *4* & *5*, I agree with at face value.

Regarding *3*:

> ETag headers are used for representations, not resources

Strong validators are used with respect to __payloads__, rather than __representations. This includes response headers, etc... See [RFC 7232, sec. 2.1](https://tools.ietf.org/html/rfc7232#section-2.1):

>   A "strong validator" is representation metadata that changes value whenever a change occurs to the representation data that would be observable in the payload body of a 200 (OK) response to GET.
>
> A strong validator might change for reasons other than a change to the representation data, such as when a semantically significant part of the representation metadata is changed (e.g., Content-Type), but it is in the best interests of the origin server to only change the value when it is necessary to invalidate the stored responses held by remote caches and authoring tools.

Further requirements for strong validators are listed the same section, for example:

> Likewise, a validator is weak if it is shared by two or more representations of a given resource at the same time, unless those representations have identical representation data.

---

> Pulling all of these together, if a client is trying to modify a resource (not a representation) the strong form of entity tag must be used.

And since there _is no strong entity tag_, that leaves us with two options: ignore the SHOULD in LDP, or ignore the MUST in RFC 7232. I don't see any other way.

escowles:
Closed after discussion at https://wiki.duraspace.org/display/FF/2016-08-11+-+Fedora+Tech+Meeting

PR #1093, opened 2016-08-26
Fix bug in HTML UI due to upgrade to Jena 3

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-2124

Related to: https://jira.duraspace.org/browse/FCREPO-1672

### DIFF
--- a/fcrepo-http-api/src/main/resources/views/common.vsl
+++ b/fcrepo-http-api/src/main/resources/views/common.vsl
@@ -46,7 +46,7 @@
 #end
 
 #macro( isObject $sub )
-#set($type = $anynode.createURI("http://fedora.info/definitions/v4/repository#mixinTypes"))
+#set($type = $helpers.createURI("http://fedora.info/definitions/v4/repository#mixinTypes"))
 #set($object = $anynode.createLiteral("fedora:object"))
 $rdf.find($sub, $type, $object).hasNext()
 #end
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
@@ -19,7 +19,6 @@ package org.fcrepo.http.commons.responses;
 
 import static com.google.common.collect.Lists.newArrayList;
 import static org.apache.jena.graph.GraphUtil.listObjects;
-import static org.apache.jena.graph.NodeFactory.createURI;
 import static org.apache.jena.rdf.model.ResourceFactory.createResource;
 import static org.apache.jena.rdf.model.ResourceFactory.createProperty;
 import static java.util.Arrays.asList;
@@ -53,6 +52,7 @@ import java.util.StringJoiner;
 import javax.ws.rs.core.UriInfo;
 
 import org.apache.jena.graph.Graph;
+import org.apache.jena.graph.NodeFactory;
 import org.apache.jena.graph.Triple;
 import org.apache.jena.graph.impl.LiteralLabel;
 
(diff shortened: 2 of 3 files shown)


fcrepo-http-api/src/main/resources/views/common.vsl
```diff
@@ -46,7 +46,7 @@
 #end
 
 #macro( isObject $sub )
-#set($type = $anynode.createURI("http://fedora.info/definitions/v4/repository#mixinTypes"))
+#set($type = $helpers.createURI("http://fedora.info/definitions/v4/repository#mixinTypes"))
```
acoburn: mixinTypes? wasn't that removed?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
```diff
@@ -399,7 +399,17 @@ public Resource rdfsClass() {
      * @return content-bearing node for the given subject
      */
     public static Node getContentNode(final Node subject) {
-        return subject == null ? null : createURI(subject.getURI().replace("/" + FCR_METADATA, ""));
+        return subject == null ? null : NodeFactory.createURI(subject.getURI().replace("/" + FCR_METADATA, ""));
```
acoburn: why not just use the static method below: `createURI(...)`?

acoburn:
I don't even pretend to understand how the velocity templates work, so I don't know why `$nodeany` is changed to `$helpers`, but if that's what's needed, then sure, this looks good.

ajs6f:
I'm perfectly willing to believe that what @awoods is doing here is good and needed. But I'll be walking away with the memory that one of our most thorough and thoroughly engaged committers, @acoburn, could only write "I don't even pretend to understand how the velocity templates work".

We should probably fix the problem that gave him the "opportunity" to write that.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
ajs6f: It seems like the content of this change is exactly the opposite of what would seem to be a very natural suggestion from @acoburn. I wouldn't be all that surprised if @awoods had some good reason to do this, but the very whimsicality of our arrangements here bespeaks an ominous lack of gravity in their foundations.

fcrepo-http-api/src/main/resources/views/common.vsl
awoods: Although it does not appear that this macro is being used, I wanted to make sure that it at least worked.
https://jena.apache.org/documentation/javadoc/jena/org/apache/jena/graph/Node_ANY.html no longer has the `createURI` method in Jena 3.

acoburn: So perhaps a separate PR could remove that block?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
awoods: @acoburn, I chose to use `NodeFactory.createURI` as a direct replacement for the previous `Node_ANY.createURI` in order to minimize the change in logic as well as keep the `getContentNode` and local `createURI` methods decoupled.
But, as you observed, the functionality is the same either way.
If there is a good reason to do so, I would be fine with changing this to use the local `createURI`.

awoods:
The root to this PR is changing from using the now non-existent:
https://jena.apache.org/documentation/javadoc/jena/org/apache/jena/graph/Node_ANY.html - `createURI`
to using:
https://jena.apache.org/documentation/javadoc/jena/org/apache/jena/graph/NodeFactory.html#createURI-java.lang.String-

fcrepo-http-api/src/main/resources/views/common.vsl
awoods: Absolutely.
https://jira.duraspace.org/browse/FCREPO-2125

ajs6f: We could remove [a lot more](https://jira.duraspace.org/browse/FCREPO-1594). :)

awoods:
Is there any reason not to merge this? 
I do not believe there are any outstanding comments.

PR #1094, opened 2016-08-27
Remove deprecated checksum query param

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-1851


acoburn:
LGTM
Will merge early next week in case there are other comments

awoods:
Is there any reason not to merge this?

PR #1095, opened 2016-08-29
Enable multiple user-provided content digests

awoods:
- Enable users to configure the default digest algorithm
** This is set with the property: fedoraconfig:defaultDigestAlgorithm
- Note: This update modifies the main CND:
** from: premis:hasMessageDigest (URI) COPY
** to: premis:hasMessageDigest (URI) COPY multiple

Resolves: https://jira.duraspace.org/browse/FCREPO-1842

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -598,7 +598,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
                                                    final ContentDisposition contentDisposition,
                                                    final MediaType contentType,
                                                    final Collection<String> checksums) throws InvalidChecksumException {
-        final Collection<URI> checksumURIs = null == checksums ?
+        final Collection<URI> checksumURIs = checksums == null ?
                 new HashSet<>() : checksums.stream().map(checksum -> checksumURI(checksum)).collect(Collectors.toSet());
         final String originalFileName = contentDisposition != null ? contentDisposition.getFileName() : "";
         final String originalContentType = contentType != null ? contentType.toString() : "";
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -56,10 +56,13 @@ import java.io.InputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Date;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import javax.inject.Inject;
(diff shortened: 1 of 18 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -594,14 +597,15 @@ protected void replaceResourceBinaryWithStream(final FedoraBinary result,
                                                    final InputStream requestBodyStream,
                                                    final ContentDisposition contentDisposition,
                                                    final MediaType contentType,
-                                                   final String checksum) throws InvalidChecksumException {
-        final URI checksumURI = checksumURI(checksum);
+                                                   final Collection<String> checksums) throws InvalidChecksumException {
+        final Collection<URI> checksumURIs = null == checksums ?
+                new HashSet<>() : checksums.stream().map(checksum -> checksumURI(checksum)).collect(Collectors.toSet());
```
ajs6f: - `null == checksums` is a bit unusual, better `checksums == null`.
- `checksum -> checksumURI(checksum)` should be a method reference.

awoods: * `null == checksums` prevents the accidental `checksums = null`... but I am happy to please.

* Initially I had a method reference, but have decided to keep it as-is since the method reference is actually longer in this case: `ContentExposingResource::checksumURI`.

ajs6f: Longer but clearer. but it's not a great difference either way.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -594,14 +597,15 @@ protected void replaceResourceBinaryWithStream(final FedoraBinary result,
                                                    final InputStream requestBodyStream,
                                                    final ContentDisposition contentDisposition,
                                                    final MediaType contentType,
-                                                   final String checksum) throws InvalidChecksumException {
-        final URI checksumURI = checksumURI(checksum);
+                                                   final Collection<String> checksums) throws InvalidChecksumException {
+        final Collection<URI> checksumURIs = checksums == null ?
+                new HashSet<>() : checksums.stream().map(checksum -> checksumURI(checksum)).collect(Collectors.toSet());
```
acoburn: @awoods, but if you make Collectors::toSet a `static import`, you'll still end up with a shorter line (even with the method reference)

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -716,20 +719,20 @@ private static void checkLinkForLdpResourceCreation(final String link) {
      * @return the sha1 checksum value
      * @throws InvalidChecksumException if an unsupported digest is used
      */
-    private static String parseDigestHeader(final String digest) throws InvalidChecksumException {
+    private static Collection<String> parseDigestHeader(final String digest) throws InvalidChecksumException {
         try {
             final Map<String,String> digestPairs = RFC3230_SPLITTER.split(nullToEmpty(digest));
-            final boolean checksumTypeIncludeSHA1 = digestPairs.keySet().stream().anyMatch("sha1"::equalsIgnoreCase);
+            final boolean isSupportedAlgorithm = digestPairs.keySet().stream().anyMatch(
```
birkland: If I'm reading this right, then any supported checksum in the request header will match.  Further down a few lines, we encounter `.filter(entry -> ContentDigest.DIGEST_ALGORITHM.isSupportedAlgorithm(entry.getKey()))`, so any *non* matching algorithm will be dropped?  

The way I interpret the wording of the ticket is that the presence of *any* non-supported algorithm should fail:
>  any user-provided digest algorithms that are not supported by Fedora should respond with a 400 status or some other indicator to the client that the digest can not be used. 

It's not exactly clear to me where that happens.  Would it be possible to add an IT where the user specifies a SHA-1 and a non-supported digest (like a MD4, or SHA-3, or BADBADNOEXIST-257) and verifies the presence of a 400 error, or other indicator?

whikloj:
@awoods I've tested and this works as expected. There is one code review comment from @acoburn that appears pending. If you two have touched base and are happy as it is, please let me know and I'll merge this.

acoburn:
@whikloj these are just nit-picky details: I'd also prefer that @awoods used `Collections.emptySet()` instead of `new HashSet<>()`, but that's just a stylistic concern. If @awoods is happy with the code now, then feel free to merge it.

awoods:
@acoburn, I ran into issues with `Collections.emptySet()` is that it is immutable. In this case, I need a mutable Set.

acoburn:
@awoods ahh, ok. in that case, merge away @whikloj

PR #1096, opened 2016-08-31
Added path-based locking in LDP PATH, POST, PUT and DELETE methods.

mikedurbin:
I think this is one way to address https://jira.duraspace.org/browse/FCREPO-2086.


acoburn:
How will this work in a multi-master arrangement? Is it possible to push this locking code into the modeshape impl code rather than making it apply to all future (and potentially distributed) implementations?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/PathLockManager.java
```diff
@@ -0,0 +1,131 @@
...
+
+        private Semaphore wrappedSemaphore;
+
+        private int count;
+
+        /**
+         * Default constructor.
+         * @param key the locking key
+         */
+        public TransientSemaphore(final String key) {
```
ajs6f: Why is this better than just using a plain `Semaphore` with `Map::remove` ?

ajs6f:
What are the implications for the API out of this? Are you expecting this kind of concurrency control to be a requirement of all impls?

fcrepo-http-api/src/test/java/org/fcrepo/http/api/PathLockManagerTest.java
```diff
@@ -0,0 +1,109 @@
...
+    public void testNonConflicting() throws InterruptedException {
+        final PathLockManager m = new PathLockManager();
+        final String path = "path1";
+        final String otherPath = "path2";
+
+        m.getSemaphoreForPath(path);
+        final Actor a = new Actor(otherPath, m);
+        final Thread t = new Thread(a);
+        t.start();
+        t.join(2000);
```
ajs6f: I hate these kinds of "wait and hope" tests. Why not use a proper flag with Awaitility's timed assertion methods for this?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -2525,6 +2524,38 @@ public void testPathsWithBrackets() {
...
+        t2.join();
+        t3.join();
+        t4.join();
+
+        try (final CloseableDataset dataset = getDataset(getObjMethod(parent))) {
+            final DatasetGraph graphStore = dataset.asDatasetGraph();
+            final Iterator<Quad> children = graphStore.find(ANY, ANY, CONTAINS.asNode(), ANY);
+            assertTrue("One of the PUTs should have resulted in a child.", children.hasNext());
+            children.next();
+            if (children.hasNext()) {
```
ajs6f: It might be clear to just `assertFalse` here, instead of `if-fail`.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -141,6 +140,8 @@
 
     protected FedoraResource resource;
 
+    protected static final PathLockManager PATH_LOCKS = new PathLockManager();
```
ajs6f: If we end up using something like this, I would prefer it be injected, not a `static`.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/PathLockManager.java
```diff
@@ -0,0 +1,131 @@
...
+ * method), perform some mutations to the resource at that path, then
+ * invoke release on that semaphore.
+ *
+ * @author md5wz
+ */
+public class PathLockManager {
+
+    private static final Logger LOGGER = getLogger(PathLockManager.class);
+
+    private Map<String, TransientSemaphore> pool = new HashMap<String, TransientSemaphore>();
```
ajs6f: I think this might ought to be a `ConcurrentHashMap`.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/PathLockManager.java
```diff
@@ -0,0 +1,131 @@
...
+            LOGGER.trace("Created a new lock for {} (Thread {})", key, Thread.currentThread().getId());
+        }
+
+        /**
+         * Releases this semaphore.  Also, if there is no one waiting on
+         * it, also removes it from the pool so that it may be garbage collected
+         * (when the caller no longer retains a reference to it).
+         */
+        public void release() {
+            synchronized (pool) {
```
ajs6f: I don't understand this. Why must a thread acquire the lock for the global pool in order to release _one_ semaphore?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/PathLockManager.java
mikedurbin: I don't know which methods on the Semaphore instance are synchronized and on what.  We need to ensure that a user releasing the Semaphore and that semaphore being removed from the map (if eligible) be synchronized with other operations that would add, remove or check whether a Semaphore existed on the map for a given path.

ajs6f: Only for that semaphore. Not for the whole pool. So get the semaphore in question from the map and then synchronize on it. But see my next comment.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/PathLockManager.java
mikedurbin: Because releasing the Semaphore may result in its removal from the pool.

ajs6f: That's why you use a `ConcurrentHashMap`. That's not a panacea, but this is pretty much the use case for it.

mikedurbin:
@acoburn it sounds like you're talking about a future where the fcrepo-http-api code is used but over a different implementation.  Right now there's still a jcr.session that's integral to the functioning of the fcrepo-http-api code.  If we generalized that to something that behaved the same way as a session, but wasn't JCR, we'd still need something like this solution, since our implementation of certain REST methods involve multiple operations against the kernel that are meant to be atomic (like checking to see if something exists and creating it if it doesn't when we PUT).   We could certainly expose an explicit locking API in our kernel (indeed I implemented one before that was since removed), but to avoid having undefined outcomes from Fedora REST API calls, we'd need to bake that locking into our fcrepo-http-api implementation.

acoburn:
@mikedurbin yes, I am referring to a yet-to-be-written future implementation. And all of my concern would be addressed if the locking/coordination code was injected. That would allow me to use (for example) a zookeeper-based coordination service for a multi-node impl but for modeshape, your code here would be entirely appropriate. Also, I am aware that the JCR session is present in the HTTP code -- that is one of the last (major) JCR constructs in the Fcrepo JAVA API, and it will eventually be replaced with a [Fedora-specific interface](https://jira.duraspace.org/browse/FCREPO-1868).

mikedurbin:
Great... if I get consensus to move forward, I'll make the locking injectable.  I only made it static because that was simple and worked and I'm pretty incompetent with these complex frameworks.  A singleton-scoped injected bean for the lock manager would work for me.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
mikedurbin: I really wanted to include the URI of the unexpected child because it's really enlightening (ie, if it has a [1] in it).  I could use an inline if to avoid a NPE when trying to generate the message String, but it feels wierd to make a String that isn't correct and you know is never going to be seen.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/DefaultPathLockManager.java
```diff
@@ -0,0 +1,356 @@
...
+    private class ActivePath {
+
+        private String path;
+
+        private ReadWriteLock rwLock;
+
+        /**
+         * A list with references to every thread that has requested
+         * (though not necessary acquired) a read or write lock through
+         * the mechanism of the greater class' contract and not yet
```
ajs6f: "greater"? Do you mean "outer"?

ajs6f:
I'm still not very enthused about the "one global map of locks" approach here (I'd rather see the JCR locks used to support this), and I think we're doing way too much to organize concurrency for clients, but more importantly, I'd like to establish as part of this PR's process that absolutely no guarantees are being made at the API level about how simultaneous requests are handled. This is all just a particular set of choices that this Modeshape-based implementation has made.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
ajs6f: Okay, I see that. This (lazy messages) is solved in [JUnit 5](http://junit.org/junit5/docs/current/api/org/junit/jupiter/api/Assertions.html#assertFalse-boolean-java.util.function.Supplier-), but that'll be a while to come.

mikedurbin:
@ajs6f, I addressed your inline comments in the latest commit.

Regarding your last comment... I think you're mischaracterizing the nature of this work.  I'm not adding new commitments we need to include in the API, I'm addressing old ones we failed at because our code wasn't thread-safe.  Specifically PUT requests to a URI should create a a resource at that URI or fail.  It shouldn't create a resource at a different URI that has a [1] in it some of the time.  

JCR locks *could* accomplish that goal, but they'd require a significant change to the kernel API but this seemed cleaner and less likely to introduce undesirable behavior.  Given some of the feedback I've received on this fix, and the fact that JCR locking infrastructure once existed in Fedora4 and was subsequently yanked, I don't think I'd be able to make any headway fixing the problem that way.

ajs6f:
@mikedurbin Now we are perhaps both mischaracterizing each other's remarks. I did not suggest that we should not address the ticket. I did and do suggest that this PR is not the only way to do that, and more to the point, that there are simpler ways. For example, throwing a 409 quickly; as far as goes the API, we need to be careful to avoid giving the impression that PUTs do not throw 409s in this situation, because some other impl might choose to do just that.

I don't understand why using JCR locks would require any changes at all to the kernel API. They would just require blocking in the kernel impl.

ajs6f:
What once existed (to which I suppose you to be referring) was an exposure of the actual JCR locks via HTTP, which is not what I am suggesting.

mikedurbin:
That was part of it.  The other part, if I recall correctly, was kernel API calls that could lock or unlock a resource.  Which we'd need to be able to do from the HTTP API in order to eliminate this bug.

ajs6f:
I don't see any need to expose the locking in the kernel API, and I'm not sure why you would claim that. All that is needed is for mutating methods in the impl classes in the `fcrepo-kernel-modeshape` to _use_ the JCR locks.

ajs6f:
To be clear, I am not going to -1 this PR.

ajs6f:
If I understand correctly, the `[1]` indicates that a same-name sibling has been created in the JCR tree. Is this not prevented by Modeshape, based on our node definition file? If this can happen in spite of our node definition file, we should file a bug on it. If we haven't correctly configured this in our node definition file, shouldn't we?

mikedurbin:
Well, then I won't argue any more, but FYI: 

This is the culprit...
https://github.com/fcrepo4/fcrepo4/blob/fcrepo-4.5.1/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java#L349-L355

We could confine locking to the kernel if we added a "findOrCreateFedoraResource()" method that used JCR locking internally and fix this one issue, but I chose to try to fix this entire class of issues.

mikedurbin:
As for same named siblings, we extend a type that supports them, and you can't disallow them with a mixin if they've been allowed.  If modeshape *did* disallow that, without this fix we wouldn't get those same named siblings, we'd get some sort of error coming up from modeshape, which is also an unacceptable outcome for an apparently valid REST API call.

ajs6f:
Can we not extend `nt:folder` instead? Obviously, we would catch that Modeshape exception and rethrow an appropriate JAXRS-mappable exception, just as we do for dozens of other conditions.

ajs6f:
I may very well be reading [this](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/main/resources/fedora-node-types.cnd) wrong, but I don't see that we explicitly  inherit from _any_ particular JCR construct. What's more, I don't see us defining a primary type at all, only mixins. (Except for tombstones, which are just weird all over.)

ajs6f:
Okay, I've looked briefly into this question, and I have an alternative suggestion that shouldn't cost very much to try: currently, as far as I can tell (based on the issue report and the codebase), we are not enforcing a same-name sibling restriction on the pair tree nodes that make up most of the JCR node hierarchy, which seems to me to be what allows this bug to occur. Can we not enable that restriction and see if it blocks this bug?

mikedurbin:
Like my last comment stated, even if the kernel properly disallowed same name siblings, that's still no solution.  Getting an error is just as unacceptable as getting an improperly created sns.

Also, it's not just peartree nodes, the integration test in the PR creates SNSs for non peartree nodes.

ajs6f:
Obviously, @mikedurbin, I think you are mistaken in claiming that SNS restrictions would prevent this problem. I think that it would in fact do exactly what we want done. As far as the different node types involved, that is in no way against my basic point. We can enable SNS restrictions for user-created nodes just as well as for pair tree nodes.

ajs6f:
I don't have a great deal of time this week, but I will try to take a look at this issue again sometime on Thursday. Until then, I would appreciate it if we could hold off on merging this PR.

mikedurbin:
@ajs6f It's been my assertion that the same name siblings appearance is caused by a failure to properly synchronize certain concurrent operations.  I concede that there should be a way to prevent SNS in modeshape and admit I don't rightly know why they're allowed now.  I believe that's irrelevent though, because it's my contention that the appearance of SNS siblings is merely a symptom of the problem.  As further evidence of what I believe the problem to be, I've submitted a commit that includes an additional integration test that I think fails for the same reason the SNS are created.  

At your leisure, could you take a look at the test and verify whether you think a) it's a valid test of expected repository behavior and if so b) what means you would prefer to fix it.

As I've stated before, if we ultimately agree that failure to meet our obligations with regards to concurrency is a problem, I don't have any objections to whether its fixed at the http-api layer or the kernel-impl layer, and chose the former only because I could get to a testable fix sooner and more cleanly.

ajs6f:
I now realize that my expectation of having time to look at this today (or next week) was radically unrealistic. {sigh}

I don't think we actually agree on "our obligations with regards to concurrency", so the rest is a wash. I think that what you've got here certainly fixes the problem-- but I also think it does a great many other things we don't have any responsibility to do.

But what's more important is how to move forward. If no one else has objections to this PR, then let's merge it and if we have to revisit it we will. Otherwise, I'd just like to see a note accompanying the ticket for this PR explaining that the choices made here are completely incidental and occur because of the stupid pair tree stuff. Other implementations incur nothing at all on this account.

mikedurbin:
One more behavior I think we can agree is wrong:

- 4 patches come in simultaneously, each attempting to add a different dc:identifier
- Each patch gets a response indicating success
- when all are complete, not all of the values are present for the resource

acoburn:
👍  Let's get this into master sooner rather than later.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/DefaultPathLockManager.java
```diff
@@ -0,0 +1,353 @@
...
+    @Override
+    public AcquiredLock lockForWrite(final String path, final Session session, final NodeService nodeService)
+            throws InterruptedException {
+        final List<ActivePath.PathScopedLock> locks = new ArrayList<>();
+
+        synchronized (this) {
+            for (String currentPath = normalizePath(path) ;
+                    currentPath == null || currentPath.length() > 0;
+                    currentPath = getParentPath(currentPath)) {
+                if (currentPath == null || (currentPath != path && nodeService.exists(session, currentPath))) {
```
awoods: Please add inline comments.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/DefaultPathLockManager.java
```diff
@@ -0,0 +1,353 @@
...
+
+    @Override
+    public AcquiredLock lockForWrite(final String path, final Session session, final NodeService nodeService)
+            throws InterruptedException {
+        final List<ActivePath.PathScopedLock> locks = new ArrayList<>();
+
+        synchronized (this) {
+            for (String currentPath = normalizePath(path) ;
+                    currentPath == null || currentPath.length() > 0;
+                    currentPath = getParentPath(currentPath)) {
```
awoods: `getParentPath` will hit an NPE if `currentPath == null`, no?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/DefaultPathLockManager.java
awoods: Are you intentionally using `!=` instead of `! currentPath.equals(path)` for comparing these strings?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/PathLockManager.java
```diff
@@ -0,0 +1,96 @@
...
+     * that until release() is invoked on the returned Lock that no changes are made
+     * to the resource that would affect its display.
+     *
+     * @param path the path to a resource to be viewed
+     * @param session the current session
+     * @param nodeService the repository NodeService implementation
+     * @return an acquired Lock on the relevant resources
+     * @throws InterruptedException if interrupted while waiting to acquire the lock
+     */
+    public AcquiredLock lockForRead(String path, Session session, NodeService nodeService) throws InterruptedException;
```
awoods: The `InterruptedException` would make more sense to me being a CheckedException if we were doing something with it at the `FedoraLdp` or `ContentExposingResource` level. But we are just allowing it to flow up through the JAX-RS endpoints. 
Would it make more sense to have it be an UncheckedException and additionally create an `InterruptedExceptionMapper` which maps to the appropriate response code?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/DefaultPathLockManager.java
```diff
@@ -0,0 +1,353 @@
...
+    static String normalizePath(final String path) {
+        if (path.endsWith("/")) {
+            return path.substring(0, path.length() - 1);
+        } else {
+            return path;
+        }
+    }
+
+    @Override
+    public AcquiredLock lockForRead(final String path, final Session session, final NodeService nodeService)
```
mikedurbin: It's been a while, but I think my reason was primarily because I could imagine an alternative implementation that might need it.

awoods: @mikedurbin, are you opposed to removing the dead code and letting alternative implementations shape things in the future?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/DefaultPathLockManager.java
mikedurbin: I was, but then when I added the "normalizePath()" method I introduced a bug.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/PathLockManager.java
mikedurbin: Sure... not sure what status code we should associate with this occurrence, which I can only imagine results from an impatient shutdown of the servlet container.

awoods: 503 Service Unavailable ?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/DefaultPathLockManager.java
mikedurbin: Done.

mikedurbin: It is now.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/DefaultPathLockManager.java
mikedurbin: I think this is clearer now, but if you mean the case where a caller invokes the method with a null parameter, yes, that case would end up with a NPE, but I think that's reasonable.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/DefaultPathLockManager.java
```diff
@@ -0,0 +1,353 @@
...
+                }
+                locks.add(getActivePath(currentPath).getWriteLock());
+            }
+        }
+
+        return new AcquiredMultiPathLock(locks);
+    }
+
+    @Override
+    public AcquiredLock lockForDelete(final String path, final Session session, final NodeService nodeService)
```
mikedurbin: I removed these too.

awoods:
:+1: This looks good to go. @acoburn ? @ajs6f ? Any reason not to push this now?

acoburn:
👍

awoods:
@mikedurbin , If we get confirmation from @scossu that this patch addresses their issue, we will want to back-port this patch to the https://github.com/fcrepo4/fcrepo4/tree/4.6.0-maintenance branch.

mikedurbin:
@awoods, neat... I'll ask follow up questions about the process when we're ready to go.

awoods:
Additionally, @mikedurbin , If we get confirmation from @scossu that this patch addresses their issue, we will want to get this patch into the 4.7.0-RC.
@whikloj ^^

scossu:
We have run 10K ingests without an error with the patched 4.7.0 version. However, this new version added a 50% increase in ingest times compared to a vanilla 4.6.0 install. 

If we can test on a patched 4.6.0 we will be able to establish whether it is the patch or the 4.7 upgrade that caused the performance penalty.

awoods:
Thank you, @scossu. @mikedurbin, are you in a position to create a patch targeting the 4.6.0-maintenance branch?

awoods:
@mikedurbin : fyi, I have created three tickets to track the improvement of the unit tests and back-porting this patch to 4.6.1 and 4.7.0.
https://jira.duraspace.org/browse/FCREPO-2270
https://jira.duraspace.org/browse/FCREPO-2271
https://jira.duraspace.org/browse/FCREPO-2272

PR #1098, opened 2016-09-16
Update lastModified when child resources are deleted

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-2213

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -873,6 +873,109 @@ public class FedoraLdpIT extends AbstractResourceIT {
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


whikloj:
The issue (FCREPO-2213) was one of Last Modified date not changing, should the tests also assert that? It does check for changes in eTags, and I think last modified date is used to generate eTags... 

So we should be okay.

I'm sorry if this is overly picky. The actual change is so simple, I want to make sure I don't miss anything.

whikloj:
Oh wait, the date wouldn't change in an integration test because the time difference would be very small. But if the eTag changes then in theory the date is changing, if you waited. My bad..

whikloj:
Ok I verified this works as described.

acoburn:
@whikloj you are correct -- I am testing ETag headers in the iTests because the last-modified header is rounded to the nearest second, which doesn't work in this context (there are only millisecond differences) but the ETag -- that always changes. I could also have employed the sleep(1000) method, but I find that to be kind of gross.

PR #1101, opened 2016-09-21
Add a Queue-based JMS message producer

acoburn:
Addresses: https://jira.duraspace.org/browse/FCREPO-2005

As a follow-on to this, we should add some documentation about how to set this up.

**Note:** I left the topic-based producer as the default. If there is interest in changing that behavior, it can be addressed in a separate PR.


ajs6f:
👍

awoods:
@acoburn "As a follow-on to this, we should add some documentation about how to set this up."
Would you please create a ticket for this if it has not already been done?

acoburn:
I would be happy to, though it will be some time before this becomes part of a released version of Fedora (i.e. 4.7.1). Is there benefit to doing this now? or should it wait.

awoods:
May as well create the documentation ticket now so the idea does not get lost with time. Thanks!

PR #1104, opened 2016-09-26
Remove DELETE/MOVE methods from Allow header on RepositoryRoot resources

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-2158

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -612,7 +612,12 @@ public class FedoraLdp extends ContentExposingResource {
             servletResponse.addHeader("Accept-Patch", contentTypeSPARQLUpdate);
 
         } else if (resource() instanceof Container) {
-            options = "MOVE,COPY,DELETE,POST,HEAD,GET,PUT,PATCH,OPTIONS";
+            // MOVE/COPY/DELETE operations are not available on a RepositoryRoot resource
+            if (resource().getContainer() == null) {
+                options = "POST,HEAD,GET,PUT,PATCH,OPTIONS";
+            } else {
+                options = "MOVE,COPY,DELETE,POST,HEAD,GET,PUT,PATCH,OPTIONS";
+            }
             servletResponse.addHeader("Accept-Patch", contentTypeSPARQLUpdate);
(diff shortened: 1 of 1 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -612,7 +612,12 @@ private void addLinkAndOptionsHttpHeaders() {
             servletResponse.addHeader("Accept-Patch", contentTypeSPARQLUpdate);
 
         } else if (resource() instanceof Container) {
-            options = "MOVE,COPY,DELETE,POST,HEAD,GET,PUT,PATCH,OPTIONS";
+            // MOVE/COPY/DELETE operations are not available on a RepositoryRoot resource
+            if (resource().getContainer() == null) {
+                options = "POST,HEAD,GET,PUT,PATCH,OPTIONS";
```
ajs6f: Could we make these into constants?

ajs6f: Is it strictly forbidden to COPY root? It seems daft, but do we actually have a reason it cannot be done?

ajs6f: Could we make these into constants?

acoburn: It's certainly not _forbidden_, but it's not supported, so it seems wrong to advertise that it is supported.

ajs6f: Oh, I didn't realize that it doesn't work. I wonder why? Oh, well, it's certainly better not to advertise it if it doesn't work.

acoburn: so, it _is_ possible to COPY the repository root, but the results are strange. Since it's not strictly prohibited, I'll add the COPY operation back to the allowable methods.

ajs6f: Fedora 4: We just hand you the shotgun. You have to pull the trigger yourself.

PR #1105, opened 2016-09-27
Exclude internal (jcr, mode) prefixes from namespace map

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-2244


awoods:
:+1: Looks good to me. 

I believe in these cases where elements are being removed from responses, we have leaned towards not creating integration tests demonstrating the absence of those elements... which is what you have here.

ajs6f:
I think the policy rests on whether we are actually _forbidding_ the appearance of something (which is pretty rare) or whether we are just saying that you can't count on it appearing (which is much more common).

PR #1107, opened 2016-09-30
Add core types to be used to replace the use of jcr.Session in the kernel API

acoburn:
I have a complete implementation for this, but so as not to make the PR too massive, these are the new interfaces being added which will be used to replace instances of `jcr.Session`, `jcr.Repository`, and the existing `org.fcrepo.kernel.api.services.TransactionService`.

Begins to address: https://jira.duraspace.org/browse/FCREPO-1868

Supersedes #968

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
@@ -21,6 +21,7 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.Collection;
 import java.util.Map;
+import java.util.Optional;
 
 /**
  * The Fedora Session abstraction
@@ -54,9 +55,9 @@ public interface FedoraSession {
 
     /**
      * Get the date this session expires
-     * @return expiration date
+     * @return expiration date, if one exists
      */
-    Instant getExpires();
+    Optional<Instant> getExpires();
 
     /**
      * Get the session identifier
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
@@ -17,6 +17,7 @@
  */
 package org.fcrepo.kernel.api;
 
+import static java.lang.Long.parseLong;
 import static java.time.Duration.ofMinutes;
 import static java.time.Duration.ofMillis;
 
@@ -33,20 +34,9 @@ import java.util.Optional;
 public interface FedoraSession {
 
     // The default timeout is 3 minutes
-    public static final Duration DEFAULT_TIMEOUT = ofMinutes(3);
+    public static final String DEFAULT_TIMEOUT = Long.toString(ofMinutes(3).toMillis());
 
-    public static final String TIMEOUT_SYSTEM_PROPERTY = "fcrepo.transactions.timeout";
-
-    public interface State {}
-
-    public enum RequiredState implements State {
-        BATCH, EXPIRED, ACTIVE
-    }
-
-    /**
-     * Begin a BATCH session
-     */
-    void beginBatch();
+    public static final String TIMEOUT_SYSTEM_PROPERTY = "fcrepo.session.timeout";
 
     /**
      * Expire the session
@@ -59,24 +49,11 @@ public interface FedoraSession {
     void commit();
 
     /**
-     * Rollback any batch operations
-     */
-    void rollback();
-
-    /**
      * Update the expiry by the provided amount
      * @param amountToAdd the amount of time to add
      * @return the new expiration date
      */
     Instant updateExpiry(Duration amountToAdd);
-
-    /**
-     * Determine whether the session has the provided state
-     * @param <T> an enum that extends the default State enum
-     * @param state the state to test
-     * @return whether the session is in the given state
-     */
-    <T extends State> boolean hasState(T state);
 
     /**
      * Get the date this session was created
@@ -103,20 +80,6 @@ public interface FedoraSession {
     String getUserId();
 
     /**
-     * Set a namespace prefix
-     * @param prefix the prefix
-     * @param uri the URI
-     */
-    void setNamespacePrefix(String prefix, String uri);
-
-    /**
-     * Get a namespace for a given prefix.
-     * @param prefix the prefix
-     * @return the namespace URI
-     */
-    Optional<String> getNamespace(String prefix);
-
-    /**
      * Get a mapping of registered namespaces
      * @return the namespace mapping
      */
@@ -141,10 +104,6 @@ public interface FedoraSession {
      * @return the default timeout value
      */
     public static Duration operationTimeout() {
-       if (System.getProperty(TIMEOUT_SYSTEM_PROPERTY) != null) {
-            return ofMillis(Long.parseLong(System.getProperty(TIMEOUT_SYSTEM_PROPERTY)));
-        } else {
-            return DEFAULT_TIMEOUT;
-        }
+       return ofMillis(parseLong(System.getProperty(TIMEOUT_SYSTEM_PROPERTY, DEFAULT_TIMEOUT)));
     }
 }
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
@@ -26,9 +26,6 @@ import java.util.Map;
  * The Fedora Session abstraction
  *
  * @author acoburn
- *
- * Note: This interface does not make any guarantees about thread-safety. Any
- * synchronization behavior is left to the discretion of the implementation.
  */
 public interface FedoraSession {
 
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
@@ -1,0 +1,150 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.kernel.api;
+
+import static java.time.Duration.ofMinutes;
+import static java.time.Duration.ofMillis;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * The Fedora Session abstraction
+ *
+ * @author acoburn
+ */
+public interface FedoraSession {
+
+    // The default timeout is 3 minutes
+    public static final Duration DEFAULT_TIMEOUT = ofMinutes(3);
+
+    public static final String TIMEOUT_SYSTEM_PROPERTY = "fcrepo.transactions.timeout";
+
+    public interface State {}
+
+    public enum RequiredState implements State {
+        BATCH, EXPIRED, ACTIVE
+    }
+
+    /**
+     * Begin a BATCH session
+     */
+    void beginBatch();
+
+    /**
+     * Expire the session
+     */
+    void expire();
+
+    /**
+     * Commit any batch operations
+     */
+    void commit();
+
+    /**
+     * Rollback any batch operations
+     */
+    void rollback();
+
+    /**
+     * Update the expiry by the provided amount
+     * @param amountToAdd the amount of time to add
+     * @return the new expiration date
+     */
+    Instant updateExpiry(Duration amountToAdd);
+
+    /**
+     * Determine whether the session has the provided state
+     * @param <T> an enum that extends the default State enum
+     * @param state the state to test
+     * @return whether the session is in the given state
+     */
+    <T extends State> boolean hasState(T state);
+
+    /**
+     * Get the date this session was created
+     * @return creation date
+     */
+    Instant getCreated();
+
+    /**
+     * Get the date this session expires
+     * @return expiration date
+     */
+    Instant getExpires();
+
+    /**
+     * Get the session identifier
+     * @return the session id
+     */
+    String getId();
+
+    /**
+     * Get the user identifier associated with this session
+     * @return the user id
+     */
+    String getUserId();
+
+    /**
+     * Set a namespace prefix
+     * @param prefix the prefix
+     * @param uri the URI
+     */
+    void setNamespacePrefix(String prefix, String uri);
+
+    /**
+     * Get a namespace for a given prefix.
+     * @param prefix the prefix
+     * @return the namespace URI
+     */
+    Optional<String> getNamespace(String prefix);
+
+    /**
+     * Get a mapping of registered namespaces
+     * @return the namespace mapping
+     */
+    Map<String, String> getNamespaces();
+
+    /**
+     * Set session-specific data
+     * @param key the key
+     * @param value the value
+     */
+    void setSessionData(String key, String value);
+
+    /**
+     * Retrieve the session data for a given key
+     * @param key the key
+     * @return the value
+     */
+    Optional<String> getSessionData(String key);
+
+    /**
+     * Retrieve the default operation timeout value
+     * @return the default timeout value
+     */
+    public static Duration operationTimeout() {
+       if (System.getProperty(TIMEOUT_SYSTEM_PROPERTY) != null) {
+            return ofMillis(Long.parseLong(System.getProperty(TIMEOUT_SYSTEM_PROPERTY)));
+        } else {
+            return DEFAULT_TIMEOUT;
+        }
+    }
+}
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
@@ -83,6 +83,9 @@ public interface FedoraSession {
      * Add session-specific data
      * @param key the key
      * @param value the value
+     *
+     * Note: it is up to the particular implementation as to whether multi-valued session data
+     * is allowed.
      */
     void addSessionData(String key, String value);
 
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
@@ -17,10 +17,6 @@
  */
 package org.fcrepo.kernel.api;
 
-import static java.lang.Long.parseLong;
-import static java.time.Duration.ofMinutes;
-import static java.time.Duration.ofMillis;
-
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Map;
@@ -32,11 +28,6 @@ import java.util.Optional;
  * @author acoburn
  */
 public interface FedoraSession {
-
-    // The default timeout is 3 minutes
-    public static final String DEFAULT_TIMEOUT = Long.toString(ofMinutes(3).toMillis());
-
-    public static final String TIMEOUT_SYSTEM_PROPERTY = "fcrepo.session.timeout";
 
     /**
      * Expire the session
@@ -98,12 +89,4 @@ public interface FedoraSession {
      * @return the value
      */
     Optional<String> getSessionData(String key);
-
-    /**
-     * Retrieve the default operation timeout value
-     * @return the default timeout value
-     */
-    public static Duration operationTimeout() {
-       return ofMillis(parseLong(System.getProperty(TIMEOUT_SYSTEM_PROPERTY, DEFAULT_TIMEOUT)));
-    }
 }
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
@@ -19,13 +19,16 @@ package org.fcrepo.kernel.api;
 
 import java.time.Duration;
 import java.time.Instant;
+import java.util.Collection;
 import java.util.Map;
-import java.util.Optional;
 
 /**
  * The Fedora Session abstraction
  *
  * @author acoburn
+ *
+ * Note: This interface does not make any guarantees about thread-safety. Any
+ * synchronization behavior is left to the discretion of the implementation.
  */
 public interface FedoraSession {
 
@@ -77,16 +80,31 @@ public interface FedoraSession {
     Map<String, String> getNamespaces();
 
     /**
-     * Set session-specific data
+     * Add session-specific data
      * @param key the key
      * @param value the value
      */
-    void setSessionData(String key, String value);
+    void addSessionData(String key, String value);
 
     /**
      * Retrieve the session data for a given key
      * @param key the key
      * @return the value
      */
-    Optional<String> getSessionData(String key);
+    Collection<String> getSessionData(String key);
+
+    /**
+     * Remove a particular session key-value pair
+     * @param key the data key
+     * @param value the data value
+     */
+    void removeSessionData(String key, String value);
+
+    /**
+     * Remove all session data for a particular key
+     * @param key the data key
+     */
+    default void removeSessionData(String key) {
+        getSessionData(key).forEach(v -> removeSessionData(key, v));
+    }
 }
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
@@ -28,7 +28,7 @@ public interface BatchService {
     /**
      * Check for expired batch operations and remove them
      */
-    void removeAndRollbackExpired();
+    void removeExpired();
 
     /**
      * Create a new batch operation with a FedoraSession and add it to the currently open ones
@@ -36,15 +36,15 @@ public interface BatchService {
      * @param session The session to use for this batch operation
      * @param username the name of the {@link java.security.Principal}
      */
-    void beginBatch(FedoraSession session, String username);
+    void begin(FedoraSession session, String username);
 
     /**
      * Create a new FedoraSession for the anonymous user and add it to the currently open ones
      *
      * @param session The session to use for this batch operation
      */
-    default void beginBatch(FedoraSession session) {
-        beginBatch(session, null);
+    default void begin(FedoraSession session) {
+        begin(session, null);
     }
 
     /**
@@ -54,7 +54,7 @@ public interface BatchService {
      * @param username the name of the {@link java.security.Principal}
      * @return the {@link FedoraSession} with this user
      */
-    FedoraSession getBatchSession(String txId, String username);
+    FedoraSession getSession(String txId, String username);
 
     /**
      * Retrieve an open {@link FedoraSession} for an anonymous user
@@ -62,8 +62,8 @@ public interface BatchService {
      * @param txId the Id of the {@link FedoraSession}
      * @return the {@link FedoraSession}
      */
-    default FedoraSession getBatchSession(String txId) {
-        return getBatchSession(txId, null);
+    default FedoraSession getSession(String txId) {
+        return getSession(txId, null);
     }
 
     /**
@@ -108,14 +108,14 @@ public interface BatchService {
      * @param txid the id of the {@link FedoraSession}
      * @param username the name of the {@link java.security.Principal}
      */
-    void rollback(String txid, String username);
+    void abort(String txid, String username);
 
     /**
      * Roll back any uncommited changes during a {@link FedoraSession} for the anonymous user
      *
      * @param txid the id of the {@link FedoraSession}
      */
-    default void rollback(String txid) {
-        rollback(txid, null);
+    default void abort(String txid) {
+        abort(txid, null);
     }
 }
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
@@ -39,7 +39,7 @@ public interface BatchService {
     void begin(FedoraSession session, String username);
 
     /**
-     * Create a new FedoraSession for the anonymous user and add it to the currently open ones
+     * Create a new batch operation with a FedoraSession for the anonymous user and add it to the currently open ones
      *
      * @param session The session to use for this batch operation
      */
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
@@ -86,6 +86,23 @@ public interface BatchService {
     }
 
     /**
+     * Refresh an existing session using an implementation-defined default
+     *
+     * @param sessionId the Id of the {@link FedoraSession}
+     * @param username the name of the {@link java.security.Principal}
+     */
+    void refresh(String sessionId, String username);
+
+    /**
+     * Refresh an existing anonymous session using an implementation-defined default
+     *
+     * @param sessionId the Id of the {@link FedoraSession}
+     */
+    default void refresh(String sessionId) {
+        refresh(sessionId, null);
+    }
+
+    /**
      * Commit any changes during a {@link FedoraSession} with the given id and username
      *
      * @param sessionId the id of the {@link FedoraSession}
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
@@ -1,0 +1,121 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.kernel.api.services;
+
+import org.fcrepo.kernel.api.FedoraSession;
+
+/**
+ * @author acoburn
+ * @since Sept 30, 2016
+ */
+public interface BatchService {
+
+    /**
+     * Check for expired batch operations and remove them
+     */
+    void removeAndRollbackExpired();
+
+    /**
+     * Create a new batch operation with a FedoraSession and add it to the currently open ones
+     *
+     * @param session The session to use for this batch operation
+     * @param username the name of the {@link java.security.Principal}
+     */
+    void beginBatch(FedoraSession session, String username);
+
+    /**
+     * Create a new FedoraSession for the anonymous user and add it to the currently open ones
+     *
+     * @param session The session to use for this batch operation
+     */
+    default void beginBatch(FedoraSession session) {
+        beginBatch(session, null);
+    }
+
+    /**
+     * Retrieve an open {@link FedoraSession} for a given user
+     *
+     * @param txId the Id of the {@link FedoraSession}
+     * @param username the name of the {@link java.security.Principal}
+     * @return the {@link FedoraSession} with this user
+     */
+    FedoraSession getBatchSession(String txId, String username);
+
+    /**
+     * Retrieve an open {@link FedoraSession} for an anonymous user
+     *
+     * @param txId the Id of the {@link FedoraSession}
+     * @return the {@link FedoraSession}
+     */
+    default FedoraSession getBatchSession(String txId) {
+        return getBatchSession(txId, null);
+    }
+
+    /**
+     * Check if a FedoraSession exists for a particular user
+     *
+     * @param txid the Id of the {@link FedoraSession}
+     * @param username the name of the {@link java.security.Principal}
+     * @return the {@link FedoraSession} object for the defined user
+     */
+    boolean exists(String txid, String username);
+
+    /**
+     * Check if a FedoraSession exists for the anonymous user
+     *
+     * @param txid the Id of the {@link FedoraSession}
+     * @return the {@link FedoraSession} object
+     */
+    default boolean exists(String txid) {
+        return exists(txid, null);
+    }
+
+    /**
+     * Commit any changes during a {@link FedoraSession} with the given id and username
+     *
+     * @param txid the id of the {@link FedoraSession}
+     * @param username the name of the {@link java.security.Principal}
+     */
+    void commit(String txid, String username);
+
+    /**
+     * Commit any changes during a {@link FedoraSession} with the given id for the anonymous user
+     *
+     * @param txid the id of the {@link FedoraSession}
+     */
+    default void commit(String txid) {
+        commit(txid, null);
+    }
+
+    /**
+     * Roll back any uncommited changes during a {@link FedoraSession}
+     *
+     * @param txid the id of the {@link FedoraSession}
+     * @param username the name of the {@link java.security.Principal}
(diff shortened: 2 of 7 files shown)


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
```diff
@@ -0,0 +1,150 @@
...
+public interface FedoraSession {
+
+    // The default timeout is 3 minutes
+    public static final Duration DEFAULT_TIMEOUT = ofMinutes(3);
+
+    public static final String TIMEOUT_SYSTEM_PROPERTY = "fcrepo.transactions.timeout";
+
+    public interface State {}
+
+    public enum RequiredState implements State {
```
ajs6f: Wait, what does it mean for a `FedoraSession` to be in `ACTIVE` as opposed to `BATCH`?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
```diff
@@ -0,0 +1,150 @@
...
+    /**
+     * Update the expiry by the provided amount
+     * @param amountToAdd the amount of time to add
+     * @return the new expiration date
+     */
+    Instant updateExpiry(Duration amountToAdd);
+
+    /**
+     * Determine whether the session has the provided state
+     * @param <T> an enum that extends the default State enum
```
ajs6f: You cannot extend a Java `enum`.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
```diff
@@ -0,0 +1,150 @@
...
+     * @return the user id
+     */
+    String getUserId();
+
+    /**
+     * Set a namespace prefix
+     * @param prefix the prefix
+     * @param uri the URI
+     */
+    void setNamespacePrefix(String prefix, String uri);
```
ajs6f: We don't have any real data-handling methods, but we're still going to demand that you manage namespaces? I don't think this makes sense here.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
acoburn: `ACTIVE` is something that is used in the current impl to test `jcr.Session::isLive`. `BATCH` is a marker to indicate that a particular session persists across multiple HTTP requests.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
```diff
@@ -0,0 +1,150 @@
...
+     * @return the value
+     */
+    Optional<String> getSessionData(String key);
+
+    /**
+     * Retrieve the default operation timeout value
+     * @return the default timeout value
+     */
+    public static Duration operationTimeout() {
+       if (System.getProperty(TIMEOUT_SYSTEM_PROPERTY) != null) {
```
ajs6f: How about just`java.lang.System.getProperty(String, String)`?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
acoburn: the comment is wrong -- `State` is an interface, not an `enum`.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
acoburn: yes, that part can go.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
ajs6f: Do we need to do that here? Can't we just offer 'begin', 'commit', etc., and leave it at that? HTTP notions don't belong here, to my mind.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
acoburn: yes, that would be much better

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
ajs6f: Then what is the point of the generic? Doesn't it come to:
```
boolean hasState(State state);
```
?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
acoburn: you mean remove `ACTIVE`? I have that here b/c the http layers test for that.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
```diff
@@ -0,0 +1,121 @@
...
+ */
+package org.fcrepo.kernel.api.services;
+
+import org.fcrepo.kernel.api.FedoraSession;
+
+/**
+ * @author acoburn
+ * @since Sept 30, 2016
+ */
+public interface BatchService {
```
ajs6f: Is this a kind of "manage the connections between users and open transactions" thing?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
ajs6f: I'm saying I don't see the need for this state machinery at all.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
acoburn: yes, that's correct, I can simplify that signature.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
acoburn: There is currently an interface called `TransactionService`. This interface is a slight modification of that, mostly to remove some of the baggage of using the word "transaction"

ajs6f: Yeah. I wonder if we want this stuff in the kernel at all.

Partly, this is about the fact that I don't think we have any clear sense of why there are all these `*Service` guys.

acoburn: The reason they are in the kernel is so that the HTTP layers can access these "services" via spring injection. Moving those out would be a structural change that goes way beyond the scope of this effort.

ajs6f: No, I'm not asking you to do that. I just don't want to make any effort to preserve them if we can avoid it.

acoburn: Yes, but thinking about a possible OSGi-based impl, these "service" interfaces could come in really handy.

ajs6f: Okay, leave it be for now. I'm certainly not trying to block this.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
```diff
@@ -0,0 +1,121 @@
...
+     */
+    void removeExpired();
+
+    /**
+     * Create a new batch operation with a FedoraSession and add it to the currently open ones
+     *
+     * @param session The session to use for this batch operation
+     * @param username the name of the {@link java.security.Principal}
+     */
+    void begin(FedoraSession session, String username);
```
awoods: From the FedoraSession interface, it would appear that a `userId` is already associated with sessions. How is this arg `username` reconciled with the session userId?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * Retrieve an open {@link FedoraSession} for a given user
+     *
+     * @param txId the Id of the {@link FedoraSession}
+     * @param username the name of the {@link java.security.Principal}
+     * @return the {@link FedoraSession} with this user
+     */
+    FedoraSession getSession(String txId, String username);
```
awoods: What is the expected result if the `txId` exists but not with the provided arg `username`?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /**
+     * Check if a FedoraSession exists for a particular user
+     *
+     * @param txid the Id of the {@link FedoraSession}
+     * @param username the name of the {@link java.security.Principal}
+     * @return the {@link FedoraSession} object for the defined user
+     */
+    boolean exists(String txid, String username);
```
awoods: Minor: change `txid` to `txId` for consistency.
..or the other way around.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Check if a FedoraSession exists for the anonymous user
+     *
+     * @param txid the Id of the {@link FedoraSession}
+     * @return the {@link FedoraSession} object
+     */
+    default boolean exists(String txid) {
```
awoods: Minor: change `txid` to `txId` for consistency.
..or the other way around.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
```diff
@@ -0,0 +1,121 @@
...
+        return exists(txid, null);
+    }
+
+    /**
+     * Commit any changes during a {@link FedoraSession} with the given id and username
+     *
+     * @param txid the id of the {@link FedoraSession}
+     * @param username the name of the {@link java.security.Principal}
+     */
+    void commit(String txid, String username);
```
awoods: Minor: change `txid` to `txId` for consistency.
..or the other way around.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Commit any changes during a {@link FedoraSession} with the given id for the anonymous user
+     *
+     * @param txid the id of the {@link FedoraSession}
+     */
+    default void commit(String txid) {
```
awoods: ...

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
```diff
@@ -0,0 +1,121 @@
...
+        commit(txid, null);
+    }
+
+    /**
+     * Roll back any uncommited changes during a {@link FedoraSession}
+     *
+     * @param txid the id of the {@link FedoraSession}
+     * @param username the name of the {@link java.security.Principal}
+     */
+    void abort(String txid, String username);
```
awoods: ...

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * Roll back any uncommited changes during a {@link FedoraSession} for the anonymous user
+     *
+     * @param txid the id of the {@link FedoraSession}
+     */
+    default void abort(String txid) {
```
awoods: ...

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
acoburn: I did not change the behavior of the existing TransactionServiceImpl, namely an Exception is thrown: https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/TransactionServiceImpl.java#L137-L138

Personally, I'd prefer to return an Optional<FedoraSession>, but this PR is about removing the jcr.Session, not about changing everything.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
acoburn: You are correct: the jcr.Session interface has a getUserID method, and the new FedoraSession interface has a getUserId method -- which in the impl just proxies to the jcr.Session::getUserID. The issue is that in the case of an anonymous user, the result of the jcr.Session::getUserID is not `null`, but rather "<anonymous>". But we can't actually rely on that string, because that, too, is configurable in the repository.json file. This way, we are relying on the value of the java.security.Principal::getName value for looking up FedoraSession values.

Believe me -- my first stab at this excluded the `username` parameter on all of these methods, but relying on that conflicts with how transaction values are looked up.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
acoburn: I'd go further and remove the "tx" part altogether, just using `sessionId`.

acoburn: @awoods I changed this and all subsequent `txid` names to `sessionId`.

awoods:
:+1:

acoburn:
@awoods, @ajs6f I added the modeshape implementation for these new interfaces. Once I get the green light on this, I'll add the rest of the commits (which actually remove the use of jcr.Session from the kernel-api).

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraRepositoryImpl.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+        throw new IllegalArgumentException("Credentials are of the wrong type");
+    }
+
+    /**
+     * Retrieve the internal JCR Repository object
+     *
+     * @return the JCR Repository
+     */
+    public Repository getRepository() {
```
ajs6f: This naming is confusing; maybe `getJcrRepository`?

acoburn: That works for me. Also, I should double check the use of this method, because we may be able to do away with this altogether.

acoburn: @ajs6f I renamed the method here and in `FedoraSessionImpl` (`getSession` -> `getJcrSession`). The `FedoraRepositoryImpl::getJcrRepository` method is only used for unit tests -- I don't know if that's a good reason to keep it, but generally, I tried to keep existing unit/integration tests in place. All things being equal, I would probably remove that test and therefore that method, too. @awoods: thoughts?

acoburn:
@awoods @ajs6f let me know when you are ready for the full implementation. Barring any changes with the code so far, the next commit will be `76 files changed, 1019 insertions(+), 2175 deletions(-)`

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
acoburn: @ajs6f: related to the `org.fcrepo.kernel.api.services.*Service` types, I wonder whether they might not be better situated in a different package -- say: `org.fcrepo.spi` or `org.fcrepo.kernel.spi` (which could be independent of the `fcrepo-kernel-api` module). Once we have a full specification and remove all of the JCR concepts from the API, I think there's a certain amount of re-organization that can happen w/r/t the modules and package namespaces, but that's a conversation for later.

ajs6f: Yeah, let's get through this. I have yet to hear anyone give a _purpose_ for those types. (I know you have OSGi in mind, but I don't think could have been our original purpose. If we decide to have these kinds of types for OSGi, they should be impl'd in a common OSGi-related module, not in the kernel-api forcing everyone to impl them directly.)

acoburn: I'd argue that the _purpose_ of these types is to make it possible for the HTTP layer (fcrepo-http-api and fcrepo-http-commons) to access those implementation-specific services that are injected via Spring (but that could be substituted with Blueprint for OSGi). Either way, I agree that they are not a necessary part of the kernel-api.

ajs6f: They do that, but they are only _needed_ to do that on the assumption that model types (like `FedoraResource`) aren't managed by the bean container or other outer system. @awoods and I have talked in the past about whether that's a step we ought to contemplate taking. My point here is that whatever we do in the JCR impl, other people may want to play it out differently elsewhere and we shouldn't make their lives any harder.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
```diff
@@ -0,0 +1,92 @@
...
+    Instant updateExpiry(Duration amountToAdd);
+
+    /**
+     * Get the date this session was created
+     * @return creation date
+     */
+    Instant getCreated();
+
+    /**
+     * Get the date this session expires
```
ajs6f: We've talked about immortal sessions. Either there should be a note explaining what to use for that "value" or this should be `Optional<Instant>` (I like the latter).

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
```diff
@@ -0,0 +1,92 @@
...
+     * @return the namespace mapping
+     */
+    Map<String, String> getNamespaces();
+
+    /**
+     * Set session-specific data
+     * @param key the key
+     * @param value the value
+     */
+    void setSessionData(String key, String value);
```
ajs6f: What about multi-valued keys?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraRepositoryImpl.java
```diff
@@ -0,0 +1,87 @@
...
+    @Override
+    public FedoraSession login(final Object credentials) {
+        if (credentials instanceof Credentials) {
+            try {
+                return new FedoraSessionImpl(repository.login((Credentials) credentials));
+            } catch (final RepositoryException ex) {
+                throw new RepositoryRuntimeException(ex);
+            }
+        }
+        throw new IllegalArgumentException("Credentials are of the wrong type");
```
ajs6f: Can this be a more specific subtype?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraRepositoryImpl.java
```diff
@@ -0,0 +1,87 @@
...
+     * Retrieve the internal JCR Repository from a FedoraRepository object
+     *
+     * @param repository the FedoraRepository
+     * @return the JCR Repository
+     */
+    public static Repository getJcrRepository(final FedoraRepository repository) {
+        if (repository instanceof FedoraRepositoryImpl) {
+            return ((FedoraRepositoryImpl)repository).getJcrRepository();
+        }
+        throw new IllegalArgumentException("FedoraRepository is of the wrong type");
```
ajs6f: Ditto-- maybe a more specific exception?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
```diff
@@ -0,0 +1,182 @@
...
+ * @author acoburn
+ */
+public class FedoraSessionImpl implements FedoraSession {
+
+    // The default timeout is 3 minutes
+    public static final String DEFAULT_TIMEOUT = Long.toString(ofMinutes(3).toMillis());
+
+    public static final String TIMEOUT_SYSTEM_PROPERTY = "fcrepo.session.timeout";
+
+    private final Session session;
```
ajs6f: This name is a little confusing: maybe call it `jcrSession` or something?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
```diff
@@ -0,0 +1,182 @@
...
+import org.fcrepo.kernel.api.FedoraSession;
+import org.fcrepo.kernel.api.exception.AccessDeniedException;
+import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
+import org.fcrepo.kernel.modeshape.utils.NamespaceTools;
+
+/**
+ * An implementation of the FedoraSession abstraction
+ * @author acoburn
+ */
+public class FedoraSessionImpl implements FedoraSession {
```
ajs6f: If this class is not thread-safe (and it does not appear to be) we should note that in the doc.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * Get the internal JCR session from an existing FedoraSession
+     * @param session the FedoraSession
+     * @return the JCR session
+     */
+    public static Session getJcrSession(final FedoraSession session) {
+        if (session instanceof FedoraSessionImpl) {
+            return ((FedoraSessionImpl)session).getJcrSession();
+        }
+        throw new IllegalArgumentException("FedoraSession is of the wrong type");
```
ajs6f: Maybe a more specific exception type?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BatchServiceImpl.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Override
+    public void begin(final FedoraSession session, final String username) {
+        transactions.put(getTxKey(session.getId(), username), session);
+    }
+
+    @Override
+    public FedoraSession getSession(final String sessionId, final String username) {
+        final FedoraSession tx = transactions.computeIfAbsent(getTxKey(sessionId, username), s -> {
```
ajs6f: I like Java 8 but this seems a bit fancy-shmancy. How about
```
String txKey = getTxKey(sessionId, username);
if (!transactions.containsKey(txKey)) {
    throw new SessionMissingException("Batch session with id: " + txKey + " is not available!");
}
return transactions.get(txKey);
```

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BatchServiceImpl.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Override
+    public boolean exists(final String sessionId, final String username) {
+        return transactions.containsKey(getTxKey(sessionId, username));
+    }
+
+    @Override
+    public void commit(final String sessionId, final String username) {
+        final FedoraSession session = transactions.remove(getTxKey(sessionId, username));
```
ajs6f: If the `session.commit()` fails, don't you not want to have removed it from the transactions? The client might want to retry.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BatchServiceImpl.java
```diff
@@ -0,0 +1,135 @@
...
+
+    @Override
+    public void refresh(final String sessionId, final String username) {
+        final FedoraSession session = getSession(sessionId, username);
+        session.updateExpiry(operationTimeout());
+    }
+
+    @Override
+    public void abort(final String sessionId, final String username) {
+        final FedoraSession tx = transactions.remove(getTxKey(sessionId, username));
```
ajs6f: Same as above-- if the expire fails, shouldn't the session still be available for further detective work?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BatchServiceImpl.java
```diff
@@ -0,0 +1,135 @@
...
+        if (tx == null) {
+            throw new SessionMissingException("Batch session with id " + sessionId +
+                    " is not available for " + username);
+        }
+        tx.expire();
+    }
+
+    private static String getTxKey(final String sessionId, final String username) {
+        return username == null ? sessionId : username + ":" + sessionId;
```
ajs6f: Maybe it's just me, but I would actually prefer anonymous sessions to have the colon.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BatchServiceImpl.java
ajs6f: The Javadocs for `Map::computeIfAbsent` say `The default implementation makes no guarantees about synchronization or atomicity properties of this method.` so you are relying on `ConcurrentHashMap`s impl-specific behavior here.... but do we have any understanding anyway of the atomicity of `*Service`? I don't think this is coherent.

acoburn:
@ajs6f I addressed the latest round of comments in the latest commit.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
```diff
@@ -77,16 +80,31 @@
     Map<String, String> getNamespaces();
 
     /**
-     * Set session-specific data
+     * Add session-specific data
```
ajs6f: Can we have a note here that it is up to the impl whether multiple values per key are afforded, and a note on the immediate impl that they aren't?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
```diff
@@ -53,10 +55,10 @@
 
     public static final String TIMEOUT_SYSTEM_PROPERTY = "fcrepo.session.timeout";
 
-    private final Session session;
+    private final Session jcrSession;
     private final String id;
     private final Instant created;
-    private final Map<String, String> sessionData;
+    private final ConcurrentHashMap<String, String> sessionData;
```
ajs6f: If we aren't claiming that this is thread-safe, what's the point of `ConcurrentHashMap`?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BatchServiceImpl.java
```diff
@@ -92,10 +93,13 @@ public void begin(final FedoraSession session, final String username) {
 
     @Override
     public FedoraSession getSession(final String sessionId, final String username) {
-        final FedoraSession tx = transactions.computeIfAbsent(getTxKey(sessionId, username), s -> {
-            throw new SessionMissingException("Batch session with id: " + s + " is not available");
-        });
-        return tx;
+        System.out.println(transactions.keySet().toString());
```
ajs6f: Looks like some debugging juice?

acoburn: Ooops!

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
ajs6f: I don't get that. If multiple threads are going to hit a single instance of this class, how is it okay for this class to not be thread-safe?

ajs6f: https://docs.jboss.org/author/display/MODE/Repository+and+Session?_sscc=t#RepositoryandSession-Threadsafety

Grumble...

ajs6f: I don't think we should make guarantees, and if we don't, we should use the cheapest `Map` that does the job.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
acoburn: I updated the FedoraSession interface to return an `Optional<Instant>`

acoburn:
@ajs6f more comments addressed in the latest commit.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Create a new batch operation with a FedoraSession and add it to the currently open ones
+     *
+     * @param session The session to use for this batch operation
+     * @param username the name of the {@link java.security.Principal}
+     */
+    void begin(FedoraSession session, String username);
+
+    /**
+     * Create a new FedoraSession for the anonymous user and add it to the currently open ones
```
awoods: I am confused by: "Create a new FedoraSession..." since a `FedoraSession` is being passed into this method.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * Retrieve an open {@link FedoraSession} for a given user
+     *
+     * @param sessionId the Id of the {@link FedoraSession}
+     * @param username the name of the {@link java.security.Principal}
+     * @return the {@link FedoraSession} with this user
+     */
+    FedoraSession getSession(String sessionId, String username);
```
awoods: What happens if no such session for the provided username exists?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * @param sessionId the Id of the {@link FedoraSession}
+     * @return the {@link FedoraSession} object
+     */
+    default boolean exists(String sessionId) {
+        return exists(sessionId, null);
+    }
+
+    /**
+     * Refresh an existing session using an implementation-defined default
```
awoods: What is the difference between "Refresh an existing session" here and `FedoraSession.updateExpiry`? https://github.com/fcrepo4/fcrepo4/pull/1107/files#diff-2a974e96c1ebe0bcd554be7d704b6955R48

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
```diff
@@ -0,0 +1,138 @@
...
+        refresh(sessionId, null);
+    }
+
+    /**
+     * Commit any changes during a {@link FedoraSession} with the given id and username
+     *
+     * @param sessionId the id of the {@link FedoraSession}
+     * @param username the name of the {@link java.security.Principal}
+     */
+    void commit(String sessionId, String username);
```
awoods: There seems to be some overlap between this interface and `FedoraSession`. How are we reconciling this commit with `FedoraSession.commit`?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BatchServiceImpl.java
```diff
@@ -0,0 +1,133 @@
...
+    private static final Logger LOGGER = getLogger(BatchServiceImpl.class);
+
+    /**
+     * TODO since sessions have to be available on all nodes, they have to
+     * be either persisted or written to a distributed map or sth, not just this
+     * plain hashmap that follows
+     */
+    private static Map<String, FedoraSession> transactions = new ConcurrentHashMap<>();
+
+    public static final long REAP_INTERVAL = 1000;
```
awoods: This can be `private`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BatchServiceImpl.java
```diff
@@ -0,0 +1,133 @@
...
+public class BatchServiceImpl extends AbstractService implements BatchService {
+
+    private static final Logger LOGGER = getLogger(BatchServiceImpl.class);
+
+    /**
+     * TODO since sessions have to be available on all nodes, they have to
+     * be either persisted or written to a distributed map or sth, not just this
+     * plain hashmap that follows
+     */
+    private static Map<String, FedoraSession> transactions = new ConcurrentHashMap<>();
```
awoods: Would `sessions` be a clearer name for this map? It seems like we are otherwise trying to get away from the term  "transaction".

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
```diff
@@ -0,0 +1,205 @@
...
+import org.fcrepo.kernel.modeshape.utils.NamespaceTools;
+
+/**
+ * An implementation of the FedoraSession abstraction
+ * @author acoburn
+ */
+public class FedoraSessionImpl implements FedoraSession {
+
+    // The default timeout is 3 minutes
+    public static final String DEFAULT_TIMEOUT = Long.toString(ofMinutes(3).toMillis());
```
awoods: Can be `private`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
```diff
@@ -0,0 +1,205 @@
...
+
+    public static final String TIMEOUT_SYSTEM_PROPERTY = "fcrepo.session.timeout";
```
awoods: Can be `private`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+     * A key for looking up the transaction id in a session key-value pair
+     */
+    public static final String FCREPO_TX_ID = "fcrepo.tx.id";
+
+    private static final ObjectMapper mapper = new ObjectMapper();
+
+    /**
+     * Create a Fedora session with a JCR session and a username
```
awoods: This javadoc should probably be updated to remove reference to `username`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
```diff
@@ -0,0 +1,205 @@
...
+    public Map<String, String> getNamespaces() {
+        return NamespaceTools.getNamespaces(jcrSession);
+    }
+
+    /**
+     *  Add session data
+     *  @param key the data key
+     *  @param value the data value
+     *
+     *  Note: while the FedoraSession inteface permits multi-valued
```
awoods: Typos: `inteface` and `assciated`

awoods: Have you verified that the current codebase does not require appending values to an existing UserData key?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
acoburn: A `SessionMissingException` is thrown.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/BatchService.java
acoburn: The BatchService implementation will typically involve an in-memory (possibly distributed) map of FedoraSessions. `BatchService::refresh` will update the expiry of one of those batch sessions. I.e., BatchService::refresh will probably call FedoraSession::updateExpiry, but as for how that is done is up to the implementation.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BatchServiceImpl.java
acoburn: sure

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BatchServiceImpl.java
acoburn: this is public (I believe) to facilitate testing. I can add the appropriate guava annotation.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
acoburn: same as above...

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
acoburn: same as above...

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
acoburn: yes

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
```diff
@@ -51,8 +52,10 @@
 public class FedoraSessionImpl implements FedoraSession {
 
     // The default timeout is 3 minutes
+    @VisibleForTesting
     public static final String DEFAULT_TIMEOUT = Long.toString(ofMinutes(3).toMillis());
```
awoods: This is not actually used in any testing code... and can presently be safely `private`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
```diff
@@ -51,8 +52,10 @@
...
 
+    @VisibleForTesting
```
awoods: This is not actually used in any testing code... and can presently be safely `private`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BatchServiceImpl.java
```diff
@@ -60,8 +61,9 @@
      * be either persisted or written to a distributed map or sth, not just this
      * plain hashmap that follows
      */
-    private static Map<String, FedoraSession> transactions = new ConcurrentHashMap<>();
+    private static Map<String, FedoraSession> sessions = new ConcurrentHashMap<>();
 
+    @VisibleForTesting
```
awoods: This is not actually used in any testing code... and can presently be safely `private`.

awoods:
This is ready to go from my perspective. @ajs6f, have all of your comments been addressed?

ajs6f:
@awoods I have no idea. That's not @acoburn's fault at all-- I find this new Github "review" mode terribly confusing and unhelpful and I can no longer keep track of this conversation. Frankly, if @acoburn gives us his word that he addressed everything, that's fine with me.

acoburn:
And I find that this "review mode" makes it difficult to keep track of issues that I've resolved. As far as I know, I have addressed all of the issues brought up in this discussion, but it has also involved a lot of back and forth, and it is entirely possible that something has fallen between the cracks.

awoods:
It sounds like, to the best of our collective knowledge, this PR is ready to go. Feel free to merge it @ajs6f .

ajs6f:
Am I to close and reopen the ticket?

acoburn:
@ajs6f just reopen the Jira ticket.

PR #1110, opened 2016-10-05
Resolves https://jira.duraspace.org/browse/FCREPO-1454 +

dbernstein:
Also a one line modification to  fcrepo-webapp/pom.xml in order to support passing jvm args to jetty for debugging purposes etc.


fcrepo-http-api/src/main/resources/views/common-node-actions.vsl
```diff
@@ -50,13 +50,24 @@
     <input type="text" id="new_id" placeholder="(auto-generated identifier)" name="slug"  class="form-control"/>
         </div>
 
+    <div class="form-group" id="turtle_payload_container">
+    	<label for="rdf"  class="control-label">
+	        RDF (turtle)
+	    </label>
+		<textarea rows="10" id="turtle_payload" name="turtle" class="form-control" 
+		placeholder="@prefix pcdm: <http://pcdm.org/models#>
```
acoburn: Is PCDM the best vocabulary to use as "placeholder text"? Perhaps something more general such as `skos:Concept` or `dcmitype:Collection` would be better.

fcrepo-http-api/src/main/resources/views/common.js
```diff
@@ -80,7 +80,11 @@
       };
       reader.readAsArrayBuffer(update_file);
     } else {
-      fn(method, url, headers, null);
+      const turtle = document.getElementById('turtle_payload').value;
```
acoburn: Is the `turtle_payload` element available for _all_ views? If so, this is fine; otherwise, you may need

    const turtle = document.getElementById('turtle_payload')
    if (turtle && turtle.value) {
        headers.push(['Content-Type', 'text/turtle'])
        fn(method, url, headers, turtle.value)
    } else {
        fn(method, url, headers, null)
    }

For example, I wonder if this would be available on NonRDFSourceDescription pages, and also on Fixity check pages -- there may be others.

fcrepo-http-api/src/main/resources/views/common-node-actions.vsl
```diff
@@ -50,13 +50,24 @@
...
         </div>
 
+    <div class="form-group" id="turtle_payload_container">
+    	<label for="rdf"  class="control-label">
+	        RDF (turtle)
+	    </label>
+		<textarea rows="10" id="turtle_payload" name="turtle" class="form-control" 
+		placeholder="@prefix pcdm: <http://pcdm.org/models#> 
+<> a pcdm:Object . ">
+	    </textarea>
```
acoburn: And given the whitespace preceding the `</textarea>` element, the placeholder text never actually appears (unless one removes the whitespace)

fcrepo-http-api/src/main/resources/views/common.js
acoburn: This is not an issue, since it appears only in the `addChild` function.

fcrepo-webapp/pom.xml
```diff
@@ -365,6 +365,7 @@
               <value>classpath:/config/file-simple/repository.json</value>
             </systemProperty>
           </systemProperties>
+          <jvmArgs>${jetty.jvm.args}</jvmArgs>
```
acoburn: How does this relate to the FCREPO-1454 ticket?

dbernstein: It doesn't directly.  I dropped that in there because helpful for debugging.  In the future I'll create a ticket and an appropriately named branch.

dbernstein: okay - will do.

fcrepo-http-api/src/main/resources/views/common-node-actions.vsl
```diff
@@ -50,13 +50,23 @@
     <input type="text" id="new_id" placeholder="(auto-generated identifier)" name="slug"  class="form-control"/>
         </div>
 
+    <div class="form-group" id="turtle_payload_container">
+        <label for="rdf"  class="control-label">
+	        RDF (turtle)
+        </label>
+        <textarea rows="10" id="turtle_payload" name="turtle" class="form-control" 
+                  placeholder="@prefix dcmitype: <http://purl.org/dc/dcmitype/> 
+<> a dcmitype:Collection ."></textarea>
```
acoburn: Another point here -- but Firefox (I believe) doesn't allow newlines in attributes (it just ignores them). So, I wonder if that implies anything about the example provided.

acoburn: Actually, the W3C states that newlines (CR & LF) are not permitted in placeholder attributes: https://www.w3.org/TR/html5/forms.html#the-placeholder-attribute

dbernstein: Perhaps I should remove the placeholder text altogether?

acoburn: I've been looking into this, too, and all the mechanisms I've seen for including multi-line text in placeholder attributes seem pretty flaky. I'd either remove the attribute or put in something like this: `<> a <http://purl.org/dc/dcmitype/Collection>` which doesn't require newlines.

fcrepo-http-api/src/main/resources/views/common.js
```diff
@@ -80,7 +80,13 @@
       };
       reader.readAsArrayBuffer(update_file);
     } else {
-      fn(method, url, headers, null);
+    	const turtle = document.getElementById('turtle_payload')
```
acoburn: Minor point: you should have semicolons at the end of statements here. They aren't technically needed, but we should at least be consistent with the javascript. Also check the indentation to make it consistent -- most indentation is two spaces, not three or four as you have here.

acoburn:
👍

PR #1111, opened 2016-10-06
Make it possible to set the location for velocity.log

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-2260

I hold no brief for the use of Velocity templates, but this makes it possible to at least configure the location of `velocity.log`. Otherwise, it shows up in all sorts of inconvenient locations.

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
@@ -17,6 +17,7 @@
  */
 package org.fcrepo.http.api.responses;
 
+import static java.lang.System.getProperty;
 import static java.util.stream.Stream.of;
 import static javax.ws.rs.core.MediaType.APPLICATION_XHTML_XML;
 import static javax.ws.rs.core.MediaType.APPLICATION_XHTML_XML_TYPE;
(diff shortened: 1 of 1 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
```diff
@@ -121,6 +122,15 @@ void init() throws IOException {
 
         LOGGER.trace("Velocity engine initializing...");
         final Properties properties = new Properties();
+        final String fcrepoHome = getProperty("fcrepo.home");
+        final String velocityLog = getProperty("fcrepo.velocity.runtime.log");
```
ajs6f: This is a bit shorter with just
```
final String velocityLog = getProperty("fcrepo.velocity.runtime.log", getProperty("fcrepo.home"));
LOGGER.debug("Using {} for the velocity log", velocityLog);
```

acoburn: the issue here is that `fcrepo.home` isn't always set (in which case it defaults to the current directory), and if `fcrepo.velocity.runtime.log` isn't set, it can't be just a directory -- it needs to be a full path to a file, so it would need to be more like this:

    final String velocityLog = getProperty("fcrepo.velocity.runtime.log", getProperty("fcrepo.home", "") + "velocity.log")

but there, if the `fcrepo.home` doesn't end with a file.separator, then you get an incorrect path for the log. And if you put in the separator but you have an empty `fedora.home` value, you may end up with `velocity.log` in the root of the filesystem.

Which is to say, that I could shorten this, but this is pretty readable as it is.

ajs6f: Okay, the hell with it. Just do whatever needs doing.

PR #1115, opened 2016-10-10
Ensures that the rest api will return a 507 response with a message body

dbernstein:
that describes the problem (ie "No space left on device") when creating or
updating a resource that exhausts the available temp space.

    Resolves: https://jira.duraspace.org/browse/FCREPO-1931


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java.orig
```diff
@@ -0,0 +1,743 @@
+/*
```
acoburn: I believe this file should not be part of the PR.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/exception/InsufficientStorageException.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.fcrepo.kernel.api.exception;
+
+/**
+ * Exception thrown when the device has insufficient storage to complete the operation.
+ *
+ * @author Daniel Bernstein
+ * @since Oct 7, 2016
+ */
+public class InsufficientStorageException extends Exception {
```
acoburn: Should this be a RuntimeException?

fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
```diff
@@ -787,9 +789,32 @@ public void testCreateNewBinary() throws RepositoryException, MalformedRdfExcept
...
+            final Response actual = testObj.createObject(null, APPLICATION_OCTET_STREAM_TYPE, "b", content, null,
+                    null);
+            verify(mockBinary).setContent(content, APPLICATION_OCTET_STREAM, Collections.emptySet(), "", null);
+        }
+    }
+
     @Test
     public void testCreateNewBinaryWithContentTypeWithParams() throws RepositoryException, MalformedRdfException,
-            InvalidChecksumException, IOException {
+    InsufficientStorageException, InvalidChecksumException, IOException {
```
acoburn: Minor point: I believe the style of wrapped lines like this is to indent the additional lines 8 spaces.

fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
```diff
@@ -803,7 +828,7 @@ public void testCreateNewBinaryWithContentTypeWithParams() throws RepositoryExce
 
     @Test
     public void testCreateNewBinaryWithChecksumSHA() throws RepositoryException, MalformedRdfException,
-            InvalidChecksumException, IOException {
+    InvalidChecksumException, InsufficientStorageException, IOException {
```
acoburn: See above about formatting.

fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
```diff
@@ -837,7 +862,7 @@ public void testCreateNewBinaryWithChecksumMD5() throws RepositoryException, Mal
 
     @Test
     public void testCreateNewBinaryWithChecksumSHAandMD5() throws RepositoryException, MalformedRdfException,
-            InvalidChecksumException, IOException {
+    InsufficientStorageException, InvalidChecksumException, IOException {
```
acoburn: formatting...

fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
```diff
@@ -863,20 +888,21 @@ public void testCreateNewBinaryWithChecksumSHAandMD5() throws RepositoryExceptio
 
     @Test(expected = ClientErrorException.class)
     public void testPostToBinary() throws MalformedRdfException, InvalidChecksumException,
-            IOException, RepositoryException {
+    InsufficientStorageException, IOException, RepositoryException {
```
acoburn: formatting...

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -296,6 +297,7 @@ public Response deleteObject() {
      * @return 204
      * @throws InvalidChecksumException if invalid checksum exception occurred
      * @throws MalformedRdfException if malformed rdf exception occurred
+     * @throws InsufficientStorageException if operation fails due to insufficient storage on the device.
```
acoburn: If this becomes a runtime exception (see below), there will be no need to declare it here or in the method signature.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java.orig
```diff
@@ -0,0 +1,743 @@
...
+
+/**
+ * @author cabeer
+ * @author ajs6f
+ * @since 9/25/14
+ */
+
+@Scope("request")
+@Path("/{path: .*}")
+public class FedoraLdp extends ContentExposingResource {
```
awoods: This file should be removed from the PR.

fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
```diff
@@ -787,9 +789,32 @@ public void testCreateNewBinary() throws RepositoryException, MalformedRdfExcept
         }
     }
 
+    @Test(expected = InsufficientStorageException.class)
+    public void testCreateNewBinaryWithInsufficientResources() throws RepositoryException, MalformedRdfException,
```
awoods: Minor: spreading the `throws` classes over three lines seems unnecessary. Will then all fit on line `794`?

fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
```diff
@@ -787,9 +789,32 @@ public void testCreateNewBinary() throws RepositoryException, MalformedRdfExcept
...
+            final RuntimeException ex = new RuntimeException(new IOException("root exception", new IOException(
+                    "No space left on device")));
+            doThrow(ex).when(mockBinary).setContent(content, APPLICATION_OCTET_STREAM_TYPE.toString(),
+                    Collections
+                    .emptySet(),
+                    "", null);
+
+            final Response actual = testObj.createObject(null, APPLICATION_OCTET_STREAM_TYPE, "b", content, null,
+                    null);
+            verify(mockBinary).setContent(content, APPLICATION_OCTET_STREAM, Collections.emptySet(), "", null);
```
awoods: Does this line ever get invoked? It seems like the exception will be thrown at line `809`.

fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/InsufficientStorageExceptionMapperTest.java
```diff
@@ -0,0 +1,53 @@
...
+    public void setUp() {
+        testObj = new InsufficientStorageExceptionMapper();
+    }
+
+    @Test
+    public void testToResponse() {
+        final InsufficientStorageException input = new InsufficientStorageException(
+                "No space left on device.", null);
+        final Response actual = testObj.toResponse(input);
+        assertEquals(507, actual.getStatus());
```
awoods: May as well use the `public` field:
https://github.com/fcrepo4/fcrepo4/pull/1115/files#diff-81224db45f13bf2e73f8c601576b40bcR45

..instead of the `507` String.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -356,6 +358,8 @@ public Response createOrReplaceObjectRdf(
                 }
                 throw new ClientErrorException("Invalid Content Type " + requestContentType, UNSUPPORTED_MEDIA_TYPE);
             }
+        } catch (final Exception e) {
```
awoods: It seems like we only care about `IOExceptions` when looking for storage issues. Is there a reason we are catching all `Exception`s here?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -506,13 +511,44 @@ public Response createObject(@HeaderParam("Content-Disposition") final ContentDi
             session.save();
         } catch (final RepositoryException e) {
             throw new RepositoryRuntimeException(e);
+        } catch (final Exception e) {
```
awoods: It seems like we only care about `IOExceptions` when looking for storage issues. Is there a reason we are catching all `Exception`s here?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
dbernstein: The IOException is the root cause but the exception that bubbles up is actually a BinaryStoreException which is not a subclass of IOException.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
dbernstein: The IOException is the root cause but the exception that bubbles up is actually a BinaryStoreException which is not a subclass of IOException.

fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
dbernstein: yes - good point  -I'll remove it.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
awoods: The question is, which exception are we catching here? `IOException` or `BinaryStoreException` or `could be anything`?

dbernstein: BinaryStoreException.  But I was concerned that because the exception bubbles up from the modeshape layer,  future versions of modeshape my not sure the BinaryStoreException. Furthermore,  naming the BinaryStoreException would tie create an unnecessary dependency on modeshape/jcr.  There has been talk of trying to decouple jcr from fedora.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
dbernstein: In this case,  there is a problem I now see.  BinaryStoreException is actually a subclass of RepositoryException, so I should really remove lines 512 and 513.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
acoburn: @dbernstein is exactly correct here -- we don't want to introduce a JCR dependency here. Since the `BinaryStoreException` inherits from `jcr.RepositoryException`, which inherits from `java.lang.Exception`, I think the `catch (Exception)` will need to stay.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -504,15 +507,44 @@ public Response createObject(@HeaderParam("Content-Disposition") final ContentDi
...
+        }
+
+        if (throwable.getCause() != null) {
+            checkForInsufficientStorageException(rootThrowable, throwable.getCause());
+        } else
+
+        if (rootThrowable instanceof InvalidChecksumException) {
+            throw (InvalidChecksumException) rootThrowable;
+        } else if (rootThrowable instanceof RuntimeException) {
+            throw (RuntimeException) rootThrowable;
```
acoburn: Are these casts necessary? Can't you just do: `throw rootThrowable;`?

dbernstein: It is necessary for compilation if that's what you're asking.

acoburn:
@dbernstein can you please rebase this PR off of the current master?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -506,13 +509,43 @@ public Response createObject(@HeaderParam("Content-Disposition") final ContentDi
...
 
         LOGGER.debug("Finished creating resource with path: {}", newObjectPath);
         return createUpdateResponse(resource, true);
     }
 
     /**
+     * @param rootThrowable The original throwable
+     * @param throwable The throwable under direct scrutiny.
+     * @throws RepositoryRuntimeException
+     * @throws InsufficientStorageException
```
acoburn: It is not necessary to include runtime exceptions in javadocs (RepositoryRuntimeException and InsufficientStorageException)

acoburn:
@dbernstein just a note here, but your rebasing effort wasn't the proper way to do this. You will notice that this pull request now involves ~1,400 lines of code added, and it adds in commits from others, as well as duplicate commits from you.

When rebasing from master, this is the way I do things (I typically have a remote called `upstream` set to `https://github.com/fcrepo4/fcrepo4.git` and a remote called `origin` set to `https://github.com/acoburn/fcrepo4.git`:

    git fetch upstream
    git rebase upstream/master
    # resolve any conflicts, i.e.
    vim path/to/conflicting/file.java
    git add path/to/conflicting/file.java
    git rebase --continue
    # once the conflicts, if any, have been resolved, push it back to origin
    git push --force origin fcrepo-branchnumber

@awoods may have some suggestions about what to do at this point.

awoods:
Locally, I have @dbernstein's PR-1115 branch prior to this rebase. I will go ahead and do the rebase and submit a new PR.

awoods:
Superceded by: https://github.com/fcrepo4/fcrepo4/pull/1122

dbernstein:
Ack! Sorry Aaron + Andrew.  I had Andrew walk me through the flow you described above.  But I ran into trouble at the last step trying to push my changes back into the branch.  I was wary of using --force not being confident that doing so would wreak havoc on the pull request.

PR #1116, opened 2016-10-11
Return the Fedora version number in the Server header

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1774

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -122,7 +122,6 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
     public static final MediaType MESSAGE_EXTERNAL_BODY = MediaType.valueOf("message/external-body");
 
     @Context protected Request request;
-    @Context protected HttpServletResponse servletResponse;
 
     @Inject
     @Optional
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
@@ -34,12 +34,17 @@ import org.fcrepo.kernel.api.models.Tombstone;
 import org.slf4j.Logger;
 
 import java.security.Principal;
+import javax.annotation.PostConstruct;
 import javax.inject.Inject;
+import javax.servlet.http.HttpServletResponse;
+
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.SecurityContext;
 import javax.ws.rs.core.UriInfo;
 
+import static org.fcrepo.http.commons.domain.ServerHeader.SERVER;
+import static org.fcrepo.http.commons.domain.ServerHeader.serverVersion;
 import static org.fcrepo.kernel.api.observer.OptionalValues.BASE_URL;
 import static org.fcrepo.kernel.api.observer.OptionalValues.USER_AGENT;
 import static org.slf4j.LoggerFactory.getLogger;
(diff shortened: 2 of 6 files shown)


ajs6f:
This looks okay to me but Travis is failing?

escowles:
I spun this up locally, and the tests all passed, but I didn't see the updated Server header in either Tomcat or using `mvn jetty:run`.

acoburn:
@escowles I also saw that when using jetty:run the server header is overwritten by jetty. In a full jetty environment, however, I saw the server header come through (provided that the request was for /rest/*). It also worked for me with a Tomcat 7 deployment -- which version of tomcat were you using?

escowles:
I was using Tomcat 8.0.30 — I just downloaded 7.0.72 and I see the FedoraCommons Server header with that version.  I downloaded the most recent (8.0.38), and it continues to send `Coyote/1.1`.

The [docs](https://tomcat.apache.org/tomcat-8.0-doc/config/http.html#Common_Attributes) say that it should use the header set by the application, but it can be overridden by setting a `server` attribute on the Connector.  I was able to do that, though it would obviously be better if the application can set it.

acoburn:
Honestly, I have no idea why the tests fail on Travis. I have been entirely unable to get the tests to fail on my local machine (despite repeated attempts -- like ~25). Do they fail for everyone else?

I was a little reluctant to push the `@Context HttpServletResponse` into the `FedoraBaseResource` -- I'm assuming that's the cause of the error. It is also possible to implement this with the `@Context HttpServletResponse` left in the `ContentExposingResouce` class -- it will require more code and (arguably) more code duplication, but if these tests keep failing, that may be the only way to approach this.

mikedurbin:
I get errors, but only when I run all the integration tests in fcrepo-http-api.  (2 out of 2 times they appeared when running all the tests, they didn't show up when I ran the failing tests in isolation).  I'll get back to you when I know the cause...

awoods:
The build fails for me in apparently the same fashion as Travis.

acoburn:
@awoods: it would be helpful if you could run the `fcrepo-http-api` integration tests with DEBUG mode turned on (and put the results in a gist).

mikedurbin:
@acoburn, here's a snippet... it's hard to tell if I included too much... but it includes the information about a 500 response to testInvalidVersionReversion...
https://gist.github.com/mikedurbin/172650d3427de5b8a07bd04ab42d8992

acoburn:
@mikedurbin thanks! I can definitely work with that.

acoburn:
I was never able to reproduce the error that everyone else seemed to get (I tried two flavors of linux and OS X), but with some minor refactoring of the `FedoraVersions` class, this now passes the travis test. It seems that the error had something to do with the `PostConstruct` annotation on the `FedoraVersions` class.

whikloj:
Maybe I am doing something wrong, but I also can't see the new Server header in Tomcat. 
```
vagrant@fedora4:/var/lib/tomcat7/webapps$ curl -i -G http://localhost:8080/fcrepo/rest/test
HTTP/1.1 200 OK
Server: Apache-Coyote/1.1
ETag: W/"b426caaaf6690672270453bcbb14cd63ec62c659"
Last-Modified: Fri, 04 Nov 2016 21:16:37 GMT
Link: <http://www.w3.org/ns/ldp#Resource>;rel="type"
Link: <http://www.w3.org/ns/ldp#Container>;rel="type"
Link: <http://www.w3.org/ns/ldp#BasicContainer>;rel="type"
Accept-Patch: application/sparql-update
Accept-Post: text/turtle,text/rdf+n3,text/n3,application/rdf+xml,application/n-triples,application/ld+json,multipart/form-data,application/sparql-update
Allow: MOVE,COPY,DELETE,POST,HEAD,GET,PUT,PATCH,OPTIONS
Preference-Applied: return=representation
Vary: Prefer
Vary: Accept, Range, Accept-Encoding, Accept-Language
Content-Type: text/turtle
Transfer-Encoding: chunked
Date: Tue, 08 Nov 2016 14:43:37 GMT

```

acoburn:
I am going to close this PR. If someone else wants to revive it at some point, they can feel free.

PR #1117, opened 2016-10-12
Set the tombstone response type as text/plain

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-2123


acoburn:
This does not change the serialization of tombstone resources, but it does change the content-type to reflect the actual response body.

PR #1120, opened 2016-10-12
Add warning that the backup and restore endpoints will be externalized

acoburn:
Addresses: https://jira.duraspace.org/browse/FCREPO-2044

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/repository/FedoraRepositoryBackup.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/repository/FedoraRepositoryBackup.java
@@ -107,7 +107,8 @@ public class FedoraRepositoryBackup extends AbstractResource {
             throw new WebApplicationException(serverError().entity(output).build());
 
         }
-        return ok().header("Warning", "This endpoint is deprecated and will be removed in a future Fedora release")
+        return ok()
+            .header("Warning", "This endpoint will be moving to an extension module in a future release of Fedora")
             .entity(backupDirectory.getCanonicalPath()).build();
     }
 }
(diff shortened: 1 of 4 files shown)


awoods:
Although /fcr:backup and /fcr:restore are not a part of the Fedora Specification, they serve a very real and practical purpose in the community implementation. I would rather we not deprecate them until we have a time-tested alternative in the import-export utility.

acoburn:
@awoods: I think the better approach is to have a command-line utility that _directly_ invokes the methods on [this service](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/RepositoryServiceImpl.java). Wrapping that service in an HTTP interface makes very little sense to me.

acoburn:
@awoods: or, put another way, once someone starts to develop a non-modeshape implementation (while re-using the existing API classes), are you saying that that implementation _must_ implement a backup and restore endpoint?

awoods:
A command-line tool that interacts with the repository directly has its appeal, but will likely suffer from locking issues with the webapp and the command-line tool trying to access the underlying repository at the same time.

Regarding other implementations of the Fedora API, absolutely not. It seems perfectly reasonable to return a "501 Not Implemented" or another suitable response.

ajs6f:
> Although /fcr:backup and /fcr:restore are not a part of the Fedora Specification, they serve a very real and practical purpose in the community implementation. I would rather we not deprecate them until we have a time-tested alternative in the import-export utility.

Please give specific, concrete, preferably quantifiable conditions that must be reached. Otherwise this is not a real plan, it's just kicking a can down an increasingly narrow alley.

acoburn:
I don't see why we couldn't move the backup/restore functionality into an extension module not unlike how the old `fcrepo-transform` module worked -- i.e. by adding new endpoints to the core API.

awoods:
@acoburn, I was thinking the same thing. Externalizing backup/restore into fcrepo-webapp-plus vs deprecating it makes a lot of sense.

ajs6f:
👍 to moving this to `fcrepo-exts` to begin with. We can move it farther out from there.

acoburn:
@awoods: as for process: should we add a warning header now (i.e. "This endpoint will be moving to an extension module in a future release of Fedora") or simply move the related code into an extension module without any associated warning header? I am inclined to think that a warning header would be appropriate.

awoods:
:+1: to Warning header now.

PR #1121, opened 2016-10-13
Clean up RdfLexicon

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-2009

Mostly, this moves JCR-related properties into the modeshape layer. It also removes tests from the HTTP layer, testing for the absence of the JCR properties. Those tests do not belong in the HTTP layer.

This also removes non-existent or unused properties and classes from the kernel-api, such as LDP properties that are not part of the published specification. Also, rather than defining common (RDF, RDFS, SKOS, DC, DCTerms) terms here, we start using Jena's vocabulary package.


ajs6f:
Using Jena's vocab classes is concise, but won't it make it harder to move to Commons RDF if we do want to do that?

acoburn:
@ajs6f that is a good point. Here is a counter-argument: if we use CommonsRDF, we will in all likelihood still use Jena in fcrepo-kernel-modeshape. If you look at the code base (excluding tests and excluding the modeshape module -- given this PR), the `org.apache.jena.vocabulary` is used in the following locations:

1) The ViewHelpers class. I hold no brief for our use of Velocity templates, and I do not see the use of `jena.vocabulary` to, in any way, block future use of CommonsRDF here.

2) Three times in the `fcrepo-event-serialization` module. Those can simply be replaced with an appropriate string, if necessary

3) Twice in `fcrepo-http-api`, the term `RDF::type` is used.

I don't see any of these being impediments to our use of CommonsRDF.

ajs6f:
Yeah, that's pretty darn minor. Okay, no prob then.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
```diff
@@ -180,11 +149,11 @@
             createProperty(LDP_NAMESPACE + "constrainedBy");
     public static final Property MEMBER_SUBJECT =
             createProperty(LDP_NAMESPACE + "MemberSubject");
+    public static final Property INBOX =
```
awoods: We should hold off on add the `INBOX` property until it is used. Currently, it is not referenced in the codebase.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
```diff
@@ -180,11 +149,11 @@
...
+            createProperty(LDP_NAMESPACE + "inbox");
 
-    public static final Set<Property> ldpProperties = of(PAGE, PAGE_OF,
-            FIRST_PAGE, NEXT_PAGE, CONTAINS, LDP_MEMBER);
-    public static final Set<Resource> ldpResources = of(CONTAINER,
-            DIRECT_CONTAINER);
+    public static final Set<Property> ldpProperties = of(CONTAINS, LDP_MEMBER);
```
awoods: This can be `private`.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
```diff
@@ -180,11 +149,11 @@
...
+    public static final Set<Resource> ldpResources = of(CONTAINER, DIRECT_CONTAINER);
```
awoods: `ldpResources` is not referenced in the codebase... remove?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/RdfJcrLexicon.java
```diff
@@ -0,0 +1,68 @@
...
+     */
+    public static final String JCR_NAMESPACE = "http://www.jcp.org/jcr/1.0";
+
+    public static final String MODE_NAMESPACE = "http://www.modeshape.org/1.0";
+
+    public static final String MIX_NAMESPACE = "http://www.jcp.org/jcr/mix/1.0";
+
+    public static final String JCR_NT_NAMESPACE = "http://www.jcp.org/jcr/nt/1.0";
+
+    public static final Set<String> managedNamespaces = of(
```
awoods: `managedNamespaces` has not references... remove?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/RdfJcrLexicon.java
```diff
@@ -0,0 +1,68 @@
...
+    public static final Property HAS_NODE_TYPE =
+            createProperty(REPOSITORY_NAMESPACE + "hasNodeType");
+    public static final Property HAS_MIXIN_TYPE =
+            createProperty(REPOSITORY_NAMESPACE + "mixinTypes");
+
+    public static final Set<Property> jcrProperties = of(
+            HAS_PRIMARY_IDENTIFIER, HAS_PRIMARY_TYPE, HAS_NODE_TYPE,
+            HAS_MIXIN_TYPE);
+
+    private static Predicate<Property> hasJcrNamespace =
```
awoods: `hasJcrNamespace` has no references... remove?

PR #1123, opened 2016-10-13
Adieu jcr.Session, it's been nice knowing you

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-1868


ajs6f:
Lot of line changes here, but is the basic story JCR `Session` => `FedoraSession`?

acoburn:
Yes, and in addition to: `Session` => `FedoraSession`
there is also: `Repository` => `FedoraRepository` and
`TransactionService` => `BatchService` (which is just a wrapper around a `Map` of `FedoraSession`s)

This changes a lot of code in the test suites b/c many of the tests mock up the entire JCR subsystem, even when they mocks don't have to (that should change, but not here). The real substantive change in the HTTP layers involves the `http.commons.session.SessionFactory` and `http.commons.session.SessionProvider`. This allows the Spring configuration to remain exactly the same.

fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/AbstractJmsIT.java
```diff
@@ -119,61 +125,68 @@ public void testIngestion() throws RepositoryException {
 
         LOGGER.debug("Expecting a {} event", RESOURCE_CREATION.getType());
 
-        final Session session = repository.login();
+        final FedoraSession session = repository.login();
+        session.addSessionData(BASE_URL, TEST_BASE_URL);
```
awoods: In many of the tests in this class, it is not obvious to me why the addition of session-data is required. Do you have any insights?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -280,13 +277,9 @@ public Response deleteObject() {
 
         try {
             resource().delete();
-
-            try {
-                session.save();
-            } catch (final RepositoryException e) {
-                throw new RepositoryRuntimeException(e);
+            if (!batchService.exists(session.getId(), getUserPrincipal())) {
```
acoburn: The initial implementation of the `FedoraSession` interface held state about such things -- whether it was part of a batch operation, etc, but it was decided to remove that. For a single-node implementation, I can see how holding that state would be very useful and fairly trivial to implement, but if we plan to have a multi-node Fedora implementation at some future point, holding such state becomes really challenging. Which is to say, by pushing that logic into a check on the `BatchService` object, this opens the door to much more robust distributed implementations.

To that I would add that the extensive amount of [reflection](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/TxAwareSession.java#L65-L82) used in the current implementation to accomplish what you describe seemed weird.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/SimpleObserver.java
```diff
@@ -184,7 +185,7 @@
     @PostConstruct
     public void buildListener() throws RepositoryException {
         LOGGER.debug("Constructing an observer for JCR events...");
-        session = repository.login();
+        session = getJcrSession(repository.login());
         session.getWorkspace().getObservationManager()
                 .addEventListener(this, EVENT_TYPES, "/", true, null, null, false);
         session.save();
```
acoburn: Not really. The short explanation is that the jcr-based event machinery is very closely tied to jcr Sessions, so the event listener needs direct access to a `jcr.Session`

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraNodes.java
```diff
@@ -125,7 +124,9 @@ public Response copyObject(@HeaderParam("Destination") final String destinationU
             LOGGER.info("Copy from '{}' to '{}'", source, destination);
             nodeService.copyObject(session, source, destination);
 
-            session.save();
+            if (!batchService.exists(session.getId(), getUserPrincipal())) {
```
acoburn: My response here is the same as above.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
awoods: Thanks for highlighting the distributed scenario. As for the reflection, that weirdness held the beauty of encapsulating the logic of whether to `Session.save` or not.

fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/AbstractJmsIT.java
acoburn: This has to do with the following: at present, we are passing data from the HTTP layer into the event system, particularly `User-Agent` and `Host` header values. Those values are passed into the event system via the current `jcr.Session`. In this PR, we replace the `jcr.Session` with a `FedoraSession`, but the mechanism is the same -- this user data is passed into the FedoraSession in the HTTP layers and then retrieved by the eventing machinery.

In these integration tests, however, there is no HTTP layer, so I'm adding that data manually. I _could_ omit that data and a warning would be issued (given that no `baseUrl` or `userAgent` values are available), but I wanted to avoid those warnings.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/observer/SimpleObserver.java
ajs6f: @awoods I think this is kosher-- there is no conceivable way any outside-the-kernel-impl process could ever become aware of this session, and inside the kernel-impl is exactly where we expect to find platform-bound constructs like this.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
awoods: @acoburn , this comment may have slipped through the cracks:
https://github.com/fcrepo4/fcrepo4/pull/1123#discussion_r83494182

acoburn: @awoods: yes, we could factor that into a convenience method.

acoburn: That makes perfect sense to me

acoburn: ...by which I mean: it would make sense to make this more obvious to future developers

acoburn: @awoods I have been thinking about this, and I think you make an excellent point: that much of this could be handled via runtime polymorphism. This would eliminate the awkward `if` construct, and make the implementation significantly cleaner. This will require some changes to the underlying `FedoraSession` interface, but I will add that change in a forthcoming commit.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraNodes.java
acoburn: As indicated above, with runtime polymorphism, this awkward construct will go away.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
awoods: @acoburn, thanks for giving this some thought. Are you planning on the "forthcoming commit" to be included in this PR or a subsequent PR?

acoburn: @awoods I was planning for that commit to be part of this PR.

awoods: :+1:

acoburn:
The latest commit makes a number of changes:

1. The `FedoraSession` interface now has a `save` and `commit` method. It also has a `expire` and `rollback` method. While this may appear redundant, it is fundamentally important for the runtime polymorphism. You will notice that a second implementation of `FedoraSession` now exists: `BatchSessionImpl`, which extends `FedoraSessionImpl`. Similar to how the current implementation works, the `save` and `expire` methods in `BatchSessionImpl` are no-op methods, and the actual `save` doesn't happen until `commit` is called. This is basically how the `TxAwareSession` works, just without all the `instanceof` checks and the downcasting.

2. The `FedoraSession` also has a method called `getBatchId`, which is like `getId` except that it is wrapped in an `Optional`. This is necessary for the `HttpResourceConverter` class in `fcrepo-http-commons`

3. This commit eliminates most of the `BatchService::exists` checks from earlier, which is in line with the suggestion from @awoods.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraSession.java
```diff
@@ -36,9 +38,26 @@
...
+     * Note: use of the save method in the context of a batch operation may not persist
+     * changes to the repository until the commit method has been issued.
+     */
+    void save();
+
+    /**
      * Commit any batch operations
      */
-    void commit();
+    default void commit() {
```
acoburn: Perhaps we could remove the `commit` method and replace it with something like `finalize` or `persist`? Either way, using a default method is wrong here.

ajs6f: Yeah, I already made two; in one case, stop making batches and "sessions" the same supertype. No polymorphism. Completely independent types. In the other (preferably, to my mind), do not have two different types at all. After all, the distinction between them is based on "how many HTTP requests built this changeset? one, or more than one?" That's purely an HTTP-layer-concern. At the level of this code, we should just have one "session"y thing. How many HTTP requests built it up is of no concern down here.

ajs6f: > Either way, using a default method is wrong here.

It _is_ weird/wrong, but you are being lead to do it because the type system is wrong. We have to get that right and the impl details will fit together properly, easily.

acoburn:
@awoods, in discussing this with @ajs6f, it appears that there are differing opinions on whether the HTTP layers should rely on polymorphism. What that means in a practical sense is that there are two possible ways forward for this effort:

1) Manage batch operations in the `FedoraSession` type (this is the current state of the PR) and keep the HTTP layer mostly independent of managing batch operations vs. single operations.

2) Manage batch operations in the HTTP layer (this is the state of the PR at https://github.com/fcrepo4/fcrepo4/pull/1123/commits/adcf0ddd939186c068b582bcc31da0e79cdbc54f)

My preference is for the second option (i.e. reverting the last commits), but mostly I want to see the JCR notions out of the kernel.

acoburn:
@awoods: in further consideration of the two options already outlined (1. manage batch operations in the modeshape kernel or 2. manage batch operations in the HTTP layer), I would like to _strongly_ advocate for the second option.

We could still achieve the sort of clean design you were hinting at by introducing a new class in `fcrepo-http-commons` called `HTTPSession` (or something similar). That class would, effectively, wrap around a `FedoraSession` and control how batch operations function (i.e. at what point changes are persisted to the repository). I would also advocate for _not_ implementing that as part of this PR. This PR is already a monster, and I don't want to make it any larger.

I would add that even this implementation, which pushes the batch handling into the kernel still has some leakage of batch operation management in the HTTP layer. I am not convinced that it is possible to entirely encapsulate batch handling in the present structure.

Unless I hear significant objections to this, I will plan to roll back the recent commits that push batch operation handling into the kernel layer so that we can pursue the second option outlined here and above.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
acoburn: As stated below, I have rolled back the commits that implement this. There are ways to encapsulate the behavior of batch sessions, but that will be for a separate effort.

ajs6f:
+1 to no notion of "batch" in the kernel layers.

awoods:
:+1: Let's move forward with this PR by making related updates to downstream modules... then circle back on this topic of encapsulating the `batchService.exists()` logic.

whikloj:
Sorry, to jump in late. This may be outside the scope of this work (as it is already huge), but I was wondering about the FedoraAuthorizationDelegates use of jcr sessions? 

Will that be changed to the new FedoraSession in later work (I apologize if this seems like a criticism, it is not... this is a huge undertaking and massive 👏  👏  to @acoburn)

The pull against WebAC got me thinking about the core interface here:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/FedoraAuthorizationDelegate.java#L83

acoburn:
This PR does not change how the authentication delegate classes work--those are all tightly bound to MODE, so they are outside the scope of this.

PR #1124, opened 2016-10-14
Fixed a bug that cause intermitted test failures.

mikedurbin:
fixes https://jira.duraspace.org/browse/FCREPO-2270


acoburn:
I tested this on my system, where previously the `fcrepo-http-api` module tests failed about 50% of the time. After running the test suite 40 times with no failures, I am confident that this change fixes the issue.

awoods:
:+1:

PR #1127, opened 2016-10-25
Prevents whitespace characters from being used in version labels.

dbernstein:
Also adds test coverage for invalid version labels.

Resolves: https://jira.duraspace.org/browse/FCREPO-1984:

### DIFF
--- a/fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/services/VersionServiceImplTest.java
+++ b/fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/services/VersionServiceImplTest.java
@@ -37,6 +37,8 @@ import org.fcrepo.kernel.api.FedoraSession;
 import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
 import org.fcrepo.kernel.api.services.VersionService;
 import org.fcrepo.kernel.modeshape.FedoraSessionImpl;
+
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -178,6 +180,25 @@ public class VersionServiceImplTest {
         final Node unversionedNode = mockSession.getNode(EXAMPLE_UNVERSIONED_PATH);
         verify(unversionedNode).isNodeType(VERSIONABLE);
         verify(unversionedNode).addMixin(VERSIONABLE);
+    }
+
+    @Test
+    public void testCreateWithInvalidVersionLabels() throws RepositoryException {
+        // [\\s~#@*+%{}<>\\[\\]|\"^]
+        final String[] invalidChars = { " ", "~", "#", "@", "*", "+", "%", "{", "}",
+            "<", ">", "[", "]", "|", "\"", "^" };
+        for (String s : invalidChars) {
+            testLabel(s);
+        }
+    }
+
+    private void testLabel(final String label) {
+        try {
+            testObj.createVersion(testSession, EXAMPLE_UNVERSIONED_PATH, label);
+            Assert.fail("The previous call should have failed.");
+        } catch (Exception ex) {
+            Assert.assertTrue(ex.getMessage().contains("Invalid label"));
+        }
     }
 
 }
--- a/fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/services/VersionServiceImplTest.java
+++ b/fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/services/VersionServiceImplTest.java
@@ -184,7 +184,11 @@ public class VersionServiceImplTest {
         testCreateWithValidLabel("version v1234");
     }
 
-
+    /**
+     * This test designed to be pass so long as no exception is thrown.
+     * @param label
+     * @throws RepositoryException
+     */
     private void testCreateWithValidLabel(final String label) throws RepositoryException {
         final VersionManager mockVersionManager = mock(VersionManager.class);
         final VersionHistory mockHistory = mock(VersionHistory.class);
--- a/fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/services/VersionServiceImplTest.java
+++ b/fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/services/VersionServiceImplTest.java
@@ -18,6 +18,7 @@
 package org.fcrepo.kernel.modeshape.services;
 
 import static org.fcrepo.kernel.modeshape.FedoraJcrConstants.VERSIONABLE;
+import static org.junit.Assert.fail;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
@@ -38,7 +39,6 @@ import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
 import org.fcrepo.kernel.api.services.VersionService;
 import org.fcrepo.kernel.modeshape.FedoraSessionImpl;
 
-import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -171,11 +171,26 @@ public class VersionServiceImplTest {
 
     @Test
     public void testMixinCreationWhenExplicitlyVersioning() throws RepositoryException {
+        testCreateWithValidLabel("LABEL");
+    }
+
+    @Test
+    public void testCreateWithLabelv001() throws RepositoryException {
+        testCreateWithValidLabel("v0.0.1");
+    }
+
+    @Test
+    public void testCreateWithLabelWithSpace() throws RepositoryException {
+        testCreateWithValidLabel("version v1234");
+    }
+
+
+    private void testCreateWithValidLabel(final String label) throws RepositoryException {
         final VersionManager mockVersionManager = mock(VersionManager.class);
         final VersionHistory mockHistory = mock(VersionHistory.class);
         when(mockWorkspace.getVersionManager()).thenReturn(mockVersionManager);
         when(mockVersionManager.getVersionHistory(EXAMPLE_UNVERSIONED_PATH)).thenReturn(mockHistory);
-        testObj.createVersion(testSession, EXAMPLE_UNVERSIONED_PATH, "LABEL");
+        testObj.createVersion(testSession, EXAMPLE_UNVERSIONED_PATH, label);
 
         final Node unversionedNode = mockSession.getNode(EXAMPLE_UNVERSIONED_PATH);
         verify(unversionedNode).isNodeType(VERSIONABLE);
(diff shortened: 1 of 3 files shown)


fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/services/VersionServiceImplTest.java
```diff
@@ -37,6 +37,8 @@
 import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
 import org.fcrepo.kernel.api.services.VersionService;
 import org.fcrepo.kernel.modeshape.FedoraSessionImpl;
+
+import org.junit.Assert;
```
acoburn: use static imports for the Assert class.

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/services/VersionServiceImplTest.java
```diff
@@ -180,4 +182,23 @@ public void testMixinCreationWhenExplicitlyVersioning() throws RepositoryExcepti
...
+            "<", ">", "[", "]", "|", "\"", "^" };
+        for (String s : invalidChars) {
+            testLabel(s);
+        }
+    }
+
+    private void testLabel(final String label) {
+        try {
+            testObj.createVersion(testSession, EXAMPLE_UNVERSIONED_PATH, label);
+            Assert.fail("The previous call should have failed.");
```
acoburn: static import for `fail`

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/services/VersionServiceImplTest.java
```diff
@@ -180,4 +182,23 @@ public void testMixinCreationWhenExplicitlyVersioning() throws RepositoryExcepti
...
+        } catch (Exception ex) {
+            Assert.assertTrue(ex.getMessage().contains("Invalid label"));
```
acoburn: static import for `assertTrue`

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/services/VersionServiceImplTest.java
ajs6f: That's what I do, but some people do just `extend Assert`, which I don't think is wrong. The important thing is to tighten it up.

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/services/VersionServiceImplTest.java
```diff
@@ -180,4 +182,23 @@ public void testMixinCreationWhenExplicitlyVersioning() throws RepositoryExcepti
...
+    public void testCreateWithInvalidVersionLabels() throws RepositoryException {
+        // [\\s~#@*+%{}<>\\[\\]|\"^]
+        final String[] invalidChars = { " ", "~", "#", "@", "*", "+", "%", "{", "}",
+            "<", ">", "[", "]", "|", "\"", "^" };
+        for (String s : invalidChars) {
+            testLabel(s);
+        }
+    }
+
+    private void testLabel(final String label) {
```
ajs6f: Is there no way we can `expect` a specific exception here? Testing for the message in a `Throwable` is often brittle.

ajs6f:
Is a version label truly a label (never dereferenced) or does it appear as a segment of an URI?

acoburn:
@ajs6f the version label becomes part of a URI -- does the versioning spec (or the memento spec) have anything to say about whether spaces are legitimate parts of a timegate URI?

ajs6f:
Spaces are no good in the path part of any URI. They should be %encoded..

acoburn:
@ajs6f I agree, and there is definitely some inconsistent behavior w/r/t version labels and spaces. For instance:

    curl -XPOST -H"Slug: new version name" http://localhost:8080/fcrepo/rest/test/fcr:versions -i

Returns the correct header:

    HTTP/1.1 201 Created
    Location: http://localhost:8080/fcrepo/rest/test/fcr:versions/new%20version%20name

But when retrieving the version triples, they aren't %encoded:

    curl http://localhost:8080/fcrepo/rest/test/fcr:versions

    @prefix fedora:  <http://fedora.info/definitions/v4/repository#> .
    <http://localhost:8080/fcrepo/rest/test/fcr:versions/new version name>
        fedora:hasVersionLabel  "new version name" .

As a note -- this response doesn't respect the single-subject rule. I don't know if the behavior here is correct or not, but it does make me wonder.

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/services/VersionServiceImplTest.java
```diff
@@ -180,4 +182,23 @@ public void testMixinCreationWhenExplicitlyVersioning() throws RepositoryExcepti
...
+        try {
+            testObj.createVersion(testSession, EXAMPLE_UNVERSIONED_PATH, label);
+            Assert.fail("The previous call should have failed.");
+        } catch (Exception ex) {
```
ajs6f: Yes, as suggested above, use `@Test(expected=)`.

ajs6f:
I think you're correct about the RDF being wrong, @acoburn , but I no longer care. I would rather work on the spec (using Memento) and clear up such matters that way.

acoburn:
FWIW, I'm completely fine with prohibiting whitespace in version labels, but as @ajs6f points out, any change in behavior (prohibiting whitespace in labels, changing response triples) ought to be based on what is defined in the relevant specification. Absent that, I'd want to be very careful about making such changes.

awoods:
@acoburn / @ajs6f : If we do not want to make any substantive changes in the absence of a spec, we should at a minimum make changes that prevent the exception that is reported in the ticket:
https://jira.duraspace.org/browse/FCREPO-1984

Updating the HTML, as suggested, may be the most direct way of doing that... although fixing the non-URL-encoded RDF also seems reasonable, potentially as a separate ticket.

escowles:
I agree with @acoburn and @ajs6f that any changes to the rules for version labels should be based on the spec.  Beyond that, I think it's a bad idea to introduce breaking changes in the REST API to fix a bug in the HTML views.  The HTML UI is a convenience, and we should never break the REST API to fix its bugs.

ajs6f:
What I said was what @escowles said. Not "do nothing", but "do the minimum".

acoburn:
@escowles++

dbernstein:
Thank you for the feedback everyone.    I'm not sure I understand why we would prevent adding a version label with a space in the UI but not via the rest api.   The UI breaks not due to the input, but rather the output.   Preventing people from entering spaces in the version label through the UI just masks the problem and assumes that UI users will not be using the API.   Am I missing something? 

Anyway the root cause can be traced back to how the local name of the URI is determined.

org.apache.jena.rdf.model.impl.Util.splitNamespaceXML(uri).  Even when the label is url encoded,  the local name split point is not what we are expecting.

When  String uri = "http://example.com/fcr:versions/version%201"
both 
uri.substring(Util.splitNamespaceXML(uri)) return "" 
while  SplitIRI.localname(uri)   returns "version%201" as we would expect. 

The reason it returns "" in this case is that the algorithm treats any character that is NOT a NameChar as a split point.  
According to the docs: 
NameChar           ::= NameStartChar | "-" | "." | [0-9] | #xB7 |
                                [#x0300-#x036F] | [#x203F-#x2040]
That being the case,  anything %-encoded characters will be treated a split point. 

My conclusion:  Spaces in URIs may cause trouble.  Spaces in the version label should never be allowed.

escowles:
@dbernstein: the rationale is this: we've gotten a lot of feedback that we should reduce the frequency of releases with breaking changes.  The API spec should help address that, and absent guidance from the spec that the current behavior is wrong, I'm in favor of keeping the current behavior (allowing spaces in version labels).  So I wasn't suggesting that we prevent creating labels with spaces in the UI — I was suggesting that the UI be updated to handle them correctly.

I agree that spaces in URIs can cause trouble — so I'm all in favor of rigorously testing the behavior to make sure we're handling them correctly.  I would also be fine with fixing the Jena namespace-parsing behavior if it can't handle spaces, and/or implementing our own namespace-parsing (we are talking about just a couple lines of code here).  That all seems preferable to a breaking change to me.

dbernstein:
@escowles :  thanks for the clear explanation.  I'm not sure that we're on the same page yet though.  The UI is handling version label creation process correctly as far as I can tell.   The problem is that allowing spaces in the version labels does not conform to the "XML QName rules" which appear to be faithfully implemented by jena. 

So I'm okay with looking for a way to by pass the jena namespace parsing.  I hope I'm not beating a dead horse here - just want to make sure I'm clear on what's what.

acoburn:
Just to be clear, it's not the _spaces_ that are an issue here. It is entirely possible in the HTML UI to create a version label with spaces right now without errors. In fact, the following version labels are all just fine:

"version string"
"version string2"
"version 2string"
"version 2 string"

What isn't fine is a label that ends with a number separated by spaces:

"version 2"

What also isn't fine is a label that is entirely numerical (and excluding whitespace won't help this at all):

"2"

IMO, the "whitespace" issue is a red herring here -- what needs to be handled is numerical labels, preferably by throwing an exception on updates. The HTML UI can enforce this with a little javascript (i.e. `if (label.match(/(?:^|\s)\d+$/)) { ... }`), but _whitespace is not the issue here_.

escowles:
Is this another instance of the XML QName rules being more restrictive than RDF 1.1 rules, like we saw in https://jira.duraspace.org/browse/FCREPO-1799?

IIRC, the decision that came out of that was that we wouldn't enforce XML compatibility, but instead would log a warning letting people know that their data may be incompatible with RDF/XML.

dbernstein:
@acoburn  - good point there re the red herring.  If we prevent users from creating version labels with numbers only  or labels ending in a number separated by a space, then the rendering will be fixed for now.  That said,  shouldn't I fix it server-side? If an api user adds a numeric version label,  the UI will still broken be. Or am I missing something?

acoburn:
@escowles I _think_ it is a slightly different issue, but I could be wrong on that (I'm encountering the error w/ TTL serializations), and it seems that the source of the error is here: https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/VersionsRdfContext.java#L97-L98

In fact, I'd be inclined to suggest that @dbernstein focus his efforts there. It _may_ be that adding appropriate %encoding will address most cases -- there's still the issue of a label being entirely numerical -- I'm not sure what to suggest there. One option would be to throw an exception; another would be to modify the value of the `Slug` header to make it into a "valid label" -- (IIRC, Slug is just advisory, and a server can ignore it and still be in line with the HTTP spec).

acoburn:
@dbernstein I'd try to fix this server-side so long as it doesn't change existing behavior (i.e. the non-broken part of the existing behavior). If the server rejects invalid labels with an exception (4xx response), then the HTML UI (and any REST client for that matter) will see some sort of appropriate error message.

dbernstein:
@acoburn :  okay.  So here's what I propose: 

I will: 
1.  create a new JIRA for preventing invalid version labels from being entered into the system (ie no number only labels or labels ending in space number.
2. work the above newly created ticket by throwing an exception (4xx response + appropriate error message) on adding a slug (personally I like this better than modifying slug header). 
3. ensure that slugs are URL encoded on the way out (ie right here:  https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/VersionsRdfContext.java#L97-L98)

dbernstein:
actually - forget about the new jira.

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/services/VersionServiceImplTest.java
```diff
@@ -169,15 +171,54 @@ public void testRemoveUnknownVersion() throws RepositoryException {
...
+        testCreateWithValidLabel("v0.0.1");
+    }
+
+    @Test
+    public void testCreateWithLabelWithSpace() throws RepositoryException {
+        testCreateWithValidLabel("version v1234");
+    }
+
+
+    private void testCreateWithValidLabel(final String label) throws RepositoryException {
```
ajs6f: One way here is to wrap the lines where an exception might be thrown with `try {...} catch (Exception e) {fail(e.getMessage());}`.

awoods:
Are there any other comments/reservations on this PR other then the [one minor suggestion](https://github.com/fcrepo4/fcrepo4/pull/1127#discussion_r85332647)?
If not, let's move this forward.

escowles:
👍  all of my concerns are addressed.

PR #1129, opened 2016-10-26
Use a new prefix definition for info:fedoraconfig/

acoburn:
Addresses: https://jira.duraspace.org/browse/FCREPO-2247

The goal here is to avoid making an [earlier change](https://github.com/fcrepo4/fcrepo4/commit/48010a20a13645977d4641cb987c939f6e2522b8) non-breaking for users who had previously made use of the `fedoraconfig` prefix.


awoods:
@acoburn, What are your thoughts on the risk of the new namespace prefix `config` clashing with existing use of that same prefix by current Fedora installations?

acoburn:
I am not at all concerned about this. If, however, I were concerned, the only sensible approach would be to remove that namespace declaration altogether, which has other implications.

In fact making any change at all to the namespaces in the CND file has a possibility of conflicting with some existing repository somewhere. But I do not believe that is a legitimate reason to not ever make a change.

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/f4f06ffe37f4229ec3c9414410907ca85c3d3401

PR #1130, opened 2016-10-27
Remove JCR from the kernel-api

acoburn:
Remove jcr-based Version references
Remove miscellaneous JCR references from the kernel-api

Resolves: https://jira.duraspace.org/browse/FCREPO-1867
Resolves: https://jira.duraspace.org/browse/FCREPO-1847

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
@@ -157,9 +157,9 @@ public interface FedoraResource {
     FedoraResource getBaseVersion();
 
     /**
-     * Get a stream of version identifiers associated with this resource
+     * Get a stream of versions associated with this resource
      *
-     * @return version identifiers
+     * @return the versions associated with this resource
      */
     Stream<FedoraVersion> getVersions();
 
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
@@ -24,9 +24,6 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Stream;
 
-import javax.jcr.version.Version;
-import javax.jcr.version.VersionHistory;
-
 import org.apache.jena.rdf.model.Resource;
 
 import org.fcrepo.kernel.api.exception.AccessDeniedException;
@@ -44,7 +41,7 @@ import org.apache.jena.rdf.model.Model;
 public interface FedoraResource {
 
     /**
-     * Get the path to the JCR node
+     * Get the path to the resource
      * @return path
      */
     String getPath();
@@ -121,11 +118,7 @@ public interface FedoraResource {
 
     /**
      * Update the provided properties with a SPARQL Update query. The updated
-     * properties may be serialized to the JCR store.
-     *
-     * After applying the statement, clients SHOULD check the result
-     * of #getDatasetProblems, which may include problems when attempting to
-     * serialize the data to JCR.
+     * properties may be serialized to the persistence layer.
      *
      * @param idTranslator the property of idTranslator
      * @param sparqlUpdateStatement sparql update statement
@@ -156,18 +149,18 @@ public interface FedoraResource {
                          final Set<? extends TripleCategory> contexts);
 
     /**
-     * Get the JCR Base version for the node
+     * Get the base version for the node
      *
      * @return base version
      */
-    public Version getBaseVersion();
+    FedoraResource getBaseVersion();
 
     /**
-     * Get JCR VersionHistory for the node.
+     * Get a stream of version labels associated with this resource
      *
-     * @return version history
+     * @return version labels
      */
-    public VersionHistory getVersionHistory();
+    Stream<String> getVersionLabels();
 
     /**
      * Check if a resource was created in this session
@@ -189,10 +182,7 @@ public interface FedoraResource {
                                 final RdfStream originalTriples) throws MalformedRdfException;
 
     /**
-         * Construct an ETag value from the last modified date and path. JCR has a
-     * mix:etag type, but it only takes into account binary properties. We
-     * actually want whole-object etag data. TODO : construct and store an ETag
-     * value on object modify
+     * Construct an ETag value for the resource.
      *
      * @return constructed etag value
      */
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
@@ -26,11 +26,12 @@ import java.util.stream.Stream;
 
 import org.apache.jena.rdf.model.Resource;
 
+import org.fcrepo.kernel.api.FedoraVersion;
+import org.fcrepo.kernel.api.RdfStream;
+import org.fcrepo.kernel.api.TripleCategory;
 import org.fcrepo.kernel.api.exception.AccessDeniedException;
 import org.fcrepo.kernel.api.exception.MalformedRdfException;
 import org.fcrepo.kernel.api.identifiers.IdentifierConverter;
-import org.fcrepo.kernel.api.TripleCategory;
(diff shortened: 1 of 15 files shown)


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
```diff
@@ -156,18 +149,18 @@ RdfStream getTriples(final IdentifierConverter<Resource, FedoraResource> idTrans
...
+     * Get the base version for the node
      *
      * @return base version
      */
-    public Version getBaseVersion();
+    FedoraResource getBaseVersion();
 
     /**
-     * Get JCR VersionHistory for the node.
+     * Get a stream of version labels associated with this resource
```
ajs6f: Are we still calling these labels just because JCR does, or are they actually identifiers?

acoburn: the labels become part of the identifier.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/VersionsRdfContext.java
```diff
@@ -66,44 +62,27 @@
...
-                        Arrays.stream(labels).map(x -> create(versionSubject, HAS_VERSION_LABEL.asNode(),
-                                createLiteral(x))),
-
-                        Stream.of(create(topic(), HAS_VERSION.asNode(), versionSubject),
-                            create(versionSubject, CREATED_DATE.asNode(),
-                                createTypedLiteral(v.getCreated()).asNode())));
-            }));
+                        = createProperty(topic() + "/" + FCR_VERSIONS + "/" + label).asNode();
+                // TODO - convert this to java.time.* classes once the kernel-api interface changes
+                final Calendar cal = getInstance();
```
ajs6f: Maybe use `Calendar.Builder`?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
ajs6f: I would rather not use the word "label" for that. It's an identifier, no? An identifier for a version within the context of a versioned resource, and it helps compose the identifier of that version outside that context.

acoburn: Actually, we need to review _all_ of the names for these methods. If you'd like, though, I could change the method name to `getVersionIdentifiers`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/VersionsRdfContext.java
acoburn: Sure? I'd like to change all of the `java.util.Date` and `java.util.Calendar` types to the newer `java.time` package in the near term, which is part of https://jira.duraspace.org/browse/FCREPO-1995

ajs6f: It's up to you, whether to spend any more time with this or just leave it be since it won't be around for that long anyway.

acoburn: I'll leave this as it is for now, but @bseeger has kindly volunteered to change the legacy Date classes to the newer Java8 classes, based on joda-time.

acoburn: Something like that. @bseeger's name is on the FCREPO-1995 ticket.

ajs6f: Good by me.

bseeger: I'm currently working on it FCREPO-1995.  Does java.time.Instance (which I'm using) use/reflect joda-time?

bseeger: Nevermind - in doing some reading and I see that joda-time users are encouraged to go to java.time in Java 8. And that's what I'm using.

acoburn: @bseeger: the design of java.time.Instance, etc is based on the joda-time project. It's an example of an excellent external project becoming part of the core JDK.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/VersionsRdfContext.java
```diff
@@ -69,45 +65,29 @@
...
-
-                        Stream.of(create(topic(), HAS_VERSION.asNode(), versionSubject),
-                            create(versionSubject, CREATED_DATE.asNode(),
-                                createTypedLiteral(v.getCreated()).asNode())));
-            }));
+                        = createProperty(topic() + "/" + FCR_VERSIONS + "/" + urlEncode(label)).asNode();
+
+                // TODO - convert this to java.time.* classes once the kernel-api interface changes
+                final Calendar cal = getInstance();
+                cal.setTime(resource.getVersion(label).getCreatedDate());
```
awoods: This logic sets the `cal` to the created date of the resource. What we want, however, is what the previous logic provided: the created date of the version itself.

acoburn: @awoods: you are entirely correct. I have added a new interface (`FedoraVersion`) that encapsulates this information in a new [commit](https://github.com/fcrepo4/fcrepo4/pull/1130/commits/e7ece1380f700786e3fd1bd0cfcb6f0bcfd79fea).

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
```diff
@@ -156,18 +150,18 @@ RdfStream getTriples(final IdentifierConverter<Resource, FedoraResource> idTrans
...
+     * Get the base version for the node
      *
      * @return base version
      */
-    public Version getBaseVersion();
+    FedoraResource getBaseVersion();
 
     /**
-     * Get JCR VersionHistory for the node.
+     * Get a stream of version identifiers associated with this resource
```
awoods: Comment should be updated to reflect the fact that `FedoraVersion`s are returned, not identifiers.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
```diff
@@ -156,18 +150,18 @@ RdfStream getTriples(final IdentifierConverter<Resource, FedoraResource> idTrans
                          final Set<? extends TripleCategory> contexts);
 
     /**
-     * Get the JCR Base version for the node
+     * Get the base version for the node
      *
      * @return base version
      */
-    public Version getBaseVersion();
+    FedoraResource getBaseVersion();
```
awoods: I don't want to hold this PR up on this point, but for symmetry, it is a little odd that `getBaseVersion` returns a `FedoraResource` whereas `getVersions` returns `FedoraVersion`s.

acoburn: I agree. We _really_ need to review the names of these methods.

acoburn: What I'd like to see is these methods renamed to reflect the memento spec. For instance, we might have `getTimegates` instead of `getVersions`, but that's for a separate effort (ideally after the versioning spec has been finalized.

PR #1131, opened 2016-10-28
Fcrepo 2248: 	Adds missing Content-Length header on HEAD request for non binary obj

dbernstein:


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -189,11 +200,33 @@ public Response head() {
 
             // we set the content-type explicitly to avoid content-negotiation from getting in the way
             builder.type(mediaType.toString());
+        } else {
+            final MediaType mediaType = RDFMediaType.TURTLE_TYPE;
+            final RdfStream stream = getResourceAsStream();
```
escowles: I like the other changes in this PR, but I'm worried about the performance implications of this, esp. for resources with many links to other repository children.  Producing the output in order to get its size seems very counter-productive.  What is the point of doing a HEAD request if it isn't fast?

acoburn: I completely agree with @escowles

acoburn: @dbernstein personally, I think we should close the issue as "Won't Fix". IMO, the only viable alternative is to _not_ include a Content-Length header for HEAD requests on containers. When I looked into this ticket earlier, I did not see a way to do that, as Jersey will (helpfully) add a Content-Length header for you.

Even if the performance considerations have been addressed, a HEAD request should, under no circumstances, try to fetch and/or serialize an entire object -- it should only check whether the object exists along with some of the basic LDP types.

dbernstein: @acoburn :  okay.  That all seems reasonable.  In the future I'll double check on IRC if I'm unsure about how to approach an issue before diving in.   That said,  I think the other refactoring s are worthwhile.   I'll pull out the change to the HEAD behavior and reissue another pull request for the other changes.

awoods: I have created the following new ticket:
https://jira.duraspace.org/browse/FCREPO-2281

Let's:
* Close this PR
* Move the majority of the refactoring from here into a new PR against the above ticket
* Add a comment to https://jira.duraspace.org/browse/FCREPO-2248 reflecting the comments in this PR

acoburn:
Closing this, based on discussion. Most of the changes here were part of https://github.com/fcrepo4/fcrepo4/commit/b06bf6bad2346373225f4b45c7275afa3530f6c7

PR #1134, opened 2016-10-29
Disable output content buffering

acoburn:
Resolves: https://jira.duraspace.org/browse/FCREPO-2282

You will notice the difference in the lack of a `Content-Length` header for RdfSource responses. NonRdfSource documents continue to have the `Content-Length` header, since that value is set by the `fcrepo-http-api` code and not the JAX-RS runtime.

Note: ideally, HEAD requests for Containers would not have a `Content-Length` header, but that does not appear to be the case: they still contain a `Content-Length: 0` header.

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/FedoraApplication.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/FedoraApplication.java
@@ -31,6 +31,7 @@ import com.codahale.metrics.MetricRegistry;
 
 import java.util.logging.Logger;
 
+import static org.glassfish.jersey.server.ServerProperties.OUTBOUND_CONTENT_LENGTH_BUFFER;
 import static org.slf4j.LoggerFactory.getLogger;
 
 /**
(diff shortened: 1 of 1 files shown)


ajs6f:
The reason this works is not obvious. Could there be an in-line comment to explain it?

acoburn:
I don't think this actually provides the performance gain I thought it would. Given the presence of a `Content-Length` header, I was under the impression that _all_ content is buffered before being flushed to output, but upon closer inspection, it seems that, for large containers, that is already happening: those containers do not have the `Content-Length` header. And in looking at the Jersey source code, if the complete `OutputStream` doesn't fit into Jersey's internal buffer (8K by default), the content is flushed without the `Content-Length` header. The change I proposed here merely writes that first 8K of data to the output stream without buffering it at all. But that's such a minor "optimization" that I am inclined to close this PR and _not_ make this change. Thoughts?

ajs6f:
-1 to optimizations that are written against Jersey impl code (which could change out from under us).

PR #1141, opened 2016-11-18
Insure 304 responses do not contain an entity body. (master)

emetsger:

Responses from Fedora *do* have an entity body. When Fedora accessed via clients such as Apache Httpclient, this can cause apparent corruption of responses from Fedora. HttpClient does not consume entity bodies from requests that return a 304. The bytes from the errant entity body returned from Fedora can remain in internal HttpClient buffers, corrupting reads from subsequent responses. 

* Addresses https://jira.duraspace.org/browse/FCREPO-2313

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -115,6 +115,7 @@ import org.glassfish.jersey.media.multipart.ContentDisposition;
 import org.jvnet.hk2.annotations.Optional;
 
 import com.fasterxml.jackson.core.JsonParseException;
+import com.google.common.annotations.VisibleForTesting;
 
 /**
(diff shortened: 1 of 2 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -586,7 +583,7 @@ private void evaluateRequestPreconditions(final Request request,
             // here we are implicitly emitting a 304
             // the exception is not an error, it's genuinely
             // an exceptional condition
-            builder = builder.cacheControl(cc).lastModified(Date.from(date)).tag(etag);
```
emetsger: `date` could be `null`, so it was changed to `roundedDate`

awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/5173f6aa5b188dec0d5cb80a7b98ce299045c51f

PR #1142, opened 2016-11-18
Insure 304 responses do not contain an entity body. (4.7.0-maint)

emetsger:

Responses from Fedora *do* have an entity body. When Fedora accessed via clients such as Apache Httpclient, this can cause apparent corruption of responses from Fedora. HttpClient does not consume entity bodies from requests that return a 304. The bytes from the errant entity body returned from Fedora can remain in internal HttpClient buffers, corrupting reads from subsequent responses. 

* Addresses https://jira.duraspace.org/browse/FCREPO-2313

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -107,6 +107,7 @@ import org.glassfish.jersey.media.multipart.ContentDisposition;
 import org.jvnet.hk2.annotations.Optional;
 
 import com.fasterxml.jackson.core.JsonParseException;
+import com.google.common.annotations.VisibleForTesting;
 
 /**
  * An abstract class that sits between AbstractResource and any resource that
@@ -528,7 +529,8 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
         evaluateRequestPreconditions(request, servletResponse, resource, session, false);
     }
 
-    private static void evaluateRequestPreconditions(final Request request,
+    @VisibleForTesting
+    static void evaluateRequestPreconditions(final Request request,
                                                      final HttpServletResponse servletResponse,
                                                      final FedoraResource resource,
                                                      final Session session,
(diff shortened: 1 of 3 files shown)


escowles:
I would like to see a failing integration test — when I do this on the command-line with curl, I do not see a response body in the 304 response.

acoburn:
@escowles The failing integration test will be difficult, since it depends on the container being used. E.g. using tomcat 7 (as I am), I do not see the entity body; using certain versions of Jetty or Tomcat 9, however, the error appears.

Given that a 304 shouldn't have a entity body, though, I think it's reasonable to have a PR that removes explicitly _setting_ the entity body.

acoburn:
My only concern about this PR (as opposed to the PR on master and on 4.6-maintenance) is that, at present, all of the commits on master _could_ be used to create a 4.7.1 release; if we merge this, then we'll have to start being very careful about cherry-picking from master to form a 4.7.1 release. We should discuss this process before moving forward with this (I'm happy to see the others merged at any time).

escowles:
@acoburn I'm fine with the functional updates here — I agree it looks like the only way we would set the entity body is if the response was a 304 which explicitly forbids it.  But I'm not sure about including an IT that doesn't actually work.

acoburn:
@escowles I agree with you about the integration test -- would you suggest simply removing the part of this PR that includes the integration test? That seems to make sense to me.

escowles:
@acoburn Yes, that would be my preference — the unit test seems sufficient to me.

emetsger:
@escowles @acoburn Yes, the IT is really meant to be optional, and to demonstrate that this issue is a nuanced one, dependent on the behavior of the container.  So I totally agree with not including it in the ultimate PR.

emetsger:
@acoburn I did have to perform some small gymnastics to get PRs that apply to 4.6, 4.7, and master, so I appreciate the concern regarding branch maintenance with these PRs.  As you say, I don't believe you will be able to cherry pick from master to the 4.7 branch.

emetsger:
@escowles if you look at the Jira issue, you'll see examples of the bad behavior, and a script that can generate that behavior.  It is dependent on your servlet container; try using the self-executing jar from 4.6.0 or 4.7.0.  Both of those exhibit bad behavior.

awoods:
Thanks for surfacing the issue, @emetsger, and providing this fix. 
As suggested, would you please:
- Remove the integration test from the 4.6-maintenance and master PRs
- Close the 4.7-maintenance PR

emetsger:
@awoods PRs for 4.6-maint and master updated to remove the IT.  I'll close this (the 4.7-main) PR.

PR #1143, opened 2016-11-18
Insure 304 responses do not contain an entity body. (4.6.0-maint)

emetsger:

Responses from Fedora do have an entity body. When Fedora accessed via clients such as Apache Httpclient, this can cause apparent corruption of responses from Fedora. HttpClient does not consume entity bodies from requests that return a 304. The bytes from the errant entity body returned from Fedora can remain in internal HttpClient buffers, corrupting reads from subsequent responses.

Addresses https://jira.duraspace.org/browse/FCREPO-2313


awoods:
Resolved with: https://github.com/fcrepo4/fcrepo4/commit/65c806d10e6d52cd65ab32beacd549bbf1a69271

PR #1146, opened 2016-11-22
Ensure XSDType not in Content-Disposition filename response header

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-2317


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -422,7 +424,7 @@ protected void addResourceHttpHeaders(final FedoraResource resource) {
             final Date modDate = binary.getLastModifiedDate() != null ? Date.from(binary.getLastModifiedDate()) : null;
 
             final ContentDisposition contentDisposition = ContentDisposition.type("attachment")
-                    .fileName(binary.getFilename())
+                    .fileName(removeXSDType(binary.getFilename()))
```
acoburn: +1 on this approach

PR #1147, opened 2016-12-06
Fcrepo 2323

awoods:
The first commit recursively removes inbound references to child of deleted containers.
The second commit ignores non-existent referents.

Resolves: https://jira.duraspace.org/browse/FCREPO-2323

Note: These two commits should not be squashed.

### DIFF
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -365,33 +365,8 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
     @Override
     public void delete() {
         try {
-            @SuppressWarnings("unchecked")
-            final Iterator<Property> references = node.getReferences();
-            @SuppressWarnings("unchecked")
-            final Iterator<Property> weakReferences = node.getWeakReferences();
-            concat(iteratorToStream(references), iteratorToStream(weakReferences)).forEach(prop -> {
-                try {
-                    final List<Value> newVals = property2values.apply(prop).filter(
-                            UncheckedPredicate.uncheck(value ->
-                                !node.equals(getSession().getNodeByIdentifier(value.getString()))))
-                        .collect(toList());
-
-                    if (newVals.size() == 0) {
-                        prop.remove();
-                    } else {
-                        prop.setValue(newVals.toArray(new Value[newVals.size()]));
-                    }
-                } catch (final RepositoryException ex) {
-                    // Ignore error from trying to update properties on versioned resources
-                    if (ex instanceof javax.jcr.nodetype.ConstraintViolationException &&
-                            ex.getMessage().contains(JCR_VERSION_STORAGE)) {
-                        LOGGER.debug("Ignoring exception trying to remove property from versioned resource: {}",
-                                ex.getMessage());
-                    } else {
-                        throw new RepositoryRuntimeException(ex);
-                    }
-                }
-            });
+            // Remove inbound references to this resource and, recursively, any of its children
+            removeReferences(node);
 
             final Node parent = getNode().getDepth() > 0 ? getNode().getParent() : null;
 
@@ -430,6 +405,52 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
         } catch (final RepositoryException e) {
             throw new RepositoryRuntimeException(e);
         }
+    }
+
+    private void removeReferences(final Node n) {
+        try {
+            // Remove references to this resource
+            doRemoveReferences(n);
+
+            // Recurse over children of this resource
+            if (n.hasNodes()) {
+                @SuppressWarnings("unchecked")
+                final Iterator<Node> nodes = n.getNodes();
+                nodes.forEachRemaining(this::removeReferences);
+            }
+        } catch (RepositoryException e) {
+            throw new RepositoryRuntimeException(e);
+        }
+    }
+
+    private void doRemoveReferences(final Node n) throws RepositoryException {
+        @SuppressWarnings("unchecked")
+        final Iterator<Property> references = n.getReferences();
+        @SuppressWarnings("unchecked")
+        final Iterator<Property> weakReferences = n.getWeakReferences();
+        concat(iteratorToStream(references), iteratorToStream(weakReferences)).forEach(prop -> {
+            try {
+                final List<Value> newVals = property2values.apply(prop).filter(
+                        UncheckedPredicate.uncheck(value ->
+                                !n.equals(getSession().getNodeByIdentifier(value.getString()))))
+                        .collect(toList());
+
+                if (newVals.size() == 0) {
+                    prop.remove();
+                } else {
+                    prop.setValue(newVals.toArray(new Value[newVals.size()]));
+                }
+            } catch (final RepositoryException ex) {
+                // Ignore error from trying to update properties on versioned resources
+                if (ex instanceof javax.jcr.nodetype.ConstraintViolationException &&
+                        ex.getMessage().contains(JCR_VERSION_STORAGE)) {
+                    LOGGER.debug("Ignoring exception trying to remove property from versioned resource: {}",
+                            ex.getMessage());
+                } else {
+                    throw new RepositoryRuntimeException(ex);
+                }
+            }
+        });
     }
 
     private void createTombstone(final Node parent, final String path) throws RepositoryException {
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/SkolemNodeRdfContext.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/SkolemNodeRdfContext.java
@@ -26,6 +26,7 @@ import org.fcrepo.kernel.modeshape.rdf.impl.mappings.PropertyValueIterator;
 import org.slf4j.Logger;
 
 import javax.jcr.AccessDeniedException;
+import javax.jcr.ItemNotFoundException;
 import javax.jcr.Node;
 import javax.jcr.Property;
 import javax.jcr.RepositoryException;
@@ -93,6 +94,11 @@ public class SkolemNodeRdfContext extends NodeRdfContext {
             return null;
 
         } catch (final RepositoryException e) {
+            if (e instanceof ItemNotFoundException) {
+                // Effectively skip resources that do not exist
+                LOGGER.warn("Reference to non-existent resource encounterd: {}", v);
+                return null;
+            }
             throw new RepositoryRuntimeException(e);
         }
     };
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/mappings/PropertyToTriple.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/mappings/PropertyToTriple.java
@@ -22,6 +22,7 @@ import static org.apache.jena.graph.NodeFactory.createLiteral;
 import static org.apache.jena.graph.Triple.create;
 import static org.fcrepo.kernel.modeshape.identifiers.NodeResourceConverter.nodeToResource;
 import static org.fcrepo.kernel.modeshape.utils.StreamUtils.iteratorToStream;
+import static org.slf4j.LoggerFactory.getLogger;
 
 import java.util.function.Function;
 import java.util.stream.Stream;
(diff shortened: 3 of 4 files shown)


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -432,6 +407,50 @@ public void delete() {
...
     }
 
+    private void removeReferences(final Node n) {
+        try {
+            // Remove references to this resource
+            doRemoveReferences(n);
+
+            // Recurse over children of this resource
+            if (n.hasNodes()) {
+                iteratorToStream(n.getNodes()).forEach(child -> removeReferences((Node) child));
```
ajs6f: Maybe just `n.getNodes().forEachRemaining(this::removeReferences);`

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/SkolemNodeRdfContext.java
```diff
@@ -93,6 +94,10 @@ public SkolemNodeRdfContext(final FedoraResource resource,
             return null;
 
         } catch (final RepositoryException e) {
+            if (e instanceof ItemNotFoundException) {
+                // Effectively skip resources that do not exist
```
ajs6f: Does this introduce any "integrity" changes?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/mappings/PropertyToTriple.java
```diff
@@ -71,7 +73,7 @@ public PropertyToTriple(final Session session, final Converter<Resource, FedoraR
             final org.apache.jena.graph.Node propPredicate = propertyConverter.convert(p).asNode();
             final String propertyName = p.getName();
 
-            return iteratorToStream(new PropertyValueIterator(p)).map(v -> {
+            return iteratorToStream(new PropertyValueIterator(p)).filter(v -> valueCanBeConverted(v)).map(v -> {
```
ajs6f: Just `filter(this::valueCanBeConverted)`?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/SkolemNodeRdfContext.java
awoods: Ideally the scenario of having references to non-existent resources would not happen in the first place... but that is the core of the bug that the first commit in this PR addresses.
What "integrity" changes do you have in mind?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
awoods: I gave that a try... but it does not compile.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/SkolemNodeRdfContext.java
awoods: That depends on how you define "intra-repository links". If you define them as links between to existing resources... then that precludes the notion of not resolving.
If you define them as links between an existing resource and one that does not exist... then, yes, that is exactly what this commit does.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
awoods: That makes sense, @acoburn. Is that a better approach than what we currently have in this PR?

ajs6f: Yes. You are currently creating a bunch of useless short-lived objects by using the stream API.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/SkolemNodeRdfContext.java
ajs6f: I don't know anyone who would define "intrarepository" as "between to existing resources". Of course I mean "links between identifiers that belong to the repository".

whikloj: I'm not sure I totally understand the concern here. @awoods is now removing links from resources to deleted resources, but those links were inaccessible from either the side of the link. @ajs6f have your concerns been addressed?

ajs6f: No, nothing is getting removed here. This function is skipping things, as the comment says. I want to know why anything needs to get skipped.

whikloj:
I have tested the behaviour before and after this PR and it does what it proposes to. Which is delete the inbound references *from* other resources to the resource being deleted and recurses down to all child nodes (of the resource being deleted) to also delete any inbound references.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/impl/SkolemNodeRdfContext.java
awoods: Links to non-existent resources are being skipped because they otherwise result in 404 errors... which is the essence of the ticket associated with this PR.
https://jira.duraspace.org/browse/FCREPO-2323

ajs6f: The question I am asking is why it is possible to create links to non-existent resources. That is not the behavior that, e.g. @scossu, expects.

awoods: Indeed, one should not be able to create links to non-existent resources... you might even consider it a bug. Actually, there is a ticket for that bug:
https://jira.duraspace.org/browse/FCREPO-2323

This PR is the resolution.

ajs6f: IF THIS PR MAKES IT IMPOSSIBLE TO CREATE SUCH LINKS, WHY DO YOU HAVE CODE ABOVE COMMENTED "skip resources that do not exist".

ajs6f: Either your PR works, in which case there is no need to skip non-existent resources, or your PR does not work.

whikloj: I think this skipping is because with this patch @awoods is deleting links to non-existent resources. But for other resources created before the patch this recursive deletion did not take place, so we skip to allow access to those resources. Otherwise we could run the inbound link deletion on a GET, but that seems wrong for some reason.

awoods: Exactly. Thank you, @whikloj .

birkland: @awoods how bad is it to have a repo with these dangling resources?  Should the code emit a WARN message?  WIll/should there be a way to clean/fsck an existing repo that has such damage already?

birkland: Thinking about this a bit more, I can see the desire for code that compensates for data "corrupted" by this bug - that would allow this fix to be dropped over an existing repository and Just Work.  However, would it be then be necessary (or worth it) to add some tests to verify that the desired compensation works as intended?

whikloj: @ajs6f I agree that a migration utility or someway to fix the data would be great, but without this patch I'm unsure how to can do that as you can't get the linking resource at all?

ajs6f: Not via HTTP. But can you not via JCR?

whikloj: No idea, I'm actually not that comfortable with my knowledge of Fedora once we get to Modeshape. Much less dealing with the resources directly inside. I would assume yes, I guess you'd start at the root and traverse every resource and check if you can HTTP GET them. If not then check the references in the JCR. Can you have a tool that does both?

awoods: @birkland, the downside of potential, "invisible", dangling references (not resources) is not obvious. The only scenario that I can think of is if there is a single-valued WEAKREFERENCE property that is in the state of referencing a non-existent resource. Note, I say "WEAKREFERENCE" type because strong "REFERENCE" type already prevents such a scenario. I have tested this on the one REFERENCE type in Fedora's CND:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/main/resources/fedora-node-types.cnd#L65

When I create a resource in the following situation:
`</a> ldp:membershipResource </x/y>`
..and attempt to delete </x>, a referential integrity exception is thrown. That means that this issue only relates to WEAKREFERENCEs, and since the Fedora CND does not define any WEAKREFERENCE types (moreless single-valued ones), I do not see a situation where invisible, dangling references will be a problem.

That said, a utility along the lines of [TechnicalMetadataMigrator](https://github.com/fcrepo4-exts/fcrepo4-upgrade-utils/blob/master/src/main/java/org/fcrepo/upgrade/utils/TechnicalMetadataMigrator.java#L51) would be useful as a "clean-up" tool.
I have created a ticket for implementing such a utility (noting that at the present, this ticket is a lower priority).
https://jira.duraspace.org/browse/FCREPO-2331

Additionally, I have pushed another commit with more logging.

awoods:
@whikloj / @ajs6f / @acoburn / @birkland : There is nothing else on my todo list wrt this PR. It is a blocker for the 4.7.1-RC. Let me know if you have any other comments... and if not, I will go ahead and do some surgical commit squashing... and we can move this and the 4.7.1-RC forward.

acoburn:
@awoods I have no further comments.

birkland:
@awoods I can't say that I fully understand the consequences of dropping this fix over a repository that once had this issue, but don't let that hold up this PR.   To a client of the repository, it looks like this fix does what it needs to do.

whikloj:
@awoods I'm fine with that, are you squashing your commits on this PR. Or do you want that done as part of the merge?

scossu:
@awoods I think this can be a quick patch to address the issue ASAP and make inaccessible resources available again for fixing (the broken links will still need to be fixed). So :+1: from me on an RC1 that enables us to fix our production repo. 

That said, I agree with @ajs6f 's and @birkland 's concerns about removing a check for broken links a permanent part of the code. Without that, the referential integrity of the repository is solely relying on the assumption that Fedora will always create valid in-repo references—which this bug confirmed to be a not valid assumption. The error we found is actually a protection for referential integrity. 

For an RC2 I would suggest to make ignoring broken links a configurable option that is turned off by default, and can be turned on in exceptional cases such as this one, for maintenance purposes. 

Thoughts?

awoods:
@whikloj , I'll squash in just a bit.

kefo:
@scossu's suggestion to make ignoring broken links a config option, which defaults to 'false', seems to me a way to isolate this for what it is - an exception in exceptional circumstances. 

Like @birkland, I too feel uncomfortable that it is hard to fully understand the ramifications - now or in the future - of ignoring (if that is the correct word) these aberrant resources.  I certainly would not want to persist a problem, especially one that could complicate things in the future.  It may be worth exploring how widespread the issue is.

awoods:
PR is squashed and ready for commit to master.

awoods:
@whikloj / @ajs6f / @acoburn / @birkland : This is ready to go. Please merge and update: https://jira.duraspace.org/browse/FCREPO-2323
..so that @bseeger and I can push the 4.7.1-RC tomorrow.
Thanks!

PR #1149, opened 2016-12-19
Explicitly encode test objects as utf8

birkland:
Resolves [FCREPO-2370](https://jira.duraspace.org/browse/FCREPO-2370) for 4.7.1-RC

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -1916,7 +1916,7 @@ public class FedoraLdpIT extends AbstractResourceIT {
     }
 
     @Test
-    public void testLinkedDeletion() throws IOException {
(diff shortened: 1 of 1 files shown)


acoburn:
@birkland, as this only affects test code, am I correct in understanding that this change is _not_ intended to result in a RC-2?

birkland:
@acoburn Right, this only affects test code.  However, it's necessary for Windows testers in order to build the RC in the first place without disabling tests.  So they can build the RC from this PR, or the 4.7.1-RC branch should this PR be merged into it.

bseeger:
@birkland - would you mind running a build with the 4.7.1 RC candidate code adding this to your java opts:

`-Dfile.encoding=UTF-8`

And seeing if the tests pass then?  

Thanks!

birkland:
After a little bit of trial and error, this seemed to do the trick:
    
    export JAVA_TOOL_OPTIONS="-Dfile.encoding=UTF8"

It seems like a plausible workaround for evaluating 4.7.1-RC without changes to any files

PR #1152, opened 2017-01-05
Add a patch to the RDF output framework to serialize RDF with their S…

awoods:
…tring and Boolean type

information ala RDF 1.0.

Resolves: https://jira.duraspace.org/browse/FCREPO-2374

The RDF writers in this module are auto-injected into the "fcrepo-http-commons/RdfStreamStreamingOutput" framework
via Spring's annotation-based instantiation of RdfWriterHelper.

This module should be removed with the release of Fedora 4.8.0, at which point, standard RDF 1.1 output syntax
should be used.

The only other evidence of this patch that should also be removed is:
1. jena-patch module definition in top-level pom.xml
2. jena-patch dependency declaration in fcrepo-webapp/pom.xml

Most of the classes in this module are exact or near copies of Jena classes. The differences from the original
classes can be found by searching for comments prefixed with:
"// NOTE, Fedora change"

### DIFF
--- a/fcrepo-webapp/pom.xml
+++ b/fcrepo-webapp/pom.xml
@@ -68,6 +68,11 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
+      <groupId>org.fcrepo</groupId>
+      <artifactId>jena-patch</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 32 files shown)


escowles:
I just built and tested this locally, and it works as expected.

bseeger:
I just built and tested it locally as well. Looks good to me.

PR #1155, opened 2017-01-10
Ensure Content-Type header for HEAD requests on containers mirrors wh…

dbernstein:
…at we see a GET.

Resolves https://jira.duraspace.org/browse/FCREPO-2159

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -150,6 +150,8 @@ import org.apache.jena.rdf.model.StmtIterator;
 import org.apache.jena.sparql.core.DatasetGraph;
 import org.apache.jena.sparql.core.Quad;
 import org.apache.jena.vocabulary.DC_11;
+
+import org.fcrepo.http.commons.domain.RDFMediaType;
 import org.fcrepo.http.commons.test.util.CloseableDataset;
 import org.glassfish.jersey.media.multipart.ContentDisposition;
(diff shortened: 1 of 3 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -236,6 +237,21 @@ public void testHeadBasicContainer() throws IOException {
...
+    public void testHeadDefaultContentType() throws IOException {
+        final String id = getRandomUniqueId();
+        createObjectAndClose(id);
+        addMixin(id, CONTAINER.getURI());
+
+        final HttpHead headObjMethod = headObjMethod(id);
+        try (final CloseableHttpResponse response = execute(headObjMethod)) {
+            final Collection<String> contentType = getHeader(response, CONTENT_TYPE);
+            assertTrue(contentType.size() == 1);
+            assertTrue("Didn't find LDP valid content-type header: " + contentType, !contentType.iterator().next()
```
awoods: I would suggest changing this test to look for the expected content-type instead of looking for the absence of an unwanted value.

dbernstein: I'm having a hard time getting jersey to return a deterministic result when no accept header is specified.  How about I add additional tests that verify that the using an appropriate accept header gives the correct content-type in the response?.

dbernstein: Nevermind - I think I figured it out.

PR #1156, opened 2017-01-13
Ensures that the error content type matches the media type of the ret…

dbernstein:
…urned representation.

All the exception handlers have been refactored to
   1) reduce duplicate code
   2) ensure that there is only one place were the default content type for exceptions
   3) smooth the way for supporting rdf error responses.

Resolves https://jira.duraspace.org/browse/FCREPO-1888

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/AccessControlExceptionMapper.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/AccessControlExceptionMapper.java
@@ -17,16 +17,10 @@
  */
 package org.fcrepo.http.commons.exceptionhandlers;
 
-import static javax.ws.rs.core.Response.status;
-import static javax.ws.rs.core.Response.Status.FORBIDDEN;
-import static org.slf4j.LoggerFactory.getLogger;
-
 import javax.jcr.security.AccessControlException;
-import javax.ws.rs.core.Response;
-import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.core.Response.ResponseBuilder;
+import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.ext.Provider;
-
-import org.slf4j.Logger;
 
 /**
  * Translate JCR AccessControlExceptions into HTTP 403 Forbidden errors
@@ -36,15 +30,15 @@ import org.slf4j.Logger;
  * @author gregjan
  */
 @Provider
-public class AccessControlExceptionMapper implements
-        ExceptionMapper<AccessControlException>, ExceptionDebugLogging {
-
-    private static final Logger LOGGER =
-        getLogger(AccessControlExceptionMapper.class);
+public class AccessControlExceptionMapper extends FedoraExceptionMapper<AccessControlException> {
 
     @Override
-    public Response toResponse(final AccessControlException e) {
-        debugException(this, e, LOGGER);
-        return status(FORBIDDEN).build();
+    protected ResponseBuilder status(final AccessControlException e) {
+        return super.status(Status.FORBIDDEN);
+    }
+
+    @Override
+    protected ResponseBuilder entity(final ResponseBuilder builder, final AccessControlException e) {
+        return builder.entity("This resource is forbidden.");
     }
 }
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/AccessControlJavaSecurityExceptionMapper.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/AccessControlJavaSecurityExceptionMapper.java
@@ -15,19 +15,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.fcrepo.http.commons.exceptionhandlers;
 
-import static javax.ws.rs.core.Response.status;
-import static javax.ws.rs.core.Response.Status.FORBIDDEN;
-import static org.slf4j.LoggerFactory.getLogger;
+package org.fcrepo.http.commons.exceptionhandlers;
 
 import java.security.AccessControlException;
 
-import javax.ws.rs.core.Response;
-import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.core.Response.ResponseBuilder;
+import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.ext.Provider;
-
-import org.slf4j.Logger;
 
 /**
  * Translate Java Security AccessControlExceptions into HTTP 403 Forbidden errors
@@ -37,16 +32,16 @@ import org.slf4j.Logger;
  * @author gregjan
  */
 @Provider
-public class AccessControlJavaSecurityExceptionMapper implements
-        ExceptionMapper<AccessControlException>, ExceptionDebugLogging {
-
-    private static final Logger LOGGER =
-        getLogger(AccessControlJavaSecurityExceptionMapper.class);
+public class AccessControlJavaSecurityExceptionMapper extends FedoraExceptionMapper<AccessControlException> {
 
     @Override
-    public Response toResponse(final AccessControlException e) {
-        debugException(this, e, LOGGER);
-        return status(FORBIDDEN).build();
+    protected ResponseBuilder status(final AccessControlException e) {
+        return super.status(Status.FORBIDDEN);
+    }
+
+    @Override
+    protected ResponseBuilder entity(final ResponseBuilder builder, final AccessControlException e) {
+        return builder.entity("This resource is forbidden..");
     }
 
 }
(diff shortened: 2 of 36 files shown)


fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/AccessControlJavaSecurityExceptionMapper.java
```diff
@@ -37,16 +32,16 @@
...
+    @Override
+    protected ResponseBuilder status(final AccessControlException e) {
+        return super.status(Status.FORBIDDEN);
+    }
 
     @Override
-    public Response toResponse(final AccessControlException e) {
-        debugException(this, e, LOGGER);
-        return status(FORBIDDEN).build();
+    protected ResponseBuilder entity(final ResponseBuilder builder, final AccessControlException e) {
```
awoods: No entity needed.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/AccessDeniedExceptionMapper.java
```diff
@@ -15,36 +15,29 @@
...
 
-    /*
-     * (non-Javadoc)
-     * @see javax.ws.rs.ext.ExceptionMapper#toResponse(java.lang.Throwable)
-     */
     @Override
-    public Response toResponse(final AccessDeniedException e) {
-        debugException(this, e, LOGGER);
-        return status(FORBIDDEN).build();
+    protected ResponseBuilder entity(final ResponseBuilder builder, final AccessDeniedException e) {
```
awoods: No entity needed.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/BadRequestExceptionMapper.java
```diff
@@ -15,36 +15,18 @@
...
-    private static final Logger LOGGER = getLogger(BadRequestExceptionMapper.class);
-
-    @Override
-    public Response toResponse(final BadRequestException e) {
-        LOGGER.error("BadRequestExceptionMapper caught an exception: {}", e.getMessage());
-        debugException(this, e, LOGGER);
-        return status(BAD_REQUEST).entity(e.getMessage()).build();
-    }
-
+public class BadRequestExceptionMapper extends FedoraExceptionMapper<BadRequestException> {
```
awoods: If we are removing the default error logging (which I would suggest), we should re-introduce it in a local `toResponse` method here.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/ClientErrorExceptionMapper.java
```diff
@@ -15,31 +15,24 @@
...
-public class ClientErrorExceptionMapper implements
-        ExceptionMapper<ClientErrorException>, ExceptionDebugLogging {
-
-    private static final Logger LOGGER = getLogger(ClientErrorExceptionMapper.class);
+public class ClientErrorExceptionMapper extends FedoraExceptionMapper<ClientErrorException> {
 
     @Override
-    public Response toResponse(final ClientErrorException e) {
-        debugException(this, e, LOGGER);
-        return fromResponse(e.getResponse()).entity(e.getMessage()).build();
```
awoods: Looks like we will be losing the `e.getMessage()` since `fromResponse` does not use the message.

awoods:
An integration test should likely be added to verify the specific issue raised in the ticket:
https://jira.duraspace.org/browse/FCREPO-1888

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/AccessControlExceptionMapper.java
```diff
@@ -36,15 +30,15 @@
...
+    protected ResponseBuilder status(final AccessControlException e) {
+        return super.status(Status.FORBIDDEN);
+    }
 
     @Override
-    public Response toResponse(final AccessControlException e) {
-        debugException(this, e, LOGGER);
-        return status(FORBIDDEN).build();
+    protected ResponseBuilder entity(final ResponseBuilder builder, final AccessControlException e) {
+        return builder.entity("This resource is forbidden.");
```
dbernstein: The problem is that if the entity is not set,  then the content-type comes back as text/html.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/FedoraExceptionMapper.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    /**
+     * Creates a response builder with the status defined in the constructor (or the default value, ie BAD_REQUEST, if
+     * the default constructor was used). Subclasses may override to specify a custom status.
+     * 
+     * @param e The exception passed into the mapper.
+     * @return The response builder.
+     */
+    protected ResponseBuilder status(final T e) {
```
dbernstein: The idea is that subclasses can inspect the T e and change the status accordingly.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/AccessControlJavaSecurityExceptionMapper.java
dbernstein: Same as comment about null entities resulting in the wrong content-type.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/AccessDeniedExceptionMapper.java
dbernstein: same as above.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/BadRequestExceptionMapper.java
dbernstein: The reason I added the default error logging is that most of the mappers used it.   So just to confirm the intention - we don't want "error" level logging in the mappers?

awoods: We want the behavior to be the same as before this PR, with the exception of the issue raised in:
https://jira.duraspace.org/browse/FCREPO-1888

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/ClientErrorExceptionMapper.java
dbernstein: Right - but the e.getMessage() is added to the entity by default in FedoraExceptionMapper.entity().

dbernstein:
A new integration test seems like a good idea.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/FedoraExceptionMapper.java
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * Sets the content on the builder. Subclasses can override this method in order to define custom logic for
+     * resolving the appropriate contentType.
+     * 
+     * @param builder The response builder.
+     * @param e The exception passed into the mapper.
+     * @return response builder
+     */
+    protected ResponseBuilder contentType(final ResponseBuilder builder, final T e) {
```
awoods: Until this method is used outside of this class, it can probably be made `private`.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/FedoraExceptionMapper.java
```diff
@@ -0,0 +1,163 @@
...
+        return status(status);
+    }
+
+    /**
+     * Creates a response builder with the specified status
+     * 
+     * @param status http status
+     * @return The response builder.
+     */
+    protected ResponseBuilder status(final Response.Status status) {
```
awoods: Until this method is used outside of this class, it can probably be made `private`.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/FedoraExceptionMapper.java
```diff
@@ -0,0 +1,163 @@
...
+        return status(status.getStatusCode());
+    }
+
+    /**
+     * Creates a response builder with the specified http status code
+     * 
+     * @param statusCode http status code
+     * @return response builder
+     */
+    protected ResponseBuilder status(final int statusCode) {
```
awoods: Until this method is used outside of this class, it can probably be made `private`.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/BadRequestExceptionMapper.java
dbernstein: Okay - I think I understand: if the mapper previously logged the error, then we log it as before using a LOGGER defined in the concrete subclass.

PR #1158, opened 2017-01-20
Fedora properly deletes triples when two or more WHERE patterns each …

dbernstein:
…match multiple triples.

Resolves https://jira.duraspace.org/browse/FCREPO-2383

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -597,10 +597,7 @@ public class FedoraLdpIT extends AbstractResourceIT {
         deleteQuery.addHeader(CONTENT_TYPE, "application/sparql-update");
 
         final String deleteQueryStatement = "" +
-                "DELETE { \n" +
-                " <> <http://example.org/test/a> ?a . \n" +
-                " <> <http://example.org/test/b> ?b . \n" +
-                "} \n" +
+                "DELETE \n" +
                 "WHERE \n" +
                 "{ \n" +
                 "  <> <http://example.org/test/a> ?a . \n" +
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -24,12 +24,12 @@ import static java.util.Arrays.asList;
 import static java.util.Arrays.stream;
 import static java.util.regex.Pattern.compile;
 import static java.util.stream.Collectors.toList;
-import static javax.ws.rs.core.Link.fromUri;
 import static javax.ws.rs.core.HttpHeaders.ACCEPT;
 import static javax.ws.rs.core.HttpHeaders.CONTENT_DISPOSITION;
 import static javax.ws.rs.core.HttpHeaders.CONTENT_LENGTH;
 import static javax.ws.rs.core.HttpHeaders.CONTENT_TYPE;
 import static javax.ws.rs.core.HttpHeaders.LINK;
+import static javax.ws.rs.core.Link.fromUri;
 import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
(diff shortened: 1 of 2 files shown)


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NodePropertiesTools.java
```diff
@@ -234,11 +234,6 @@ public void removeNodeProperty(final Node node, final String propertyName, final
                         LOGGER.debug("Removing value '{}' from property '{}'", strValueToRemove, propertyName);
                         property.setValue(newValues);
                     }
-                } else {
-                    LOGGER.debug("Value not removed from property name '{}' (value '{}')", propertyName,
```
bseeger: I think there is some value to leaving the debug statement in here, if that part doesn't cause issues.  That way it's really clear that the function failed to remove the triple(s).  Else it looks like it will be silent failure, where even turning on debug won't help.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -64,14 +64,14 @@
 import static org.apache.jena.riot.WebContent.contentTypeRDFXML;
 import static org.apache.jena.riot.WebContent.contentTypeSPARQLUpdate;
 import static org.apache.jena.riot.WebContent.contentTypeTurtle;
+import static org.apache.jena.vocabulary.DC_11.title;
```
bseeger: I don't see this even being used at all. Does it need to be here?

bseeger: Never mind - I see it now. Was looking at the wrong view.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -560,6 +560,72 @@ public void testUpdateObjectGraph() throws IOException {
...
+        final String deleteQueryStatement = "" +
+                "DELETE { \n" +
+                " <> <http://example.org/test/a> ?a . \n" +
+                " <> <http://example.org/test/b> ?b . \n" +
+                "} \n" +
+                "WHERE \n" +
+                "{ \n" +
+                "  <> <http://example.org/test/a> ?a . \n" +
+                "  <> <http://example.org/test/b> ?b . \n" +
+                "} ";
```
whikloj: This does not seem to be the same SPARQL syntax as was described in the ticket?

whikloj: Perhaps a second test with the `DELETE WHERE` syntax?
```
DELETE WHERE 
{ 
<> <http://example.org/test/a> ?a . 
<> <http://example.org/test/b> ?b . 
} 
```

awoods:
Running the scripts in the ticket against this PR, I still see the error... and a hint in the output:
`This method cannot be called on a property with multiple values`

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NodePropertiesTools.java
dbernstein: okay - I'll put that back.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
dbernstein: Good point - I was going off of the script.  I'll change that up.  I think one test is sufficient since they should be be functionally equivalent, no?

awoods:
Recompiling and re-running the original scripts in the ticket indicate all is well. You were correct, @dbernstein, I had a stale local copy.
@whikloj / @bseeger , this looks good to me if you want to squash-merge.

whikloj:
👍  built and tested and works as described.

bseeger:
Looks good to me.

PR #1159, opened 2017-01-20
Ensures that the error content type matches the media type of the ret…

dbernstein:
…urrned representation.

Resolves https://jira.duraspace.org/browse/FCREPO-1888


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
```diff
@@ -268,9 +270,14 @@ protected static String getLocation(final HttpResponse response) {
     }
 
     protected String getContentType(final HttpUriRequest method) throws ClientProtocolException, IOException {
+        return getContentType(method, OK);
+    }
+
+    protected String getContentType(final HttpUriRequest method, final Status httpStatus) throws ClientProtocolException,
```
awoods: Minor checkstyle error: https://travis-ci.org/fcrepo4/fcrepo4/builds/193642642#L6783

awoods: This exception can be removed from the `throws` clause: 'ClientProtocolException'
..and its `import` declaration.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
```diff
@@ -268,9 +270,14 @@ protected static String getLocation(final HttpResponse response) {
     }
 
     protected String getContentType(final HttpUriRequest method) throws ClientProtocolException, IOException {
```
awoods: This exception can be removed from the `throws` clause: 'ClientProtocolException'

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/BadRequestExceptionMapper.java
```diff
@@ -44,7 +45,7 @@
     public Response toResponse(final BadRequestException e) {
         LOGGER.error("BadRequestExceptionMapper caught an exception: {}", e.getMessage());
         debugException(this, e, LOGGER);
-        return status(BAD_REQUEST).entity(e.getMessage()).build();
+        return status(BAD_REQUEST).entity(e.getMessage()).type(DEFAULT_CONTENT_TYPE).build();
```
awoods: I think the clarity of the code would be increased by replacing `DEFAULT_CONTENT_TYPE` with the actual "TEXT_PLAIN_WITH_CHARSET".
I realize that by abstracting the actual content-type with a global variable allows for changing the value of that variable in the future... but in that theoretical scenario, there is an equally good chance that these classes will change as well and my sense would be the same then, too; namely, that seeing the actual content-type is more clear.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/ExceptionMapperConstants.java
```diff
@@ -0,0 +1,36 @@
...
+
+package org.fcrepo.http.commons.exceptionhandlers;
+
+import static org.fcrepo.http.commons.domain.RDFMediaType.TEXT_PLAIN_WITH_CHARSET;
+
+/**
+ * @author Daniel Bernstein
+ * @since Jan 19, 2017
+ */
+class ExceptionMapperConstants {
```
awoods: Suggest removing this class.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/BadRequestExceptionMapper.java
dbernstein: Fair enough.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/ExceptionMapperConstants.java
dbernstein: It's no longer necessary - agreed.

PR #1165, opened 2017-01-23
Add scm and developers to fcrepo-parent pom

awoods:
- Required by release plugin


acoburn:
Are you sure this belongs in the parent pom and not in the individual project configs? I would encourage you to put this data in the project configs and _not_ in the parent pom.

bseeger:
Hi @acoburn, we're trying to get past this issue where we're failing sonatype validation because that info is not in the parent pom:

`Invalid POM: /org/fcrepo/fcrepo-parent/4.7.1/fcrepo-parent-4.7.1.pom: SCM URL missing, Developer information missing`

acoburn:
what projects are failing? All that data is present in the main fcrepo/pom.xml

bseeger:
Working on fcrepo4 - that fcrepo-parent is the only one failing that POM validation step.

acoburn:
OK, whatever you need to do. But that information isn't in any previous release of `fcrepo-parent`, even if you run `mvn help:effective-pom` on one of those previous releases.

awoods:
I am guessing that each pom needs an SCM section for sonatype/maven-central. The fcrepo-parent/pom.xml used to get that from the sonatype-parent... but we have discontinued using the sonatype parent.

acoburn:
The `<developers>` section is optional and can be omitted. The `scm` section is required. Here's that value from the `fcrepo-parent/4.7.0` release:

```
  <scm>
    <connection>scm:svn:http://svn.sonatype.org/spice/trunk/oss/oss-parenti-9/fcrepo-parent</connection>
    <developerConnection>scm:svn:https://svn.sonatype.org/spice/trunk/oss/oss-parent-9/fcrepo-parent</developerConnection>
    <tag>fcrepo-4.7.0</tag>
    <url>http://svn.sonatype.org/spice/trunk/oss/oss-parent-9/fcrepo-parent</url>
  </scm>
```

which is clearly wrong and which is why we removed the sonatype parent.

PR #1168, opened 2017-01-26
HTML UI: External Content redirects to binary description

dbernstein:
Resolves https://jira.duraspace.org/browse/FCREPO-2387

### DIFF
--- a/fcrepo-http-api/src/main/resources/views/common.js
+++ b/fcrepo-http-api/src/main/resources/views/common.js
@@ -96,9 +96,11 @@
         if (res.status >= 400 || res.getResponseHeader('Link') == null) {
           var newLocation = url;
           // Note: HEADing an external resource returns a temporary redirect to the external resource:
-          // therefore there is no Link header. However what we want is to see is the metadata 
+          // therefore there is no Link header. However what we want to see is the metadata 
           // for the external reference rather than  the external object itself.
           // (c.f. https://jira.duraspace.org/browse/FCREPO-2387)
+          // WARNING: Fragile code relying on magic suffix '/fcr:metadata' and absence of 'Link' header 
+          // on external resource.
           if(!url.match(/.*fcr:metadata/)){
             newLocation = url + "/fcr:metadata";
           }
--- a/fcrepo-http-api/src/main/resources/views/common.js
+++ b/fcrepo-http-api/src/main/resources/views/common.js
@@ -94,7 +94,15 @@
       const url = this.href;
       http('HEAD', url, function(res) {
         if (res.status >= 400 || res.getResponseHeader('Link') == null) {
-          location.href = url;
+          var newLocation = url;
+          // Note: HEADing an external resource returns a temporary redirect to the external resource:
+          // therefore there is no Link header. However what we want is to see is the metadata 
+          // for the external reference rather than  the external object itself.
+          // (c.f. https://jira.duraspace.org/browse/FCREPO-2387)
+          if(!url.match(/.*fcr:metadata/)){
+            newLocation = url + "/fcr:metadata";
+          }
+          location.href = newLocation ;
           return;
         }
 


fcrepo-http-api/src/main/resources/views/common.js
```diff
@@ -94,7 +94,15 @@
       const url = this.href;
       http('HEAD', url, function(res) {
         if (res.status >= 400 || res.getResponseHeader('Link') == null) {
-          location.href = url;
+          var newLocation = url;
```
awoods: Ideally, we would be able to more directly detect that a redirect took place, such as comparing the original `url` to the response `Location`, or check for a `307` status.
But in the absence of that possibility, this looks good.
@acoburn : thoughts?

acoburn: There are a number of issues with this code, and I don't actually see a good resolution.

First, to @awoods point -- the `XMLHttpRequest` will transparently follow 3xx redirects, so there is no way to test for, say, a 307 response. 

Secondly, with the redirect in place, if the remote resource doesn't have CORS headers set up correctly for the request, you'd get a preflight error anyway.

Third, the comment:

> ... to the external resource: therefore there is no Link header.

How can you guarantee that there is no `Link` header on a resource that is outside of Fedora's domain? That is very shaky ground to stand on.

Fourth, testing for the presence of `/fcr:metadata` is not a particularly "LDP" way of doing things. Yes, it works with this particular implementation of Fedora, but that's all one could claim.

That said, I don't want to stand in the way of addressing this ticket, so feel free to merge this code as-is.

awoods: I shared the same thoughts when reviewing this PR... but was also unable to determine a more reliable, LDP-based solution.
I would suggest add to the inline comments:
```
// WARNING: Fragile code relying on magic suffix '/fcr:metadata' and absence of 'Link' header on external resource.
```

escowles: In theory, this check could be a little more robust — handling both the absence of Link headers or simply not having any of the Fedora Link headers that are checked in the next section.  But I would be fine with just updating the comment about how this relies on the magic /fcr:metadata suffix.

birkland: Would the [fetch API](https://fetch.spec.whatwg.org/) be a possibility?  I know almost nothing about it, but the spec and [this blog post]( https://davidwalsh.name/fetch) suggest that it's possible to use different redirect strategies (e.g. `manual`).  [Browser support](http://caniuse.com/#feat=fetch) is possibly adequate.

acoburn: The fetch API _would_ be perfect for this, unless we care about supporting Safari and IE. One option would be to polyfill `fetch` with `XMLHttpRequest` for those browsers, but doing so might make this code a lot more complicated than it needs to be, especially for a feature that has nothing to do with the API spec.

dbernstein: I don't like the fact that there's no guarantee that the Link header won't be present on the external resource.   I could explore the fetch API option.

awoods: @dbernstein, although `fetch` sounds like a  non-starter if it does not work on Safari (and IE).

awoods: I am ok with settling on the current approach (with inclusion of warning comments) unless anyone feels strongly otherwise.
@escowles ? @acoburn ? @birkland ?

escowles: 👍  that seems like the best thing we can do here

dbernstein: okay - I'll fix the typo and we'll move on.

PR #1169, opened 2017-01-27
FCREPO-2391: buffer unique statement operation when processing PATCH

barmintor:
- buffer unique statement operations to prevent OOM on large sparql update solution products


awoods:
Could you add an integration test that verifies the fix?

barmintor:
do you want me to use the sample requests from the ticket?

awoods:
That would be perfect.

barmintor:
The examples I added differ from the ticket by leaving one of the non-problematic triples in- I wanted to be able to demonstrate that the updates actually ran.

PR #1170, opened 2017-02-06
A fcrepo.jms.baseUrl with no port is unable to override the port in the request URL

emetsger:
The system property `fcrepo.jms.baseUrl` may be defined in order to override the URIs that are used by messages emitted from the repository. This is useful, for example, when URIs used by backend infrastructure are different from URIs used by external, user-facing services. 

By default, the `fcrepo.jms.baseUrl` is equal to the incoming request's base URL. 

The URI provided by `fcrepo.jms.baseUrl` must be an http(s) URL, but more importantly the implementation recognizes the URI as a template. This allows an administrator to override some or all of the request URL as local needs dictate. Example values for `fcrepo.jms.baseUrl` include: 

Host: `http://external.org` 
Host and port: `http://external.org:8080/` 
Host and path: `http://external.org/fcrepo/rest` 
etc. 

However, there is a bug in Fedora whereby a `fcrepo.jms.baseUrl` with no port (e.g. `http://external.org/fcrepo/rest`) cannot be used to override a request url that contains a port (e.g. `http://internal.org:8080/fcrepo/rest/foo`). 

When `fcrepo.jms.baseUrl` is set equal to `http://external.org/fcrepo/rest`, the expectation is that repository messages for `http://internal.org:8080/fcrepo/rest/foo` will be emitted as `http://external.org/fcrepo/rest/foo` (no port), but currently are emitted as `http://external.org:8080/fcrepo/rest/foo` (with a port).

Note that  447728c contains unit tests which verify the expected behavior and expose the bug.

Jira: https://jira.duraspace.org/browse/FCREPO-2395

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
@@ -52,6 +52,8 @@ abstract public class FedoraBaseResource extends AbstractResource {
 
     private static final Logger LOGGER = getLogger(FedoraBaseResource.class);
 
+    static final String JMS_BASEURL_PROP = "fcrepo.jms.baseUrl";
+
     @Inject
     protected HttpSession session;
 
@@ -103,7 +105,7 @@ abstract public class FedoraBaseResource extends AbstractResource {
      **/
     protected void setUpJMSInfo(final UriInfo uriInfo, final HttpHeaders headers) {
         try {
-            String baseURL = getBaseUrlProperty();
+            String baseURL = getBaseUrlProperty(uriInfo);
             if (baseURL.length() == 0) {
                 baseURL = uriInfo.getBaseUri().toString();
             }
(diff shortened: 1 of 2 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
```diff
@@ -122,13 +121,30 @@ protected void setUpJMSInfo(final UriInfo uriInfo, final HttpHeaders headers) {
...
+    }
+
+    /**
+     * Produce a baseURL for JMS events using the system property fcrepo.jms.baseUrl of the form http[s]://host[:port],
+     * if it exists.
+     *
+     * @param uriInfo used to build the base url
+     * @return String the base Url
+     */
+    protected String getBaseUrlProperty(final UriInfo uriInfo) {
```
awoods: It would appear to be safe to reduce the access to this new method from `protected` to `private`. 
Minimizing the contract available to other classes is recommended.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
```diff
@@ -122,13 +121,30 @@ protected void setUpJMSInfo(final UriInfo uriInfo, final HttpHeaders headers) {
     /**
      * Produce a baseURL for JMS events using the system property fcrepo.jms.baseUrl of the form http[s]://host[:port],
      * if it exists.
-     *
+     * <p>
+     * Implementation note: forwards to {@link #getBaseUrlProperty(UriInfo)}, using the internal {@code UriInfo}.
+     * </p>
      * @return String the base Url
      */
     protected String getBaseUrlProperty() {
```
awoods: This method is never used... and can be removed.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
emetsger: Happy to reduce the visibility and remove the other method, but that would also break any binary compat (e.g. any labs or exts that may have overridden the method).  If that's OK I'll make the updates!

awoods: I suspect it is safe to tighten the access to `private`. We can reverse that if anyone indicates a need... but I have never heard of anyone working in the area of extending this class outside of the core codebase.

awoods:
Resolved with:
https://github.com/fcrepo4/fcrepo4/commit/447728cf46ef5daa328462934ca7f9feb33e7548 and
https://github.com/fcrepo4/fcrepo4/commit/da93346865472045763b2630af7640c4ceb83116

Resolves: https://jira.duraspace.org/browse/FCREPO-2395

PR #1172, opened 2017-02-07
WebApplicationExceptionMapper adds entity bodies on 304 responses

emetsger:
FCREPO-2313 addressed an issue where entity bodies were being included in responses that MUST NOT carry entity bodies, per RFC 2616. 

Unfortunately, the fix for FCREPO-2313 was incomplete. The WebApplicationExceptionMapper adds entity bodies to responses without regard for their status code. In particular ContentExposingResource#evaluateRequestPreconditions(...) may lead to an exception being thrown with a 304 response code. 

The consequences of this issue are the same as the consequences for FCREPO-2313- that is, application servers and clients get confused when processing responses that contain entity bodies when they otherwise shouldn't. 

In addition, the request logic is a bit unclear to me, but it seems that the intent is that if preconditions fail, an exception should be thrown by ContentExposingResource#evaluateRequestPreconditions(...). In that case shouldn't the response be a 412?

Jira issue: https://jira.duraspace.org/browse/FCREPO-2397


awoods:
As for the question of 304 or 412 in the case of ContentExposingResource#evaluateRequestPreconditions(...), the preconditions being evaluated are etag and last-modified.
Depending on the request and associated headers, sometimes 304 is appropriate and sometimes 412 is appropriate.
For 304, this may be the case: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.25

PR #1176, opened 2017-03-03
Adds default garbage collection configuration to repository.json files

dbernstein:
Resolves:  https://jira.duraspace.org/browse/FCREPO-2406

### DIFF
--- a/fcrepo-configs/src/main/resources/config/clustered-mysql/repository.json
+++ b/fcrepo-configs/src/main/resources/config/clustered-mysql/repository.json
@@ -35,5 +35,10 @@
             { "classname" : "org.fcrepo.auth.common.BypassSecurityServletAuthenticationProvider" }
         ]
     },
+    "garbageCollection" : {
+        "threadPool" : "modeshape-gc-pool",
+        "initialTime" : "00:00",
+        "intervalInHours" : 24
+    },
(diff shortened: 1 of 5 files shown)


fcrepo-configs/src/main/resources/config/clustered-mysql/repository.json
```diff
@@ -35,5 +35,10 @@
             { "classname" : "org.fcrepo.auth.common.BypassSecurityServletAuthenticationProvider" }
         ]
     },
+    "garbageCollection" : {
+        "threadPool" : "modeshape-gc-pool",
```
awoods: Since the purpose of this configuration addition is to make it clear that binary garbage collection is configurable and has default behavior, we should probably stick to the default values... like you have done with `initialTime` and `intervalInHours`.
It looks like `threadPool`'s default is `modeshape-gc`.
See: 
- https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/resources/org/modeshape/jcr/repository-config-schema.json#L54
- https://github.com/ModeShape/modeshape/blob/master/modeshape-jcr/src/main/java/org/modeshape/jcr/RepositoryConfiguration.java#L551

dbernstein: Sure - that seems reasonable.

PR #1184, opened 2017-03-22
Correctly parse 'described-by' Link header. (FCREPO-2341)

emetsger:
@whikloj's [patch](https://jira.duraspace.org/browse/FCREPO-2341?focusedCommentId=53173&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-53173) for [FCREPO-2341](https://jira.duraspace.org/browse/FCREPO-2341) seems to work.

This PR corrects the parsing of HTTP `Link` headers in the Fedora HTML UI to be more rigorous when multiple `Link` headers are present in the response (as is the case when API-X is set up as a proxy to the Fedora repository).  The `describedBy` link header will always be used, even in the presence of multiple link headers in the response.

When multiple `Link` headers are present in the response, the browser seems to pick one of them at random.  This is problematic, for example, when API-X adds additional `Link` headers, such as a link to a service document.  Users may be randomly redirected to an API-X service discovery document after uploading a binary via (an API-X proxied) Fedora HTML UI.

### DIFF
--- a/fcrepo-http-api/src/main/resources/views/common.js
+++ b/fcrepo-http-api/src/main/resources/views/common.js
@@ -55,10 +55,10 @@
       http(method, url, headers, data, function(res) {
         if (res.status == 201) {
           const loc = res.getResponseHeader('Location');
-          const link = res.getResponseHeader('Link');
-          if (link != null && link.match('rel="describedby"')) {
-            const str = link.match(/<[^>]+>/)[0];
-            window.location = str.slice(1, str.length - 1);
+          const linkheaders = (res.getResponseHeader('Link') != null) ? res.getResponseHeader('Link').split(", ") : null;
+          const link = (linkheaders != null) ? linkheaders.filter(function(h) { return h.match(/rel="describedby"/)}) : "";
+          if (linkheaders != null && link.length > 0) {
+            window.location = link[0].substr(1, link[0].indexOf('>') - 1);
           } else if (loc != null) {
             window.location = loc;
           } else {


awoods:
Unrelated to the content of this pull-request and in an attempt to make pull-requests more transparent at a glance, in the future, please ensure the subject text of the pull-request is a brief but meaningful description:
https://wiki.duraspace.org/display/FF/Git+Workflow#GitWorkflow-Committing

A JIRA number as the subject is opaque for all except the person/people who are actively thinking about a give JIRA ticket.

fcrepo-http-api/src/main/resources/views/common.js
```diff
@@ -55,10 +55,10 @@
       http(method, url, headers, data, function(res) {
         if (res.status == 201) {
           const loc = res.getResponseHeader('Location');
-          const link = res.getResponseHeader('Link');
-          if (link != null && link.match('rel="describedby"')) {
-            const str = link.match(/<[^>]+>/)[0];
-            window.location = str.slice(1, str.length - 1);
+          const linkheaders = res.getResponseHeader('Link').split(", ");
+          const link = linkheaders.filter(function(h) { return h.match(/rel="describedby"/)});
+          if (linkheaders != null && link.length > 0) {
```
awoods: It seems like the check for `linkheaders != null` should be before line:58.

awoods:
@ruebot : absolutely!

fcrepo-http-api/src/main/resources/views/common.js
```diff
@@ -55,10 +55,10 @@
       http(method, url, headers, data, function(res) {
         if (res.status == 201) {
           const loc = res.getResponseHeader('Location');
-          const link = res.getResponseHeader('Link');
-          if (link != null && link.match('rel="describedby"')) {
-            const str = link.match(/<[^>]+>/)[0];
-            window.location = str.slice(1, str.length - 1);
+          const linkheaders = res.getResponseHeader('Link').split(", ");
```
awoods: I am hitting a `null` in this line when creating a new resource container resource.

emetsger:
@awoods I don't receive a `null` on my platform.  I've tried this patch against master and against 4.7.1.  I'm using Chrome on Mac OS version 56.0.2924.87 (64-bit).

The other platform I have available to test on is Safari version 9.1.2 (11601.7.7), but that just seems totally broken (unrelated to this patch).

I pushed an update that I hope guards against `null` objects.   Please test and let me know if it works for you.

PR #1187, opened 2017-04-05
Skolemizing blank nodes with hash URIs

escowles:
Skolemizing blank nodes to hash URI resources (prefixed with "#genid") instead of external nodes prefixed with "/.well-known/genid/".

Fixes https://jira.duraspace.org/browse/FCREPO-2108
See also https://github.com/fcrepo/fcrepo-specification/issues/82

### DIFF
--- a/fcrepo-integration-rdf/src/test/java/org/fcrepo/integration/rdf/AbstractIntegrationRdfIT.java
+++ b/fcrepo-integration-rdf/src/test/java/org/fcrepo/integration/rdf/AbstractIntegrationRdfIT.java
@@ -125,7 +125,7 @@ public abstract class AbstractIntegrationRdfIT extends AbstractResourceIT {
 
             Triple replacement = next;
 
-            if (replacement.getSubject().toString().contains(".well-known")) {
+            if (replacement.getSubject().toString().contains("#genid")) {
                 if (!bnodeMap.containsKey(replacement.getSubject())) {
                     bnodeMap.put(replacement.getSubject(), createBlankNode());
                 }
@@ -135,7 +135,7 @@ public abstract class AbstractIntegrationRdfIT extends AbstractResourceIT {
                         replacement.getObject());
             }
 
-            if (replacement.getObject().toString().contains(".well-known")) {
+            if (replacement.getObject().toString().contains("#genid")) {
 
                 if (!bnodeMap.containsKey(replacement.getObject())) {
                     bnodeMap.put(replacement.getObject(), createBlankNode());
--- a/fcrepo-integration-rdf/src/test/java/org/fcrepo/integration/rdf/AbstractIntegrationRdfIT.java
+++ b/fcrepo-integration-rdf/src/test/java/org/fcrepo/integration/rdf/AbstractIntegrationRdfIT.java
@@ -39,6 +39,7 @@ import org.fcrepo.integration.http.api.AbstractResourceIT;
 import javax.ws.rs.core.Response;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.net.URI;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -125,7 +126,7 @@ public abstract class AbstractIntegrationRdfIT extends AbstractResourceIT {
 
             Triple replacement = next;
 
-            if (replacement.getSubject().toString().contains("#genid")) {
+            if (isSkolemizedBnode(replacement.getSubject())) {
                 if (!bnodeMap.containsKey(replacement.getSubject())) {
                     bnodeMap.put(replacement.getSubject(), createBlankNode());
                 }
@@ -135,7 +136,7 @@ public abstract class AbstractIntegrationRdfIT extends AbstractResourceIT {
                         replacement.getObject());
             }
 
-            if (replacement.getObject().toString().contains("#genid")) {
+            if (isSkolemizedBnode(replacement.getObject())) {
 
                 if (!bnodeMap.containsKey(replacement.getObject())) {
                     bnodeMap.put(replacement.getObject(), createBlankNode());
@@ -157,6 +158,14 @@ public abstract class AbstractIntegrationRdfIT extends AbstractResourceIT {
             betterGraph.add(replacement);
         }
         return betterGraph;
+    }
+
+    private static boolean isSkolemizedBnode(final Node node) {
+        if (!node.isURI()) {
+            return false;
+        }
+        final URI uri = URI.create(node.toString());
+        return uri.getFragment() != null && uri.getFragment().startsWith("genid");
     }
 
     protected void checkResponse(final HttpResponse response, final Response.StatusType expected) {
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/JcrRdfTools.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/JcrRdfTools.java
@@ -18,17 +18,17 @@
 package org.fcrepo.kernel.modeshape.rdf;
 
 import static org.apache.jena.rdf.model.ModelFactory.createDefaultModel;
+import static org.apache.jena.rdf.model.ResourceFactory.createResource;
+import static java.util.UUID.randomUUID;
 import static javax.jcr.PropertyType.REFERENCE;
 import static javax.jcr.PropertyType.STRING;
 import static javax.jcr.PropertyType.UNDEFINED;
 import static javax.jcr.PropertyType.WEAKREFERENCE;
-import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_SKOLEM;
 import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_PAIRTREE;
 import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_RESOURCE;
 import static org.fcrepo.kernel.api.RdfLexicon.isManagedPredicate;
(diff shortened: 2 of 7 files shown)


fcrepo-integration-rdf/src/test/java/org/fcrepo/integration/rdf/AbstractIntegrationRdfIT.java
```diff
@@ -125,7 +125,7 @@ private static Graph getTidiedGraph(final Graph graph) {
 
             Triple replacement = next;
 
-            if (replacement.getSubject().toString().contains(".well-known")) {
+            if (replacement.getSubject().toString().contains("#genid")) {
```
ajs6f: Might want to use a stronger test parsing out the pieces of an URI.

fcrepo-integration-rdf/src/test/java/org/fcrepo/integration/rdf/AbstractIntegrationRdfIT.java
```diff
@@ -135,7 +135,7 @@ private static Graph getTidiedGraph(final Graph graph) {
                         replacement.getObject());
             }
 
-            if (replacement.getObject().toString().contains(".well-known")) {
+            if (replacement.getObject().toString().contains("#genid")) {
```
ajs6f: Ditto.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/JcrRdfTools.java
```diff
@@ -410,33 +405,20 @@ private void findOrCreateHashUri(final IdentifierConverter<Resource, FedoraResou
...
             skolemizedBnodeMap.put(id, skolemizedSubject);
         }
 
         return skolemizedBnodeMap.get(id);
     }
 
-    private static String skolemizedPrefix() {
-        return "/.well-known/genid/";
+    private String blankNodeIdentifier() {
+        return "genid" + randomUUID().toString();
```
escowles: I dithered about the generated ids here:

* UUIDs seem obnoxiously long, but on the other hand we need to be sure they are unique to the record without being able to inspect the other triples
* "genid" doesn't follow the whole "/.well-known/genid/" spec, but maybe it's enough to be clear that these are generated?

ajs6f: UUIDs are fine for my money. Alternatively, if it really bugs you, keep a counter property on the unhashed resource. Also makes sure concurrency is fine for that. Sounds like a pain in the neck.

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/kernel/modeshape/rdf/JcrRdfToolsTest.java
```diff
@@ -349,37 +363,14 @@ public void shouldCreateHashUriSubjects() throws RepositoryException {
...
-        when(mockHashNode.isNew()).thenReturn(true);
-        when(FedoraTypesUtils.getClosestExistingAncestor(mockSession,"/.well-known/genid/"))
-                .thenReturn(mockChildNode);
-        final Statement statement = testObj.skolemize(testSubjects, x);
-        assertEquals("info:fedora/x", statement.getSubject().toString());
-        assertEquals("info:fedora/x", statement.getObject().toString());
-        verify(testObj.jcrTools).findOrCreateNode(mockSession, "/.well-known/genid/");
-        verify(mockNode).addMixin(FEDORA_SKOLEM);
-        verify(mockNode.getParent()).addMixin(FEDORA_PAIRTREE);
-    }
```
escowles: This test seemed redundant to me, so I just moved the `FEDORA_PAIRTREE` check to the `shouldSkolemizeBlankNodeObjects()` test above.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/JcrRdfTools.java
ajs6f: `/.well-known/genid/` is cool, but necessarily (given the recommendation from which it is drawn) it does not consider the question o linked lifecycles between resources named by the URIs in play. We _have_ to think about that.

fcrepo-integration-rdf/src/test/java/org/fcrepo/integration/rdf/AbstractIntegrationRdfIT.java
escowles: I could make this very specific to the implementation and make sure that the URI fragment was "genid" plus something that matches the UUID spec.  Is that what you're thinking of here?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/rdf/JcrRdfTools.java
escowles: Managing a shared counter seems like overkill — I'm fine with UUIDs, but it would be nice if we could have shorter ids.  Though given the use of UUIDs elsewhere, users should be accustomed to them by now.

awoods:
Nice work @escowles ! 
Other than potentially addressing @ajs6f's comments, this looks good to go.

fcrepo-integration-rdf/src/test/java/org/fcrepo/integration/rdf/AbstractIntegrationRdfIT.java
ajs6f: No, more that you're just using `String::contains` on the whole URI. You could parse into `URI` and use the specific methods there.

escowles:
@awoods @ajs6f: I think the outstanding questions are:

1. Is "#genid" + UUID a reasonable hash URI resource identifier?  I think it is in keeping with the use of UUIDs for generated IDs elsewhere, so I think we should go with that (instead of trying to generate shorter IDs using shared state or something).

2. Is it safe to treat hash URIs that start with "#genid" as blank nodes for the purposes of computing graph equivalence for the integration test.  Since we have control over the input RDF file, I think it's safe to do this, but I could always verify that the hash URIs were "#genid" + a UUID with a regex, if anyone feels strongly we should do this.

So, I think this ready to go, but please let me know if either of you disagree about either of those or anything else here.

escowles:
@ajs6f Oh, I see what you mean now — I'll parse the URIs and verify that the hash URI identifier, and not some other random part of the URI contains the expected string.

ajs6f:
@escowles Right, exactly, about the "where did the hash URI identifier appear" thing.

For my money, UUIDs are just fine. It's absurd to send blank nodes into the system and then whine when they don't have nice names. You specifically abdicated right or responsibility for names. You can't complain.

christopher-johnson:
In my implementation, I have observed that #genid" + UUID yields a rather unsightly URI.  e.g.
`http://dor.blumenbach-online.de/fcrepo/rest/collection/AIG/773/sequence/normal#genidf6a16832-7c97-457c-af4f-50fd4acc5f2b`
While this is functionally valid, it may have been more aesthetically pleasing to delimit the #genid from the UUID with a "/" so that the skolem identifier prefix is clearly delimited from the UUID, so that it would look like this:
`http://dor.blumenbach-online.de/fcrepo/rest/collection/AIG/773/sequence/normal#genid/f6a16832-7c97-457c-af4f-50fd4acc5f2b`

No big deal, but it might be worthwhile to revisit this semantic while it is still new.  I believe that the "/" remains a reserved delimiter per rfc3986, so it is legal in a hash as well.  (have not tested in fcrepo, though)

awoods:
That is a good point, @christopher-johnson. As you likely saw in this pull-request, there was some discussion on the aesthetics of this hash-URI. 
Having a `/` would have the effect of introducing another layer of ModeShape node resources... but there might be another delimiter that would achieve the same goal you are raising... maybe `-`?

christopher-johnson:
OK, the "/" has a special meaning for the Fedora implementation of ModeShape.  

How about `#genid=UUID`?  this seems quite clear and follows existing conventions. 

Alternatively, I have seen that there was (is) an [old proposal](http://zesty.ca/crit/draft-yee-url-textsearch-00.txt) to allow an "extended fragment identifier scheme" that introduced "namespace like" identifiers in a fragment, like `#:genid:UUID` ... I guess it could then be possible to allow user-defined skolem namespaces.

awoods:
@christopher-johnson : Based on your [original comment](https://github.com/fcrepo4/fcrepo4/pull/1187#issuecomment-296425968), the objective of this brainstorming is looking for a solution to the aesthetics of the hash-URI, correct?

christopher-johnson:
@awoods.  Correct.  And I suppose it is generally true to assert that for a programmer, good aesthetics also equates to well-designed functionality in the form follows function mindset.

christopher-johnson:
@awoods I have been reading over the [Web Annotation Data Model](https://www.w3.org/TR/annotation-model/#fragment-selector) "fragment selector" specification.  In view of the clarity of this document regarding their intention to provide a standard for fragment implementations, it occurs to me that the new meaning of "hash URI identifier" for this pull request does not refer to any current specifications for what that term actually means in fcrepo and its new association with Skolemization.  

I think that there is sufficient ambiguity in fcrepo's concept of "hash uri identifier" to merit a formal definition referencing known standards.  One spec that applies to a very limited extent is [RFC 3870](https://tools.ietf.org/html/rfc3870#page-4).  I believe that what is being done with this pull request makes sense, but it just lacks formal semantic definition, unlike its predecessor that depended on rfc5785.

PR #1195, opened 2017-05-18
User-supplied namespace prefixes

escowles:
Fixes https://jira.duraspace.org/browse/FCREPO-2376


escowles:
I believe this is related to build failures I am seeing when doing a master build with parallelization enabled — see https://jira.duraspace.org/browse/FCREPO-2466

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/iterators/RdfAdder.java
```diff
@@ -53,7 +56,38 @@
      */
     public RdfAdder(final IdentifierConverter<Resource, FedoraResource> idTranslator, final Session session,
```
awoods: This method is only used in tests... and can therefore be removed.

escowles: OK, I've removed that constructor and updated the unit test to use the new constructor instead.

PR #1198, opened 2017-06-01
Fixes parallel build (ie -Dfcrepo.streams.parallel=true) Integration …

dbernstein:
…test failures.  The problem stemmed from the fact that the GraphDifferencer wasn't taking into account the possibility of parallel RDF streams.

Resolves: https://jira.duraspace.org/browse/FCREPO-2466


awoods:
Me too... and the numbers when running the `10k-members.sh` test with `-Dfcrepo.streaming.parallel=true` is as fast as ever.

PR #1201, opened 2017-06-15
Prevents users from specifying a server managed property as the objec…

dbernstein:
…t of an

ldp:hasMemberRelation predicate.

Resolves: https://jira.duraspace.org/browse/FCREPO-2519

# Notes
This pull request to my knowledge does not introduce breaking changes.

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -95,12 +95,12 @@ import org.fcrepo.http.commons.domain.ldp.LdpPreferTag;
 import org.fcrepo.http.commons.responses.RangeRequestInputStream;
 import org.fcrepo.http.commons.responses.RdfNamespacedStream;
 import org.fcrepo.http.commons.session.HttpSession;
-import org.fcrepo.kernel.api.RdfLexicon;
 import org.fcrepo.kernel.api.RdfStream;
 import org.fcrepo.kernel.api.TripleCategory;
 import org.fcrepo.kernel.api.exception.InvalidChecksumException;
 import org.fcrepo.kernel.api.exception.MalformedRdfException;
 import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
+import org.fcrepo.kernel.api.exception.ServerManagedPropertyException;
 import org.fcrepo.kernel.api.models.Container;
 import org.fcrepo.kernel.api.models.FedoraBinary;
 import org.fcrepo.kernel.api.models.FedoraResource;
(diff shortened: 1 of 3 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -636,9 +642,38 @@ protected void replaceResourceWithStream(final FedoraResource resource,
             throw new RepositoryRuntimeException(e);
         }
 
+        ensureConsistencyOfInput(inputModel);
+
         resource.replaceProperties(translator(), inputModel, resourceTriples);
     }
 
+    private void ensureConsistencyOfInput(final Model inputModel) throws BadRequestException {
+        LOGGER.info("ensuring consistency!");
```
escowles: Can you trim down this logging? I think `info` is too much here, and we probably want only `debug` if there is no exception.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -636,9 +642,38 @@ protected void replaceResourceWithStream(final FedoraResource resource,
...
+        LOGGER.info("iterator.hasNext()={}", iterator.hasNext());
+
+        while (iterator.hasNext()) {
+            final Statement s = iterator.next();
+            LOGGER.debug("statement: s={}, p={}, o={}", s.getSubject(), s.getPredicate(), s.getObject());
+            if (s.getPredicate().equals(HAS_MEMBER_RELATION)) {
+                final RDFNode obj = s.getObject();
+                if (obj.isResource()) {
+                    final String uri = obj.asResource().getURI();
+                    RdfLexicon.serverManagedProperties.forEach(p -> {
```
escowles: I suspect there is a snappy one-liner using the Predicate, but can't this at least be shortened to: `if (RdfLexicon.serverManagedProperties.contains(p)) ...`

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
dbernstein: good catch.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
awoods: Also, we want to be checking more than `serverManagedProperties`... which only includes:
```
    public static final Set<Property> serverManagedProperties = of(
            CREATED_DATE, CREATED_BY, LAST_MODIFIED_DATE, LAST_MODIFIED_BY);
```

awoods:
@dbernstein : here is a pull on pull:
https://github.com/dbernstein/fcrepo4/pull/3

awoods:
@escowles : If you are good with this PR, please feel free to squash/merge.

PR #1202, opened 2017-06-20
Add Link headers to /fcr:versions listing and add IT

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2511 & https://jira.duraspace.org/browse/FCREPO-2415

Adds 2 Link headers to the response to a `{path}/fcr:versions` request and adds an Integration Test to look for the RDFSource header.

# How should this be tested?

1. Create a container.
2. Create a version of the container.
3. `curl -i -I "http://path/to/the/container/fcr:versions"`

You should see 
```
Link: <http://www.w3.org/ns/ldp#Resource>;rel="type"
Link: <http://www.w3.org/ns/ldp#RDFSource>;rel="type"
```

This also solved FCREPO-2415 as described in my duplicate ticket [FCREPO-2525](https://jira.duraspace.org/browse/FCREPO-2525)

Basically, we assume that if a HEAD request gets a `200 OK` status code and there **aren't** any Link headers then it is a NonRdfSource resource. Then it appends `/fcr:metadata` to the end of the URL and redirects. 

Which in this case leads to `http://localhost:8080/rest/your_thing/fcr:versions/fcr:metadata` and a stack trace.

_Note_: Sorry for the import re-organizing. I thought I was using the Fedora style settings for Eclipse, but maybe I messed something up.

Example:

# Interested parties
@dbernstein @awoods

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersioning.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersioning.java
@@ -17,15 +17,16 @@
  */
 package org.fcrepo.http.api;
 
+import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
 import static javax.ws.rs.core.HttpHeaders.LINK;
 import static javax.ws.rs.core.Response.created;
 import static javax.ws.rs.core.Response.noContent;
 import static javax.ws.rs.core.Response.status;
-import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
 import static org.apache.commons.lang3.StringUtils.isBlank;
+import static org.fcrepo.kernel.api.RequiredRdfContext.VERSIONS;
 import static org.fcrepo.http.commons.domain.RDFMediaType.JSON_LD;
-import static org.fcrepo.http.commons.domain.RDFMediaType.N3_ALT2_WITH_CHARSET;
 import static org.fcrepo.http.commons.domain.RDFMediaType.N3_WITH_CHARSET;
+import static org.fcrepo.http.commons.domain.RDFMediaType.N3_ALT2_WITH_CHARSET;
 import static org.fcrepo.http.commons.domain.RDFMediaType.NTRIPLES;
 import static org.fcrepo.http.commons.domain.RDFMediaType.RDF_XML;
 import static org.fcrepo.http.commons.domain.RDFMediaType.TEXT_HTML_WITH_CHARSET;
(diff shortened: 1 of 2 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersionsIT.java
```diff
@@ -247,6 +247,36 @@ public void testVersionHeaders() throws IOException {
...
+        try (final CloseableHttpResponse response = execute(head)) {
+            assertEquals(OK.getStatusCode(), getStatus(response));
+
+            final Collection<String> linkHeaders = getLinkHeaders(response);
+
+            final Set<Link> resultSet = linkHeaders.stream().map(Link::valueOf).flatMap(link -> {
+                final String linkRel = link.getRel();
+                final URI linkUri = link.getUri();
+                if (linkRel.equals(RDF_SOURCE_LINK.getRel()) && linkUri.equals(RDF_SOURCE_LINK.getUri())) {
+                    // Found nonRdfSource!
```
whikloj: Just noticed some copy/pasta. I'll wait for any code review comments before I push a commit to correct this.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersionsIT.java
```diff
@@ -17,24 +17,24 @@
  */
 package org.fcrepo.integration.http.api;
 
-import static org.apache.jena.graph.Node.ANY;
```
awoods: Is it possible for you to revert these `import` shuffles?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersioning.java
```diff
@@ -159,6 +162,9 @@ public RdfNamespacedStream getVersionList() {
             throw new RepositoryVersionRuntimeException("This operation requires that the node be versionable");
         }
 
+        servletResponse.addHeader(LINK, "<" + LDP_NAMESPACE + "Resource>;rel=\"type\"");
```
awoods: Could you use the `Link` methods to create the header instead of manually:
```
Link.Builder builder = Link.fromUri(LDP_NAMESPACE + "Resource").rel("type");
servletResponse.addHeader(LINK, builder.build().toString());

PR #1208, opened 2017-07-11
Fixing relative URL issues for HTTPS

c1mckay:
* * *

**JIRA Ticket**: (link)
https://opensource.ncsa.illinois.edu/jira/browse/NDS-644

This pull requests replaces the absolute urls that are built in:
_'fcrepo-http-api/src/main/resources/views/common-head.vsl'_
with the relative counterpart.  Otherwise, when loading this over HTTPS, those links are served as http://<url> instead of https:// which results in mixed content errors.

# What's new?
The code to build absolute URLs is replaced with only the relative URLs.

Example:
If the site is hosted on :
https://sbwf58-fedoracommons.workbench.nationaldataservice.org/fcrepo/rest/

The location of the css page (among other files) according to the html page would be:
http://sbwf58-fedoracommons.workbench.nationaldataservice.org/fcrepo/rest/../static/css/bootstrap-3.0.0.min.css

This results in mixed content errors.  The url should be built with https, or be a relative link.

# How should this be tested?

I imagine the scenario can be created by just hosting the endpoint over HTTPS.  I've set up 2 instances that use the docker listed at [yinlinchen/fcrepo4-docker](https://hub.docker.com/r/yinlinchen/fcrepo4-docker/~/dockerfile/), except in the second one, I point to a war file that I've built using this repo.
That docker is listed at [c1mckay/ndslabs-fedora-commons](https://hub.docker.com/r/c1mckay/ndslabs-fedora-commons/builds/)

Unchanged Mix Content Error Fedora:
https://sgki99-fedoraerror.workbench.nationaldataservice.org/fcrepo/rest/

Changed c1mckay Repo Fedora:
https://saux4l-fedoracommons.workbench.nationaldataservice.org/fcrepo/rest/

# Additional Notes:
I'm available to be reached at c1mckay@sdsc.edu for further questions.

Example:
* Does this change require documentation to be updated? 
-No
* Does this change add any new dependencies? 
-No
* Does this change require any other modifications to be made to the repository (ie. Regeneration activity, etc.)? 
-No
* Could this change impact execution of existing code?
-I don't think so.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/main/resources/views/common-head.vsl
+++ b/fcrepo-http-api/src/main/resources/views/common-head.vsl
@@ -1,6 +1,6 @@
-<link rel="icon" type="image/png" href="../static/images/fcrepo-favicon.png">
-<script src="../static/js/jquery-1.9.1.min.js" ></script>
-<script src="../static/js/bootstrap-3.0.0.min.js"></script>
-<link href="../static/css/bootstrap-3.0.0.min.css" rel="stylesheet">
-<script type="text/javascript" src="fcr:assets/common.js"></script>
-<link rel="stylesheet" type="text/css" href="fcr:assets/common.css">
+<link rel="icon" type="image/png" href="$uriInfo.baseUriBuilder.build()../static/images/fcrepo-favicon.png">
+<script src="$uriInfo.baseUriBuilder.build()../static/js/jquery-1.9.1.min.js" ></script>
+<script src="$uriInfo.baseUriBuilder.build()../static/js/bootstrap-3.0.0.min.js"></script>
+<link href="$uriInfo.baseUriBuilder.build()../static/css/bootstrap-3.0.0.min.css" rel="stylesheet">
+<script type="text/javascript" src="$uriInfo.baseUriBuilder.build()fcr:assets/common.js"></script>
+<link rel="stylesheet" type="text/css" href="$uriInfo.baseUriBuilder.build()fcr:assets/common.css">
--- a/fcrepo-http-api/src/main/resources/views/common-head.vsl
+++ b/fcrepo-http-api/src/main/resources/views/common-head.vsl
@@ -1,6 +1,6 @@
-<link rel="icon" type="image/png" href="$uriInfo.baseUriBuilder.build()../static/images/fcrepo-favicon.png">
-<script src="$uriInfo.baseUriBuilder.build()../static/js/jquery-1.9.1.min.js" ></script>
-<script src="$uriInfo.baseUriBuilder.build()../static/js/bootstrap-3.0.0.min.js"></script>
-<link href="$uriInfo.baseUriBuilder.build()../static/css/bootstrap-3.0.0.min.css" rel="stylesheet">
-<script type="text/javascript" src="$uriInfo.baseUriBuilder.build()fcr:assets/common.js"></script>
-<link rel="stylesheet" type="text/css" href="$uriInfo.baseUriBuilder.build()fcr:assets/common.css">
+<link rel="icon" type="image/png" href="../static/images/fcrepo-favicon.png">
+<script src="../static/js/jquery-1.9.1.min.js" ></script>
+<script src="../static/js/bootstrap-3.0.0.min.js"></script>
+<link href="../static/css/bootstrap-3.0.0.min.css" rel="stylesheet">
+<script type="text/javascript" src="fcr:assets/common.js"></script>
+<link rel="stylesheet" type="text/css" href="fcr:assets/common.css">
--- a/fcrepo-http-api/src/main/resources/views/common-head.vsl
+++ b/fcrepo-http-api/src/main/resources/views/common-head.vsl
@@ -1,7 +1,6 @@
-<meta http-equiv="Content-Security-Policy" content="upgrade-insecure-requests">
-<link rel="icon" type="image/png" href="$uriInfo.baseUriBuilder.build()../static/images/fcrepo-favicon.png">
-<script src="$uriInfo.baseUriBuilder.build()../static/js/jquery-1.9.1.min.js" ></script>
-<script src="$uriInfo.baseUriBuilder.build()../static/js/bootstrap-3.0.0.min.js"></script>
-<link href="$uriInfo.baseUriBuilder.build()../static/css/bootstrap-3.0.0.min.css" rel="stylesheet">
-<script type="text/javascript" src="$uriInfo.baseUriBuilder.build()fcr:assets/common.js"></script>
-<link rel="stylesheet" type="text/css" href="$uriInfo.baseUriBuilder.build()fcr:assets/common.css">
+<link rel="icon" type="image/png" href="../static/images/fcrepo-favicon.png">
+<script src="../static/js/jquery-1.9.1.min.js" ></script>
+<script src="../static/js/bootstrap-3.0.0.min.js"></script>
+<link href="../static/css/bootstrap-3.0.0.min.css" rel="stylesheet">
+<script type="text/javascript" src="./fcr:assets/common.js"></script>
+<link rel="stylesheet" type="text/css" href="./fcr:assets/common.css">
--- a/fcrepo-http-api/src/main/resources/views/common-head.vsl
+++ b/fcrepo-http-api/src/main/resources/views/common-head.vsl
@@ -1,6 +1,6 @@
-<link rel="icon" type="image/png" href="$uriInfo.baseUriBuilder.build()../static/images/fcrepo-favicon.png">
-<script src="$uriInfo.baseUriBuilder.build()../static/js/jquery-1.9.1.min.js" ></script>
-<script src="$uriInfo.baseUriBuilder.build()../static/js/bootstrap-3.0.0.min.js"></script>
-<link href="$uriInfo.baseUriBuilder.build()../static/css/bootstrap-3.0.0.min.css" rel="stylesheet">
+<link rel="icon" type="image/png" href="../static/images/fcrepo-favicon.png">
+<script src="../static/js/jquery-1.9.1.min.js" ></script>
+<script src="../static/js/bootstrap-3.0.0.min.js"></script>
+<link href="../static/css/bootstrap-3.0.0.min.css" rel="stylesheet">
 <script type="text/javascript" src="$uriInfo.baseUriBuilder.build()fcr:assets/common.js"></script>
-<link rel="stylesheet" type="text/css" href="$uriInfo.baseUriBuilder.build()fcr:assets/common.css">
+<link rel="stylesheet" type="text/css" href="fcr:assets/common.css">
--- a/fcrepo-http-api/src/main/resources/views/common-head.vsl
+++ b/fcrepo-http-api/src/main/resources/views/common-head.vsl
@@ -3,4 +3,4 @@
 <script src="../static/js/bootstrap-3.0.0.min.js"></script>
 <link href="../static/css/bootstrap-3.0.0.min.css" rel="stylesheet">
 <script type="text/javascript" src="$uriInfo.baseUriBuilder.build()fcr:assets/common.js"></script>
-<link rel="stylesheet" type="text/css" href="$uriInfo.baseUriBuilder.build()fcr:assets/common.css">
+<link rel="stylesheet" type="text/css" href="./fcr:assets/common.css">
--- a/fcrepo-http-api/src/main/resources/views/common-head.vsl
+++ b/fcrepo-http-api/src/main/resources/views/common-head.vsl
@@ -2,5 +2,5 @@
 <script src="../static/js/jquery-1.9.1.min.js" ></script>
 <script src="../static/js/bootstrap-3.0.0.min.js"></script>
 <link href="../static/css/bootstrap-3.0.0.min.css" rel="stylesheet">
-<script type="text/javascript" src="./fcr:assets/common.js"></script>
-<link rel="stylesheet" type="text/css" href="./fcr:assets/common.css">
+<script type="text/javascript" src="fcr:assets/common.js"></script>
+<link rel="stylesheet" type="text/css" href="fcr:assets/common.css">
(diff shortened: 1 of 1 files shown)


awoods:
Thank you for this, @c1mckay .
Have you seen the build errors? https://travis-ci.org/fcrepo4/fcrepo4/builds/252591025#L9367

c1mckay:
seems like the checks work now?

awoods:
@fcrepo4/committers : Please give this a look.

escowles:
This does not work for resources at two or more levels of hierarchy deep (e.g., `http://localhost:8080/rest/a` works but `http://localhost:8080/rest/a/b` does not), because the relative URL assumes the CSS/JS are always one level up.  Similarly, we can't assume the path is at `/` because the webapp can be run either at the root of the servlet container or under a path.

So I think we need to use code to build the URI, not try to use relative links that will work in some situations but not all of them.

bseeger:
Yes - what @escowles said.  I think this needs to be fixed with a more robust solution.

awoods:
The issue that is trying to be resolved by this PR is that the Fedora resources are served under `https` but the /fcr:assets are served under `http`... for some unknown reason.

Is there a proxy involved? Is there some webserver configuration that would address this issue?

Fedora simply uses the requesting URI to construct the /fcr:assets URI... so ensuring that the requesting URI is `https` may be a reasonable approach.

craig-willis:
Thanks all.  A little context -- @c1mckay and I work on the development of the [NDS Labs Workbench](http://www.nationaldataservice.org/projects/labs.html).   In short, this is a service to allow folks in the research data management community to experiment with a variety of data management, analysis and visualization tools, all deployed via Docker (on Kubernetes).  @c1mckay has been helping to add support for Fedora Commons and resolve the above HTTP/HTTPS problem.

In our system, all services run via Docker with requests proxied by an NGINX reverse proxy,  provided by [Kubernetes](https://github.com/kubernetes/ingress). When running Fedora Commons, we're seeing mixed content errors (that we don't see for most other services).

The relevant configuration for nginx is as follows:
```
location / {
            # enforce ssl on server side
            if ($scheme = http) {
                return 301 https://$host$request_uri;
            }
            proxy_set_header Host                   $host;
            proxy_set_header X-Real-IP              $remote_addr;
            proxy_set_header                        Upgrade           $http_upgrade;
            proxy_set_header                        Connection        $connection_upgrade;
            proxy_set_header X-Forwarded-For        $proxy_add_x_forwarded_for;
            proxy_set_header X-Forwarded-Host       $host;
            proxy_set_header X-Forwarded-Port       $server_port;
            proxy_set_header X-Forwarded-Proto      $pass_access_scheme;
            ... 
            proxy_pass http://c1mckay-sgki99-fedoraerror-8080;
        }
```

I'm not familiar with how UriInfo/BaseUriBuilder constructs the base URI, but it sounds like NGINX isn't sending the right information.  We'll try to look at the request sent from NGINX to Tomcat, but would welcome any advice.

craig-willis:
Running tcpdump, I see the following request:

```
	Host: sgki99-fedoraerror.workbench.nationaldataservice.org
	X-Real-IP: 130.126.255.78
	Connection: close
	X-Forwarded-For: 130.126.255.78
	X-Forwarded-Host: sgki99-fedoraerror.workbench.nationaldataservice.org
	X-Forwarded-Port: 443
	X-Forwarded-Proto: https
	Content-Length: 0
	pragma: no-cache
	cache-control: no-cache
	upgrade-insecure-requests: 1
	user-agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36
	accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8
	referer: https://sgki99-fedoraerror.workbench.nationaldataservice.org/fcrepo/
	accept-encoding: gzip, deflate, sdch, br
	accept-language: en-US,en;q=0.8
	cookie: _ga=GA1.2.1234430866.1486585784; _gid=GA1.2.1123225320.1499879362
```

awoods:
@craig-willis : Would it be possible to change:
>  proxy_pass http://c1mckay-sgki99-fedoraerror-8080;

to:
>  proxy_pass https://c1mckay-sgki99-fedoraerror-8080;

craig-willis:
Unfortunately, no.  We have TLS/SSL termination at the proxy.  To enable HTTPS for the proxy_pass we'd have to add certificates to every application.

awoods:
That is good to know, and the crux of the current issue.

Would it be possible to intercept requests to `http://.../fcr:assets` and redirect them to `https://.../fcr:assets` at the nginx level?

craig-willis:
It's possible, but will require modification of the Kubernetes NGINX controller to add per-application rules.  It also looks like Jersey has an [open issue](https://github.com/jersey/jersey/issues/2716) to add support for the X-Forwarded-* headers in getBaseBuilder.  

Since the X-Forward-Proto header is present in the request, perhaps it would be better to for the header and use it to explicitly set the scheme?

For example from [Stack Overflow](https://stackoverflow.com/questions/34862344/how-can-i-force-uribuilder-to-use-https):
```
getUriInfo().getRequestUriBuilder().scheme("https").build()
```

Obviously, this would be a different PR.

awoods:
The `getRequestUriBuilder().scheme()` option could be a possibility... but would need to be a configurable option, since many sites will not have the `https` -> `http` pattern found in your environment.

craig-willis:
@awoods -- thanks for your help and feedback.  I understand that this sort of proxied configuration is unusual.  I'm glad we were able to identify the root problem.

acoburn:
I would like to add that the proxy configuration that @craig-willis describes is decidedly _not_ unusual. It is almost exactly what we have configured right now at Amherst: i.e. SSL termination via nginx to a regular HTTP port where Fedora runs. _This is an extremely common pattern for web applications._ And while the solution proposed here is not the right solution (I would echo what @escowles says), the mixed content issue described here _is_ a real problem with Fedora that should be fixed.

birkland:
I just took a quick look at this.  

Using a proxy and SSL termination is common.  It'd be the most flexible, straightforward, and secure way to put fedora on the internet.   `getRequestUriBuilder().scheme()` sounds like a good avenue to explore.

@craig-willis  or @c1mckay : I haven't had time to dig in, but I'm a bit surprised it doesn't work with plain http URIs, since you're using HSTS.   Links to the Fedora resources are [plain http URIs](https://www.w3.org/blog/2016/05/https-and-the-semantic-weblinked-data/), and the browser does the right thing by upgrading to https directly (rather than starting with http and following the redirect).  I don't quite understand why the browser isn't doing the same thing for the assets. Do you need a `Content-Security-Policy` for active content?

craig-willis:
@birkland Good question -- thank you.  I hadn't noticed that the mixed-content errors are only for js/css, so it does seem to be an active content problem.  CSP is new territory for me, so we'll take a look.

whikloj:
Ok, so I tested changing `common-head.vsl` to

```
<link rel="icon" type="image/png" href="/static/images/fcrepo-favicon.png">
<script src="/static/js/jquery-1.9.1.min.js" ></script>
<script src="/static/js/bootstrap-3.0.0.min.js"></script>
<link href="/static/css/bootstrap-3.0.0.min.css" rel="stylesheet">
<script type="text/javascript" src="/static/js/common.js"></script>
<link rel="stylesheet" type="text/css" href="/static/css/common.css">
```

This seems to work, I also moved the `common.js` and `common.css` to the static location also (didn't make sense to serve it from Jersey). If this works for you @craig-willis, then we could also remove the [ViewAssets](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ViewAssets.java) class

whikloj:
@c1mckay, @craig-willis I made a branch with my  changes for you to try out. 
https://github.com/whikloj/fcrepo4/tree/FCREPO-2534

awoods:
Thanks, @whikloj .
On a side note, @c1mckay and @craig-willis , if you provide me your email addresses, I will send you an invitation to create a JIRA/Wiki account. (awoods@duraspace.org)

acoburn:
Jared, that will only work if Fedora is deployed in the root of the servlet container. And there are plenty of people who will deploy it with some context (fcrepo), in which case this won't work

Sent from my iPhone

On Jul 14, 2017, at 10:45 AM, Andrew Woods <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks, @whikloj<https://github.com/whikloj> .
On a side note, @c1mckay<https://github.com/c1mckay> and @craig-willis<https://github.com/craig-willis> , if you provide me your email addresses, I will send you an invitation to create a JIRA/Wiki account. (awoods@duraspace.org<mailto:awoods@duraspace.org>)

whikloj:
Hmm @acoburn good point. I'll keep looking.

whikloj:
Ok, so I can get the base context using the `$uriInfo.getBaseUri().getPath()`, this gives me `/rest/`, `/fcrepo/rest/`, etc. What is a good way to determine that `/rest` is the jersey path? 

Then I can just change:
```
<script src="$uriInfo.baseUriBuilder.build()../static/js/jquery-1.9.1.min.js" ></script>
```

to
```
<script src="$uriInfo.getBaseUri().getPath().replace(jersey_path, "")/static/js/jquery-1.9.1.min.js" ></script>
```

acoburn:
Why not just get the base path, split it on the "/" char, pop the last element (which will be "rest" or whatever is defined in web.xml -- which may very well be something other than "rest"), and then join the remainder of the list into string, using "/" as a delimiter.

I definitely wouldn't assume that everyone uses "rest". Plus, that value is defined in the fcrepo-webapp, and making assumptions in fcrepo-http-api on what exists in fcrepo-webapp is entirely backwards.

craig-willis:
Thanks, all.  

I just ran a quick test and it looks like using ``$uriInfo.baseUriBuilder.scheme(null).build()`` in place of ``$uriInfo.baseUriBuilder.build()`` could also solve the problem.  In this case, the URLs are constructed [protocol-less](https://stackoverflow.com/questions/4831741/can-i-change-all-my-http-links-to-just):

```
 <script src="//suiuq5-fedoracommons3.workbench.nationaldataservice.org/rest/../static/js/jquery-1.9.1.min.js" ></scrip
```

From my understanding, in this scenario the browser will use the scheme/protocol of the current page.  

Thoughts?

whikloj:
I have take @acoburn's advice and added a new variable to the velocity template context which is the baseUri minus one directory if there is at least one directory (in some of the tests we seem to end up with a baseUri of `http://localhost:8080/`)

@c1mckay @craig-willis Could you see if a relative path to these resources will work with your proxy?

https://github.com/fcrepo4/fcrepo4/compare/master...whikloj:FCREPO-2534?expand=1

awoods:
As of https://github.com/fcrepo4/fcrepo4/pull/1208/commits/6b27613d7086fd2525add8a5d43e8e305fef73f4 , this works with `http://` deployment. I assume it also does with `https://`...since @c1mckay produced the pull-request.
I would be happy to move this pull-request forward unless there are other comments / suggestions.

...after we get an [iCLA and/or cCLA](https://wiki.duraspace.org/display/DSP/Contributor+License+Agreements) on file from @c1mckay .

c1mckay:
The PR consists of craig-willis's proposition to use '.scheme(null)'.  Also, I submitted the legal document to legal@duraspace.org

awoods:
@fcrepo4/committers : any additional comments?

escowles:
+1

whikloj:
I'm not really a fan of leaving the relative `/../` dance in there, but if stripping down to `/<context>/static/...` for static resources is not desirable I'm not going to  hold this up.

whikloj:
@awoods @c1mckay @craig-willis 
https://github.com/fcrepo4/fcrepo4/pull/1215

c1mckay:
I'm investigating it right now

c1mckay:
I can confirm that @whikloj 's solution also works.

awoods:
I like the idea of removing the `/fcr:assets` endpoint offered in https://github.com/fcrepo4/fcrepo4/pull/1215 . 

This change introduces an additional dependency from fcrepo-http-api on fcrepo-webapp in so far as the former would now rely on `common.*` files existing in the `static` directory of the latter... along the lines of the [concern raised](https://github.com/fcrepo4/fcrepo4/pull/1208#issuecomment-315449550) by @acoburn . However, this is not the first introduction of such a dependency.

I am glad to see we have two working solutions.

Does anyone have comments that would inform the choice of going with one solution or the other?

whikloj:
@awoods There is already a dependency in that all of the other CSS and JS files included in the `common-head.vsl` are already served from the `/static` directory and only the `common.js` and `common.css` were served via the `/fcr:assets` endpoint.

whikloj:
@awoods @acoburn however if having the CSS and JS required for the velocity templates in  fcrepo-http-api but located in the fcrepo-webapp is not desirable, then I would suggest modifying the `fcr:assets` endpoint to support all the css and js files and move them **all** from fcrepo-webapp to fcrepo-http-api.

acoburn:
I have very little opinion here. If I did, though, I would suggest moving all static resource into the `fcrepo-webapp` module. In principle, it creates a circular dependency, but it is a pretty loose one at that, and it also allows alternate deployment packages (e.g. https://gitlab.amherst.edu/acdc/amherst-fedora-webapp) to customize those static assets, if needed. If you really wanted to get fancy, you could make the location of the static resources configurable, and that would effectively eliminate the circular dependency.

whikloj:
@acoburn about removing the circular dependency is this because we would we leave the css and js resources (or a version of them) in the fcrepo-http-api and then keep the the current ones in the fcrepo-webapp but also allows for easier replacements?

acoburn:
@whikloj, no, you'd entirely remove the `fcr:assets` endpoint from `fcrepo-http-api`. The velocity template will need to point to some URL for those assets, but if you make that location configurable, then the `fcrepo-webapp` configuration can define that location (e.g. `/static/js/....`). And voila, there is no circular dependency. There would be nothing wrong with setting a default location as `/static/js/.....`.

awoods:
@whikloj and @c1mckay : I am inclined to move forward with https://github.com/fcrepo4/fcrepo4/pull/1215 to address this issue... unless there is a reason not to.

p.s. @c1mckay and @craig-willis , thanks again for motivated this.

c1mckay:
That solution also works with our configuration, so we're okay with that!

awoods:
Superseded by: https://github.com/fcrepo4/fcrepo4/pull/1215

PR #1220, opened 2017-08-21
Add root path to velocity context

awoods:
Add root path to velocity context
    
- Remove ViewAssets
- Remove variables to common.js and common.css locations
    
Resolves: https://jira.duraspace.org/browse/FCREPO-2534


birkland:
I've tested this behind an SSL termination proxy, and it does fix the problem where the browser refuses to load static assets.  

There still seems to be an issue with the UI in Chrome, though.  The Chrome flags the page as insecure, but does not explain why.  It still proceeds with an https connection and renders the page well, but displays an "i" in the corner, instead of the green padlock.   Firefox seems to be perfectly fine.

awoods:
@birkland : This PR has already been committed to `master`.
Is the behavior you are seeing in this port of the fix in the `4.7-maintenance` branch the same as what is seen in `master`?

birkland:
@awoods  Yes.  I think I'm going to approve this anyway (it's a definite improvement), but there's still something wrong in both  master and maintenance-post-this-PR with respect to SSL compatibility of the UI.  Unfortunately, I can't be any more specific than "it makes Chrome unhappy".

birkland:
Ah, found it!

> Mixed Content: The page at 'https://localhost/fcrepo/rest/test' was loaded over a secure connection, but contains a form which targets an insecure endpoint 'http://localhost/fcrepo/rest/test'. This endpoint should be made available over a secure connection.

PR #1228, opened 2017-09-12
Ensures correct response codes for creating LDP-NR

harringj:
- Adds new UnsupportedAlgorithmException and mapper
- Updates FedoraLdp class to throw new exception appropriately
- Updates FedoraLdpTest and FedoraLdpIT

Resolves: https://jira.duraspace.org/browse/FCREPO-2593

This PR is a contribution toward the API Alignment sprint (9/11/17-9/22/17).

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -20,13 +20,13 @@ package org.fcrepo.http.api;
 
 import static com.google.common.base.Strings.nullToEmpty;
 import static java.nio.charset.StandardCharsets.UTF_8;
-import static javax.ws.rs.core.MediaType.TEXT_HTML;
-import static javax.ws.rs.core.MediaType.TEXT_PLAIN_TYPE;
-import static javax.ws.rs.core.MediaType.WILDCARD;
 import static javax.ws.rs.core.HttpHeaders.ACCEPT;
 import static javax.ws.rs.core.HttpHeaders.CONTENT_DISPOSITION;
 import static javax.ws.rs.core.HttpHeaders.CONTENT_TYPE;
 import static javax.ws.rs.core.HttpHeaders.LINK;
+import static javax.ws.rs.core.MediaType.TEXT_HTML;
+import static javax.ws.rs.core.MediaType.TEXT_PLAIN_TYPE;
+import static javax.ws.rs.core.MediaType.WILDCARD;
 import static javax.ws.rs.core.Response.created;
 import static javax.ws.rs.core.Response.noContent;
 import static javax.ws.rs.core.Response.notAcceptable;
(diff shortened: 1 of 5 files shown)


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/exception/UnsupportedAlgorithmException.java
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.kernel.api.exception;
+
+/**
+ * @author harring
+ * @since 2017-09-12
+ */
+public class UnsupportedAlgorithmException extends Exception {
```
birkland: Would it be possible to just use the built-in `java.security.NoSuchAlgorithmException` here?  It's used "when a particular cryptographic algorithm is requested but is not available in the environment" -- do the semantics of `UnsupportedAlgorithmException` diverge much from that?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -595,7 +598,6 @@ private void checkForInsufficientStorageException(final Throwable rootThrowable,
      * @return 204 No Content (for updated resources), 201 Created (for created resources) including the resource
      *    URI or content depending on Prefer headers.
      */
-    @SuppressWarnings("resource")
```
birkland: With this annotation removed, there appears to be a new warning of a potential resource leak.

fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
```diff
@@ -692,7 +691,7 @@ public void testGetWithExternalMessageBinary() throws Exception {
 
 
     @Test
-    @SuppressWarnings({"resource", "unchecked"})
```
birkland: There is now a warning about a potential resource leak

fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
```diff
@@ -839,7 +838,7 @@ public void testPatchObject() throws Exception {
...
```
birkland: There is now a warning about a potential resource leak

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/exception/UnsupportedAlgorithmException.java
harringj: Hmm, I guess I'm not sure whether we'd want a security-related exception to be thrown here. Are we thinking of checksums as a security feature here or as more related to fixity?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/UnsupportedAlgorithmExceptionMapper.java
```diff
@@ -0,0 +1,56 @@
...
+public class UnsupportedAlgorithmExceptionMapper implements
+        ExceptionMapper<UnsupportedAlgorithmException>, ExceptionDebugLogging {
+
+    private static final Logger LOGGER =
+            getLogger(UnsupportedAlgorithmExceptionMapper.class);
+
+    @Override
+    public Response toResponse(final UnsupportedAlgorithmException e) {
+
+        LOGGER.error("InvalidChecksumException intercepted by UnsupportedAlgorithmExceptionMapper: {}\n",
```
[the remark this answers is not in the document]
harringj: Ok, removed that logging statement.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
harringj: Sorry, Eclipse was complaining about this not being needed. I'll just leave these sorts of things alone in future :-)

fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
harringj: Fixed

fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
harringj: Fixed

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/exception/UnsupportedAlgorithmException.java
birkland: Ah, I see.  It's because it is being used in a very specific circumstance, which demands a very specific response.  `NoSuchAlgorithmException` may be too general - i.e if it's thrown while doing a non-fixity-related operation, we might want a different response.

birkland:
@harringj Looks good from the comments.  Once you push the changes, we should be good to merge.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/exception/UnsupportedAlgorithmException.java
birkland: @harringj  Would you be able to update the Javadoc for this class, with something like 

<pre>
/** Client uses an unsupported hash algorithm 
* &lt;p&gt;
* Thrown in circumstances where a client has used an unknown or unsupported algorithm in a request,
* e.g. with `Digest` or `Want-Digest`.
* &lt;/p&gt;
* @author harring 
* @since 2017-09-12  
*/ 
public class UnsupportedAlgorithmException extends Exception { 
</pre>

harringj: Oh, my bad for overlooking the javadoc on this class!

PR #1229, opened 2017-09-12
Use Link header to determine interaction model, instead of mime type sniffing

escowles:
Use `Link rel="type"` header to determine interaction type, instead of checking the MIME type
and `Content-Disposition` headers.

Fixes https://jira.duraspace.org/browse/FCREPO-2584

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -59,8 +59,6 @@ import static org.fcrepo.http.commons.domain.RDFMediaType.TEXT_PLAIN_WITH_CHARSE
 import static org.fcrepo.http.commons.domain.RDFMediaType.TURTLE;
 import static org.fcrepo.http.commons.domain.RDFMediaType.TURTLE_WITH_CHARSET;
 import static org.fcrepo.http.commons.domain.RDFMediaType.TURTLE_X;
-import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_BINARY;
-import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_CONTAINER;
 import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_PAIRTREE;
 import static org.fcrepo.kernel.api.RdfLexicon.BASIC_CONTAINER;
 import static org.fcrepo.kernel.api.RdfLexicon.DIRECT_CONTAINER;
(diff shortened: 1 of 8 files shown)


escowles:
@birkland My reading of [3.3 Resource Management, HTTP POST](https://fcrepo.github.io/fcrepo-specification/#httpPOST) is that containers must specify their default interaction model in a constraints document (see PR #1230), and the default interaction model specified there should be used if the client doesn't specify one.

I'm open to the idea of falling back on the old behavior (I suppose since the constraints document is a note, we could describe the fallback process there).  If we do update the constraints document, I think it would be something like: 

> If no interaction model is supplied via a Link header, then the &lt;http://www.w3.org/ns/ldp#BasicContainer&gt; interaction model will be used if the `Content-Type` value is an RDF type (application/ld+json, application/n-triples, application/rdf+xml, text/n3, text/rdf+n3, text/turtle, or application/x-turtle) and a `Content-Disposition: attachment` header is not present.  If there is no content, a non-RDF `Content-Type`, or if a `Content-Disposition: attachment` header is present, the &lt;http://www.w3.org/ns/ldp#NonRDFSource&gt; interaction model will be used.

I think this is one area where the specified behavior is very different from the current implementation, so I'm not sure whether it's helpful or just confusing.  Just writing that text was hard and confusing for me, and I'm not at all sure that an implementer would be able to understand the precedence there.

For the HTML UI, I think we can just add the interaction model header in https://github.com/fcrepo4/fcrepo4/blob/container-constraints/fcrepo-webapp/src/main/webapp/static/js/common.js#L76.

birkland:
@escowles  How about this:
> If no interaction model is provided by the client, the server will choose a default value implied by `Content-Type`.  RDF Content-Types (e.g application/ld+json, application/n-triples, application/rdf+xml, text/n3, text/rdf+n3, text/turtle, or application/x-turtle) will default to &lt;http://www.w3.org/ns/ldp#BasicContainer&gt;.  Non-RDF content-types will default to &lt;http://www.w3.org/ns/ldp#NonRDFSource&gt;.   Empty `POST`s or `PUT`s that do not specify a `Content-Type` will default to a &lt;http://www.w3.org/ns/ldp#BasicContainer&gt;

If the spec is implying that a server _shouldn't_ do this, then perhaps we should clarify with the broader spec group.  Providing a default answer of RDFSource vs NonRDFSource based on Content-Type is something that pretty much every read/write LDP server does, as far as I am aware.

escowles:
@birkland Yes, I think that text is better.  Crucially, I think we can leave out the Content-Disposition (since you can now specify the interaction model to override the default behavior).

I'll update this PR to fall back on Content-Type, and then revise the constraints document in #1230.

PR #1231, opened 2017-09-13
Support Want-Digest header with fixity check for NonRDFSource.

lsitu:
Fixes https://jira.duraspace.org/browse/FCREPO-2601

Add support for Want-Digest header with fixity check on NonRDFSource.

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -794,7 +794,6 @@ public class FedoraLdp extends ContentExposingResource {
         // handle the Want-Digest header with fixity check
         final Collection<String> preferredDigests = parseWantDigestHeader(wantDigest);
         if (preferredDigests.isEmpty()) {
-            LOGGER.warn("Unsupported digest algorithm provided in 'Want-Digest' header: {}", wantDigest);
             throw new UnsupportedAlgorithmException(
                     "Unsupported digest algorithm provided in 'Want-Digest' header: " + wantDigest);
         }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -19,6 +19,7 @@ package org.fcrepo.http.api;
 
 
 import static com.google.common.base.Strings.nullToEmpty;
+import static com.google.common.base.Strings.isNullOrEmpty;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static javax.ws.rs.core.HttpHeaders.ACCEPT;
 import static javax.ws.rs.core.HttpHeaders.CONTENT_DISPOSITION;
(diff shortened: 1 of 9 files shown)


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/BasicCacheEntry.java
```diff
@@ -17,6 +17,7 @@
  */
 package org.fcrepo.kernel.modeshape.utils;
 
+import org.apache.http.impl.auth.UnsupportedDigestAlgorithmException;
```
birkland: I believe this should be `org.fcrepo.kernel.api.exception.UnsupportedAlgorithmException`.

When I do:
`curl -H "Want-Digest: sha256" -I http://localhost:8080/rest/img.jpg`

I get a 500 error, caught by the wildcard exception mapper.  The exception is:
<pre>
DEBUG 16:53:31.225 (WildcardExceptionMapper) WildcardExceptionMapper intercepted exception:{}

org.apache.http.impl.auth.UnsupportedDigestAlgorithmException: Unsupported digest algorithm: sha256
        at org.fcrepo.kernel.modeshape.utils.BasicCacheEntry.checkFixity(BasicCacheEntry.java:111)
        at org.fcrepo.kernel.modeshape.FedoraBinaryImpl.checkFixity(FedoraBinaryImpl.java:392)
        at org.fcrepo.http.api.FedoraLdp.handleWantDigestHeader(FedoraLdp.java:802)
        at org.fcrepo.http.api.FedoraLdp.head(FedoraLdp.java:217)
</pre>

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -770,6 +789,22 @@ private String mintNewPid(final String slug) {
         return pid;
     }
 
+    private String handleWantDigestHeader(final FedoraBinary binary, final String wantDigest)
+            throws UnsupportedAlgorithmException {
+        // handle the Want-Digest header with fixity check
+        final Collection<String> preferredDigests = parseWantDigestHeader(wantDigest);
+        if (preferredDigests.isEmpty()) {
+            LOGGER.warn("Unsupported digest algorithm provided in 'Want-Digest' header: {}", wantDigest);
```
birkland: `warn` seems to be too high of a log level.  The exception will already be logged at the `DEBUG` level by the exception handler for `UnsupportedAlgorithmException`, so I think you can just remove this

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/BasicCacheEntry.java
lsitu: Good catch! Thanks. I've added a commit to change the exception to UnsupportedAlgorithmException with a test. See https://github.com/fcrepo4/fcrepo4/pull/1231/commits/c306e2f9d6cfa2f7241becbe7f29f1cd1ed6f824.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
lsitu: Sure, the warn logging is removed. See commit c306e2f.

PR #1232, opened 2017-09-13
PR is a two in one:  fcrepo-2602 and fcrepo-2603

dbernstein:
ActivityStreams JSON is now used for message bodies of transmitted events.  
The new ActivityStreams based JSON-LD uses  proper actor values for emitted messages

# Resolves:
https://jira.duraspace.org/browse/FCREPO-2602
   https://jira.duraspace.org/browse/FCREPO-2603

# What does the PR do? 

This commit retools the message bodies to use the ActivityStreams vocabulary.  In addition to reworking the body by enabling the -Dfcrepo.auth.webac.userAgent.baseUri=http://your-user-agent-base-uri/ flag on your fcrepo4 instance you can ensure that the value of the "actor" property will conform to the URLs used by SOLID WebAC. 

# What's new?
Since the format of the message body changed, any consumers of the messages will need to be retooled accordingly.

# How should this be tested?
Unit tests cover the serialization and deserialization processes. If you would like to verify the contents of the generated messages you can use the following wiki to view the events generated by fedora:
https://wiki.duraspace.org/display/FEDORA4x/How+to+Inspect+Event+Messages+Generated+by+Fedora
If you do not use the -Dfcrepo.auth.webac.userAgent.baseUri flag the actor value should appear as "#<username>".
# Additional Notes:
* Breaking change:  Downstream consumers of messages will need to be updated to work with new message format.
* [Documentation will need to be updated] (https://jira.duraspace.org/browse/FCREPO-2606)
# Interested parties


fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/JsonLDEventMessage.java
```diff
@@ -34,85 +33,29 @@
...
-    static class Activity {
         @JsonProperty("type")
         public List<String> type;
 
-        @JsonProperty("identifier")
-        public String identifier;
+        @JsonProperty("id")
+        public String id;
 
-        @JsonProperty("atTime")
```
birkland: Is there some way we can keep the timestamps in the message, to convey when the activity happened?  

There are some options in the [activityStreams vocabulary](https://www.w3.org/TR/activitystreams-vocabulary), but maybe we should bounce this off @escowles , to see if the spec should say something about it.

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/JsonLDEventMessage.java
```diff
@@ -122,53 +65,50 @@ public Agent(final String type, final String name) {
...
-        final List<String> types = evt.getTypes().stream().map(rdfType -> rdfType.getType()).collect(toList());
-        if (!types.contains(PROV_NAMESPACE + "Activity")) {
-            types.add(PROV_NAMESPACE + "Activity");
-        }
-        msg.wasGeneratedBy = new Activity(evt.getEventID(), evt.getDate(), types);
-
-        if (baseUrl == null) {
-            msg.id = "info:fedora" + evt.getPath();
-        } else {
-            msg.isPartOf = baseUrl;
```
birkland: It's OK to keep an `isPartOf` in the message, as is illustrated in the [basic example](https://fcrepo.github.io/fcrepo-specification/#basic-notification-example).  I wish this was in the spec as a `SHOULD`.   Your call @dbernstein, we could also just create another ticket to add this in if it makes sense in principle to do so, but there are questions, etc

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/JsonLDEventMessage.java
```diff
@@ -122,53 +65,50 @@ public Agent(final String type, final String name) {
...
-            msg.id = baseUrl + evt.getPath();
-        }
+        final String userAgent = evt.getUserID();
+        msg.id = evt.getEventID();
+        final String userAgentBaseUri = System.getProperty(USER_AGENT_BASE_URI_PROPERTY, "#");
```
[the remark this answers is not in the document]
awoods: I do not believe we will be requiring actors be identified by WebIDs:
https://github.com/fcrepo/fcrepo-specification/pull/207/files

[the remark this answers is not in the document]
birkland: @awoods but the question of "where are URIs for users coming from"in order to establish the proper identity of the actor is still open, right?

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/JsonLDEventMessage.java
escowles: I think we can use the https://www.w3.org/TR/activitystreams-vocabulary/#dfn-starttime and https://www.w3.org/TR/activitystreams-vocabulary/#dfn-endtime properties (possibly with the same value ).

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/JsonLDEventMessage.java
dbernstein: @birkland : Leaving it in makes sense.  I'll do that.

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/JsonLDEventMessage.java
dbernstein: 👍

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/JsonLDEventMessage.java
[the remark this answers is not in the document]
dbernstein: @awoods  @birkland  : so I'll leave what's in there now.  FCREPO-2608 will take care of the rest.

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/JsonLDEventMessage.java
```diff
@@ -122,53 +65,50 @@ public Agent(final String type, final String name) {
...
+        msg.actor = userAgentBaseUri  + userAgent;
```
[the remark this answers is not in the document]
dbernstein: @birkland :  okay - this makes sense. I'll restore those values.

fcrepo-event-serialization/src/test/java/org/fcrepo/event/serialization/JsonLdEventTest.java
```diff
@@ -104,43 +111,29 @@ public void testJsonSerializationAsModel() {
...
+        assertTrue(model.contains(eventSubject, type, createResource(EventType.RESOURCE_MODIFICATION.getType())));
+        assertTrue(model.contains(eventSubject, createProperty(ACTIVITY_STREAMS_NAMESPACE + "actor"), createResource(
+                getAgentIRI())));
+        assertTrue(model.contains(eventSubject, createProperty(ACTIVITY_STREAMS_NAMESPACE + "object"),
+                resourceSubject));
+
+        final AtomicInteger name = new AtomicInteger();
+        model.listStatements(new SimpleSelector(null, createProperty(ACTIVITY_STREAMS_NAMESPACE + "name"), blankNode))
+                .forEachRemaining(statement -> {
+                    assertEquals(EventType.RESOURCE_MODIFICATION.getName(), statement.getString());
```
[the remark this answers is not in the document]
dbernstein: @birkland:  cool -  thanks for the clarification.

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/JsonLDEventMessage.java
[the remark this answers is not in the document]
birkland: @dbernstein Yes.  To be precise, 2608 will make a method from `fcrepo-kernel-api` available, and 2603 will be to use that in order to populate the actor of messages (removing/obsoleting the code here)

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/JsonLDEventMessage.java
dbernstein: @escowles :  I'm leaning towards "published"  - seems more clear - the time the event was published (think pub/sub)  looks like endtime and starttime are more geared towards events that span time (e.g. social events).

birkland: @dbernstein "published" seems like a better fit to the kind of data we have available.  The Fedora impl does not keep track of actions' duration, so to say that the internal timestamp Fedora is using equates to start/end times may be misleading.   This will allow us to later on add start and end times when/if it makes sense to do so (like for long-running operations where it might matter).

escowles: @dbernstein Yes, "published" on the event object sounds like a better fit.

fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/JsonLDEventMessage.java
```diff
@@ -61,114 +71,138 @@ public ContextElement(final String id, final String type) {
...
+        actor.add(new Person(userAgentBaseUri + userAgent, Arrays.asList("Person")));
+        final String softwareAgent = evt.getInfo().get(USER_AGENT);
+        if (softwareAgent != null) {
+            actor.add(new Application(softwareAgent, Arrays.asList("Application")));
+        }
+
+        final JsonLDEventMessage msg = new JsonLDEventMessage();
+
+        msg.id = evt.getEventID();
+        msg.isPartOf = baseUrl;
```
birkland: This results in a message that states that the _activity_ is part of the parent resource, or a triple like:

    <urn:uuid:291d809e-cbd5-41c9-939e-5819332bc3b8> dct:isPartOf <http://localhost:8080/rest> .

Instead, we want to say that _the resource relevant to the event_ is `partOf` its parent container, or 

    <http://localhost:8080/rest/test> dct:isPartOf <http://localhost:8080/rest>

So the `isPartOf` goes in the `object`

dbernstein: ah - good catch - will fix that right up.

birkland: Sorry to bug you once more, but the resource `isPartOf` points to doesn't seem to be correct.  In my testing, I got a message:
<pre>
{
  "id": "urn:uuid:daa486e3-a8ac-43ec-9ac8-f511acb3dc80",
  "type": [
    "Delete"
  ],
  "name": "delete resource",
  "published": "2017-09-21T19:51:19.877Z",
  "actor": [
    {
      "type": [
        "Person"
      ],
      "id": "#bypassAdmin"
    },
    {
      "type": [
        "Application"
      ],
      "name": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36 Edge/15.15063"
    }
  ],
  "object": {
    "type": [
      "http://fedora.info/definitions/v4/repository#Binary",
      "http://www.w3.org/ns/ldp#NonRDFSource",
      "http://fedora.info/definitions/v4/repository#Resource",
      "http://fedora.info/definitions/v4/repository#NonRdfSourceDescription",
      "http://www.w3.org/ns/prov#Entity"
    ],
    <b>"id": "http://localhost:8080/rest/objects/raven/test.sh",
    "isPartOf": "http://localhost:8080/rest"</b>
....
</pre>

So `isPartOf` in this example should point to its parent, `http://localhost:8080/rest/objects/raven` instead of the root repository resource

birkland: Actually, never mind!  The old code does this too.  I think it's a bug, though.  Will create a separate issue to address it.

PR #1236, opened 2017-09-15
Adds support for depth header to DELETE action

harringj:
- Adds logic to FedoraLdp.deleteObject() to check depth header for
containers
- Adds two new IT test cases
- Adds helper method to abstract IT class

Resolves: https://jira.duraspace.org/browse/FCREPO-2600

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -326,7 +326,7 @@ public class FedoraLdp extends ContentExposingResource {
     @DELETE
     @Timed
     public Response deleteObject() {
-        FedoraResource resource = resource();
+        final FedoraResource resource = resource();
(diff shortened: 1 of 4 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -296,14 +297,24 @@ private int getChildrenLimit() {
     @DELETE
     @Timed
     public Response deleteObject() {
-        evaluateRequestPreconditions(request, servletResponse, resource(), session);
+        FedoraResource resource = resource();
```
harringj: Just a note on storing resource() in a variable for this method. This change was not strictly speaking in scope for this ticket, but it seemed like an easy and reasonable small gain in efficiency not to call resource() three times, when once would do :-)

birkland:
Obsoleted by https://github.com/fcrepo4/fcrepo4/pull/1237

PR #1237, opened 2017-09-15
Fcrepo 2600 new

harringj:
Adds support for depth headers to http DELETE method

- Adds logic to FedoraLdp.deleteObject() to check depth header for
containers
- Adds two new IT test cases
- Adds helper method to abstract IT class

Resolves: https://jira.duraspace.org/browse/FCREPO-2600

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -327,10 +327,14 @@ public class FedoraLdp extends ContentExposingResource {
     @Timed
     public Response deleteObject() {
 <<<<<<< HEAD
+<<<<<<< HEAD
         final FedoraResource resource = resource();
 =======
         FedoraResource resource = resource();
 >>>>>>> 406cd50... Adds support for depth header to DELETE action
+=======
+        final FedoraResource resource = resource();
+>>>>>>> d6882f6... corrects checkstyle errors
         if (resource instanceof ContainerImpl) {
             final String depth = headers.getHeaderString("Depth");
             System.out.println(depth);
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -326,7 +326,11 @@ public class FedoraLdp extends ContentExposingResource {
     @DELETE
     @Timed
     public Response deleteObject() {
+<<<<<<< HEAD
         final FedoraResource resource = resource();
+=======
+        FedoraResource resource = resource();
+>>>>>>> 406cd50... Adds support for depth header to DELETE action
         if (resource instanceof ContainerImpl) {
             final String depth = headers.getHeaderString("Depth");
             System.out.println(depth);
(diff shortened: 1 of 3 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -319,14 +318,24 @@ private int getChildrenLimit() {
     @DELETE
     @Timed
     public Response deleteObject() {
-        evaluateRequestPreconditions(request, servletResponse, resource(), session);
+        final FedoraResource resource = resource();
+        if (resource instanceof ContainerImpl) {
```
birkland: Unfortunately, this brings a modeshape implementation class into `fcrepo-http-api`.   `ContainerImpl` should be removed and replace with something else.  I think  plain ol' `org.fcrepo.kernel.api.models.Container` would work.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -319,14 +318,24 @@ private int getChildrenLimit() {
...
+            final String depth = headers.getHeaderString("Depth");
+            System.out.println(depth);
```
birkland: Remove the println, or log as `DEBUG`

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -319,14 +318,24 @@ private int getChildrenLimit() {
...
+                throw new ClientErrorException("Depth header, if present, must be set to 'infinity' for containers",
+                        SC_BAD_REQUEST);
+            }
+        }
+
+        evaluateRequestPreconditions(request, servletResponse, resource, session);
 
         LOGGER.info("Delete resource '{}'", externalPath);
 
-        final AcquiredLock lock = lockManager.lockForDelete(resource().getPath());
```
birkland: Actually, probably leave it as `resource()` for consistency sake with everything else in this class.  The implementation looks like it simply retrieves a field anyway (i.e. only does a lookup once).

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -319,14 +318,24 @@ private int getChildrenLimit() {
...
+        final AcquiredLock lock = lockManager.lockForDelete(resource.getPath());
 
         try {
-            resource().delete();
+            resource.delete();
```
birkland: ditto, just leave as `resource()`

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
harringj: Fixed

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
harringj: Sorry, meant to delete this

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
harringj: Ok

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
harringj: Ok

PR #1240, opened 2017-09-19
Fcrepo 2609

harringj:
- copies AccessRolesProvider and AbstractRolesAuthorizationDelegate
classes into module

- updates refs to these classes in existing code

Resolves: https://jira.duraspace.org/browse/FCREPO-2609

### DIFF
--- a/fcrepo-auth-webac/pom.xml
+++ b/fcrepo-auth-webac/pom.xml
@@ -51,12 +51,6 @@
 
     <dependency>
       <groupId>org.fcrepo</groupId>
-      <artifactId>fcrepo-auth-roles-common</artifactId>
-      <version>5.0.0-SNAPSHOT</version>
-    </dependency>
-
-    <dependency>
-      <groupId>org.fcrepo</groupId>
       <artifactId>fcrepo-kernel-api</artifactId>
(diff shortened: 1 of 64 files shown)


harringj:
Please note that one test case in WebAcRecipesIT currently fails, namely, testNoAccessRolesResource. It expects a 404 SC_NOT_FOUND, but returns 204 SC_NO_CONTENT. I am not quite sure what the desired behavior is here. All other unit and IT test cases pass, which makes me think maybe this test case is not expecting the right value.

FYI @birkland @dbernstein

birkland:
@harringj I think the problem is that the rbacl module had a class `AccessRoles` which defined a `fcr:accessroles` endpoint.  That's not part of the API, and looks like it might take some effort to get working properly, due to the fact that userids are now URIs.  

So, I propose _deleting_ that test, and flagging this issue as a breaking change.  Does that seem logical?

harringj:
I see. Yes, @birkland , that sounds reasonable to me.

harringj:
Thanks for reviewing and for the feedback, @birkland! I made those changes you requested, and when Travis is done, it sounds like we should be good to go :-)

PR #1241, opened 2017-09-19
Unsupported external-body access-types now result in 415 Unsupported …

dbernstein:
…Media Type.

Resolves: https://jira.duraspace.org/browse/FCREPO-2586

**JIRA Ticket**: (link)

It ensures that unsupported external-body access-types now result in 415 Unsupported Media Type.

# How should this be tested?
0. build and fire up fedora.
1. run `curl -XPUT -H"Content-Type: message/external-body; access-type=ftp; NAME=\"/some/file\"; site=\"example.com\"" -i localhost:8080/rest/external`
2. Verify that 415 http is returned.

# Additional Notes:
If a fedora binary has an invalid external-body content type associated with it,  a GET request for that resource will fail with a 415.  Previously GET requests would not fail.  

# Interested parties
@escowles

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -22,8 +22,8 @@ import static java.util.EnumSet.of;
 import static java.util.stream.Stream.concat;
 import static java.util.stream.Stream.empty;
 import static javax.ws.rs.core.HttpHeaders.CACHE_CONTROL;
-import static javax.ws.rs.core.HttpHeaders.CONTENT_LENGTH;
 import static javax.ws.rs.core.HttpHeaders.CONTENT_DISPOSITION;
+import static javax.ws.rs.core.HttpHeaders.CONTENT_LENGTH;
 import static javax.ws.rs.core.HttpHeaders.CONTENT_LOCATION;
 import static javax.ws.rs.core.HttpHeaders.CONTENT_TYPE;
 import static javax.ws.rs.core.HttpHeaders.LINK;
@@ -45,9 +45,9 @@ import static org.fcrepo.kernel.api.RdfLexicon.BASIC_CONTAINER;
 import static org.fcrepo.kernel.api.RdfLexicon.CONSTRAINED_BY;
 import static org.fcrepo.kernel.api.RdfLexicon.CONTAINER;
 import static org.fcrepo.kernel.api.RdfLexicon.DIRECT_CONTAINER;
+import static org.fcrepo.kernel.api.RdfLexicon.HAS_MEMBER_RELATION;
 import static org.fcrepo.kernel.api.RdfLexicon.INDIRECT_CONTAINER;
 import static org.fcrepo.kernel.api.RdfLexicon.LDP_NAMESPACE;
-import static org.fcrepo.kernel.api.RdfLexicon.HAS_MEMBER_RELATION;
 import static org.fcrepo.kernel.api.RdfLexicon.isManagedNamespace;
 import static org.fcrepo.kernel.api.RdfLexicon.isManagedPredicate;
 import static org.fcrepo.kernel.api.RequiredRdfContext.EMBED_RESOURCES;
@@ -70,6 +70,7 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -80,6 +81,7 @@ import javax.servlet.http.HttpServletResponse;
 import javax.servlet.ServletContext;
 import javax.ws.rs.BadRequestException;
 import javax.ws.rs.BeanParam;
+import javax.ws.rs.NotSupportedException;
 import javax.ws.rs.core.CacheControl;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.EntityTag;
@@ -136,6 +138,7 @@ import com.google.common.annotations.VisibleForTesting;
  */
 public abstract class ContentExposingResource extends FedoraBaseResource {
 
+    private static final String URL_ACCESS_TYPE = "URL";
     private static final Logger LOGGER = getLogger(ContentExposingResource.class);
     public static final MediaType MESSAGE_EXTERNAL_BODY = MediaType.valueOf("message/external-body");
 
@@ -196,7 +199,8 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
             final MediaType mediaType = MediaType.valueOf(((FedoraBinary) resource()).getMimeType());
 
             if (isExternalBody(mediaType)) {
-                return externalBodyRedirect(URI.create(mediaType.getParameters().get("URL"))).build();
+                checkExternalBody(mediaType);
+                return externalBodyRedirect(getURLAccessType(mediaType)).build();
             }
 
             return getBinaryContent(rangeValue);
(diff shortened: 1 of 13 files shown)


dbernstein:
Once this one is in,  https://jira.duraspace.org/browse/FCREPO-2585 is ready to go (but based on this PR).

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -136,6 +138,11 @@
  */
 public abstract class ContentExposingResource extends FedoraBaseResource {
 
+    /**
```
awoods: Remove dead comment section.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -214,11 +222,58 @@ protected Response getContent(final String rangeValue,
...
+
+    protected URI getURLAccessType(final MediaType mediaType) {
+        return URI.create(getAccessTypeValue(mediaType.getParameters(), URL_ACCESS_TYPE));
+    }
+
+    /**
+     * Throws a NotSupportedException if media type is an invalid message/external-body type. To be valid it must
+     * contain the access-type parameter that equals a valid access type and a URL parameter that contains a URL.
+     * 
+     * @param mediaType
```
awoods: Need `@param` description.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -214,11 +219,58 @@ protected Response getContent(final String rangeValue,
...
+
+    /**
+     * Throws a NotSupportedException if media type is an invalid message/external-body type. To be valid it must
+     * contain the access-type parameter that equals a valid access type and a URL parameter that contains a URL.
+     * 
+     * @param mediaType The media type to be validated
+     * @throws NotSupportedException
+     */
+    protected void checkExternalBody(final MediaType mediaType) {
+        if (isExternalBody(mediaType)) {
```
whikloj: All uses of this function seem to have been  in a `if (isExternalBody(foo)) {` already, so this seems to be a redundant check.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -214,11 +219,58 @@ protected Response getContent(final String rangeValue,
         return ok(outputStream).build();
     }
 
+
+    protected URI getURLAccessType(final MediaType mediaType) {
+        return URI.create(getAccessTypeValue(mediaType.getParameters(), URL_ACCESS_TYPE));
```
whikloj: `getAccessTypeValue()` seems to only get called here and can be replaced with a stream. So maybe this function should just be
```
protected URI getURLAccessType(final MediaType mediaType) {
    final String accessType = mediaType.getParameters().entrySet().stream()
            .filter(t -> t.getKey().equalsIgnoreCase(URL_ACCESS_TYPE)).map(Map.Entry::getValue)
            .findFirst().orElse(null);
    return URI.create(accessType);
}
```
However `URI.create(null)` is probably not helpful, so maybe we should throw an Exception there?

Also a unit test of this function might be helpful?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -214,11 +219,58 @@ protected Response getContent(final String rangeValue,
...
+
+            if (getAccessTypeValue(params, accessType) == null) {
+                throw new BadRequestException("When specifying the  " + accessType +
+                        " access type you must include a " + URL_ACCESS_TYPE + " key/value pair: " + mediaType
+                                .toString());
+            }
+        }
+    }
+
+    protected String getAccessTypeValue(final Map<String, String> params, final String accessType) {
```
whikloj: As described in `getURLAccessType()` I think this can be moved there and removed, unless there are other uses?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -214,11 +219,58 @@ protected Response getContent(final String rangeValue,
...
+                        MESSAGE_EXTERNAL_BODY + " content type:" +
+                        mediaType.toString());
+            }
+
+            if (!accessType.equalsIgnoreCase(URL_ACCESS_TYPE)) {
+                throw new NotSupportedException("Unsupported media type:  the " + accessType +
+                        " access type  is not supported: " + mediaType.toString());
+            }
+
+            if (getAccessTypeValue(params, accessType) == null) {
```
whikloj: Oh my bad, here is a call. 

So you could replace this check with the stream above but instead of `orElse(null);` use
`.orElseThrow(() -> new BadRequestException("When specifying the  " + accessType +
                        " access type you must include a " + URL_ACCESS_TYPE + " key/value pair: " + mediaType
                        .toString()));`


I dunno, I'm told streams are better than loops.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
whikloj: Or perhaps `isExternalBody()` and `checkExternalBody()` should be one combined function?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
whikloj: Actually isn't this check the same as the one on line 238?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -214,11 +219,58 @@ protected Response getContent(final String rangeValue,
...
+                throw new BadRequestException("When specifying the  " + accessType +
+                        " access type you must include a " + URL_ACCESS_TYPE + " key/value pair: " + mediaType
+                                .toString());
+            }
+        }
+    }
+
+    protected String getAccessTypeValue(final Map<String, String> params, final String accessType) {
+        for (String key : params.keySet()) {
```
awoods: There seems to be a fair amount of overlap the @lsitu's [recent PR](https://github.com/fcrepo4/fcrepo4/commit/6c07231df4395065d206843fd2d08f131a1e5bb6).

Most notably: 
https://github.com/fcrepo4/fcrepo4/commit/6c07231df4395065d206843fd2d08f131a1e5bb6#diff-12a9a89ed2d5d4efac9920aeb1579a1aR74

@dbernstein , would you mind reviewing the above PR and refactoring this current work to reduce redundant code?

dbernstein: @awoods - sure thing.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
dbernstein: good points @whikloj.  I'll address as part of the refactoring requested below.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraBinaryImpl.java
```diff
@@ -394,18 +395,10 @@ public RdfStream getFixity(final IdentifierConverter<Resource, FedoraResource> i
 
             final String mimeType = getMimeType();
             if (mimeType.contains(CacheEntryFactory.MESSAGE_EXTERNAL_BODY)) {
```
awoods: The constant `message/external-body` is defined in two classes:
- CacheEntryFactory.java , and
- MessageExternalBodyContentType.java

I would suggest we refactor that into one location.

dbernstein:
@awoods :  all good feedback. I'll make the mods shortly.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -214,11 +216,28 @@ protected Response getContent(final String rangeValue,
         return ok(outputStream).build();
     }
 
+
+    protected URI getExternalResourceLocation(final MediaType mediaType) throws UnsupportedAccessTypeException {
```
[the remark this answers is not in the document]
dbernstein: @awoods It would be, if it didn't mean adding the javax.ws.rs-api dependency to the fcrepo-kernel-api pom in order to make use of MediaType.   It seems to me that since the method is only used in the fcrepo-http-api layer.  maybe we should keep it as is.   I'm willing to change it if you think the added encumbrance is worth it.  Your call.

fcrepo-kernel-api/src/test/java/org/fcrepo/kernel/api/utils/MessageExternalBodyContentTypeTest.java
```diff
@@ -45,25 +44,39 @@ public void testParseExternalBody() throws UnsupportedAccessTypeException {
...
         MessageExternalBodyContentType.parse(
-                "invalid/mimetype;" + " access-type=url; URL=\"\"");
+                "invalid/mimetype; access-type=url; URL=\"\"");
     }
 
     @Test(expected = UnsupportedAccessTypeException.class)
     public void testParseExternalBodyMissingURL() throws UnsupportedAccessTypeException {
         MessageExternalBodyContentType.parse(
-                "invalid/mimetype;" + " access-type=url;");
+                "message/mimetype; access-type=url;");
```
awoods: This should be `message/external-body`

dbernstein:
@awoods :  I think this one is finally ready - would you mind merging this one when you get a chance?  fcrepo-2585 is more or less ready once this once goes in.

whikloj:
All previous comments seem to have been addressed. This works as described.

PR #1242, opened 2017-09-20
Make user identities available as URIs in kernel API.

lsitu:
Fixes #https://jira.duraspace.org/browse/FCREPO-2608

Make user identities available as URIs in kernel API through method getUserAgent().

### DIFF
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
@@ -25,6 +25,7 @@ import static java.time.Instant.now;
 import static java.util.Collections.singleton;
 import static java.util.Optional.of;
 import static java.util.UUID.randomUUID;
+import static org.slf4j.LoggerFactory.getLogger;
 
 import java.net.URI;
 import java.time.Duration;
@@ -44,15 +45,16 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.annotations.VisibleForTesting;
 import org.fcrepo.kernel.api.FedoraSession;
 import org.fcrepo.kernel.api.exception.AccessDeniedException;
-import org.fcrepo.kernel.api.exception.RepositoryConfigurationException;
 import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
 import org.fcrepo.kernel.modeshape.utils.NamespaceTools;
+import org.slf4j.Logger;
 
 /**
  * An implementation of the FedoraSession abstraction
  * @author acoburn
  */
 public class FedoraSessionImpl implements FedoraSession {
+    private static final Logger LOGGER = getLogger(FedoraSessionImpl.class);
 
     // The default timeout is 3 minutes
     @VisibleForTesting
@@ -63,6 +65,9 @@ public class FedoraSessionImpl implements FedoraSession {
 
     @VisibleForTesting
     public static final String USER_AGENT_BASE_URI_PROPERTY = "fcrepo.auth.webac.userAgent.baseUri";
+
+    @VisibleForTesting
+    public static final String DEFAULT_USER_AGENT_BASE_URI = "info:fedora/local-user#";
 
     private final Session jcrSession;
     private final String id;
@@ -236,11 +241,15 @@ public class FedoraSessionImpl implements FedoraSession {
      */
     private URI buildDefaultURI(final String userId) {
         // Construct the default URI for the user ID that is not a URI.
-        final String userAgentBaseUri = System.getProperty(USER_AGENT_BASE_URI_PROPERTY);
+        String userAgentBaseUri = System.getProperty(USER_AGENT_BASE_URI_PROPERTY);
         if (isNullOrEmpty(userAgentBaseUri)) {
-            throw new RepositoryConfigurationException("User agent base URI prioperty is missing."
-                + " Please configure it with variable fcrepo.auth.webac.userAgent.baseUri.");
+            // use the default local user agent base uri
+            userAgentBaseUri = DEFAULT_USER_AGENT_BASE_URI;
         }
-        return URI.create(userAgentBaseUri + userId);
+
+        final String userAgentUri = userAgentBaseUri + userId;
+
+        LOGGER.warn("Default URI is created for user {}: {}", userId, userAgentUri);
+        return URI.create(userAgentUri);
     }
 }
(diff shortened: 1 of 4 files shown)


fcrepo-kernel-modeshape/src/test/java/org/fcrepo/integration/kernel/modeshape/FedoraSessionImplIT.java
```diff
@@ -0,0 +1,131 @@
...
+    @Test
+    public void testGetIdExceptionWithUserIdNonURI() throws RepositoryException {
+        when(request.getRemoteUser()).thenReturn(FEDORA_USER);
+        when(request.getUserPrincipal()).thenReturn(new BasicUserPrincipal(FEDORA_USER));
+        when(request.isUserInRole(eq("admin"))).thenReturn(true);
+
+        final ServletCredentials credentials = new ServletCredentials(request);
+        final FedoraSession session = repo.login(credentials);
+
+        // should thrown for unsupported digest algorithm sha256
```
birkland: Is this the correct comment?  I don't see how sha256 is involved

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraSessionImpl.java
```diff
@@ -205,4 +228,19 @@ public static Session getJcrSession(final FedoraSession session) {
...
+    /**
+     * Build default URI with the configured base uri for agent
+     * @param userId
+     * @return
+     */
+    private URI buildDefaultURI(final String userId) {
+        // Construct the default URI for the user ID that is not a URI.
+        final String userAgentBaseUri = System.getProperty(USER_AGENT_BASE_URI_PROPERTY);
+        if (isNullOrEmpty(userAgentBaseUri)) {
+            throw new RepositoryConfigurationException("User agent base URI prioperty is missing."
```
birkland: This will make it _required_ to provide a system property in order to support string user names.  I believe this is a good thing.  That being said, I think we should:
* mark this as a breaking change (since somebody who deploys Fedora may have to define this new property)
* provide a default base URI for the one-click jar in `fcrepo-webapp`, since it is intended for demonstration purposes, and should be easy to use
* log (either `DEBUG` or `INFO`) that a URI was created from a string.  

Would you be able to add the last two bullets to this PR, if you agree it is appropriate?

lsitu: @birkland Thanks. I've added the logging for the default user URI created. I tried to add the default URI to jetty-console-maven-plugin but I am not able to get it works. Do you have any working examples that can point me to? 

Here is I was trying: 
<configuration>
...
    <systemProperties>
        <systemProperty>
            <name>fcrepo.auth.webac.userAgent.baseUri</name>
             <value>http://fedora.info/definitions/v4/person#</value>
         </systemProperty>
     </systemProperties>
</configuration>

birkland: @lsitu In the profile for `one-click`, there is a `<properties>` section that currently defines `fcrepo.modeshape.configuration`.  Somehow, it influences the resulting jar in such a manner that results in that property being set when it is run.  Glancing at the code, I have no idea how or why that works, but it does.  Somehow, you'll need to do the same thing with `fcrepo.auth.webac.userAgent.baseUri` as happens with ``fcrepo.modeshape.configuration`.

birkland: I wonder if any @fcrepo4/committers have any recollection on how the one-click jar gets the `fcrepo.modeshape.configuration` property, based on setting it in the pom.

escowles: I believe this is where that's configured: https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-webapp/src/main/resources/spring/fcrepo-config.xml#L28-L30

lsitu: @birkland I think the configuration for fcrepo.modeshape.configuration is different from environment variable setting, which is set for https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-webapp/src/main/resources/spring/fcrepo-config.xml#L30.
It seem like setting environment variable is an issue for the one click. 

Could we just change the strategy to provide a default Fedora user agent URI like http://fedora.info/definitions/v4/person# if it's not set instead of throw the RepositoryConfigurationException? I think this could be good in the case when the user won't care about the user agent URI, and this won't be a breaking change. I see Hyrax just provide a default URL for user agent by prefixing http://projecthydra.org/ns/auth/person# to the user name. Thoughts?

birkland: @lsitu Providing a default would be acceptable as well.  In that case, we should set it to something non-resolvable, and non-meaningful.  So I would argue for something like `info:fedora/local-user#` in that case.

@escowles Yes, I see where spring uses a place holder, the mystery to me is where Spring gets the value for `fcrepo.modeshape.configuration` in the case of the one-click jar.  Who or what sets it when the jar is executed?

escowles: @birkland Oh, I think that's being set by the oneclick profile in pom.xml: https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-webapp/pom.xml#L494

lsitu: @birkland I've updated it to use the default user URI as we discussed above instead of throwing the exception. See commit https://github.com/fcrepo4/fcrepo4/pull/1242/commits/422a95f46cbca447deb4a0bf31434c60c6314a44.

birkland: @escowles Right, it's set in the pom; but what is the  _mechanism_ by which that value affects the generated artifact in a manner such that Spring will know about its value?  In other words, the value clearly gets recorded somewhere in the jar file.  I'm wondering where, how, and by what mechanism does Spring discover it?  (this won't block the PR, but I'm really curious)

birkland: @lsitu Thanks, looks good

escowles: @birkland Sorry to be so obtuse about this: I think that the Maven resources filtering does that part: https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-webapp/pom.xml#L499-L505

PR #1248, opened 2017-09-29
FCREPO-2634. Removed versioning implementation code.

mohideen:
https://jira.duraspace.org/browse/FCREPO-2634

Removed the current versioning implementation code to have the branch in a clean state for the Memento versioning implementation.

@awoods @whikloj


whikloj:
@mohideen So I'm wondering...should we leave the skeletons (ie. function signatures, etc) in here or should we strip it all out and start fresh? What are your thoughts on this?

mohideen:
@whikloj I tried to leave in the function signatures whenever it did not seem to be specific to the previous implementation. Also, I have tried to remove all frozen/unfrozen node related version code, as we were initially thinking of this branch as a starting place for implementing versioning independent of Modeshape.

If we have decided to use Modeshape versioning (I had to miss that meeting today), I think it might be useful to preserve more of the function signatures.

whikloj:
@mohideen we are try not to use Modeshape versioning, so this works. 👍

PR #1252, opened 2017-10-04
Create LDPCv resource for TimeMap.

lsitu:
Fixes https://jira.duraspace.org/browse/FCREPO-2639

Create LDPCv resource for TimeMap.

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraTypes.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraTypes.java
@@ -41,6 +41,8 @@ public interface FedoraTypes {
 
     String FEDORA_TOMBSTONE = "fedora:Tombstone";
 
+    String FEDORA_TIME_MAP = "fedora:TimeMap";
+
     String LDP_BASIC_CONTAINER = "ldp:BasicContainer";
 
     String LDP_CONTAINER = "ldp:Container";
@@ -82,5 +84,4 @@ public interface FedoraTypes {
     String LDP_IS_MEMBER_OF_RELATION = "ldp:isMemberOfRelation";
 
     String LDP_MEMBER_RESOURCE = "ldp:membershipResource";
-
 }
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
@@ -69,6 +69,18 @@ public interface FedoraResource {
     FedoraResource getContainer();
 
     /**
+     * Get the TimeMap/LDPCv of this resource
+     * @return the container for TimeMap/LDPCv of this resource
+     */
+    FedoraResource getLDPCv();
+
+    /**
+     * Create the TimeMap/LDPCv of this resource if it doesn't exist
+     * @return the container for TimeMap/LDPCv of this resource
+     */
+    FedoraResource findOrCreateLDPCv();
+
+    /**
      * Get the child of this resource at the given path
      * @param relPath the given path
      * @return the child of this resource
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
@@ -69,16 +69,10 @@ public interface FedoraResource {
     FedoraResource getContainer();
 
     /**
-     * Get the TimeMap/LDPCv of this resource
-     * @return the container for TimeMap/LDPCv of this resource
-     */
-    FedoraResource getLDPCv();
-
-    /**
      * Create the TimeMap/LDPCv of this resource if it doesn't exist
      * @return the container for TimeMap/LDPCv of this resource
      */
-    FedoraResource findOrCreateLDPCv();
+    FedoraResource findOrCreateTimeMap();
 
     /**
      * Get the child of this resource at the given path
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -62,6 +62,7 @@ import static org.fcrepo.kernel.modeshape.utils.NamespaceTools.getNamespaceRegis
 import static org.fcrepo.kernel.modeshape.utils.StreamUtils.iteratorToStream;
 import static org.fcrepo.kernel.modeshape.utils.UncheckedFunction.uncheck;
 import static org.modeshape.jcr.api.JcrConstants.JCR_CONTENT;
+import static org.modeshape.jcr.api.JcrConstants.NT_FOLDER;
 import static org.slf4j.LoggerFactory.getLogger;
 
 import java.net.URI;
@@ -159,6 +160,8 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
     private static final long NO_TIME = 0L;
 
     private static final PropertyConverter propertyConverter = new PropertyConverter();
+
+    public static final String LDPCV_TIME_MAP = "TimeMap";
 
     // A curried type accepting resource, translator, and "minimality", returning triples.
     private static interface RdfGenerator extends Function<FedoraResource,
@@ -342,8 +345,17 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
      */
     private static Predicate<Node> nastyChildren = isInternalNode
                     .or(TombstoneImpl::hasMixin)
+                    .or(FedoraResourceImpl::isTimeMap)
                     .or(UncheckedPredicate.uncheck(p -> p.getName().equals(JCR_CONTENT)))
                     .or(UncheckedPredicate.uncheck(p -> p.getName().equals("#")));
+
+    private static boolean isTimeMap(final Node node) {
+        try {
+            return node.isNodeType(FEDORA_TIME_MAP);
+        } catch (final RepositoryException e) {
+            throw new RepositoryRuntimeException(e);
+        }
+    }
 
     private static final Converter<FedoraResource, FedoraResource> datastreamToBinary
             = new Converter<FedoraResource, FedoraResource>() {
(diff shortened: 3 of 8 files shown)


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -368,6 +380,42 @@ public FedoraResource getContainer() {
     }
 
     @Override
+    public FedoraResource getLDPCv() {
```
whikloj: This function doesn't seem to be used yet, what was your plan around this?

lsitu: @whikloj I was thinking about in cases like DELETE, HEAD, GET of the TimeMap/LDPCv, we may want to throw a Exception if the TimeMap LDPCv container is not found instead of using FindOrCreated() to create a new LDPCv. What do you think?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -160,6 +161,8 @@
 
     private static final PropertyConverter propertyConverter = new PropertyConverter();
 
+    public static final String LDPCV_TIME_MAP = "TimeMap";
```
bseeger: Is there a difference between a FEDORA_TIME_MAP and this LDPCV_TIME_MAP?  Do we need to definitions of this?  They should be one in the same...??

bseeger: Actually, Never mind. I see what's going on.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -368,6 +380,42 @@ public FedoraResource getContainer() {
...
+        } catch (RepositoryException e) {
+            throw new RepositoryRuntimeException(e);
+        }
+    }
+
+    @Override
+    public FedoraResource findOrCreateLDPCv() {
+        final Node ldpcvNode;
+        try {
+            ldpcvNode = findOrCreateChild(getNode(), LDPCV_TIME_MAP, NT_FOLDER);
```
bseeger: Is this making the LDPCv a childe of the LDPR?  If not, where does the LDPCv live?  

To be clear, the reason I'm wondering is that it's explicit in the spec that the LDPCv cannot be a child of the LDPR.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
whikloj: I think trying to DELETE/HEAD/GET a resource in the Repository that doesn't exist should throw an Exception based on normal interactions. But I could be mistaken on that.

But this function could be called from the parent LDPRv to `get` the TimeMap resource when creating the header links for example. So that's fine.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
bseeger: at least via the ldp:contains triple - it could be part of the LDPR via a different triple.

whikloj: It is explicit in the spec that the LDPRv cannot `ldp:contains` the LDPCv. We are making them children in the JCR context, but avoiding creating the `ldp:contains` connection for the LDPCv.

bseeger: Ah, okay - I don't know how the internals work very well.   I wonder why the spec folks didn't want the LDPCv to be a child of the LDPR. If it was connected to the life cycle, then this won't work either.   Did we decide that if a LDPR is deleted so is all it's version history?  I know we talked about it but can't recall what was decided.

whikloj: That is the current plan, everything is nested.
`LDPRv -> LDPCv -> LDPRm(s)`

But @lsitu noted that currently we turn off versioning by deleting the `LDPCv` which means you can't toggle versioning on a resource without destroying all previously made versions. So we may need to review that later.

whikloj:
This is a good start in my mind. 👍 

This code will need to be referenced by @dbernstein in his work on enabling versioning on resources. So that might help expose any issues? Also @awoods said he'd have a look

lsitu:
@whikloj Thanks. Glad we all agree on it and we can move forward with other tickets now.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraTypes.java
```diff
@@ -41,6 +41,8 @@
 
     String FEDORA_TOMBSTONE = "fedora:Tombstone";
 
+    String FEDORA_TIME_MAP = "fedora:TimeMap";
```
awoods: We should have a follow-on ticket to add `TimeMap` to the [Fedora ontology](https://github.com/fcrepo4/ontology/blob/master/repository.rdf).

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
awoods: Add annotation: `@VisibleForTesting`

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
```diff
@@ -69,6 +69,18 @@
     FedoraResource getContainer();
 
     /**
+     * Get the TimeMap/LDPCv of this resource
+     * @return the container for TimeMap/LDPCv of this resource
+     */
+    FedoraResource getLDPCv();
```
awoods: It would likely be more clear to change the name of this method to `getTimeMap()`;

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
```diff
@@ -69,6 +69,18 @@
...
+
+    /**
+     * Create the TimeMap/LDPCv of this resource if it doesn't exist
+     * @return the container for TimeMap/LDPCv of this resource
+     */
+    FedoraResource findOrCreateLDPCv();
```
awoods: It would likely be more clear to change the name of this method to `findOrCreateTimeMap()`;

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -368,6 +380,42 @@ public FedoraResource getContainer() {
...
+
+            if (ldpcvNode.isNew()) {
+                LOGGER.debug("Created TimeMap LDPCv {}", node.getPath() + "/" + LDPCV_TIME_MAP);
```
awoods: It seems like this log message would be less of a maintenance burden by getting the path from: `ldpcvNode.getPath()`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -368,6 +380,42 @@ public FedoraResource getContainer() {
...
+
+                // add minxin type fedora:Resource
```
awoods: Minor typo: `minxin`

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -368,6 +380,42 @@ public FedoraResource getContainer() {
...
+                if (node.canAddMixin(FEDORA_RESOURCE)) {
+                    node.addMixin(FEDORA_RESOURCE);
+                }
+
+                // add minxin type fedora:TimeMap
```
awoods: Minor typo: `minxin`

fcrepo-kernel-modeshape/src/test/java/org/fcrepo/integration/kernel/modeshape/FedoraResourceImplIT.java
```diff
@@ -1098,6 +1100,43 @@ public void testDeletePartOfMultiValueProperty() throws RepositoryException {
...
+
+        // Create TimeMap (LDPCv)
+        final FedoraResource ldpcvResource = resource.findOrCreateLDPCv();
+
+        assertNotNull(ldpcvResource);
+        assertEquals("/" + pid + "/" + LDPCV_TIME_MAP, ldpcvResource.getPath());
+
+        session.commit();
+
+        assertTrue(jcrSession.getNode("/" + pid).getNode(LDPCV_TIME_MAP).isNodeType(FEDORA_TIME_MAP));
```
awoods: You can probably just add this line to the test above and then remove this otherwise duplicative test method.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
awoods: Actually, instead of adding the `TimeMap` functionality to the existing `FedoraResource`, it may be cleaner to create the new class `FedoraTimeMap.java`.

Additionally, I believe we were planning on the timemap resource to be named `fcr:versions` and to have requests to timemaps be intercepted by a Jersey path selector:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersioning.java#L73

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraTimeMapImpl.java
```diff
@@ -0,0 +1,72 @@
...
+     * Construct a {@link org.fcrepo.kernel.api.models.FedoraResource} from an existing JCR Node
+     * @param node an existing JCR node to be treated as a fcrepo object
+     */
+    public FedoraTimeMapImpl(final Node node) {
+        super(node);
+    }
+
+
+    @Override
+    public FedoraResource getTimeMap() {
```
whikloj: Now I'm not sure if this makes sense outside the FedoraResourceImpl. Calling `getTimeMap()` on the LDPRv resource to return the TimeMap made sense. Now we are calling `getTimeMap()` on the TimeMap...unsure.

lsitu: @whikloj Yep, it's not as easy understanding as adding it FedoraResourceImpl. But the FedoraTimeMap model can be interpreted by the FedoraResourceConverter as its own. This could be addressed later when we use the fcr:version endpoint for versioning. 
Would you like me to add another commit to address it now?

whikloj: No you know what its fine. I think some of this is just my own trouble fully understanding.

PR #1257, opened 2017-10-06
Create TimeMap LDPCv for binary resource.

lsitu:
Fixes https://jira.duraspace.org/browse/FCREPO-2644
Create TimeMap LDPCv for binary resource.

Create TimeMap under the parent container of the binary resource with the name of the binary node.
Refactored codes to move the signature getTimeMap() to FedoraResource to make it easier to access the TimeMap for FedoraResource/FedoraBinary.


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraBinaryImpl.java
```diff
@@ -445,6 +448,44 @@ private static void decorateContentNode(final Node contentNode, final Collection
...
+    }
+
+    @Override
+    public FedoraResource findOrCreateTimeMap() {
+        final Node ldpcvParentNode;
+        final Node ldpcvNode;
+
+        try {
+            final Node container = getContainingNode(getNode()).get();
+            ldpcvParentNode = findOrCreateChild(container, LDPCV_TIME_MAP, NT_FOLDER);
```
awoods: As a comment to this method, could you please detail the design of the TimeMap containers?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraBinaryImpl.java
```diff
@@ -445,6 +448,44 @@ private static void decorateContentNode(final Node contentNode, final Collection
     }
 
     @Override
+    public FedoraResource getTimeMap() {
+        try {
+            final Node container = getContainingNode(getNode()).get();
```
mohideen: @lsitu Is this implementation assuming that the BinaryServiceImpl will be refactored to use a new nt:folder node as an enclosing node for the binary nt:file node.

https://github.com/lsitu/fcrepo4/blob/60d6c54bcb8ed28563fe9d8ad86814b6c69a4dc0/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BinaryServiceImpl.java#L57

lsitu: @mohideen No. I don't think so. This is retrieve the TimeMap for binary.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraBinaryImpl.java
lsitu: As we discussed during the sprint wrap-up meeting this morning, the TimeMap for binary will be created inside the TimeMap of the parent container that holds the binary resource, which has the same name as the binary node.

awoods:
Superseded by: https://github.com/fcrepo4/fcrepo4/pull/1258

PR #1258, opened 2017-10-06
Create new versionFile nodeType with nt:folder inside

whikloj:
Enable Binary TimeMap test

**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2644

Creates a new nodetype based on the nt:file but with an additional nt:folder child called fedora:timemap to hold the timemap and mementos.

# How should this be tested?

Integration and unit tests run. There are additional example actions, but with the resource called **TimeMap**, so change this to **fedora:timemap** now.

Example:
* Does this change require documentation to be updated? Hopefully not
* Could this change impact execution of existing code? Possibly

# Interested parties
SPRINTERS ASSEMBLE
@awoods

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
@@ -257,6 +257,8 @@ public final class RdfLexicon {
     public static final Predicate<Property> isManagedPredicate =
         hasFedoraNamespace.or(p -> managedProperties.contains(p));
 
+    public static final String NT_VERSION_FILE = "nt:versionFile";
+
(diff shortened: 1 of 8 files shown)


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
```diff
@@ -257,6 +257,8 @@
     public static final Predicate<Property> isManagedPredicate =
         hasFedoraNamespace.or(p -> managedProperties.contains(p));
 
+    public static final String NT_VERSION_FILE = "nt:versionFile";
```
awoods: Please add a comment to this constant indicating that it is actually a Fedora-defined type for the purpose of...

PR #1260, opened 2017-10-06
Delete a LDPCv (TimeMap ).

lsitu:
Fixes https://jira.duraspace.org/browse/FCREPO-2621

Delete a LDPCv (TimeMap ).

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
@@ -53,6 +53,7 @@ import org.fcrepo.kernel.api.exception.TombstoneException;
 import org.fcrepo.kernel.api.identifiers.IdentifierConverter;
 import org.fcrepo.kernel.api.models.NonRdfSourceDescription;
 import org.fcrepo.kernel.api.models.FedoraResource;
+import org.fcrepo.kernel.modeshape.FedoraTimeMapImpl;
 import org.fcrepo.kernel.modeshape.TombstoneImpl;
 import org.fcrepo.kernel.modeshape.identifiers.HashConverter;
 import org.fcrepo.kernel.modeshape.identifiers.NamespaceConverter;
@@ -118,6 +119,12 @@ public class HttpResourceConverter extends IdentifierConverter<Resource,FedoraRe
 
                 final boolean metadata = values.containsKey("path")
                         && values.get("path").endsWith("/" + FCR_METADATA);
+
+                final boolean timeMap = values.containsKey("path")
+                        && values.get("path").endsWith("/" + FCR_VERSIONS);
+                if (timeMap) {
+                    return new FedoraTimeMapImpl(node);
+                }
 
                 final FedoraResource fedoraResource = nodeConverter.convert(node);
 
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -418,6 +418,8 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
     @Override
     public void delete() {
         try {
+            final boolean isTimeMap = FedoraTimeMapImpl.hasMixin(node);
+
             // Remove inbound references to this resource and, recursively, any of its children
             removeReferences(node);
 
@@ -433,8 +435,8 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
 
             node.remove();
 
-            if (parent != null) {
-                createTombstone(parent, name);
+            if (!isTimeMap && parent != null) {
+                    createTombstone(parent, name);
 
                 // also update membershipResources for Direct/Indirect Containers
                 containingNode.filter(UncheckedPredicate.uncheck((final Node ancestor) ->
@@ -978,7 +980,7 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
 
   @Override
   public void disableVersioning() {
-      LOGGER.warn("Review if method (disableVersioning) can be removed after implementing Memento!");
+    getDescription().getChild(LDPCV_TIME_MAP).delete();
   }
 
   @Override
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersioningIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersioningIT.java
@@ -1,0 +1,61 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.integration.http.api;
+
+import static javax.ws.rs.core.Response.Status.NO_CONTENT;
+import static org.fcrepo.kernel.api.FedoraTypes.FCR_VERSIONS;
+import static org.fcrepo.kernel.api.RdfLexicon.VERSIONED_RESOURCE;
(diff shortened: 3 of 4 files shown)


fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
```diff
@@ -119,6 +120,12 @@ protected FedoraResource doForward(final Resource resource) {
                 final boolean metadata = values.containsKey("path")
                         && values.get("path").endsWith("/" + FCR_METADATA);
 
+                final boolean timeMap = values.containsKey("path")
+                        && values.get("path").endsWith("/" + FCR_VERSIONS);
+                if (timeMap) {
+                    return new FedoraTimeMapImpl(node);
```
awoods: Instead of the logic of determining if the `resource` is a TimeMap here, I believe it should happen in:
- https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java#L296 and
- https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/identifiers/NodeResourceConverter.java#L49

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -433,8 +435,8 @@ public void delete() {
 
             node.remove();
 
-            if (parent != null) {
-                createTombstone(parent, name);
+            if (!isTimeMap && parent != null) {
+                    createTombstone(parent, name);
```
awoods: It looks like an inadvertent tab was added here.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
lsitu: @awoods I was thinking about the incoming path is /path/to/resource/fcr:versions. So the  JCR node returned from HttpResourceConverter.java#L120 
 (https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java#L120) is the JCR node for the resource and we can't determine whether we want the TimeMap or the resource in NodeResourceConverter.java#L49 (https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/identifiers/NodeResourceConverter.java#L49).

Do I miss anything? I think we can move the signature FedoraTimeMap.getTimeMap() back to FedoraResource.getTimeMap() if you think it works better. Thoughts?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
lsitu: Right. I'll remove it. Thanks.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
whikloj: I think (and these Converters still confuse me) that we probably need the `getTimeMap()` in the `FedoraResourceImpl` and requests will have to be chained as `fcr:versions` is not an actual path.
So a request for `.../resource1/fcr:versions` gets `resource1` and then returns the result of the `getTimeMap()` function on that resource.
Which means that a request for `.../resource1/fcr:versions/1234-5678` will have to do the above and then find the appropriate child of the TimeMap.
This makes sense to me, but I'm not sure if this is a bad pattern.

awoods: I agree, @whikloj . The question that remains for me is how much of the translation work we will be doing in [FedoraVersioning](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersioning.java) and [FedoraVersions](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersions.java), and how much in the translation chain.

lsitu: @awoods @whikloj: Basing on our discussions above, I've moved signature getTimeMap() back to FedoraResource/FedoraResourceImpl. See https://github.com/fcrepo4/fcrepo4/pull/1260/commits/79502171a145a19bb6e50b650a7452108a82405a.

whikloj:
So this is more of a question than a comment. Can and if so how do I use the `hasTimeMapMixin` on a `Resource` to see if it is a TimeMap? 

Like if in the HttpResourceConverter `doBackwardPathOnly` I want to tell whether to append `/fcr:version` on to the end of the path. Previously I was going to do
```
if (resource instanceof FedoraTimeMap) {
```
if we remove that class/interface I'm just wondering how to perform the same check?

lsitu:
@whikloj Good point! I was just thinking about moving back to the original implementation we agreed on. This should work better: https://github.com/fcrepo4/fcrepo4/pull/1260/commits/a3d96324e92ceb137be320c53379f4859db07436.

lsitu:
@awoods @whikloj How about the refactoring above in commit https://github.com/fcrepo4/fcrepo4/pull/1260/commits/691edd151a9ec8e9f49c8d31cc40abc5d084b0d3 for FedoraResource.getTimeMap()?

whikloj:
@lsitu I'm not understanding what you are asking. What do you want to refactor?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
awoods: I do not think the suffix `/fcr:versions` should ever find its way into this method... because it should be intercepted in either FedoraVersioning or FedoraVersions.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraTimeMap.java
```diff
@@ -22,9 +22,5 @@
  * @since Oct. 04, 2017
  */
 public interface FedoraTimeMap {
```
awoods: Sinde FedoraTimeMapImpl extends FedoraResourceImpl, FedoraTimeMap should extend FedoraResource.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -373,6 +372,17 @@ public FedoraResource getContainer() {
     }
 
     @Override
+    public FedoraResource getTimeMap() {
```
awoods: This method only appears to be used once: in the [FedoraResourceImplIT](https://github.com/fcrepo4/fcrepo4/pull/1260/files#diff-0e3024e0f228c42d1ad0611f251af6c4R1143). Therefore, it can probably be removed.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -433,8 +445,8 @@ public void delete() {
 
             node.remove();
 
-            if (parent != null) {
-                createTombstone(parent, name);
+            if (!isTimeMap && parent != null) {
```
awoods: An inline comment here would be helpful, explaining why creating a tombstone is not done for timemaps and resources without parents.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -978,7 +990,7 @@ public void enableVersioning() {
 
   @Override
   public void disableVersioning() {
-      LOGGER.warn("Review if method (disableVersioning) can be removed after implementing Memento!");
+    getDescription().getChild(LDPCV_TIME_MAP).delete();
```
awoods: Actually, this is a situation where `FedoraResourceImpl.getTimeMap()` could be used.

lsitu:
@whikloj I think I've done with refactoring in that commit so we just need review. Thanks. I'll response to @awoods comments shortly.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
whikloj: @awoods as this defined in the interface I think it must remain here...no?

whikloj: Now that this is used for deleting the TimeMap, I'll assume this comment is resolved.

PR #1261, opened 2017-10-07
Disallow interactions on paths with fedora: in it.

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2645

Disallows any interaction if any part of the path (or Slug) uses a fedora: namespace prefix.

# How should this be tested?

Spin up a Fedora
Try to PUT a resource to http://localhost:8080/rest/fedora:resource, it should succeed
Pull in this PR
Try to PUT a resource to http://localhost:8080/rest/fedora:resource, it should fail.

Same for GET, POST, DELETE, HEAD or OPTIONS.

Example:

# Interested parties
@awoods

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -66,7 +66,6 @@ import java.net.URISyntaxException;
 import java.text.MessageFormat;
 import java.time.Instant;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Date;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -720,4 +722,16 @@ private static URI checksumURI( final String checksum ) {
         }
         return null;
     }
+
+    /**
+     * Check if a path has a segment prefixed with fedora:
+     *
+     * @param externalPath the path.
+     */
+    public static void hasRestrictedPath(final String externalPath) {
```
awoods: It seems like this method should be in FedoraLDP as a `private` method... since that is the only class that uses it.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -199,6 +200,7 @@ public void postConstruct() {
         TURTLE_X, TEXT_HTML_WITH_CHARSET })
     public Response head() throws UnsupportedAlgorithmException {
         LOGGER.info("HEAD for: {}", externalPath);
+        hasRestrictedPath(externalPath);
```
awoods: We probably only need to add this check on `PUT` and `POST`.

whikloj: @awoods I can do that, but we should remember to check later and see if you could access.
```
http://localhost:8080/fcrepo/rest/someResource/fedora:versions
```
and get/head the TimeMap. Not that it would be horrible, but you would probably not get the appropriate headers.

whikloj: I'm thinking we should also cover `DELETE`.

PR #1272, opened 2017-12-04
Improve mediatype validation on ingest

claussni:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2520

Add verification of client provided mediatype.

# What's new?
If a client provides an invalid mediatype (according to https://tools.ietf.org/html/rfc6838#section-4.2), the request is rejected with status code 400 BAD REQUEST

# How should this be tested?
See https://jira.duraspace.org/browse/FCREPO-2658

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -122,6 +122,7 @@ import javax.ws.rs.core.Link;
 import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.core.Variant;
 
+import com.google.common.net.MediaType;
 import org.apache.commons.io.IOUtils;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
@@ -419,11 +420,12 @@ public class FedoraLdpIT extends AbstractResourceIT {
 
     @Test
     public void testGetExternalDatastream() throws IOException, ParseException {
-        final String externalContentType = "message/external-body;access-type=URL;url=\"some:uri\"";
+        final MediaType expectedMediaType =
+                MediaType.parse("message/external-body;access-type=URL;url=\"some:uri\"");
 
         final String id = getRandomUniqueId();
         final HttpPut put = putObjMethod(id);
-        put.addHeader(CONTENT_TYPE, externalContentType);
+        put.addHeader(CONTENT_TYPE, expectedMediaType.toString());
         assertEquals(CREATED.getStatusCode(), getStatus(put));
 
         // Configure HEAD request to NOT follow redirects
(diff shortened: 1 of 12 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -3600,4 +3611,18 @@ public void testDigestAbsence() throws IOException {
             assertTrue(response.getHeaders(DIGEST).length == 0);
         }
     }
+
+    /*
+     * @see https://jira.duraspace.org/browse/FCREPO-2520
+     * @see https://jira.duraspace.org/browse/FCREPO-2658
+     */
+    @Test
+    public void Ingest_failes_with_invalid_mediatype() {
```
bseeger: Hello - can you follow the naming practice, please -- testIngestFailsWithInvalidMediatype() or something like that? 
Thanks

bseeger:
I do wonder, perhaps naively, if we want to go the the opposite route. The code, as it exists w/o this PR, doesn't care what media type you supply.  We could fix the other side of this equation and relax retrieving that triple, which is where it appears the issue is.   This PR seems to lock us into known media types and there might be a few out there (or new ones?) that would now fail to go in.  Maybe I'm wrong there, but I could see an issue where ingest fails because the media type is new and the google MediaType library hasn't been updated yet.

claussni:
> This PR seems to lock us into known media types

The `MediaType` class does not enforce any particular value for type or subtype of a media type literal, apart from wildcard symbols:
https://google.github.io/guava/releases/19.0/api/docs/com/google/common/net/MediaType.html

However, the documentation also states:

> Note that this specifically does not represent the value of the MIME Content-Type header and as such has no support for header-specific considerations such as line folding and comments. 

This might be a problem somewhen, since the media type is mainly used as a content-type header value.

bseeger:
Ah, okay. I see what you're saying and I was misinterpreting what was happening. I'll download the code and test this out a bit.

claussni:
I rebased this branch with master.

bseeger:
I tested this and it works fine as is. I wasn't able to test with a folded header as I wasn't able to figure out how to easily create one. But I'm not sure how many scenarios a folded header would be handed in to fedora.

Thanks for putting together this PR - I have an overall thought on the structure of it, which is a bit different from this PR's structure. 

One could put this check up a level at the HTTP layer - in the code before setContent() is called:
https://github.com/fcrepo4/fcrepo4/pull/1272/files#diff-d372035f89c8d0846920e20bd13e0145L650

Then one could use the JAX-RS MediaType method valueOf, which seems to serve the same purpose as the guava parse() method.  Also, JAX-RS exceptions are available then, so the new one might not be necessary.  

It keeps the HTTP header check out of the modeshape kernel code. @awoods - do you feel strongly about the current method versus moving the check up into the HTTP layer?

bseeger:
I'd like to close this PR as https://github.com/fcrepo4/fcrepo4/pull/1280 replaces it as a solution.

PR #1280, opened 2018-01-24
Adds code to handle invalid mime-type

bseeger:
- Allows binary to be retrieved even if mime-type is invalid

**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2520

This PR will allow a client to still be able to retrieve a binary, should the mime type field become syntactically incorrect.  This is possible solution (or a step in the right direction) for the above JIRA ticket. 

Another solution is to ensure that ebucore:hasMimeType field is not set to something invalid.  Not sure how hard that would be to implement. 

# What's new?
This change wraps up the MediaType.valueOf() call and catches an exception it might throw.  If it throws one, the exception gets logged and the code keeps going as if the mime type were 'application/octet-stream', versus returning a 500 error. 

# How should this be tested?
Start up a fcrepo4 server.  Add a binary file with a correct mime type.  Change the technical metadata triple:  ebucore:hasMimeType to something syntactically invalid - like "== invalid==".  Then try to GET the binary again.  It should allow you to retrieve the binary, which will show the content type of 'application/octet-stream'

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -17,7 +17,6 @@
  */
 package org.fcrepo.http.api;
 
-
 import static java.util.EnumSet.of;
 import static java.util.stream.Stream.concat;
 import static java.util.stream.Stream.empty;
@@ -194,8 +193,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
         final RdfNamespacedStream outputStream;
 
         if (resource() instanceof FedoraBinary) {
-
-            final MediaType mediaType = MediaType.valueOf(((FedoraBinary) resource()).getMimeType());
+            final MediaType mediaType = getBinaryResourceMediaType();
 
             if (isExternalBody(mediaType)) {
                 return externalBodyRedirect(getExternalResourceLocation(mediaType)).build();
@@ -380,7 +378,9 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
 
 
             // we set the content-type explicitly to avoid content-negotiation from getting in the way
-            return builder.type(binary.getMimeType())
+            // getBinaryResourceMediaType will try to use the mime type on the resource, falling back on
+            // 'application/octet-stream' if the mime type is syntactically invalid
+            return builder.type(getBinaryResourceMediaType().toString())
                     .cacheControl(cc)
                     .build();
 
@@ -712,6 +712,21 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
             final String requestBody,
             final RdfStream resourceTriples) {
         resource.getDescribedResource().updateProperties(translator(), requestBody, resourceTriples);
+    }
+
+    /**
+     * This method returns a MediaType for a binary resource.
+     * If the resource's media type is syntactically incorrect, it will
+     * return 'application/octet-stream' as the media type.
+     */
+    protected MediaType getBinaryResourceMediaType() {
+        try {
+            return MediaType.valueOf(((FedoraBinary) resource()).getMimeType());
+        } catch (IllegalArgumentException e) {
+            LOGGER.warn("Syntactically incorrect MediaType encountered on resource {}: '{}'",
+                    resource().getPath(), ((FedoraBinary)resource()).getMimeType());
+            return MediaType.APPLICATION_OCTET_STREAM_TYPE;
+        }
     }
 
     /**
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -209,7 +209,7 @@ public class FedoraLdp extends ContentExposingResource {
         Response.ResponseBuilder builder = ok();
 
         if (resource() instanceof FedoraBinary) {
-            final MediaType mediaType = MediaType.valueOf(((FedoraBinary) resource()).getMimeType());
+            final MediaType mediaType = getBinaryResourceMediaType();
 
             if (isExternalBody(mediaType)) {
                 builder = externalBodyRedirect(getExternalResourceLocation(mediaType));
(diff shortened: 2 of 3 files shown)


dbernstein:
I'm going to give it a test.  Assuming all goes well I will merge it in a few.

claussni:
Just to make this clear: The repository is masking an error caused by corrupted data.

> Another solution is to ensure that ebucore:hasMimeType field is not set to something invalid. Not sure how hard that would be to implement.

That would be a consistent solution.

awoods:
I have not tested https://github.com/fcrepo4/fcrepo4/pull/1272 , but it looks like a sound creation-side solution. Along the lines of @escowles and @birkland comments, is there any reason that we should not target getting https://github.com/fcrepo4/fcrepo4/pull/1272 merged as well?

escowles:
I am 👍 on merging both #1272 and #1280

awoods:
I am too (:+1: on merging https://github.com/fcrepo4/fcrepo4/pull/1272)... pending a close review of the updates.

bseeger:
My hesitation with #1272 and what lead me on this journey was to move the check up into the API layer versus having a HTTP check down in the modeshape kernel.

awoods:
That makes sense @bseeger ... although I am not sure that was exactly where the `GET`-side solution landed.

birkland:
This whole issue is rife with complexities, especially when it became apparent that there are two possible vectors (HTTP headers, and directly modifying `ebucore:hasMimeType`).  It's not clear to me if what the right approach is, or if both vectors can be handled holistically.  (In my mind, it'd be ideal to make `ebucore:hasMimeType` less special, but that'd be out of scope for this ticket)

awoods:
Based on @bseeger's earlier exploration, I was under the impression that the HTTP headers vector was taken care of by the servlet container... leaving only the `ebucore:hasMimeType` issue to be addressed.
Is your understanding different, @birkland ?

bseeger:
Http Headers:  Tomcat / Jetty Covers that vector (at least in my experience), so the code from #1272 doesn't actually do anything as it won't get a bad mimetype passed to it (because it is the one that came thru tomcat/jetty.  It's already slightly validated (at least syntax). 

ebucore:hasMimeType - that's a more complicated issue to solve.  We'd have to cover all the scenarios that could be set in.    

This PR covers the retrieval of a binary with a badly formatted ebucore:hasMimeType field, essentially.  

I do think another JIRA ticket should be created to cover the scenario of how to work with ebucore:hasMimeType and ensure that it's valid.

birkland:
@awoods I don't know the servlet spec well enough to know if Content-Type validation is compulsory by the servlet container.

escowles:
It looks like we are already parsing the input body for PUT/POST:

https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java#L665

However, PATCHes are being passed straight to the kernel:

https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java#L714

So unless we want to move the SPARQL-Update processing into the HTTP impl, it seems like the kernel would need to ultimately do the content-type validation.

bseeger:
A little more background:  In the PUT and POST handlers for fcrepo, they use MediaType as the object type in the [handler](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java#L371).  That calls MediaType.valueOf() -- that function will cause a IllegalArgumentException to be thrown if a syntactically invalid content type is passed in.  That alone is probably why tomcat/jetty do not let in syntactically invalid content types.  But I'm not sure if that's where it's being checked - as in, whether or not tomcat / jetty catch it before then.  I can't fine anything in the servlet spec that suggest validating the mime type.    In either case, a syntactically incorrect media type will not be able to get into the PUT / POST handlers.

claussni:
After following the discussion I think that #1272
<https://github.com/fcrepo4/fcrepo4/pull/1272> is not a good solution. It
does not address the problem in a concise way, yet it brings huge changes
to the Kernel API.

In my view the problem lies in the Kernel API method signatures, eg:

void setContent(InputStream content, String contentType, Collection<URI>
checksums,

 There contentType is passed as a string. This parameter should have a
proper type (eg. MediaType) leaving the parsing and error handling to
layers further up and freeing the Kernel from checking assumptions.

awoods:
Thanks, @claussni. It also appears that the method [FedoraBinary.setContent(...)](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraBinary.java#L53) is only used in one place out side of test code:
[ContentExposingResource.replaceResourceBinaryWithStream(...)](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java#L640-L655)
...which implies minimal impact with such a change.

@bseeger ? @claussni ? would you be interested in exploring this suggestion?

bseeger:
I can see the value in that parameter's type being changed from String to Media Type.  That, at a minimum, would prevent an invalid MediaType from being passed into the kernel via that method.  That change should be minor and would protect the kernel more.   However, it's not a scenario I'm particularly worried about right now, since setContent() is only called from one place right now - as in, this is good to do, but not a high priority for me. 

We still need to consider a fix for the passing in of an invalid string value for the property ebucore:hasMimeType -- I would consider this a higher priority.

awoods:
Thanks for keeping the focus, @bseeger :)

We have a hook for [validating PATCH requests](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java#L1087) and should also be able to validate [PUT requests in a similar fashion](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java#L745).

bseeger:
:) 
I can work on adding logic to validate mime type (ebucore:hasMimeType) to those areas you've identified next week.

PR #1283, opened 2018-02-15
Check RDF mimetype in ebucore:hasMimeType triple

bseeger:
**This is a continuation of the fix for FCREPO-2520. ebucore:hasMimeType triple will be syntatically validated on a sparql update or PUT request. 

***

JIRA Ticket: https://jira.duraspace.org/browse/FCREPO-2520

Introduces code that will run through the triples and make sure an object of a ebucore:hasMimeType triple is syntactically valid. 

# What's new?
This code adds on to the triple validation that is done in the kernel-modeshape module. It adds a check for syntax on mime type. 

# How should this be tested?

You could test this change by sending in a sparql request with ebucore:hasMimeType set to something syntactically invalid - like "Junk Mime Type" - that should return a 400 error with a message about the mime type. 
Also, sending RDF via a PUT command with a syntactically invalid mime type would cause a 400 to be returned with a message about mime type. 

# Additional Notes:
I'm not sure if this would be considered a breaking change -- client software that sets the mime type could now (potentially) get a 400, whereas before it would have succeeded no matter what the value was.  This might cause them to have to re-write some code.  

# Interested parties
@awoods, @whikloj, @dbernstein

### DIFF
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -777,6 +777,7 @@ public class FedoraResourceImpl extends JcrTools implements FedoraTypes, FedoraR
                 originalTriples.filter(triple -> !isRelaxed.test(createProperty(triple.getPredicate().getURI()))));
 
 
+
         try (final RdfStream replacementStream =
                 new DefaultRdfStream(idTranslator.reverse().convert(this).asNode())) {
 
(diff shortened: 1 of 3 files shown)


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -777,8 +795,25 @@ public void replaceProperties(final IdentifierConverter<Resource, FedoraResource
...
+                // do some very basic validation to catch invalid RDF
+                // this uses the same checks that updateProperties() uses
+                final Collection<IllegalArgumentException> errors = testStream
+                        .map(x -> Quad.create(x.getSubject(), x))
+                        .flatMap(FedoraResourceImpl::validateQuad)
+                        .filter(x -> x != null)
+                        .collect(Collectors.toList());
+
+                if (!errors.isEmpty()) {
+                    throw new ConstraintViolationException(
```
whikloj: I haven't tested this but, the `FedoraResourceImplIT` tests made me wonder, it seems like you might get a different error for the same problem depending on the operation. ConstrainViolationException or IllegalArgumentException... or is that just in the tests?

whikloj: Ok, so the end result is a `400 Bad Request` to either `PATCH`ing or `PUT`ing the `/fcr:metadata` endpoint or `PUT`ing with an invalid `Content-type` header. So you can disregard this.

PR #1302, opened 2018-03-14
Fcrepo 2705: ensures ACL resource exists if ACL link header is present on PUT/POST

harringj:
- adds InvalidACLException and InvalidACLExceptionMapper classes
- adds InvalidACLException.rdf
- updates FedoraLdp

NB: this PR resolves the first part of FCREPO-2705, which is "if the resourceAcl points to a non-existent resource return  a 409 indicating problem"

Resolves: https://jira.duraspace.org/browse/FCREPO-2705


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -619,6 +628,13 @@ public Response createObject(@HeaderParam(CONTENT_DISPOSITION) final ContentDisp
         }
     }
 
+    private void checkAclUriExists(final URI resourceAcl) {
+        final FedoraResource aclResource = getResourceFromPath(resourceAcl.getPath());
```
dbernstein: I think you'll need a try catch here that checks if exception itself or its cause is an instance of PathNotFoundRuntimeException.

PR #1308, opened 2018-03-16
Fix FCREPO-2687 - Restrict Cross Domain ACL

yinlinchen:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2687

This PR is to solve the issue that Fedora does not allow to create a link to an ACL resource that is outside of the repository. 

Example:
For example, the following should not be allowed: 
* create rdf resource with acl on POST 
```
curl -v -XPUT -H "Link: <http://google.com>; rel=\"acl\"" "http://localhost:8080/rest/test" 
```
# How should this be tested?

1. create rdf resource with acl on POST 
```
curl -v -XPUT -H "Link: <http://google.com>; rel=\"acl\"" "http://localhost:8080/rest/test" 
```
2. It should return 409 CONFLICT with a message "Cross Domain ACLs is not allowed" 
 
# Interested parties
Tag @awoods @dbernstein

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -643,6 +643,15 @@ public class FedoraLdp extends ContentExposingResource {
     private void checkAclUriExistsAndHasCorrectType(final URI resourceAcl) {
         FedoraResource aclResource = null;
         try {
+
+            final String aclHost = resourceAcl.getHost();
+            final String serverHost = (headers.getHeaderString("X-Forwarded-Host") == null) ? this.uriInfo
+                    .getBaseUri().getHost() : headers.getHeaderString("X-Forwarded-Host");
+
+            if (!serverHost.equals(aclHost)) {
+                throw new InvalidACLException("Cross Domain ACLs is not allowed");
+            }
+
             //extract external path
             final String contextPath = this.uriInfo.getBaseUri().getPath();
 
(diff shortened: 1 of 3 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -663,6 +663,13 @@ private void checkAclUriExists(final URI resourceAcl) {
 
     private void addResourceAcl(final URI resourceAcl) {
         if (resourceAcl != null) {
+            final String aclHost = resourceAcl.getHost();
```
awoods: It seems like the other checking of correctness for the ACL link is taking place in: 
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java#L643

Maybe changing the name of that method and performing the check there would be more consistent.
Additionally, doing this check sooner in `PUT`/`POST` provides quicker/more efficient response.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -643,6 +643,15 @@ public Response createObject(@HeaderParam(CONTENT_DISPOSITION) final ContentDisp
     private void checkAclUriExistsAndHasCorrectType(final URI resourceAcl) {
         FedoraResource aclResource = null;
         try {
+
+            final String aclHost = resourceAcl.getHost();
+            final String serverHost = (headers.getHeaderString("X-Forwarded-Host") == null) ? this.uriInfo
+                    .getBaseUri().getHost() : headers.getHeaderString("X-Forwarded-Host");
+
+            if (!serverHost.equals(aclHost)) {
+                throw new InvalidACLException("Cross Domain ACLs is not allowed");
```
awoods: Typo: "Cross Domain ACLs [are] not allowed"

PR #1320, opened 2018-04-17
Allow Datetime negotiation on versioned resources

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2613

Adds the ability for a versioned resource to take an `Accept-Datetime` header and return:
* a `302 Found` to the Memento which is the closest but older version (if more than one Memento)
* a `302 Found` to the Memento regardless of older or not (if only one Memento)
* a `404 Not Found` if there are no Mementos. 

# What's new?
Added `FedoraResource findMementoByDatetime(Instant mementoDatetime);` to `FedoraResource` interface and added an implementation to `FedoraResourceImpl`.

Added `getMemento()` to `FedoraLdp` if there is an `Accept-Datetime` header and the resource is versioned. This takes over the interaction and returns the headers for the normal resource but adds either a `302 Found` with a `Location` header to the Memento or a `404 Not Found` if there are no Mementos.

# How should this be tested?

1. Build and run it.
1. Create a non-versioned resource, send a GET with an `Accept-Datetime` header. You get the resource as normal
1. Make a versioned resource 
1. Send a GET request to the versioned resource with an `Accept-Datetime: 2018-04-17T01:01:01Z` header (ie. not RFC-1123) get a `400 Bad Request`.
1. Send a GET request to the versioned resource with an `Accept-Datetime: Tue, 17 Apr 2018 19:04:17 GMT` header (ie. is RFC-1123) get a `404 Not Found`.
1. Make a version of the resource
1. Send a GET request to the versioned resource with an `Accept-Datetime: Tue, 17 Apr 2018 19:04:17 GMT` header (ie. is RFC-1123) get a `302 Found` with a `Location` header to the Memento.

Make more Mementos and requests and ensure you always get the Memento with a datetime just older than your request.

Example:
* Does this change require documentation to be updated? Probably

# Interested parties
@dbernstein @bseeger @bbpennel @yinlinchen


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -155,6 +161,8 @@
 
     static final String ACCEPT_EXTERNAL_CONTENT = "Accept-External-Content-Handling";
 
+    public static final String ACCEPT_DATETIME = "Accept-Datetime";
```
dbernstein: I would suggest moving this constant into ContentExposing Resource and then using it to replace the string at ContentExposingResource:251

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -293,6 +312,33 @@ public Response getResource(@HeaderParam("Range") final String rangeValue)
...
+     */
+    public Response getMemento(final String datetimeHeader) {
+        try {
+            final Instant mementoDatetime = Instant.from(DateTimeFormatter.RFC_1123_DATE_TIME.parse(datetimeHeader));
+            final FedoraResource memento = resource().findMementoByDatetime(mementoDatetime);
+            final Response builder;
+            if (memento != null) {
+                builder =
+                    status(FOUND).header("Location", translator().reverse().convert(memento).toString()).build();
+
```
dbernstein: Please remove empty line.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
```diff
@@ -111,7 +111,16 @@ public void setLogger() {
     protected static CloseableHttpClient client = createClient();
 
     protected static CloseableHttpClient createClient() {
-        return HttpClientBuilder.create().setMaxConnPerRoute(MAX_VALUE).setMaxConnTotal(MAX_VALUE).build();
+        return createClient(false);
+    }
+
+    protected static CloseableHttpClient createClient(final boolean disableRedirects) {
+        if (disableRedirects) {
```
dbernstein: I think it would be cleaner to write method body with one return statement and no duplication of the create(), setMaxConnPerRoute(), setMaxConnTotal(), and build() statements.

dbernstein: But that may be a question of style. So your call.

PR #1321, opened 2018-04-17
FCREPO-2632. Use 'acl:agentClass foaf:Agent' to denote public access

mohideen:
Changes:
- The getRoles method in the roles provider class will include agentClass roles and the agentClass roles will be distinguished from the agent roles using a suffix.
- The authorization delegate will treat the EVERYONE principal as a special case and check for matching agentClass roles from the roles provider.
- Updated tests accordingly.

https://jira.duraspace.org/browse/FCREPO-2632

### DIFF
--- a/fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACRolesProvider.java
+++ b/fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACRolesProvider.java
@@ -28,6 +28,7 @@ import static java.util.stream.Stream.of;
 import static org.apache.jena.graph.NodeFactory.createURI;
 import static org.apache.jena.rdf.model.ModelFactory.createDefaultModel;
 import static org.apache.jena.riot.Lang.TTL;
+import static org.fcrepo.auth.webac.URIConstants.AGENT_CLASS_SUFFIX;
 import static org.fcrepo.auth.webac.URIConstants.FOAF_AGENT_VALUE;
 import static org.fcrepo.auth.webac.URIConstants.VCARD_GROUP;
 import static org.fcrepo.auth.webac.URIConstants.VCARD_MEMBER_VALUE;
@@ -246,6 +247,12 @@ public class WebACRolesProvider implements AccessRolesProvider {
                 concat(auth.getAgents().stream(), dereferenceAgentGroups(auth.getAgentGroups()).stream())
                     .forEach(agent -> {
                         effectiveRoles.computeIfAbsent(agent, key -> new HashSet<>())
+                            .addAll(auth.getModes().stream().map(URI::toString).collect(toSet()));
+                    });
+                auth.getAgentClasses().stream()
+                    .forEach(agentClass -> {
+                        // Use the AGENT_CLASS_SUFFIX to distinguish from agent roles
+                        effectiveRoles.computeIfAbsent(agentClass + AGENT_CLASS_SUFFIX, key -> new HashSet<>())
                             .addAll(auth.getModes().stream().map(URI::toString).collect(toSet()));
                     });
             });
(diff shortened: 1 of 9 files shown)


fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACRolesProvider.java
```diff
@@ -248,6 +249,12 @@ private FedoraResource getBaseVersion(final FedoraResource resource) {
                         effectiveRoles.computeIfAbsent(agent, key -> new HashSet<>())
                             .addAll(auth.getModes().stream().map(URI::toString).collect(toSet()));
                     });
+                auth.getAgentClasses().stream()
+                    .forEach(agentClass -> {
+                        // Use the AGENT_CLASS_SUFFIX to distinguish from agent roles
+                        effectiveRoles.computeIfAbsent(agentClass + AGENT_CLASS_SUFFIX, key -> new HashSet<>())
```
mohideen: I have used a Key suffix to distinguish agent roles from agentClass roles. I thought this will have less memory overhead. Would be better if a nested map is used instead where the agent roles and agentClass roles are on two separate HashMap and return a nested HashMap containing these two HashMaps?

whikloj: I don't think agents will have roles anymore, instead you make an `acl:agentGroup` which is just a list of `acl:agents`. 
I'd confirm with @peichman-umd and @dbernstein but I think the only user principal that will still use `acl:agentClass` is the public access `foaf:Agent`

peichman-umd: @whikloj That is correct, according to the [SOLID spec section on agents](https://github.com/fcrepo4/fcrepo4/pull/1324), the only recognized use of `acl:agentClass` is with `foaf:Agent` to indicate "Public Access (All Agents)".

peichman-umd: @whikloj Side opinion: this makes me wish the WebAC spec had just defined an individual (like `acl:everyone`) to mean "everyone", and just used `<> acl:agent acl:everyone`. I feel like `<> acl:agentClass foaf:Agent` may be a holdover from original plans to allow arbitrary agent classes, the same way that `acl:accessToClass` allows use of arbitrary classes on the protected resources.

whikloj: @peichman-umd you are probably correct. You could suggest that to the SOLID folks and see if they'd update their spec 😉

acoburn: @peichman-umd The target of `acl:agent` is a particular `foaf:Agent` (i.e. a Named Individual), so `<> acl:agent acl:everyone` would refer to one particular agent just as `<> acl:agent <http://example.com/me>` refers to one particular agent. If, on the other hand, you want to refer to a class of named individuals: for example "everyone", then you need something different than `acl:agent`, since you are referring to a "class" of named individuals rather than a particular named individual. Hence: `<> acl:agentClass foaf:Agent` or `<> acl:agentClass acl:AuthenticatedAgent` or `<> acl:agentClass ex:LocalSubDomainAgent`.

This isn't really a WebACL issue, it's an RDF issue.

mohideen: @whikloj Created another PR #1334 as an alternate implementation. I think it is a bit less complex and it is also more explicit in including the agentClass-Role mapping only when the agentClass is foaf:Agent.

PR #1331, opened 2018-04-20
Stop using Slug header for versions from HTML UI

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2697 and https://jira.duraspace.org/browse/FCREPO-2768

Allows you to use the Create Version button in the HTML UI, but not trying to send a Slug header.

# What's new?

# How should this be tested?

Before PR, make a resource and press the button. Get an error.
After PR, make a resource and press the button. Get redirected to the fcr:versions page (which is blank). But if you `curl` the resource there is a version created.

# Additional Notes:

Example:
* Does this change add any new dependencies?  no
* Does this change require any other modifications to be made to the repository (ie. Regeneration activity, etc.)?  no

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-webapp/src/main/webapp/static/js/common.js
+++ b/fcrepo-webapp/src/main/webapp/static/js/common.js
@@ -192,13 +192,8 @@
 
   function createVersionSnapshot(e) {
       const uri = document.getElementById('main').getAttribute('resource');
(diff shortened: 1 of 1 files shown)


yinlinchen:
Works as described in this PR.

PR #1333, opened 2018-04-23
Create (via PUT) ACL resources.

lsitu:
https://jira.duraspace.org/browse/FCREPO-2756

Create ACL resources via PUT.

# What's new?
Create ACL resources via PUT on the static fcr:acl URI http://resource/path/fcr:acl, and the ACL is created in container fedora:acl under the resource like TimeMap instead of linking with the acl:accessControl predicate.

# How should this be tested?
```
1. Create RDFSource
$ curl -i -XPOST -H "Slug: TestObject" -H "Content-Type: text/n3" --data-binary "<> <info:fedora/test/field> \"Test Object\"" http://localhost:8080/rest/

2. Create ACL on the RDFSource and verify status 201 with the location header URI translated to fcr:acl
$ curl -i -XPUT http://localhost:8080/rest/TestObject/fcr:acl
HTTP/1.1 201 Created
Date: Mon, 23 Apr 2018 14:35:18 GMT
ETag: W/"75a70600c3237bd40744b79377f218d444fae0eb"
Last-Modified: Mon, 23 Apr 2018 14:35:18 GMT
Link: <http://localhost:8080/static/constraints/ContainerConstraints.rdf>; rel="http://www.w3.org/ns/ldp#constrainedBy"
Location: http://localhost:8080/rest/TestObject/fcr:acl
Content-Type: text/plain
Content-Length: 45
Server: Jetty(9.2.3.v20140905)

3. Create NonRdfSource
$ curl -i -XPOST -H "Slug: binary"  http://localhost:8080/rest -H "Content-Type: application/octet-stream" --data-binary "binary content"

4. Create ACL on the NonRdfSource and verify status 201 with the location header URI translated to fcr:acl
$ curl -i -XPUT http://localhost:8080/rest/binary/fcr:acl
HTTP/1.1 201 Created
Date: Mon, 23 Apr 2018 14:50:41 GMT
ETag: W/"cb84ff95041ef51503db7a6b6b01e9db2df1ddd2"
Last-Modified: Mon, 23 Apr 2018 14:50:41 GMT
Link: <http://localhost:8080/static/constraints/ContainerConstraints.rdf>; rel="http://www.w3.org/ns/ldp#constrainedBy"
Location: http://localhost:8080/rest/binary/fcr:acl
Content-Type: text/plain
Content-Length: 41
Server: Jetty(9.2.3.v20140905)
```

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraAcl.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraAcl.java
@@ -17,8 +17,12 @@
  */
 package org.fcrepo.http.api;
 
+import static javax.ws.rs.core.Response.created;
+import static javax.ws.rs.core.Response.noContent;
 import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_WEBAC_ACL;
 import static org.slf4j.LoggerFactory.getLogger;
+
+import java.net.URI;
 
 import javax.jcr.nodetype.ConstraintViolationException;
 import javax.servlet.http.HttpServletResponse;
@@ -35,8 +39,6 @@ import org.fcrepo.http.api.PathLockManager.AcquiredLock;
 import org.fcrepo.kernel.api.models.FedoraResource;
 import org.slf4j.Logger;
 import org.springframework.context.annotation.Scope;
-
-import com.google.common.annotations.VisibleForTesting;
 
 /**
  * @author lsitu
@@ -59,15 +61,6 @@ public class FedoraAcl extends ContentExposingResource {
      */
     public FedoraAcl() {
         super();
-    }
-
-    /**
-     * Create a new FedoraNodes instance for a given path
-     * @param externalPath the external path
-     */
-    @VisibleForTesting
-    public FedoraAcl(final String externalPath) {
-        this.externalPath = externalPath;
     }
 
     /**
@@ -94,7 +87,13 @@ public class FedoraAcl extends ContentExposingResource {
             lock.release();
         }
 
-        return createUpdateResponse(aclResource, created);
+        addCacheControlHeaders(servletResponse, resource, session);
+        final URI location = getUri(aclResource);
+        if (created) {
+            return created(location).build();
+        } else {
+            return noContent().location(location).build();
+        }
     }
 
     @Override
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraAcl.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraAcl.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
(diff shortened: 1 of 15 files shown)


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -176,6 +176,9 @@
 
     public static final String LDPCV_BINARY_TIME_MAP = "fedora:binaryTimemap";
 
+    @VisibleForTesting
```
awoods: This annotation can be removed. 
`CONTAINER_WEBAC_ACL` is also used in the non-test context of `HttpResourceConverter`.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraAcl.java
```diff
@@ -0,0 +1,104 @@
...
+    public FedoraAcl() {
+        super();
+    }
+
+    /**
+     * Create a new FedoraNodes instance for a given path
+     * @param externalPath the external path
+     */
+    @VisibleForTesting
+    public FedoraAcl(final String externalPath) {
```
awoods: This constructor can be removed, as it is not used.

awoods:
When `PUT`ting to a resource (e.g. /rest/test/fcr:acl), I am not sure the response headers are correct.
I am getting the Link header:
```
Link: <http://localhost:8080/static/constraints/ContainerConstraints.rdf>; rel="http://www.w3.org/ns/ldp#constrainedBy"
```
I do not think we are considering the /fcr:acl resource a container.

Also, I would expect to get a `Location` header in the response indicating the location of the newly created resource.

lsitu:
@awoods I'll remove the constrainedBy link header from the ACL resource creation response then.

For the /fcr:acl container, to add child resources like permissions, the current code base may expect it to be a Container (see [FedoraLdp.java#L640-L641](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java#L640-L641)) and the integration tests in WebAc module([WebACRecipesIT.java](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-auth-webac/src/test/java/org/fcrepo/integration/auth/webac/WebACRecipesIT.java)) will be failed. Does it sound good to change the condition in FedoraLdp.java#L640-L641 to allow the ACL resource that is not a Container to have children for thoses tests in WebACRecipesIT?

For the Location header, do you expect the location header to be the converted URI "http://localhost:8080/rest/TestObject/fcr:acl" or the actually URI location "http://localhost:8080/rest/TestObject/fedora:acl"?

awoods:
@lsitu: when originally creating a repository resource, the response does not include:
```
Link: <http://localhost:8080/rest/obj/fcr:acl>; rel="acl"
```
However, subsequent `GET`s on the resource (/obj) do include the above header.
Would you like to resolve that in this PR or in a follow-on PR?

lsitu:
@awoods I think I can fix it here.

lsitu:
@awoods I've added commit https://github.com/fcrepo4/fcrepo4/pull/1333/commits/07cb3dbaa90071341e45b5a11ce00043f0fd92e1 for the /fcr:acl link header with the original resource creation.

PR #1337, opened 2018-04-23
FCREPO-2764. Reimplement PrincipalProviders as Servlet Filters.

mohideen:
Resolves: https://jira.duraspace.org/browse/FCREPO-2764

Reimplement the principal providers as Servlet Filters.

# What's new?
Added an AbstractPrincipalProvider class that takes care of adding principals to the session for Shiro.

# How should this be tested?
IT tests needs to be added when all the issues under https://jira.duraspace.org/browse/FCREPO-2753 are complete.

# Additional Notes:
The ServletContainerAuthenticationProvider class has been updated to make it compatible with the PrinicipalProvider interface changes to avoid build failure. The ServletContainerAuthenticationProvider class can be ultimately removed when the Shiro based implementation is completed.

# Interested parties
@fcrepo4/committers, @peichman-umd

### DIFF
--- a/fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/AbstractPrincipalProvider.java
+++ b/fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/AbstractPrincipalProvider.java
@@ -40,7 +40,7 @@ import org.slf4j.Logger;
  * This abstract class implements the Filter interface to add principals
  * to the session so that Shiro can evaluate the principals.
  * 
- * The getPrincipals methods is left as the responsibility of the classes
+ * The getPrincipals method is left as the responsibility of the classes
  * that will be extending this abstract class.
  *
  * @author Mohamed Abdul Rasheed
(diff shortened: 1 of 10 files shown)


fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/AbstractPrincipalProvider.java
```diff
@@ -0,0 +1,78 @@
...
+ * that will be extending this abstract class.
+ *
+ * @author Mohamed Abdul Rasheed
+ * @see PrincipalProvider
+ */
+abstract class AbstractPrincipalProvider implements PrincipalProvider {
+
+    private static final Logger log = getLogger(AbstractPrincipalProvider.class);
+
+    public static final String REALM_NAME = "org.fcrepo.auth.webac.WebACAuthorizingRealm";
```
awoods: This field can be `private`.

fcrepo-auth-common/src/main/java/org/fcrepo/auth/common/PrincipalProvider.java
```diff
@@ -36,19 +37,18 @@
  * @author Gregory Jansen
  * @see HttpHeaderPrincipalProvider
  */
-public interface PrincipalProvider {
+public interface PrincipalProvider extends Filter {
 
     /**
      * Extract principals from the provided credentials.
```
awoods: Change to `...from the provided HttpServletRequest`.

PR #1340, opened 2018-04-25
Return 403 on deletes when  acl:Append applies

mohideen:
Resolves: https://jira.duraspace.org/browse/FCREPO-2715

Adds unit test to verify DELETE requests fail when the user has acl:Append but no acl:Write.

# How should this be tested?
From the fcrepo-auth-webac module run `mvn clean test -Dtest=WebACFilterTest`

# Interested parties
@fcrepo4/committers, @dbernstein


mohideen:
Replaced with #1349

PR #1346, opened 2018-04-26
Adds support for configurable pairtree length and count of minted identifiers

randalldfloyd:
**Resolves FCREPO-2671**: https://jira.duraspace.org/browse/FCREPO-2671
**Resolves FCREPO-2672**: https://jira.duraspace.org/browse/FCREPO-2672

Adds a new configurable PID minting supplier that allows for the creation of either hierarchical or non-hierarchical identifiers.  Using the supplier:

* with no args returns a non-hierarchical identifier
* with length and count args returns a hierarchical identifier with a parent pairtree of (count) number of parts that are (length) long
* with length or count args of zero returns a non-hierarchical identifier

The length and count is optionally configurable via Spring XML, which in turn can also read values from system properties.  This resolves FCREPO-2672: https://jira.duraspace.org/browse/FCREPO-2672

# How should this be tested?
```
1. Create a resource without a slug:
$ curl -i -XPOST  -H "Link: <http://mementoweb.org/ns#OriginalResource>; rel=\"type\"" "http://localhost:8080/rest" \
-H "Content-Type: text/n3" --data-binary "<> <info:fedora/test/field> \"Versioned Resource\""

2. Check that resulting resource identifier does not contain a parent pairtree
Location: http://localhost:8080/rest/8530d479-33ec-4aa7-ab9f-36817c08aa4d

3. Uncomment the following from fcrepo-webapp/src/main/resources/spring/fcrepo-config.xml 
(and restart fcrepo-webapp if applicable.)
<bean class="org.fcrepo.kernel.api.services.functions.ConfigurableHierarchicalSupplier"
        c:desiredLength="${fcrepo.uuid.path.length:2}"
        c:desiredCount="${fcrepo.uuid.path.count:4}"/>

4. Create a resource without a slug:
$ curl -i -XPOST  -H "Link: <http://mementoweb.org/ns#OriginalResource>; rel=\"type\"" "http://localhost:8080/rest" \
-H "Content-Type: text/n3" --data-binary "<> <info:fedora/test/field> \"Versioned Resource\""

5. Check that resulting resource identifier contains a parent pairtree
Location: http://localhost:8080/rest/5c/06/12/b4/5c0612b4-265a-4333-8446-8abf02b25e75
```

Interested parties/reviewers:
@awoods @dbernstein @whikloj


whikloj:
@randalldfloyd Instead of switching from using the `UniqueValueSupplier` interface, you can just write a new class with a configurable constructor. I am assuming the PIDs will not get changed often as that could cause you some problems.

But then you leave the `get()` method alone instead of passing the parameters with each PID. 

Something like.
```
public class ConfigurableHierarchicalSupplier implements UniqueValueSupplier {

    private final int length;

    private final int count;

    ConfigurableHierarchicalSupplier(final int desiredLength, final int desiredCount) {
        length = desiredLength;
        count = desiredCount;
    }

    ConfigurableHierarchicalSupplier() {
        length = HierarchicalIdentifierSupplier.DEFAULT_LENGTH;
        count = HierarchicalIdentifierSupplier.DEFAULT_COUNT;
    }

    /**
     * Mint a unique identifier as a UUID
     *
     * @return uuid
     */
    @Override
    public String get() {

        final String s = randomUUID().toString();
        final StringJoiner joiner = new StringJoiner("/", "", "/" + s);

        IntStream.rangeClosed(0, count - 1)
                .forEach(x -> joiner.add(s.substring(x * length, (x + 1) * length)));

        return joiner.toString();
    }
}
```

Then you pass the parameters when you create it.
```
protected UniqueValueSupplier defaultPidMinter = new DefaultPathMinter(2, 5);

private static class DefaultPathMinter extends ConfigurableHierarchicalSupplier {}
```

randalldfloyd:
@whikloj  Thanks, that's a lot better. Now I understand your suggestion on slack yesterday, I didn't quite grok it correctly.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/functions/ConfigurableHierarchicalSupplier.java
```diff
@@ -0,0 +1,80 @@
...
+    public ConfigurableHierarchicalSupplier(final int desiredLength, final int desiredCount) {
+        length = desiredLength;
+        count = desiredCount;
+    }
+
+    /**
+     * Mint a unique identifier by default using defaults
+     *
+     */
+    protected ConfigurableHierarchicalSupplier() {
```
whikloj: Make `public` for reasons outlined above.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/AbstractResource.java
```diff
@@ -97,7 +98,12 @@
     @Optional
     protected Supplier<String> pidMinter;
 
-    protected UniqueValueSupplier defaultPidMinter = new DefaultPathMinter();
+    // TODO: Allow configuration properties to override to length and count
+    final int desiredLength = HierarchicalIdentifierSupplier.DEFAULT_LENGTH;
+    final int desiredCount = HierarchicalIdentifierSupplier.DEFAULT_COUNT;
```
whikloj: You can remove these. They are assumed if you instantiate the class without any arguments. So
```
protected UniqueValueSupplier defaultPidMinter = new ConfigurableHierarchicalSupplier(2, 4);
```
uses the value you pass and 
```
protected UniqueValueSupplier defaultPidMinter = new ConfigurableHierarchicalSupplier();
```
will use the constructor you will make from `protected` to `public` below which assumes the values from `HierarchicalIdentifierSupplier`.

whikloj:
I feel like after seeing this, we probably should have just removed the entire `HierarchicalIdentiferSupplier` as it will never get used after this. Just move the defaults directly into your new `Configurable...` class. If you want to work that in here, great. If not I can open a new JIRA and do that after.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/functions/ConfigurableHierarchicalSupplier.java
whikloj: or below if viewed on the main PR page.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/AbstractResource.java
```diff
@@ -97,7 +98,12 @@
...
+
+    protected UniqueValueSupplier defaultPidMinter =
+            new ConfigurableHierarchicalSupplier(desiredLength, desiredCount);
```
whikloj: If you directly instantiate the `ConfigurableHierarchicalSupplier` above, then this is not required and could be removed.

randalldfloyd:
@whikloj All requested changes have been made, but if they vary slightly it's because I also sneaked in resolution to https://jira.duraspace.org/browse/FCREPO-2672 (configurable pairtrees via Spring and/or system properties.) As required by 2671, default behavior is no pairtrees, and re-enabling pairtrees via configuration requires editing Spring XML as noted in PR comments.   Per your suggestion I also went ahead and got rid of `HierarchicalIdentiferSupplier`.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -1902,8 +1902,9 @@ public void testIngestOnSubtree() throws IOException {
      */
     @Test
     public void testIngestOnPairtree() throws IOException {
-        //  Following the approach undertaken for FedoraExportIT#shouldRoundTripOnePairtree
-        final String objName = getLocation(postObjMethod());
+        final String parent = "123456/789/10";
+        executeAndClose(putObjMethod(parent));
```
whikloj: This test is no longer valid with these changes.

It seems like if the test is still required then we should create a separate test class which runs with the old pidminter setup so we could still test pairtrees.

Not sure, @awoods

whikloj:
I only have a question about the one test for ingesting in pairtrees. That test could be reverted and `@Ignore`d for now or it could be (as I suggested) run with a different Pid Minter configuration.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
whikloj: Yeah, I figured just telling the PID to have a `/` in it wouldn't create a pairtree, but I'm not really sure. Hence not blocking this. I'm fine, if this test is still useful.

randalldfloyd:
@awoods @dbernstein Made minor changes as requested, will also work on updating the wiki documentation mentioned.

PR #1360, opened 2018-05-31
Fcrepo-2802 - Update content size for url binaries on 

bbpennel:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2802

Removes call to set content size property for `UrlBinary`s on read requests (GET or HEAD), to prevent potential unwanted modifications.

# What's new?
* Calls to setContentSize for LocalFile and URL binaries removed from getContentSize call, now only being invoked when the binary is updated.
* Changed how size of an http url is determined to check for a Content-Length header, to avoid memory issues resulting from loading the remote binary into memory to calculate size.

# How should this be tested?
```
##### Http resource

# create the resource 
curl -v -X PUT -H "Link: <https://duraspace.org/wp-content/themes/duraspace/assets/images/whitedura.png>; rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"; handling=\"proxy\"; type=\"image/png\"" http://localhost:8080/rest/duraspace-logo.png -ufedoraAdmin:fedoraAdmin

#verify that premis:hasSize property is set
curl http://localhost:8080/rest/duraspace-logo.png/fcr:metadata -ufedoraAdmin:fedoraAdmin
# premis:hasSize           "3361"^^<http://www.w3.org/2001/XMLSchema#long> ;

##### Local File

# Create local file
echo "file content" >> /tmp/local.txt
# Ingest local file
curl -v -X PUT -H "Link: <file:/tmp/local.txt>; rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"; handling=\"proxy\"; type=\"image/png\"" http://localhost:8080/rest/local.txt -ufedoraAdmin:fedoraAdmin

# verify size
curl http://localhost:8080/rest/local.txt/fcr:metadata -ufedoraAdmin:fedoraAdmin
# premis:hasSize           "13"^^<http://www.w3.org/2001/XMLSchema#long> ;

# Change file behind fedora's back
echo "uh oh" >> /tmp/local.txt

# verify size property hasn't changed on a GET operation
curl http://localhost:8080/rest/local.txt/fcr:metadata -ufedoraAdmin:fedoraAdmin
# premis:hasSize           "13"^^<http://www.w3.org/2001/XMLSchema#long> ;
```

# Additional Notes
Triggering unintentional modifications this way would be fairly challenging, but it is best to avoid the possibility as there may be unintended consequences in other use cases. Also, GET on a LocalFileBinary that references a file that doesn't exist currently throws a FileNotFoundException. When that is resolved it might be easier to trigger this.

# Interested parties
@bseeger

### DIFF
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/UrlBinary.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/UrlBinary.java
@@ -103,7 +103,7 @@ public class UrlBinary extends AbstractFedoraBinary {
                     try {
                         return Long.parseLong(contentLength);
                     } catch (final NumberFormatException e) {
-                        LOGGER.debug("Unable to parse Content-Length of remote file {}", resourceUri, e);
+                        LOGGER.warn("Unable to parse Content-Length of remote file {}", resourceUri, e);
                     }
                 }
             }
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/UrlBinary.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/UrlBinary.java
@@ -24,6 +24,7 @@ import static org.fcrepo.kernel.api.FedoraExternalContent.REDIRECT;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/UrlBinary.java
```diff
@@ -89,19 +88,28 @@ public InputStream getContent() {
...
+
+            final int status = httpConn.getResponseCode();
+
+            if (status == HttpURLConnection.HTTP_OK) {
+                final String contentLength = httpConn.getHeaderField("Content-Length");
+                if (contentLength != null) {
+                    try {
+                        return Long.parseLong(contentLength);
+                    } catch (final NumberFormatException e) {
+                        LOGGER.debug("Unable to parse Content-Length of remote file {}", resourceUri, e);
```
bseeger: I'm be interested in this being `info` or `warn` versus debug, since it's a pretty serious issue.

PR #1361, opened 2018-06-29
Don't return Content-Length for LDP-NR redirect responses

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2818

Removes the `Content-Length` header from responses for External-Content Redirect LDP-NRs.

# How should this be tested?

To test the actual timing of a response you need a remote URI to redirect to, but you can test the change to the headers with a local file.

1. `curl -i -ufedoraAdmin:fedoraAdmin -XPUT -H"Link: <file:///path/to/your/file>; rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"; handling=\"redirect\"" http://localhost:8080/rest/my_redirect`
1. Before the PR
  1. do a HEAD request. 
  1. See the `Content-Length:` greater than 0
1. After the PR
  1. do a HEAD request
  1. See a `Content-Length: 0` header

If you have a remote webserver, you can use that URI and actually see how much faster it is with.
```
curl -G -L -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/my_redirect
```

# Additional Notes:

Example:

# Interested parties
@awoods @bbpennel

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -564,6 +564,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
                     .build();
 
             servletResponse.addHeader(CONTENT_TYPE, binary.getMimeType());
+            // Returning content-length > 0 causes the client to wait for additional data before following the redirect.
             if (!binary.isRedirect()) {
                 servletResponse.addHeader(CONTENT_LENGTH, String.valueOf(binary.getContentSize()));
             }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -564,7 +564,9 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
                     .build();
 
             servletResponse.addHeader(CONTENT_TYPE, binary.getMimeType());
-            servletResponse.addHeader(CONTENT_LENGTH, String.valueOf(binary.getContentSize()));
+            if (!binary.isRedirect()) {
+                servletResponse.addHeader(CONTENT_LENGTH, String.valueOf(binary.getContentSize()));
+            }
             servletResponse.addHeader("Accept-Ranges", "bytes");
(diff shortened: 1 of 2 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -564,7 +564,9 @@ protected void addResourceHttpHeaders(final FedoraResource resource) {
                     .build();
 
             servletResponse.addHeader(CONTENT_TYPE, binary.getMimeType());
-            servletResponse.addHeader(CONTENT_LENGTH, String.valueOf(binary.getContentSize()));
+            if (!binary.isRedirect()) {
```
bbpennel: My only suggestion would be to add a comment here mentioning why we would want to drop the header for redirects

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -456,6 +456,7 @@ public void testHeadExternalDatastreamRedirect() throws IOException, ParseExcept
             assertEquals(TEMPORARY_REDIRECT.getStatusCode(), response.getStatusLine().getStatusCode());
             assertEquals("http://example.com/test", getLocation(response));
             assertEquals("bytes", response.getFirstHeader("Accept-Ranges").getValue());
+            assertEquals("0", response.getFirstHeader("Content-Length").getValue());
```
bbpennel: It's too bad this comes through as a Content-Length = 0, which seems slightly more misleading than not specifying a content length, but I guess that's just the way it is in order to keep clients happy.

whikloj: I can take a better look when I have a chance, but for a `GET` request I am assuming that this is correct as the body length is 0. For a `HEAD` request I think it should be removed completely.

bbpennel: I'm okay with this going in as is. If we are going to have Content-Length on other HEAD requests, then it makes sense to have a consistent response for redirect binaries across HEAD and GET. Was just lamenting that it could be confusing to see a content-length of 0 when that's not the actual length, but its just the first hop in a redirect so it should be fine.

PR #1367, opened 2018-07-11
Disambiguate content-type headers

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2827

Renames the ExternalContentHandler static `CONTENT_TYPE` to `EXT_CONTENT_TYPE` (which resolves to `type`) to avoid naming confusion with the various public statics `CONTENT_TYPE` which resolve to `Content-Type`.

# What's new?
Simple rename and a static import, along with a test.

# How should this be tested?

Before the PR, created a proxied external content resource pointing at a remote web server that **does** return a `Content-Type` header. 

This header is not used and your resource with have a `Content-Type: application/octet-stream`.

After this PR your resource should have the same `Content-Type` that the remote server returns.

# Additional Notes:

Example:

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/ExternalContentHandlerIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/ExternalContentHandlerIT.java
@@ -1,0 +1,73 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/ExternalContentHandlerIT.java
```diff
@@ -0,0 +1,70 @@
...
+    private static final String NON_RDF_SOURCE_LINK_HEADER = "<" + NON_RDF_SOURCE.getURI() + ">;rel=\"type\"";
+
+    @Test
+    public void testRemoteUriContentType() throws Exception {
+        final HttpPost method = postObjMethod();
+        method.addHeader(CONTENT_TYPE, "audio/ogg");
+        method.addHeader(LINK, NON_RDF_SOURCE_LINK_HEADER);
+        method.setEntity(new StringEntity("xyz"));
+        final String external_location;
+        final String final_location = "proxy_resource";
```
awoods: This test may be more reliable if instead of having a hard-coded resource name, you use `getRandomUniqueId()`.

whikloj: I can make this change as a follow-on PR.

PR #1373, opened 2018-07-18
PUT requests to fcr:acl now accept RDF bodies.

dbernstein:
Resolves: https://jira.duraspace.org/browse/FCREPO-2808

Allows user to specify a RDF body in a PUT request on an ACL resource.

# How should this be tested?
1. Save the following to a file call acl.ttl
```
@prefix foaf: <http://xmlns.com/foaf/0.1/> . 
@prefix ldp: <http://www.w3.org/ns/ldp#> . 
                                
<#authorization_1> a acl:Authorization ; 
                   acl:agentClass foaf:Agent ; 
                   acl:accessToClass ldp:Resource ; 
                   acl:default <https://example.org/root/> ; 
                   acl:mode acl:Read .
```

2.  create a resource
```
curl -X PUT http://localhost:8080/rest/test
```
3. create the acl using a body
```
curl -v -X PUT  http://localhost:8080/rest/test/fcr:acl -H "Content-Type: text/turtle" --data-binary "@acl.ttl"
```
4. verify that the resource was properly created:
```
curl -v -X GET  http://localhost:8080/rest/test/fcr:acl 
```

# Interested parties
@peichman-umd , @fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraAclIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraAclIT.java
@@ -60,7 +60,7 @@ public class FedoraAclIT extends AbstractResourceIT {
     }
 
     @Test
-    public void testCreateAcl() throws Exception {
+    public void testCreateAclWithoutBody() throws Exception {
         createObjectAndClose(id);
 
         final HttpPut put = new HttpPut(subjectUri + "/" + FCR_ACL);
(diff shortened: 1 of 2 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraAclIT.java
```diff
@@ -184,4 +185,31 @@ public void testGetNonExistentAcl() throws Exception {
...
+
+        put.setEntity(new StringEntity(aclBody));
+        put.setHeader("Content-Type", "text/turtle");
+
+        final String aclLocation;
+        try (final CloseableHttpResponse response = execute(put)) {
+            assertEquals(CREATED.getStatusCode(), getStatus(response));
+            aclLocation = response.getFirstHeader("Location").getValue();
+            // verify the acl container is translated to fcr:acl
+            assertEquals(subjectUri + "/" + FCR_ACL, aclLocation);
```
awoods: We should probably test the content of the returned body to make sure that the triples were indeed added.

dbernstein: Ah good catch.

peichman-umd:
👍

PR #1375, opened 2018-07-19
Return 405 on PUT, POST and PATCH to LDPCv.

lsitu:
Return 405 on PUT, POST and PATCH to LDPCv.
* * *

**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2821

Return status 405 on PUT, POST and PATCH to LDPCv,  and include the option headers for the existing memento resource.

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -81,7 +81,6 @@ import static org.fcrepo.kernel.api.RequiredRdfContext.LDP_MEMBERSHIP;
 import static org.fcrepo.kernel.api.RequiredRdfContext.MINIMAL;
 import static org.fcrepo.kernel.api.RequiredRdfContext.PROPERTIES;
 import static org.fcrepo.kernel.api.RequiredRdfContext.SERVER_MANAGED;
-import static org.fcrepo.kernel.modeshape.FedoraResourceImpl.LDPCV_TIME_MAP;
 import static org.slf4j.LoggerFactory.getLogger;
 
 import com.fasterxml.jackson.core.JsonParseException;
@@ -576,7 +575,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
      * @param path the path of the memento
      */
     protected void mementoLinkAndOptionsHttpHeaders(final String path) {
-        if (path.contains("/" + LDPCV_TIME_MAP + "/") && nodeService.exists(session.getFedoraSession(), path)) {
+        if (nodeService.exists(session.getFedoraSession(), path)) {
             addLinkAndOptionsHttpHeaders(resource());
         }
     }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -571,16 +571,6 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
     }
 
     /**
-     * Add Link and Option headers if the resource exists
-     * @param path the path of the resource
-     */
-    protected void addLinkAndOptionsHttpHeadersIfExists(final String path) {
-        if (nodeService.exists(session.getFedoraSession(), path)) {
-            addLinkAndOptionsHttpHeaders(resource());
-        }
-    }
-
-    /**
      * Utility function for building a Link.
      *
      * @param linkUri String of URI for the link.
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -81,6 +81,7 @@ import static org.fcrepo.kernel.api.RequiredRdfContext.LDP_MEMBERSHIP;
 import static org.fcrepo.kernel.api.RequiredRdfContext.MINIMAL;
 import static org.fcrepo.kernel.api.RequiredRdfContext.PROPERTIES;
 import static org.fcrepo.kernel.api.RequiredRdfContext.SERVER_MANAGED;
+import static org.fcrepo.kernel.modeshape.FedoraResourceImpl.LDPCV_TIME_MAP;
 import static org.slf4j.LoggerFactory.getLogger;
 
 import com.fasterxml.jackson.core.JsonParseException;
@@ -568,6 +569,16 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
         }
 
         servletResponse.addHeader("Allow", options);
+    }
+
+    /**
+     * Add Link and Option headers for a memento
+     * @param path the path of the memento
+     */
+    protected void mementoLinkAndOptionsHttpHeaders(final String path) {
+        if (path.contains("/" + LDPCV_TIME_MAP + "/") && nodeService.exists(session.getFedoraSession(), path)) {
+            addLinkAndOptionsHttpHeaders(resource());
+        }
     }
 
     /**
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -571,10 +571,10 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
     }
 
     /**
-     * Add Link and Option headers for a memento
-     * @param path the path of the memento
+     * Add Link and Option headers if the resource exists
+     * @param path the path of the resource
      */
-    protected void mementoLinkAndOptionsHttpHeaders(final String path) {
+    protected void addLinkAndOptionsHttpHeadersIfExists(final String path) {
         if (nodeService.exists(session.getFedoraSession(), path)) {
             addLinkAndOptionsHttpHeaders(resource());
         }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -64,7 +64,6 @@ import static org.fcrepo.kernel.api.RdfLexicon.INTERACTION_MODELS;
 import static org.fcrepo.kernel.api.RdfLexicon.NON_RDF_SOURCE;
 import static org.fcrepo.kernel.api.RdfLexicon.VERSIONED_RESOURCE;
 import static org.fcrepo.kernel.api.FedoraExternalContent.COPY;
-import static org.fcrepo.kernel.modeshape.FedoraResourceImpl.LDPCV_TIME_MAP;
 import static org.slf4j.LoggerFactory.getLogger;
 
 import java.io.IOException;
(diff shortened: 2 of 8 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -565,6 +566,16 @@ protected void addLinkAndOptionsHttpHeaders(final FedoraResource resource) {
         servletResponse.addHeader("Allow", options);
     }
 
+    /**
+     * Add Link and Option headers for a memento
+     * @param path the path of the memento
+     */
+    protected void mementoLinkAndOptionsHttpHeaders(final String path) {
+        if (path.contains("/" + LDPCV_TIME_MAP + "/") && nodeService.exists(session.getFedoraSession(), path)) {
```
whikloj: I think these two tests are the same thing?
*    `path.contains(LDPCV_TIME_MAP)`
*    `externalPath.contains("/" + FedoraTypes.FCR_VERSIONS)`

So I'm not sure this is adding anything new

whikloj:
I ran the same commands against `master` and this PR and got the same results. So I think much of this is already in master. I think we should keep the tests, but I think the rest of the changes are not having any effect.

But perhaps someone else could check to ensure I am not going crazy.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
lsitu: @whikloj I think you may have got the wrong image. We've made LDPCV_TIME_MAP (https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java#L175) the node name, which is the internal representation of FedoraTypes.FCR_VERSIONS (https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraTypes.java#L94). 
The issue here is that our code won't translate those version ID that are not in digit format (seconds), and this will produce error codes other than 405 while retrieving the node from Mode.

lsitu:
@whikloj Well, I got all those new tests failed on master branch, and commenting out those code changes got the same result. Could you try it again?
screen shot 2018-07-19 at 8 11 56 am

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
whikloj: Ok so this deals with non-existent mementos. So I think the actual problem is the ticket. An LDPCv is the `/fcr:versions` end-point. Anything under that is an LDPRm. I'm going to add a comment to the original ticket.
There is nothing in the specification about how a PUT to an LDPRm should respond [4.2.4](https://fedora.info/2018/06/25/spec/#ldprm-put)

lsitu:
@dbernstein / @whikloj I've added a commit to update the pattern for memento ID to use any characters instead of the digits only, but keeping status 405 for PUTing on LDPCv/menento path in any case since we might really want the user know that PUTing to a menemto is not allowed.

dbernstein:
@whikloj :  this looks good to me.  I checked out that the PR does in fact produce the right return codes, whereas master does not.  Any reason not to squash and merge that you can see?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -565,6 +565,16 @@ protected void addLinkAndOptionsHttpHeaders(final FedoraResource resource) {
         servletResponse.addHeader("Allow", options);
     }
 
+    /**
+     * Add Link and Option headers for a memento
+     * @param path the path of the memento
+     */
+    protected void mementoLinkAndOptionsHttpHeaders(final String path) {
```
awoods: Since there is nothing in this method that requires or checks that the param `path` is a memento, a clearer method name may be: `addLinkAndOptionsHttpHeadersIfExists`.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
```diff
@@ -90,7 +90,7 @@
     // The fifth group for allows ACL.
     // The sixth group allows for any hashed suffixes on the acls.
     private final static Pattern FORWARD_COMPONENT_PATTERN = Pattern.compile(
-            "(.*?)(/" + FCR_METADATA + ")?(/" + FCR_VERSIONS + "(/\\d+)?)?(/" + FCR_ACL + "(\\#\\w+)?)?$");
+            "(.*?)(/" + FCR_METADATA + ")?(/" + FCR_VERSIONS + "(/.+)?)?(/" + FCR_ACL + "(\\#\\w+)?)?$");
```
awoods: I am reluctant to change this regex if the reason is just for the integration test.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NamespaceTools.java
```diff
@@ -85,7 +85,7 @@ public static void validatePath(final Session session, final String path) {
             final int colonPosition = segment.indexOf(':');
             if (segment.length() > 0 && colonPosition > -1) {
                 final String prefix = segment.substring(0, colonPosition);
-                if (!prefix.equals("fedora")) {
+                if (!prefix.equals("fedora") && !prefix.equals("fcr")) {
```
awoods: Why is the addition of the `fcr` prefix now being skipped?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
lsitu: @awoods I think one of the reasons for the pattern change is @whikloj's `19/Jul/18 1:15 PM` comment https://jira.duraspace.org/browse/FCREPO-2821#comment-60168: `We do (at least I think) need to alter the FORWARD_COMPONENT_PATTERN as that would make GET requests to things like /fcr:versions/something fail with a 404 instead of a 400.` 

And I am also seeing that with the origin pattern to allow digits only for Memento ID, if an incoming path is something like http://localhost:8080/rest/test4/obj/fcr:versions/new, it won't get translated to the internal path `/test4/obj/fedora:Timemap/new`, and I need to do a special check to see whether it can be translated or not before testing for the existing of the JCR node, which will ensure status 405 return. See ContentExposingResource.java#L574 (https://github.com/fcrepo4/fcrepo4/pull/1375/commits/44b76a9d8c05c669fa149bc37d767506f415cc23#diff-d372035f89c8d0846920e20bd13e0145R574).

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NamespaceTools.java
lsitu: I think if it fails to retrieve the Fedora resource, it'll fall down to the validatePath(jcrSession, path) method above from https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java#L148, which will have the 400 error for "Prefix has not been registered" instead of status 405 we want with PUTing, POSTing or PATCHing to a LDCPv/Memento path.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
awoods: My understanding is that memento IDs are being generated by Fedora... that we do not allow users to specify their own names for mementos, and therefore the regex should never be violated.
Can you confirm @bbpennel ?

lsitu: @awoods / @bbpennel Do you want me to change it back to use the digits pattern for memento ID? I think it's correct that the memento IDs are being generated by FedoraI at this time and it's always digits. But for an incoming path like `/rest/test4/obj/fcr:versions/new`, it won't get converted properly to the corresponding internal path `/rest/test4/obj/fedora:timemap/new` when using the digits pattern.

lsitu:
@bbpennel Thanks. I've added commit https://github.com/fcrepo4/fcrepo4/pull/1375/commits/5b648e5dd91a19050ca485471e0e7a5223cc35a5 to split the tests for PUT, POST and PATCH to mementos and corrected the indent spaces.

dbernstein:
1. Change the regular expression to allow 14 digit 
2. Constraint document to explain why the 400 occurred (on  GET)
3. 405 on PUT should stand.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpResourceConverter.java
whikloj: > but it may be best to separate that from the test to verify that PUT/POST/PATCH are being disallowed, and that a resources doesn't exist.

👍 👍

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NamespaceTools.java
whikloj: Yeah this in a by-product of the regex being discussed above not matching on the a URI like `/fcrepo/rest/versioned_resource/fcr:versions/new`.

When it matches you get a PathNotFoundException before here, when it doesn't match it tries to find the `/fcr:versions`.

Alternatively we could duplicate the work in `FedoraLdp` for `GET` and move it to `FedoraVersioning` to match particularly on paths like `/fcr:versions/.*`. But that could be a lot of duplicated code.

lsitu:
@fcrepo4/committers The PR is updated basing on the discussions and decisions from yesterday's Fedora Tech meeting, and it's rebased with master. GET/HEAD/OPTIONS requests with invalid memento path are also handled with status 400 and constraint document.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NamespaceTools.java
awoods: I would suggest a comment be added to explain why the `fcr` prefix is being disregarded. In that way, we can more easily know the implications of potentially removing the `!prefix.equals("fcr")` check in the future.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -1024,4 +1031,29 @@ private void checkAclLinkHeader(final List<String> links) throws RequestWithAclL
...
+            FedoraTypes.FCR_VERSIONS, externalPath);
+    }
+
+    /*
+     * Ensure that an incoming versioning/memento path can be converted.
+     */
+    private void checkMementoPath() {
+        if (externalPath.contains("/" + FedoraTypes.FCR_VERSIONS)) {
+            final String path = toPath(translator(), externalPath);
+            if (path.contains(FedoraTypes.FCR_VERSIONS) && !LDPCV_TIME_MAP.equals(FedoraTypes.FCR_VERSIONS)) {
```
bbpennel: `!LDPCV_TIME_MAP.equals(FedoraTypes.FCR_VERSIONS)` appears to be comparing two constants which do not have the same value, so it seems like this will always evaluate to true. Is it needed?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/NamespaceTools.java
lsitu: Sure. I'll add a comment for it.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
lsitu: Yes. This can be removed.

PR #1376, opened 2018-07-19
Fcrepo 2835 - Allowed external paths configuration

bbpennel:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2835

Adds an optional configuration file for specifying what URIs external content may be created from. If no configuration is provided, then external binary creation is disabled.

# What's new?
* Adds java startup parameter "fcrepo.external.content.allowed" which allows users to optionally specify where the allowed list configuration file is.
* Introduces a ExternalContentPathValidator which verifies that a path is from the allowed list and does not contain relative modifiers.
* Introduces ExternalContentHandlerFactory which constructors ExternalContentHandlers and performs validation of links prior to returning the handler
* Validator configured by default to reload configuration file if it is modified (assuming it was provided at startup time)
* Updates tests accordingly

Example format for the configuration file:
```
file:///allowed/path/
file:/another/allowed/path/
file:///exact/match/file.txt
https://duraspace.com/
http://example.com/
```

# How should this be tested?
Adds tests for new functional, but should also be tested as follows:
```
# Start jetty without allowed list
mvn jetty:run

# Try to create external binaries, should fail with 400 status
curl -v -X PUT -H "Link: <https://duraspace.org/wp-content/themes/duraspace/assets/images/whitedura.png>; rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"; handling=\"proxy\"; type=\"image/png\"" http://localhost:8080/rest/allow_http -ufedoraAdmin:fedoraAdmin

# Try to create external binaries, should fail with 400 status
curl -v -X PUT -H "Link: <file:///Users/bbpennel/git/fcrepo4/NOTICE>; rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"; handling=\"proxy\"; type=\"text/plain\"" http://localhost:8080/rest/allow_file -ufedoraAdmin:fedoraAdmin

# Create allowed list, very permissive
echo "http://
file:/" > allowed.txt

# Stop jetty

# Restart jetty with allowed list configuration
mvn jetty:run -Dfcrepo.external.content.allowed=/Users/bbpennel/git/fcrepo4/allowed.txt

# Should succeed
curl -v -X PUT -H "Link: <file:///Users/bbpennel/git/fcrepo4/NOTICE>; rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"; handling=\"proxy\"; type=\"image/png\"" http://localhost:8080/rest/allow_file -ufedoraAdmin:fedoraAdmin

# Should fail because of ".."
curl -v -X PUT -H "Link: <file:///Users/bbpennel/git/fcrepo4/fcrepo-auth-webac/../NOTICE>; rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"; handling=\"proxy\"; type=\"text/plain\"" http://localhost:8080/rest/allow_file -ufedoraAdmin:fedoraAdmin

# Should fail because https was not permitted
curl -v -X PUT -H "Link: <https://duraspace.org/wp-content/themes/duraspace/assets/images/whitedura.png>; rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"; handling=\"proxy\"; type=\"image/png\"" http://localhost:8080/rest/allow_http -ufedoraAdmin:fedoraAdmin

# Add a new line to the allowed list WITHOUT restarting jetty
echo "https://duraspace.org/wp-content/" >> allowed.txt

# After a few seconds to wait for config to reload, this should now succeed
curl -v -X PUT -H "Link: <https://duraspace.org/wp-content/themes/duraspace/assets/images/whitedura.png>; rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"; handling=\"proxy\"; type=\"image/png\"" http://localhost:8080/rest/allow_http -ufedoraAdmin:fedoraAdmin
```

# Additional Notes:
Not sure if documentation exists for external content yet, but this will definitely need to be documented. Will also need to update the fcrepo configuration options page.

# Interested parties
@awoods @whikloj @bseeger

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentHandlerFactory.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentHandlerFactory.java
@@ -18,11 +18,13 @@
 package org.fcrepo.http.api;
 
 import static org.fcrepo.kernel.api.RdfLexicon.EXTERNAL_CONTENT;
+import static org.slf4j.LoggerFactory.getLogger;
 
 import java.util.List;
 import java.util.stream.Collectors;
 
 import org.fcrepo.kernel.api.exception.ExternalMessageBodyException;
+import org.slf4j.Logger;
 
 /**
  * Constructs ExternalContentHandler objects from link headers
@@ -30,6 +32,8 @@ import org.fcrepo.kernel.api.exception.ExternalMessageBodyException;
  * @author bbpennel
  */
 public class ExternalContentHandlerFactory {
+
+    private static final Logger LOGGER = getLogger(ExternalContentHandlerFactory.class);
 
     private ExternalContentPathValidator validator;
 
@@ -59,7 +63,12 @@ public class ExternalContentHandlerFactory {
             final String link = externalContentLinks.get(0);
             final String uri = getUriString(link);
             // Validate that the URI is valid according to allowed set of external uris
-            validator.validate(uri);
+            try {
+                validator.validate(uri);
+            } catch (final ExternalMessageBodyException e) {
+                LOGGER.warn("Rejected invalid external path {}", uri);
+                throw e;
+            }
 
             return new ExternalContentHandler(link);
         }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentHandlerFactory.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentHandlerFactory.java
@@ -1,0 +1,92 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.http.api;
+
+import static org.fcrepo.kernel.api.RdfLexicon.EXTERNAL_CONTENT;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.fcrepo.kernel.api.exception.ExternalMessageBodyException;
+
+/**
+ * Constructs ExternalContentHandler objects from link headers
+ *
+ * @author bbpennel
+ */
+public class ExternalContentHandlerFactory {
+
+    private ExternalContentPathValidator validator;
+
+    /**
+     * Looks for ExternalContent link header and if it finds one it will return a new ExternalContentHandler object
+     * based on the found Link header. If multiple external content headers were found or the URI provided in the
+     * header is not a valid external content path, then an ExternalMessageBodyException will be thrown.
+     *
+     * @param links links from the request header
+     * @return External Content Handler Object if Link header found, else null
+     * @throws ExternalMessageBodyException thrown if more than one external content link was provided, or if the URL
+     *         of the header was not a valid external content path.
+     */
+    public ExternalContentHandler createFromLinks(final List<String> links) throws ExternalMessageBodyException {
+        if (links == null) {
+            return null;
+        }
+
+        final List<String> externalContentLinks = links.stream()
+                .filter(x -> x.contains(EXTERNAL_CONTENT.toString()))
+                .collect(Collectors.toList());
+
+        if (externalContentLinks.size() > 1) {
+            // got a problem, you can only have one ExternalContent links
+            throw new ExternalMessageBodyException("More then one External Content Link header in request");
+        } else if (externalContentLinks.size() == 1) {
+            final String link = externalContentLinks.get(0);
+            final String uri = getUriString(link);
+            // Validate that the URI is valid according to allowed set of external uris
+            validator.validate(uri);
+
+            return new ExternalContentHandler(link);
+        }
+
+        return null;
+    }
+
+    private static String getUriString(final String link) {
+        final String value = link.trim();
+        if (value.startsWith("<")) {
+            final int gtIndex = value.indexOf('>');
+            if (gtIndex != -1) {
+                return value.substring(1, gtIndex).trim();
+            } else {
+                throw new IllegalArgumentException("Missing token > in " + value);
+            }
+        } else {
+            throw new IllegalArgumentException("Missing starting token < in " + value);
+        }
+    }
+
+    /**
+     * Set the external content path validator
+     *
+     * @param validator validator
+     */
+    public void setValidator(final ExternalContentPathValidator validator) {
+        this.validator = validator;
+    }
+}
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
@@ -58,7 +58,7 @@ public class ExternalContentPathValidator {
 
     private final Pattern RELATIVE_MOD_PATTERN = Pattern.compile(".*(^|/)\\.\\.($|/).*");
 
-    private String allowedListPath;
+    private String configPath;
 
     private List<String> allowedList;
 
@@ -112,7 +112,7 @@ public class ExternalContentPathValidator {
      * Initialize the allow list
      */
     public void init() throws IOException {
-        if (isEmpty(allowedListPath)) {
+        if (isEmpty(configPath)) {
             return;
         }
 
@@ -138,12 +138,12 @@ public class ExternalContentPathValidator {
      * @throws IOException
      */
     private synchronized void loadAllowedPaths() throws IOException {
-        try (final Stream<String> stream = Files.lines(Paths.get(allowedListPath))) {
+        try (final Stream<String> stream = Files.lines(Paths.get(configPath))) {
             allowedList = stream.map(line -> line.trim().toLowerCase())
                 .filter(line -> {
                     if (line.contains("../") || !SCHEME_PATTERN.matcher(line).matches()) {
                         LOGGER.error("Invalid path {} specified in external path configuration {}",
-                                line, allowedListPath);
+                                line, configPath);
                         return false;
                     }
                     return true;
@@ -159,9 +159,9 @@ public class ExternalContentPathValidator {
             return;
         }
 
-        final Path path = Paths.get(allowedListPath);
+        final Path path = Paths.get(configPath);
         if (!path.toFile().exists()) {
-            LOGGER.debug("Allow list configuration {} does not exist, disabling monitoring", allowedListPath);
+            LOGGER.debug("Allow list configuration {} does not exist, disabling monitoring", configPath);
             return;
         }
         final Path directoryPath = path.getParent();
@@ -232,10 +232,10 @@ public class ExternalContentPathValidator {
     /**
      * Set the file path for the allowed external path configuration
      *
-     * @param allowListPath
+     * @param configPath
      */
-    public void setAllowListPath(final String allowListPath) {
-        this.allowedListPath = allowListPath;
+    public void setConfigPath(final String configPath) {
+        this.configPath = configPath;
     }
 
     /**
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
@@ -22,6 +22,7 @@ import static java.nio.file.StandardWatchEventKinds.OVERFLOW;
 import static org.apache.commons.lang3.StringUtils.isEmpty;
 import static org.slf4j.LoggerFactory.getLogger;
 
+import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -36,6 +37,7 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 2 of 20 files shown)


awoods:
Is there a reason that no integration tests are included?

bbpennel:
@awoods There are existing integration tests that use external binaries, but I can add one in which they are rejected.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
```diff
@@ -0,0 +1,269 @@
...
+
+    private Thread monitorThread;
+
+    private boolean monitorRunning;
+
+    /**
+     * Validates that an external path is valid. The path must be an HTTP or file URI within the allow list of paths,
+     * be absolute, and contain no relative modifier.
+     *
+     * @param extPath
```
awoods: Please complete javadoc.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
```diff
@@ -0,0 +1,269 @@
...
+     * @throws ExternalMessageBodyException
```
awoods: Please complete javadoc.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
```diff
@@ -0,0 +1,269 @@
...
+    public void shutdown() {
+        if (monitorThread != null) {
+            monitorThread.interrupt();
+        }
+    }
+
+    /**
+     * Loads the allowed list
+     *
+     * @throws IOException
```
awoods: Please complete javadoc.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
```diff
@@ -0,0 +1,269 @@
...
+     * @param configPath
+     */
+    public void setConfigPath(final String configPath) {
+        this.configPath = configPath;
+    }
+
+    /**
+     * Get the file path for the configuration file
+     *
+     * @return
```
awoods: Please complete javadoc.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
```diff
@@ -0,0 +1,269 @@
...
+     */
+    public String getConfigPath() {
+        return configPath;
+    }
+
+    /**
+     * Set whether to monitor the configuration file for changes
+     *
+     * @param monitorForChanges
```
awoods: Please complete javadoc.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
```diff
@@ -0,0 +1,269 @@
...
+            throw new ExternalMessageBodyException("Path was not a valid URI: " + extPath);
+        }
+        if (!uri.isAbsolute()) {
+            throw new ExternalMessageBodyException("Path was not absolute: " + extPath);
+        }
+        if (!ALLOWED_SCHEMES.contains(uri.getScheme())) {
+            throw new ExternalMessageBodyException("Path did not provide an accept scheme: " + extPath);
+        }
+
+        if (allowedList.stream().anyMatch(allowed -> path.startsWith(allowed))) {
```
awoods: `path.startsWith()` may be too permissive.
With the following rule:
```
file:///tmp/awoods/test
```
The following external content can be created:
```
file:///tmp/awoods/testsuite-execution-report.html
```

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
```diff
@@ -0,0 +1,269 @@
...
+    public void setConfigPath(final String configPath) {
+        this.configPath = configPath;
+    }
+
+    /**
+     * Get the file path for the configuration file
+     *
+     * @return
+     */
+    public String getConfigPath() {
```
awoods: This method can be removed... never used.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
bbpennel: Is this an issue? when setting up the configuration you could put `file:///tmp/awoods/test/` to avoid that. Using a simple starts with allows for specific files to be allowed without extra characters or behaviors.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
```diff
@@ -0,0 +1,269 @@
...
+        try {
+            uri = new URI(path);
+        } catch (final URISyntaxException e) {
+            throw new ExternalMessageBodyException("Path was not a valid URI: " + extPath);
+        }
+        if (!uri.isAbsolute()) {
+            throw new ExternalMessageBodyException("Path was not absolute: " + extPath);
+        }
+        if (!ALLOWED_SCHEMES.contains(uri.getScheme())) {
+            throw new ExternalMessageBodyException("Path did not provide an accept scheme: " + extPath);
```
awoods: typo?
`an accept scheme` -> `an allowed scheme`?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/spring/DefaultPropertiesLoader.java
```diff
@@ -54,7 +54,8 @@
         OBJECT_STORE("fcrepo.object.directory"),
         BINARY_STORE("fcrepo.binary.directory"),
         MODE_INDEX("fcrepo.modeshape.index.directory"),
-        ACTIVE_MQ("fcrepo.activemq.directory");
+        ACTIVE_MQ("fcrepo.activemq.directory"),
+        ALLOWED_EXTERNAL_CONTENT("fcrepo.external.content.allowed");
```
awoods: When this property is specified with JAVA_OPTS to point to a local file, the Tomcat start-up print statement is correct. However, when no system property is specified, the property name is printed (`fcrepo.external.content.allowed`) instead of the expected default location.

It appears that this class (`DefaultPropertiesLoader.java`) would need to be refactored to accommodate the new scenario presented by "external content allowed". If you would prefer the refactoring take place within its own ticket/PR, I am ok moving this through as-is.
As a note, such a separate ticket would likely be similar to a suggestion @whikloj recently made regarding consolidation of system properties.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
```diff
@@ -0,0 +1,273 @@
...
+    /**
+     * Validates that an external path is valid. The path must be an HTTP or file URI within the allow list of paths,
+     * be absolute, and contain no relative modifier.
+     *
+     * @param extPath external binary path to validate
+     * @throws ExternalMessageBodyException thrown if the path is invalid.
+     */
+    public void validate(final String extPath) throws ExternalMessageBodyException {
+        if (allowedList == null || allowedList.size() == 0) {
+            throw new ExternalMessageBodyException("External content is disallowed by the server");
```
bseeger: This functions fine, but reminds me that the constrained by file needs to be updated to be clear that it may have failed because the path/domain used was not in the 'safe' file.

fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
```diff
@@ -797,11 +805,14 @@ public void testGetWithExternalMessageMissingURLBinary() throws Exception {
                 .toString()
                 .replaceAll("<.*>", "< >");
 
-        final Response actual = testObj.createObject(null, null, null, null, Arrays.asList(badExternal), null);
+        testObj.createObject(null, null, null, null, Arrays.asList(badExternal), null);
```
bseeger: This is a side change.  I don't mind this so much, but you didn't do it for line 818 below.  (and I'm not sure if there are other spots).  I'm not clear on what's acceptable in a PR when it's outside the scope of the issue - it's a question I've had in general.  I'm a fan of cleaning code up, but I'm not sure if this should be reverted or if you should fix all the spots (and it might just be line 818, I'm not sure.).

fcrepo-integration-rdf/src/test/resources/spring-test/rest.xml
```diff
@@ -19,5 +19,16 @@
   <context:annotation-config/>
 
   <context:component-scan base-package="org.fcrepo"/>
+  
+  <!-- External content configuration -->
+    <bean name="externalContentPathValidator" class="org.fcrepo.http.api.ExternalContentPathValidator"
```
bseeger: Indent is wrong.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentHandlerFactory.java
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            return new ExternalContentHandler(link);
+        }
+
+        return null;
+    }
+
+    private static String getUriString(final String link) {
+        final String value = link.trim();
```
bseeger: is there a reason to not use Link here? sort of like-so: https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentHandler.java#L112

fcrepo-http-api/src/test/java/org/fcrepo/http/api/ExternalContentPathValidatorTest.java
```diff
@@ -0,0 +1,315 @@
...
+        addAllowedPath(allowExactPath);
+
+        validator.validate(extPath);
+    }
+
+    /*
+     * Test ignored because it takes around 10+ seconds to poll for events on MacOS:
+     * https://bugs.openjdk.java.net/browse/JDK-7133447 Can be enabled for one off testing
+     */
+    @Ignore("Test is ignored due to file event timing")
```
bseeger: Do we want to make a note (JIRA) to come back to this?  Or... what are the plans for this?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
bseeger: Or something to that affect.

fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
bbpennel: I was taking the approach of fixing warnings in methods that I updated for reasons related to the PR, but I was inconsistent and should have updated line 818 too in that case.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentHandlerFactory.java
bbpennel: Yes, I had to do this processing before it went through the `Link` class because that class automatically evaluates path modifiers. So if I get a url of `file:///path/to/stuff/../file.txt` and put it through the `Link` class and call .getURL on it, the value comes back as `file:///path/to/file.txt`. This is the reason I had to change when validation happened from operating on the ExternalContentHandler to happening before the handler.

fcrepo-http-api/src/test/java/org/fcrepo/http/api/ExternalContentPathValidatorTest.java
bbpennel: Its just ignored because the test is kind of long. We could turn it on and it should work, just wasn't sure if it was worth 10-20s in every build (at least for people on macs, its probably faster in other environments). Seemed worth keeping around since its the only test of the reloading mechanism rather than deleting before making my PR

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/spring/DefaultPropertiesLoader.java
bbpennel: Hm yeah that is pretty weird looking and misleading given that there actually is not a default value. This seems like a potentially simple refactor, just add a "setDefaultValue" property for each enum value and have it false for this one case. I could do that refactor in this PR if that's all it is. But I don't recall @whikloj 's suggested changes for consolidating system properties, so that may complicate things.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentHandlerFactory.java
bseeger: Ah, okay.  Thanks for the explanation. :)

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
bseeger: I'm wondering if we said something more about the allowed path.    How would a user of the system learn what the allowed path was?  That's a larger question then this PR, really.  Should we expose that information somewhere some folks using the system know what it is, or should we point them to the system administrator for more information about allowed paths?

fcrepo-http-api/src/test/java/org/fcrepo/http/api/ExternalContentPathValidatorTest.java
bseeger: Ah, okay. I read it as not working, but I totally see what you mean now.   My bad.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
bbpennel: There could be an endpoint for retrieving the list, but my inclination at this point would be to leave it up to the administrators to expose the info in whatever way they choose to. Don't need to have fedora giving out info about how your filesystem is arranged

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/spring/DefaultPropertiesLoader.java
bbpennel: @awoods I have updated the behavior of `DefaultPropertiesLoader` to not set a default value for `fcrepo.external.content.allowed` and to not log the property when it is not set. Let me know if you want to keep this change or not.

fcrepo-webapp/src/main/webapp/static/constraints/ExternalMessageBodyException.rdf
```diff
@@ -17,10 +17,10 @@
     <owl:Ontology rdf:about="static/constraints/ExternalMessageBodyException">
         <rdfs:label xml:lang="en">External Message Body Constraint</rdfs:label>
         <rdfs:comment xml:lang="en">Part of the request for creating/updating an External Message Body was missing
-            or malformed. Link header should include a URI reference to the external resource, a link relation
-            type ('rel') that's set to "http://fedora.info/definitions/fcrepo#ExternalContent";',
-            a 'handling' attribute set to one of these three things: proxy, redirect, or copy, and an optional
-            attribute of 'type' representing the content type of the external body.
+            or malformed. Link header must include a URI reference to an external resource from an allowed path,
+            a link relation type ('rel') that is set to "http://fedora.info/definitions/fcrepo#ExternalContent";',
```
bseeger: I think you can take out the `;'` at the end of the line.  Looks like a typo that was there before you updated this code.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/spring/DefaultPropertiesLoader.java
```diff
@@ -90,16 +102,21 @@ public void loadSystemProperties() {
...
                 }
             }
         }
 
         for (final PROPERTIES prop : PROPERTIES.values()) {
-            final String val = prop.getValue();
-            LOGGER.info("{} = {}", val, getProperty(val));
+            final String name = prop.getValue();
+            final String val = getProperty(name);
+            if (val != null) {
```
awoods: I would suggest that it is actually helpful to print the `what-ever-property = null` in the case that a property has not been set. It may help show an administrator that a property was not set... when they thought it was.

bbpennel: Okay, it should now display
`INFO 08:57:27.709 (DefaultPropertiesLoader) fcrepo.external.content.allowed = null`

PR #1393, opened 2018-08-31
Allow OPTIONS when there is an acl:Read permission.

peichman-umd:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2847

Allow OPTIONS when there is an acl:Read permission.

# What's new?
Permit OPTIONS requests when the current user has acl:Read mode permissions to the requested resource.

# How should this be tested?

Includes unit and integration tests, so the standard `mvn clean test` should test.

# Interested parties
@awoods, @fcrepo4/committers


whikloj:
This is fine, but as I discovered and then discussed with @birkland in Slack. Forcing a user to include the `acl:accessTo` triple when there is only one acceptable object for that triple seems painful. If that triple doesn't exist in the submitted document we should assume they mean the resource the ACL is hanging off of. No?

PR #1394, opened 2018-09-04
Implemented support for acl:default

lsitu:
Implement support with acl:default for parent ACL inheritance.

**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2698

Apply parent/ancestor ACLs to a resource only if it contains an acl:default statement that is used for inheritance as defined in https://github.com/solid/web-access-control-spec#acl-inheritance-algorithm.

# What's new?
Those parent ACLs with no acl:default statement will be ignored for child resource authorization.

# Interested parties


dbernstein:
@lsitu :  Good thinking on the new integration test.

PR #1408, opened 2018-09-14
Ensure ldp:RDFSource type is on LDP-RSs

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2862

Adds the `http://www.w3.org/ns/ldp#RDFSource` Link header with `rel="type"` to all LDP-RSs. 

# How should this be tested?

Put a LDP-RS and when you GET or HEAD it see the new type, same with TimeMaps and Binary descriptions.

# Additional Notes:

Example:

# Interested parties
@fcrepo4/committers


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -577,7 +578,7 @@ protected void addLinkAndOptionsHttpHeaders(final FedoraResource resource) {
      * @param relation the relation string.
      * @return the string version of the link.
      */
-    private static String buildLink(final String linkUri, final String relation) {
+    public static final String buildLink(final String linkUri, final String relation) {
```
awoods: With the intention of only opening access as much as necessary, this can be `protected` instead of `public`.

awoods: In general, making methods `final` creates less flexible code. Additionally, in this case, `static` removes any functional effects of being `final`.
I would suggest not making this method `final`.

PR #1410, opened 2018-09-17
Referencing acl:group lists using Hash URIs now works properly.

dbernstein:
**ACLs that reference group list resource using a hash URI  in acl:agentGroup triples was not authorizing correctly.  This PR addresses the bug.**
* * *

**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2864

Now it is possible to use hash URIs as the object of an acl:agentGroup predicate.  Two integration tests were added to confirm that the functionality works. The first test ensures that members defined in hash-URI blocks are respected while acls referencing an unauthorized block within the same resource results in a 403.  The second test ensures that non-hash URIs work as expected.

# How should this be tested?
Integration tests cover the new functionality and provides missing testing for existing functionality.  

# Additional Notes:
* Does this change require documentation to be updated? No
* Could this change impact execution of existing code? Possibly, but unlikely.

# Interested parties

### DIFF
--- a/fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACRolesProvider.java
+++ b/fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACRolesProvider.java
@@ -27,7 +27,7 @@ import static java.util.stream.Stream.empty;
 import static java.util.stream.Stream.of;
 import static org.apache.jena.graph.NodeFactory.createURI;
 import static org.fcrepo.auth.webac.URIConstants.FOAF_AGENT_VALUE;
-import static org.fcrepo.auth.webac.URIConstants.VCARD_GROUP;
+import static org.fcrepo.auth.webac.URIConstants.VCARD_GROUP_VALUE;
 import static org.fcrepo.auth.webac.URIConstants.VCARD_MEMBER_VALUE;
 import static org.fcrepo.auth.webac.URIConstants.WEBAC_ACCESSTO_CLASS_VALUE;
 import static org.fcrepo.auth.webac.URIConstants.WEBAC_ACCESSTO_VALUE;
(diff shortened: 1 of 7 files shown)


fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACRolesProvider.java
```diff
@@ -313,12 +321,26 @@ private FedoraResource getBaseVersion(final FedoraResource resource) {
...
-            final FedoraResource resource) {
-        return resource.getTriples(translator, PROPERTIES).filter(memberTestFromTypes.apply(resource.getTypes()))
-            .map(Triple::getObject).flatMap(WebACRolesProvider::nodeToStringStream);
+                                                  final FedoraResource resource, final String hashPortion) {
+
+        //resolve list of triples, accounting for hash-uris.
+        final List<Triple> triples = resource.getTriples(translator, PROPERTIES).filter(
+            triple -> hashPortion == null || triple.getSubject().getURI().endsWith(hashPortion)).collect(toList());
+        //determine if there is a rdf:type vcard:Group
+        final boolean hasVcardGroup = triples.stream().filter(
```
bbpennel: I would suggest using `anyMatch` here instead of `count`. Also, just a suggestion, but it may be more readable to use triple.matches:
```
final boolean hasVcardGroup = triples.stream().anyMatch(
                triple -> triple.matches(triple.getSubject(), RDF_TYPE_NODE, VCARD_GROUP_NODE));
```

fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACRolesProvider.java
```diff
@@ -313,12 +321,26 @@ private FedoraResource getBaseVersion(final FedoraResource resource) {
...
+            triple -> triple.predicateMatches(RDF_TYPE_NODE) &&
+                      triple.getObject().getURI().equals(VCARD_GROUP_VALUE)).count() > 0;
```
bbpennel: It is probably worthwhile to create and use constant nodes for VCARD_GROUP_VALUE and VCARD_MEMEBER_VALUE, like you did for rdf:type:
```
    private static final org.apache.jena.graph.Node VCARD_GROUP_NODE = createURI(VCARD_GROUP_VALUE);
    private static final org.apache.jena.graph.Node VCARD_MEMBER_NODE = createURI(VCARD_MEMBER_VALUE);
```

PR #1420, opened 2018-09-24
Triples in mementos not the same as in original resource

bseeger:
There were extra / missing triples in the memento. This fixes that discrepancy between the original and it's newly created memento.
* * *

**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2848

This makes sure the memento triples are the same as the original triples (for a newly created memento).
# What's new?
* FEDORA_CONTAINER is now added explicitly to the memento.  
* When the memento is requested from the server, it doesn't deliver the server managed triples separately, as they are already part of the PROPERTY triples.  If you get the PROPERTY triples and SERVER_MANAGED triples, you'll get duplication on the server managed properties. 

# How should this be tested?
As per in the JIRA ticket: 

// create a resource to version
$> curl -X POST -H "Link: <http://mementoweb.org/ns#OriginalResource>; rel=\"type\"" http://localhost:8080/fcrepo4/rest/ -H "slug: test1"

// version it
$> curl -X POST http://localhost:8080/fcrepo4/rest/test1/fcr:versions

which creates a memento – fetch memento information by querying the fcr:versions endpoint.

Then get the Original Resource and the memento and compare them: 
$> curl http://localhost:8080/fcrepo4/rest/test1
$> curl http://localhost:8080/fcrepo4/rest/test1/fcr:versions/UNIQUEID

The triples should be the same. 

# Interested parties
@dbernstein @bbpennel


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -318,17 +318,22 @@ private RdfStream getResourceTriples(final int limit, final FedoraResource resou
...
         final List<Stream<Triple>> streams = new ArrayList<>();
 
+
         if (returnPreference.getValue().equals("minimal")) {
             streams.add(getTriples(resource, of(PROPERTIES, MINIMAL)).filter(tripleFilter));
 
-            if (ldpPreferences.prefersServerManaged()) {
+            // Mementos already have the server managed properties in the PROPERTIES category
+            // since mementos are immutable and these triples are no longer managed
+            if (!resource.isMemento() && ldpPreferences.prefersServerManaged()) {
```
bbpennel: might make sense to swap the order of these checks to prioritize the condition that doesn't hit the database, but its a minor thing

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
```diff
@@ -594,6 +594,18 @@ private static int countLinkHeader(final HttpResponse response, final String uri
         }).count();
     }
 
+    protected static String getOriginalResourceUri(final CloseableHttpResponse response) {
+        final List<String> linkList = getLinkHeaders(response).stream().filter(x -> {
```
bbpennel: Might as well just create the Link once:
```
.map(x -> Link.valueOf(x))
.filter(x -> x.getRel().equals("original"))
.collect(toList())
```

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
```diff
@@ -594,6 +594,18 @@ private static int countLinkHeader(final HttpResponse response, final String uri
...
+            final Link theLink = Link.valueOf(x);
+            return theLink.getRel().equals("original");
+        }).map(x -> Link.valueOf(x).getUri().toString()).collect(toList());
+
+        assertTrue("No original resource Link in headers", linkList.size() != 0);
```
bbpennel: I think it'd be okay to just use the second assert (equals 1), and also switch to assertEquals. Or skip the assertions and have the stream block return findAny without validating the number of headers since this is a test helper, up to you.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
```diff
@@ -594,6 +594,15 @@ private static int countLinkHeader(final HttpResponse response, final String uri
         }).count();
     }
 
+    protected static String getOriginalResourceUri(final CloseableHttpResponse response) {
+        final List<Link> linkList = getLinkHeaders(response).stream()
+            .map(x -> Link.valueOf(x))
+            .filter(x -> x.getRel().equals("original"))
+            .collect(toList());
```
bbpennel: since the assertions were removed, you could skip going to a list
```
return getLinkHeaders(response).stream()
            .map(x -> Link.valueOf(x))
            .filter(x -> x.getRel().equals("original"))
            .findFirst().get();
```

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -318,17 +318,22 @@ private RdfStream getResourceTriples(final int limit, final FedoraResource resou
 
         final List<Stream<Triple>> streams = new ArrayList<>();
 
+
         if (returnPreference.getValue().equals("minimal")) {
             streams.add(getTriples(resource, of(PROPERTIES, MINIMAL)).filter(tripleFilter));
 
-            if (ldpPreferences.prefersServerManaged()) {
+            // Mementos already have the server managed properties in the PROPERTIES category
+            // since mementos are immutable and these triples are no longer managed
```
whikloj: This probably has more to do with the Memento creation than return. But doesn't this mean you can't get a Memento without server managed triples?
I did a test and a normal (no options) request for the OriginalResource vs a Memento match up perfectly. 

But if you add a `Prefer: return=representation; omit="http://fedora.info/definitions/v4/repository#ServerManaged"` then the Memento has 3 triples while the OriginalResource has 1.

original_vs_memento

bseeger: Yes, I think you're right and it alls about what context you're in. I think this turns into a meta conversation quickly.  The memento has the server managed triples from the original, now immutable, but does it also have it's own server managed triples?  Those original triples are no longer server managed (though they used to be).   It seems the decision was made that those original server managed triples are no longer server managed (makes sense, since it's a copy in time). 
Mostly I had to "say ^^ out loud" to process it myself.  
I think you've found a good twist on this...   why does the original only return one, when it seems like it should have those other two as well? Weird.

bseeger: after thinking about this some... it seems we have the issue of what server managed  triple is part of the original resource (which seems like being in PROPERTY is okay) and what server managed triples belong to the memento.  Ie, the memento should have it's own server managed triples, but we can't distinguish, in the returned object, which server managed triples belong to the memento itself and which belong to the copy of the resource the memento is for.  ... so, should the memento have it's own server managed triples?

bseeger: And how would we represent that?  Should the returned object have a subject for the original resource and then another set of triples where the subject is the  memento, which contain the server managed triples for the memento?

bseeger: @dbernstein @awoods    what do you guys think of the above questions?

bseeger: @whikloj - I'm actually confused as to why we're not getting those other triples in the original, since I think the only server managed properties are:  
        fedora:createdBy
        fedora:created
        fedora:lastModified  
        fedora:lastModifiedBy

whikloj: I'm sorry for opening this can of metaphysical worms. I'm not sure what is considered valid here.

The Memento with `omit` server managed triples gives this:
```
<http://localhost:8080/rest/versioned>
        rdf:type  ldp:RDFSource ;
        rdf:type  ldp:Container ;
        rdf:type  ldp:BasicContainer .
```

The same Memento with `include` server managed triples gives this:
```
<http://localhost:8080/rest/versioned>
        rdf:type               fedora:Container ;
        rdf:type               ldp:RDFSource ;
        rdf:type               fedora:Resource ;
        rdf:type               ldp:BasicContainer ;
        fedora:created         "2018-09-17T15:17:22.936Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> ;
        fedora:writable        true ;
        fedora:lastModified    "2018-09-17T15:17:22.936Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> ;
        fedora:lastModifiedBy  "fedoraAdmin" ;
        fedora:createdBy       "fedoraAdmin" .
```

For the OriginalResource, `omit`ting server managed triples:
```
<http://localhost:8080/rest/versioned>
        rdf:type  ldp:BasicContainer .
```

and `include`ing server managed triples:
```
<http://localhost:8080/rest/versioned>
        rdf:type               fedora:Container ;
        rdf:type               fedora:Resource ;
        rdf:type               ldp:BasicContainer ;
        fedora:lastModifiedBy  "fedoraAdmin" ;
        fedora:createdBy       "fedoraAdmin" ;
        fedora:created         "2018-09-17T15:17:22.936Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> ;
        fedora:lastModified    "2018-09-17T15:17:22.936Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> ;
        rdf:type               ldp:RDFSource ;
        fedora:writable        true .
```

So maybe the problem is that the OriginalResource is considering these as server managed triples?
```
<>    rdf:type               ldp:RDFSource ;
      rdf:type               ldp:Container .
```

dbernstein: @whikloj :  thanks for the clarification of the current behavior.   I've put the topic on the agenda for tomorrow's call.

PR #1421, opened 2018-09-28
Enable versioning, creation of Mementos and view Mementos in HTML UI

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2730

Make the HTML UI work (mostly) with versions.

# What's new?
* Checks if an item is:
    * the repository root or 
    * a Memento or 
    * is Versionable (has /fcr:versions container) or
    *  is an a LDP-NR or its description.

It then changes the HTML interface based on the above to

* Allows you to enable versioning on LDP-RSes (but not LDP-NRs or LDP-RSs acting as descriptions)
* Allows you to make a version.
* Allows you to view the versions (but the triples appear under Other Resources, AND if you click the URL or dc:title instead of the plus sign you move away from the Memento).

# How should this be tested?

Build it and play with the HTML UI.

# Additional Notes:

Example:
* Does this change require documentation to be updated? Probably

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/main/resources/views/common-node-actions.vsl
+++ b/fcrepo-http-api/src/main/resources/views/common-node-actions.vsl
@@ -3,32 +3,32 @@
 #set ($isRoot = $helpers.isRootResource($rdf, $topic))
 
 #if ($isVersion == true)
-<div class="alert alert-warning">
+  <div class="alert alert-warning">
     <span class="glyphicon glyphicon-warning-sign"></span>
     This is a <strong>historic version</strong> and cannot be modified.
-</div>
-<form id="action_remove_version" method="DELETE" action="$uriInfo.getAbsolutePath().toString()" data-redirect-after-submit="$helpers.getVersionSubjectUrl($uriInfo, $topic)/fcr:versions" >
+  </div>
+  <form id="action_remove_version" method="DELETE" action="$uriInfo.getAbsolutePath().toString()" data-redirect-after-submit="$helpers.getVersionSubjectUrl($uriInfo, $topic)/fcr:versions" >
     <button type="submit" class="btn btn-danger standalone">Delete this Version</button>
-</form>
+  </form>
 #else
-#if ($helpers.isRdfResource($rdf, $content, "http://fedora.info/definitions/v4/repository#", "Binary"))
+  #if ($helpers.isRdfResource($rdf, $content, "http://fedora.info/definitions/v4/repository#", "Binary"))
     <a href="$content" class="btn btn-success btn-lg"><span class="glyphicon glyphicon-download"></span> Download Content</a><br/><br/>
 	<a href="$content/fcr:fixity" class="btn btn-success btn-lg">Fixity</a>
     #if ($writable == true)
-    <h3>Update Content</h3>
-    <form id="action_update_file">
-      <div class="form-group">
-        <label for="update_file" class="control-label">File</label>
-        <input type="file" id="update_file"/>
-      </div>
-      <input type="submit" id="binary_update_content" class="btn btn-primary" value="Update">
-    </form>
+      <h3>Update Content</h3>
+      <form id="action_update_file">
+        <div class="form-group">
+          <label for="update_file" class="control-label">File</label>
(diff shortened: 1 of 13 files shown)


fcrepo-http-api/src/main/resources/views/common-node-actions.vsl
```diff
@@ -86,20 +83,32 @@ WHERE { }
...
+    <hr />
+</form>
+#end
 
 <form id="action_delete" name="action_delete">
     <h3>Delete Resource</h3>
     <button name="delete-button" type="submit" class="btn btn-danger">Delete</button>
     <hr />
 </form>
+	#end
```
bseeger: is this indent correct?

whikloj: I think so, but I didn't indent the actual HTML just the Velocity template tags. I can move them around if it helps, the `#end` on line 104 should be indented as well.

bseeger:
Jared - now that $isVersion is a thing in the .vsl files, how hard would it be to change "Other Resource" to something like "Memento Resource" in the display?

fcrepo-http-api/src/main/resources/views/common-node-actions.vsl
bseeger: I'm not familiar with Velocity at all, so whatever makes the most sense.

bseeger:
@whikloj  - when I run this in tomcat 8.5.32, I get this error when I try to enable versioning on a resource by clicking the "Enable Versioning" button:  "RDF was not parsable: [line: 1, col: 2 ] Unexpected character ('<' (code 60)): expected a valid value (number, String, array, object, 'true', 'false' or 'null')"

whikloj:
@bseeger 
1. There is a bunch of stuff that could be done, some might be to display a different template. Currently we are using the same one for all resources, but it (I think) is possible to replace the resource.vsl with a memento.vsl.
1. Does that resource have any children? If it does then it falls in to the range of https://jira.duraspace.org/browse/FCREPO-2879

bseeger:
In response to 2) above, no, the resource doesn't have any children.  
`javax.ws.rs.BadRequestException: RDF was not parsable: [line: 1, col: 2 ] Unexpected character ('<' (code │
DE│60)): expected a valid value (number, String, array, object, 'true', 'false' or 'null')                   │
DE│    at org.fcrepo.http.api.ContentExposingResource.parseBodyAsModel(ContentExposingResource.java:963)     │
DE│    at org.fcrepo.http.api.ContentExposingResource.replaceResourceWithStream(ContentExposingResource.java:│
DE│935)                                                                                                      │
DE│    at org.fcrepo.http.api.FedoraLdp.createOrReplaceObjectRdf(FedoraLdp.java:478)            `

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
```diff
@@ -146,15 +150,30 @@ public String getVersionSubjectUrl(final UriInfo uriInfo, final Node subject) {
...
      *
      * @param graph the graph
      * @param subject the subject
      * @return the modification date if it exists
      */
-    public Optional<String> getVersionDate(final Graph graph, final Node subject) {
-        return getValue(graph, subject, CREATED_DATE.asNode());
+    public Instant getVersionDate(final Graph graph, final Node subject) {
+        final String[] pathParts = subject.getURI().split("/");
+        final Instant datetime = DateTimeFormatter.ofPattern("yyyyMMddHHmmss").withZone(ZoneId.of("UTC"))
```
bseeger: I wonder if we should make this pattern ("yyyyMMddHHmmss") a static variable for reuse, since it represents part of a MEMENTO's id and it's used in a few spots?     (fcrepo-kernel-modeshape -> VersionServiceImpl.java and some tests)

whikloj: Sure I can do that.

PR #1423, opened 2018-10-02
Change log level from WARN to TRACE

whikloj:
**JIRA Ticket**: No ticket

Reduces log level to remove some messages 

# What's new?

# How should this be tested?

Build and see much fewer
```
WARN 20:33:03.937 (FedoraSessionUserUtil) Default URI is created for user bypassAdmin: info:fedora/local-user#bypassAdmin
```
messages

# Additional Notes:

Example:

# Interested parties
@awoods @bbpennel @dbernstein

### DIFF
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/FedoraSessionUserUtil.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/FedoraSessionUserUtil.java
@@ -58,7 +58,7 @@ public class FedoraSessionUserUtil {
             } else {
                 return buildDefaultURI(userId);
             }
-        } catch (IllegalArgumentException e) {
+        } catch (final IllegalArgumentException e) {
             return buildDefaultURI(userId);
         }
     }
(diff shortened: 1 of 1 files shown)


birkland:
Just FYI, that warning was at one point intentional, as a way of raising a
constant warning not to use strings as a form of identity now that the spec
aligns with URIs.  That being said, maybe continuing to use strings is a
defacto consensus, and the spec should call that put as explicitly OK?

whikloj:
If that is the case maybe we should be preloading a fedoraAdmin resource to use?

birkland:
Not necessarily pre-loading, but expecting URIs for agents rather than strings.  

I could have sworn the spec once had a section that clarified that the Fedora spec does not require WebIDs as mentioned in the SOLID spec, but does require URIs with no particular expectation they are resolvable.  

The end result was punting: keeping support for string agents by way of implicitly constructing a URI by appending a prefix, but emitting a warning.  It was unclear at the time whether that was appropriate or not.  So if it _is_ appropriate, then removing the warning is fine.  It was almost a year ago, so perhaps thinking on that has advanced.

whikloj:
What you say @birkland appears correct. The SOLID spec states (though there is no `MUST`) that we should be using the WebID URI. I see nothing in the Fedora Spec to overrride this. But we are generating a URI in this code, is the problem with the URI that it is not resolvable?

birkland:
Fond it!:  https://github.com/fcrepo/fcrepo-specification/issues/403

To be honest, I believe the SOLID specs are imprecisely written.  It suggests that agent URIs are WebIDs, but I think the requirement that agent URIs `MUST` be resolvable WebIDs (if that is what SOLID is saying - I simply can't parse it) is not practical.   It would have been helpful for the spec editors to clarify this for the sake of Fedora implementations.

PR #1424, opened 2018-10-02
Ensure that all server managed triples are suppressed on omit.

mohideen:
**JIRA Ticket**: 
- https://jira.duraspace.org/browse/FCREPO-2877
- https://jira.duraspace.org/browse/FCREPO-2880

Ensure that all server managed triples are suppressed in RDF body when omitting Server Managed Triples

# How should this be tested?
1. Verify that all the relevant SMTs are being tested for in FedoraLdpIT
- testGetObjectOmitMembership
- testGetObjectOmitContainment
- testGetObjectOmitServerManaged
2. Run the test to verify all the tests pass.

## Manual Test
- Follow steps to reproduce the issue as per the description in FCREPO-2877. 
- Verify that the `ldp:BasicContainer` is no longer being returned.

# Interested parties
@dbernstein, @fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -77,7 +77,6 @@ import static org.fcrepo.kernel.api.RdfLexicon.VERSIONED_RESOURCE;
 import static org.fcrepo.kernel.api.RdfLexicon.VERSIONING_TIMEGATE_TYPE;
 import static org.fcrepo.kernel.api.RdfLexicon.VERSIONING_TIMEMAP_TYPE;
 import static org.fcrepo.kernel.api.RdfLexicon.WEBAC_NAMESPACE_VALUE;
-import static org.fcrepo.kernel.api.RdfLexicon.isManagedLDPType;
 import static org.fcrepo.kernel.api.RdfLexicon.isManagedNamespace;
 import static org.fcrepo.kernel.api.RdfLexicon.isManagedPredicate;
 import static org.fcrepo.kernel.api.RequiredRdfContext.EMBED_RESOURCES;
@@ -230,8 +229,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
     protected ExternalContentHandlerFactory extContentHandlerFactory;
 
     private static final Predicate<Triple> IS_MANAGED_TYPE = t -> t.getPredicate().equals(type.asNode()) &&
-            (isManagedNamespace.test(t.getObject().getNameSpace()) ||
-            isManagedLDPType.test(createResource(t.getObject().getURI())));
+            (isManagedNamespace.test(t.getObject().getNameSpace()));
     private static final Predicate<Triple> IS_MANAGED_TRIPLE = IS_MANAGED_TYPE
         .or(t -> isManagedPredicate.test(createProperty(t.getPredicate().getURI())));
 
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -229,7 +229,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
     protected ExternalContentHandlerFactory extContentHandlerFactory;
 
     private static final Predicate<Triple> IS_MANAGED_TYPE = t -> t.getPredicate().equals(type.asNode()) &&
-            (isManagedNamespace.test(t.getObject().getNameSpace()));
+            isManagedNamespace.test(t.getObject().getNameSpace());
     private static final Predicate<Triple> IS_MANAGED_TRIPLE = IS_MANAGED_TYPE
         .or(t -> isManagedPredicate.test(createProperty(t.getPredicate().getURI())));
 
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -77,6 +77,7 @@ import static org.fcrepo.kernel.api.RdfLexicon.VERSIONED_RESOURCE;
 import static org.fcrepo.kernel.api.RdfLexicon.VERSIONING_TIMEGATE_TYPE;
 import static org.fcrepo.kernel.api.RdfLexicon.VERSIONING_TIMEMAP_TYPE;
 import static org.fcrepo.kernel.api.RdfLexicon.WEBAC_NAMESPACE_VALUE;
+import static org.fcrepo.kernel.api.RdfLexicon.isManagedLDPType;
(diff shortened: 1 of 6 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -229,7 +230,8 @@
     protected ExternalContentHandlerFactory extContentHandlerFactory;
 
     private static final Predicate<Triple> IS_MANAGED_TYPE = t -> t.getPredicate().equals(type.asNode()) &&
-            isManagedNamespace.test(t.getObject().getNameSpace());
+            (isManagedNamespace.test(t.getObject().getNameSpace()) ||
+            isManagedLDPType.test(createResource(t.getObject().getURI())));
```
whikloj: Just wondering if this should be removed and replaced with the `isManagedPredicate` from `RdfLexicon` which contains other stuff as well?
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java#L280-L281

whikloj: Also we shouldn't be suppressing `<> rdf:type ldp:BasicContainer` when we suppress SMTs. This is not an SMT, I think

awoods: The "Are LDP Container types SMTs" question should be resolved on the 2018-10-04 Tech call.

As for the `isManagedPredicate`, the test here is:
- is the predicate `rdf:type`, and
- is the object in a managed namespace, or
- is the object a managed LDP type

@whikloj : are you suggesting the possibility of changing the logic to:
- is the predicate `rdf:type`, and
- is the predicate managed?

whikloj: I am just unclear as to what is an SMT and that makes testing for them very difficult. I think I should just wait until that is decided then I can have an opinion.

whikloj: So following the discussion on the tech call my opinion is that the existing `isManagedNamespace` predicate could be expand to include both `fedora:*` and `ldp:*` and then this change to add `isManagedType` would be unnecessary. Was that your understand as well @mohideen @awoods?

awoods: @whikloj : That sounds right based on today's call.

mohideen: @whikloj @awoods So, we won't be able to create resources with `ldp:*` rdf:type?

```
$ curl -i -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/testputldp -X PUT -H "Content-Type: text/turtle" --data-binary '<> a <http://www.w3.org/ns/ldp#DirectContainer> .'
HTTP/1.1 409 Conflict
Date: Thu, 04 Oct 2018 23:44:06 GMT
Set-Cookie: JSESSIONID=186v9jsw7ckum1qnzffgc3lp6s;Path=/
Set-Cookie: rememberMe=deleteMe; Path=/; Max-Age=0; Expires=Wed, 03-Oct-2018 23:44:06 GMT
Link: <http://localhost:8080/static/constraints/ServerManagedTypeException.rdf>; rel="http://www.w3.org/ns/ldp#constrainedBy"
Content-Length: 98
Server: Jetty(9.3.1.v20150714)

The repository type (http://www.w3.org/ns/ldp#DirectContainer) of this resource is system managed.
```

Did we say we will use `Link` headers to overcome this?

mohideen: Refactored based on today's decisions.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -2142,11 +2267,11 @@ public void testPutToChangeInteractionModelWithRdf() throws IOException {
         assertTrue(getLinkHeaders(getObjMethod(pid + "/b")).contains(DIRECT_CONTAINER_LINK_HEADER));
 
         // successful update the properties with the interaction mode
-        final String ttla = "<> a <" + DIRECT_CONTAINER.getURI()
-                + "> ; <" + MEMBERSHIP_RESOURCE + "> <" + container + ">;\n"
+        final String ttla = "<> <" + MEMBERSHIP_RESOURCE + "> <" + container + ">;\n"
                 + "<" + HAS_MEMBER_RELATION + "> <info:some/relation> .\n";
         final HttpPut puta = putObjMethod(pid + "/b", "text/turtle", ttla);
         puta.addHeader("Prefer", "handling=lenient; received=\"minimal\"");
+        puta.addHeader(LINK, "<http://www.w3.org/ns/ldp#DirectContainer>;rel=type");
```
awoods: Maybe replace this String with the constant: `DIRECT_CONTAINER_LINK_HEADER`

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -2256,8 +2381,8 @@ public void testGetObjectReferencesIndirect() throws Exception {
         // create an indirect container
         final HttpPut createContainer = new HttpPut(serverAddress + pid1 + "/members");
         createContainer.addHeader(CONTENT_TYPE, "text/turtle");
-        final String membersRDF = "<> a <http://www.w3.org/ns/ldp#IndirectContainer>; "
-            + "<http://www.w3.org/ns/ldp#hasMemberRelation> <" + memberRelation + ">; "
+        createContainer.addHeader(LINK, "<http://www.w3.org/ns/ldp#IndirectContainer>;rel=type");
```
awoods: Maybe replace this String with the constant: `INDIRECT_CONTAINER_LINK_HEADER`

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -3640,8 +3765,8 @@ public void testContainerLastModified() throws Exception {
         final long lastmod3;
         final HttpPut createContainer = new HttpPut(objURI + "/members");
         createContainer.addHeader(CONTENT_TYPE, "text/turtle");
-        final String membersRDF = "<> a <http://www.w3.org/ns/ldp#DirectContainer>; "
-            + "<http://www.w3.org/ns/ldp#hasMemberRelation> <http://pcdm.org/models#hasMember>; "
+        createContainer.addHeader(LINK, "<http://www.w3.org/ns/ldp#DirectContainer>;rel=type");
```
awoods: Maybe replace this String with the constant: `DIRECT_CONTAINER_LINK_HEADER`

PR #1432, opened 2018-10-05
Make all LDP-Rs versionable by default

dbernstein:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2890

 https://jira.duraspace.org/browse/FCREPO-2889

This PR makes all LDP-Rs versionable by default.   Now the  `Link: <http://mementoweb.org/ns#OriginalResource>; rel="type"`  header is simply ignored on creation of PUTs and POSTs.  The related integration tests were removed or modified.

# How should this be tested?
Create  resources without the versioning link header and verify that all resources are versionable regardless of whether the versioning header is present.
```
# rdf source
curl -i -u fedoraAdmin:fedoraAdmin -X POST http://localhost:8080/rest -H "Slug: test1-rdf"; 
curl -i -u fedoraAdmin:fedoraAdmin -X GET http://localhost:8080/rest/test1-rdf
# non rdf 
curl -i -u fedoraAdmin:fedoraAdmin -X POST http://localhost:8080/rest -H "Slug: test1-nr" -H "Content-Type: text/plain"; 
curl -i -u fedoraAdmin:fedoraAdmin -X GET http://localhost:8080/rest/test1-nr
# rdf source with header
curl -i -u fedoraAdmin:fedoraAdmin -X POST http://localhost:8080/rest -H "Slug: test2-rdf" -H "Link: <http://mementoweb.org/ns#OriginalResource>; rel=\"type\"";  
curl -i -u fedoraAdmin:fedoraAdmin -X GET http://localhost:8080/rest/test2-rdf
# non rdf with header
curl -i -u fedoraAdmin:fedoraAdmin -X POST http://localhost:8080/rest -H "Slug: test2-nr" -H "Content-Type: text/plain"  -H "Link: <http://mementoweb.org/ns#OriginalResource>; rel=\"type\""; 
curl -i -u fedoraAdmin:fedoraAdmin -X GET http://localhost:8080/rest/test2-nr
```

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -290,7 +290,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
         // add vary headers
         final List<String> varyValues = new ArrayList<>(VARY_HEADERS);
 
-        if (resource.isVersioned()) {
+        if (resource.isOriginalResource()) {
             varyValues.add(ACCEPT_DATETIME);
         }
 
@@ -559,16 +559,16 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
                 buildLink(constraintURI, CONSTRAINED_BY.getURI()));
         }
 
-        final boolean isVersioned = resource.isVersioned();
+        final boolean isOriginal = resource.isOriginalResource();
         // Add versioning headers for versioned originals and mementos
-        if (isVersioned || resource.isMemento() || resource instanceof FedoraTimeMap) {
+        if (isOriginal || resource.isMemento() || resource instanceof FedoraTimeMap) {
             final URI originalUri = getUri(resource.getOriginalResource());
             final URI timemapUri = getUri(resource.getTimeMap());
             servletResponse.addHeader(LINK, buildLink(originalUri, "timegate"));
             servletResponse.addHeader(LINK, buildLink(originalUri, "original"));
             servletResponse.addHeader(LINK, buildLink(timemapUri, "timemap"));
 
-            if (isVersioned) {
+            if (isOriginal) {
                 servletResponse.addHeader(LINK, buildLink(VERSIONED_RESOURCE.getURI(), "type"));
                 servletResponse.addHeader(LINK, buildLink(VERSIONING_TIMEGATE_TYPE, "type"));
             } else if (resource instanceof FedoraTimeMap) {
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersioningIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersioningIT.java
@@ -130,7 +130,6 @@ import org.junit.rules.TemporaryFolder;
  */
 public class FedoraVersioningIT extends AbstractResourceIT {
 
-    private static final String VERSIONED_RESOURCE_LINK_HEADER = "<" + VERSIONED_RESOURCE.getURI() + ">; rel=\"type\"";
     private static final String BINARY_CONTENT = "binary content";
     private static final String BINARY_UPDATED = "updated content";
 
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersioningIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersioningIT.java
@@ -1041,38 +1041,6 @@ public class FedoraVersioningIT extends AbstractResourceIT {
         }
     }
 
-    @Test
-    public void testGetUnversionedObjectVersionProfile() throws Exception {
-        final String containerUri = serverAddress + id;
-
-        createObject(id);
-
-        final HttpGet getVersion = new HttpGet(containerUri + "/" + FCR_VERSIONS);
-        assertEquals(NOT_FOUND.getStatusCode(), getStatus(getVersion));
-
-        // Verify original has no memento headers
-        assertNoMementoHeaders(containerUri);
-    }
-
-    @Test
-    public void testGetUnversionedBinaryAndDescriptionVersionProfile() throws Exception {
-        createDatastream(id, "ds", "content");
-
-        final String binaryUri = serverAddress + id + "/ds";
-        final HttpGet getBinary = new HttpGet(binaryUri + "/" + FCR_VERSIONS);
-        assertEquals(NOT_FOUND.getStatusCode(), getStatus(getBinary));
-
-        // Verify original binary has no memento headers
-        assertNoMementoHeaders(binaryUri);
-
-        final String metadataUri = binaryUri + "/" + FCR_METADATA;
-        final HttpGet getDesc = new HttpGet(metadataUri + "/" + FCR_VERSIONS);
-        assertEquals(NOT_FOUND.getStatusCode(), getStatus(getDesc));
-
-        // Verify original has no memento headers
-        assertNoMementoHeaders(metadataUri);
-    }
-
     private void assertNoMementoHeaders(final String uri) throws Exception {
         final HttpGet getOriginal = new HttpGet(uri);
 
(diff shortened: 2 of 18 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersioningIT.java
```diff
@@ -1714,7 +1633,6 @@ private void createVersionedExternalBinaryMemento(final String rescId, final Str
         final HttpPut httpPut = putObjMethod(rescId);
         httpPut.addHeader(LINK, "<" + NON_RDF_SOURCE.getURI() + ">;rel=\"type\"");
         httpPut.addHeader(LINK, getExternalContentLinkHeader(externalUri, handling, null));
-        httpPut.addHeader(LINK, VERSIONED_RESOURCE_LINK_HEADER);
```
awoods: The declaration of the constant at the top of this class can now be removed. (`VERSIONED_RESOURCE_LINK_HEADER`)

bbpennel:
With everything versioned by default, I wonder if we should clean up all the references to `resource.isVersioned()`, since it seems like unnecessary work to keep checking for timemaps all over the place when you can't get rid of them

dbernstein:
@bbpennel : I've removed the unnecessary checks, but held off on completely removing  the enable and disable logic until we get community feedback in the release process.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -406,22 +407,23 @@ public FedoraResource getTimeMap() {
...
-            if (isMemento()) {
-                timeMapNode = node.getParent();
+            if (isOriginalResource()) {
+                return findOrCreateTimeMap();
+            } else if (isMemento()) {
+                return Optional.of(node.getParent()).map(nodeConverter::convert).orElse(null);
+
             } else {
-                timeMapNode = node.getNode(LDPCV_TIME_MAP);
+                return null;
```
bbpennel: I wonder if we should throw an exception here, since it would be requesting the timemap of something that can't have a timemap?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BinaryServiceImpl.java
```diff
@@ -96,6 +96,7 @@ public FedoraBinary findOrCreateBinary(final FedoraSession session, final String
                 touch(binary.getNode());
             }
 
+
```
awoods: Why?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BinaryServiceImpl.java
```diff
@@ -112,7 +113,9 @@ public NonRdfSourceDescription findOrCreateDescription(final FedoraSession sessi
 
             initializeNewDescriptionProperties(descNode);
 
-            return new NonRdfSourceDescriptionImpl(descNode);
+            final NonRdfSourceDescription desc = new NonRdfSourceDescriptionImpl(descNode);
```
awoods: You can probably just return the `new NonRdfSourceDescriptionImpl(descNode);`.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -1229,30 +1235,14 @@ public FedoraResource getBaseVersion() {
...
-
     @Override
-    public boolean isVersioned() {
-        try {
-            return getNode(getNode(), LDPCV_TIME_MAP, false) != null;
-        } catch (final RepositoryException ex) {
-            throw new RepositoryRuntimeException(ex);
-        }
+    public boolean isOriginalResource() {
+        return !isAcl() && !isMemento() && !isTimeMap();
```
bbpennel: It looks like FedoraTimeMap and FedoraWebacAcl have their own classes that inherit from FedoraResource. What would you think about overriding this method in those classes to just return `false`, and change this method to only check `!isMemento()`?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/services/BinaryServiceImpl.java
```diff
@@ -112,7 +113,9 @@ public NonRdfSourceDescription findOrCreateDescription(final FedoraSession sessi
 
             initializeNewDescriptionProperties(descNode);
 
-            return new NonRdfSourceDescriptionImpl(descNode);
```
bbpennel: Wondering if we need to keep any of the changes to this file?

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/utils/FedoraTypesUtils.java
```diff
@@ -145,6 +145,11 @@
      */
     public static Predicate<Node> isAcl = new AnyTypesPredicate(FEDORA_WEBAC_ACL);
 
+    /**
+     * Predicate for determining whether this {@link Node} is a TimeMap.
+     */
+    public static Predicate<Node> isTimeMap = new AnyTypesPredicate(FEDORA_TIME_MAP);
```
bbpennel: If you agree with my previous comment about overriding the isOriginalResource method, then we may not need this.

fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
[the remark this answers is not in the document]
dbernstein: If you have any suggestions, I'm open to them.  We used to look for the presence of the timemap as a direct child.  Timemaps are being lazily instantiated so we can't assume that a timemap will be present when isOriginalResource() is called.    Perhaps we can add a property to the jcr node that will flag it as an original resource?  If we went that route, we would need to suppress it on the way out and add it to the server managed triples.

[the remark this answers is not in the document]
awoods: Or we leave this logic (with @bbpennel 's [suggestion](https://github.com/fcrepo4/fcrepo4/pull/1432#discussion_r224164818)), and hope for the best.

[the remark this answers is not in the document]
dbernstein: actually - ignore this:  I tried Ben's suggestion again (it didn't work the first time I tried it).   This time I made all the unrelated changes and then layered Ben's suggestions on top and it worked.  So I must have introduced something unnecessary by mistake.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -290,7 +290,7 @@ protected void setVaryAndPreferenceAppliedHeaders(final HttpServletResponse serv
         // add vary headers
         final List<String> varyValues = new ArrayList<>(VARY_HEADERS);
 
-        if (resource.isVersioned()) {
+        if (resource.isOriginalResource()) {
```
bseeger: won't every resource just need AcceptDatetime now?  Ie, every resource is an OriginalResource, no?

bseeger: So maybe that's what this check is to prevent - a memento getting accept date time?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
```diff
@@ -257,20 +251,12 @@ void replaceProperties(final IdentifierConverter<Resource, FedoraResource> idTra
...
-     * Disable versioning
-     */
-    void disableVersioning();
-
-    /**
-     * Check if a resource is versioned
-     * @return whether the resource is versioned
+     * Check if a resource is an original resource
+     * (ie versionable, as opposed to non-versionable resources
+     * like mementos, timemaps, and acls).
```
bseeger: is there ever a case for acls to be versioned?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
bbpennel: @bseeger Right, it would exclude mementos, timemaps and acls

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
dbernstein: No. Not at this time.

PR #1437, opened 2018-10-06
Adding rdf:type memento:* - whether via PUT, POST or PATCH -

dbernstein:
no longer causes resources to disappear.
Instead a MalformedRdfException is thrown.

**Adding rdf:type memento:* - whether via PUT, POST or PATCH - no longer causes resources to disappear.**

**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2884

Adding `<> a memento:*` should result in a constraint exception since there are simply no cases where that should be allowed.  Furthermore,  we use the triple internally to and thus adding it to resources through the API wreaks havoc in the form of mysteriously disappearing resources.

This PR ensures that a constraint error is thrown whenever a client attempts to add such a triple via PUT, POST, or PATCH.

Integration tests cover the three cases.

# How should this be tested?

1. POST a resource  and verify 400 response: 
` curl -i -u fedoraAdmin:fedoraAdmin -X POST http://localhost:8080/rest/test -H "Content-Type: text/turtle"  --data "<> a <http://mementoweb.org/ns#Memento> ."`

2. PUT a resource  and verify 400 response: 
` curl -i -u fedoraAdmin:fedoraAdmin -X PUT http://localhost:8080/rest/test -H "Content-Type: text/turtle"  --data "<> a <http://mementoweb.org/ns#Memento> ."`

3. Create a resource, PATCH,   and verify 400 response: 
```
curl -i -u fedoraAdmin:fedoraAdmin -X PUT  http://localhost:8080/rest/test;
curl -i -u fedoraAdmin:fedoraAdmin -X PATCH  http://localhost:8080/rest/test -H "Content-Type: application/sparql-update"  --data "INSERT { <> a <http://mementoweb.org/ns#Memento> . } WHERE {}"
```

Example:
* Does this change require documentation to be updated?  Yes
* Does this change require any other modifications to be made to the repository (ie. Regeneration activity, etc.)?  No
* Could this change impact execution of existing code? Never say never.
# Interested parties

### DIFF
--- a/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
+++ b/fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
@@ -36,6 +36,7 @@ import static org.fcrepo.kernel.api.RdfCollectors.toModel;
 import static org.fcrepo.kernel.api.RdfLexicon.INTERACTION_MODELS;
 import static org.fcrepo.kernel.api.RdfLexicon.LAST_MODIFIED_DATE;
 import static org.fcrepo.kernel.api.RdfLexicon.LDP_NAMESPACE;
+import static org.fcrepo.kernel.api.RdfLexicon.MEMENTO_NAMESPACE;
 import static org.fcrepo.kernel.api.RdfLexicon.RDF_NAMESPACE;
 import static org.fcrepo.kernel.api.RdfLexicon.isManagedNamespace;
 import static org.fcrepo.kernel.api.RdfLexicon.isManagedPredicate;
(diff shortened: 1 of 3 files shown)


fcrepo-kernel-modeshape/src/main/java/org/fcrepo/kernel/modeshape/FedoraResourceImpl.java
```diff
@@ -884,6 +890,42 @@ private void checkInteractionModel(final UpdateRequest request) {
...
+                final UpdateData op = (UpdateData) operation;
+                quads.addAll(op.getQuads());
+            } else if (operation instanceof UpdateDeleteWhere) {
+                final UpdateDeleteWhere op = (UpdateDeleteWhere) operation;
+                quads.addAll(op.getQuads());
+            }
+        }
+
+        quads.stream().forEach(e -> {
+           ensureNoMementNamespacedObjects(e.asTriple());
```
awoods: This is probably fine for now, but should likely be refactored to not be re-iterating through the request for every triple that we want to block. 
For example, we are doing something quite similar [directly above this method](https://github.com/fcrepo4/fcrepo4/pull/1437/files#diff-7aee50d5030524028026aa94cd75f350R855).

PR #1450, opened 2018-10-11
Remove dead and redundant code from fcrepo-kernel-api

awoods:
- Tighten method and field qualifiers
- Some changes spread beyond fcrepo-kernel-api... since it modules use the -api

Resolves: https://jira.duraspace.org/browse/FCREPO-2904

Non-functional clean-up of fcrepo-kernel-api

# How should this be tested?
Successful build, deployment and one-click

# Additional Notes:
Some changes spread beyond fcrepo-kernel-api... since it modules use the -api

# Interested parties
@fcrepo4/committers


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
```diff
@@ -42,15 +41,13 @@
     **/
     public static final String REPOSITORY_NAMESPACE = "http://fedora.info/definitions/v4/repository#";
 
-    public static final String EVENT_NAMESPACE = "http://fedora.info/definitions/v4/event#";
-
-    public static final String FCREPO_API_NAMESPACE = "http://fedora.info/definitions/fcrepo#";
+    private static final String FCREPO_API_NAMESPACE = "http://fedora.info/definitions/fcrepo#";
```
bbpennel: I'm a little unsure about changing the visibility of these constants. I understand they aren't used currently outside this class, so its probably fine, it just seems likely to get flipped back later when we need to use one of these namespaces in a test or something.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
```diff
@@ -81,25 +76,19 @@
...
-
     /**
      * Is this namespace one that the repository manages?
      */
     public static final Predicate<String> isManagedNamespace = p -> p.equals(REPOSITORY_NAMESPACE) ||
             p.equals(LDP_NAMESPACE);
 
     // MEMBERSHIP
-    public static final Property HAS_PARENT =
+    private static final Property HAS_PARENT =
```
bbpennel: I don't think HAS_PARENT is ever populated anymore (I believe we deleted the class that added it a few weeks ago), so it seems like we can remove this property and update `membershipProperties`

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
```diff
@@ -179,34 +168,30 @@
     private static final Set<Property> ldpManagedProperties = of(CONTAINS);
 
     // REPOSITORY INFORMATION
-    public static final Property HAS_OBJECT_COUNT =
+    private static final Property HAS_OBJECT_COUNT =
```
bbpennel: `HAS_OBJECT_COUNT` and `HAS_OBJECT_SIZE` are never populated since having removed the metrics code, so these should be removeable (I think they are already blocked from being added via user requests because of the namespace)

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
```diff
@@ -179,34 +168,30 @@
...
-    public static final Property HAS_ACCESS_ROLES_SERVICE =
-            createProperty(REPOSITORY_NAMESPACE + "hasAccessRoles");
 
-    public static final Set<Property> repositoryProperties = of(
+    private static final Set<Property> repositoryProperties = of(
             HAS_OBJECT_COUNT, HAS_OBJECT_SIZE, HAS_TRANSACTION_SERVICE);
 
     // NAMESPACES
-    public static final Property HAS_NAMESPACE_PREFIX =
+    private static final Property HAS_NAMESPACE_PREFIX =
```
bbpennel: `HAS_NAMESPACE_URI` and `HAS_NAMESPACE_PREFIX` do not appear to be used for anything but specifying that they are SMTs, perhaps we should remove them?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
```diff
@@ -263,12 +233,12 @@
         managedProperties = b.build();
     }
 
-    public static final Set<Property> relaxableProperties
+    private static final Set<Property> relaxableProperties
```
bbpennel: This is the exact same thing as `serverManagedProperties` except for the order of the properties. Perhaps we should delete `serverManagedProperties` and use this in its place?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/VersionService.java
```diff
@@ -47,13 +47,13 @@
     /**
      * To format a datetime for use as a Memento path.
      */
-    public static final DateTimeFormatter MEMENTO_LABEL_FORMATTER = DateTimeFormatter.ofPattern("yyyyMMddHHmmss")
+    DateTimeFormatter MEMENTO_LABEL_FORMATTER = DateTimeFormatter.ofPattern("yyyyMMddHHmmss")
```
bbpennel: Does java automatically assume the `static final` for variables in interfaces? I didn't realize that.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/exception/ConstraintViolationException.java
```diff
@@ -23,6 +23,7 @@
  * @author whikloj
  * @since 2015-05-29
  */
+@SuppressWarnings("WeakerAccess")
```
bbpennel: This actually gives me a warning in eclipse:
`Unsupported @SuppressWarnings("WeakerAccess")`

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
awoods: I think `HAS_CHILD` can also be safely removed. We only use it in tests along the lines of: "assert HAS_CHILD does not exist".

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
awoods: Although the values are the same, there is a semantic difference. I would be inclined to create a separate ticket the reconciles our various incarnations of "Server Managed" variables.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/VersionService.java
awoods: Yes. https://www.javahelps.com/2015/02/introduction-to-interface-with-java-8.html

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
awoods: I believe these "clean-up" PRs should try to eliminate any possibility of introducing bugs.

New ticket: https://jira.duraspace.org/browse/FCREPO-2919

PR #1457, opened 2018-10-13
Remove dead and redundant code from fcrepo-http-api

awoods:
- Tighten method and field qualifiers.

Resolves: https://jira.duraspace.org/browse/FCREPO-2909

Non-functional clean-up of fcrepo-http-api

# How should this be tested?
Successful build, deployment and one-click

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -217,10 +217,10 @@ public class FedoraLdpIT extends AbstractResourceIT {
 
     private static final Logger LOGGER = getLogger(FedoraLdpIT.class);
 
-    private static DateTimeFormatter headerFormat =
+    private static final DateTimeFormatter headerFormat =
       DateTimeFormatter.ofPattern("EEE, dd MMM yyyy HH:mm:ss zzz", Locale.US).withZone(of("GMT"));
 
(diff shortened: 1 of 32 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -217,10 +217,10 @@
 
     private static final Logger LOGGER = getLogger(FedoraLdpIT.class);
 
-    private static DateTimeFormatter headerFormat =
+    private static final DateTimeFormatter headerFormat =
       DateTimeFormatter.ofPattern("EEE, dd MMM yyyy HH:mm:ss zzz", Locale.US).withZone(of("GMT"));
```
whikloj: This looks to be an RFC-1123 date format. So I think you could use the predefined one in the DateTimeFormatter or you could use the VersionService's which has the timezone set too.

[org.fcrepo.kernel.api.services.VersionService.MEMENTO_RFC_1123_FORMATTER](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/VersionService.java#L56)

awoods:
@whikloj : on a side note, in the future, it would be helpful if you selected `Request changes` instead of `Approve` when you have suggestions. Suggestions under the umbrella of `Approve` tend to get ignored... and your suggests should not be ignored.

PR #1460, opened 2018-10-16
vcard:member predicates with URI objects are now recognized as author…

dbernstein:
…ized users

Resolves:

**Now URI objects of vcard:member predicates in lists referenced by acl:agentGroup predicates are properly authroized.**
* * *

**JIRA Ticket**:  https://jira.duraspace.org/browse/FCREPO-2921

If you have configured your Fedora instance to use a "fcrepo.auth.webac.userAgent.baseUri" system property,  you can now expect that  a <> vcard:member <${fcrepo.auth.webac.userAgent.baseUri}/your-username> triple in your group list referenced by acl:agentGroup in your acl will be properly authorized.

I have added an integration test to confirm that this works.

# How should this be tested?
1. set the system property of your fedora instance: 
`-Dfcrepo.auth.webac.userAgent.baseUri=http://example.com/`
2. create a resource 
`curl -u fedoraAdmin:fedoraAdmin -X PUT http://localhost:8080/rest/test`
3. create a group-list resource with the following body: 
group-list.ttl
```
@prefix    acl:  <http://www.w3.org/ns/auth/acl#>.
@prefix  vcard:  <http://www.w3.org/2006/vcard/ns#>.
<> a vcard:Group ;
          vcard:hasMember  <http://example.com/testuser> .
```
`curl -u fedoraAdmin:fedoraAdmin -X PUT http://localhost:8080/rest/group-list -H "Content-Type: text/turtle" --data "@group-list.ttl"`

4. create an acl for the test resource with the following body (acl.ttl)
acl.ttl
```

<#authorization>    a               acl:Authorization;
    acl:accessTo    </rest/test>;
    acl:mode        acl:Read;
    acl:mode        acl:Write;
    acl:agentGroup  </rest/group-list>.
```

`curl -u fedoraAdmin:fedoraAdmin -X PUT http://localhost:8080/rest/test/fcr:acl -H "Content-Type: text/turtle" --data "@acl.ttl"`

5. Verify the resource is read accessible with the testuser.
`curl -i -u testuser:testpass http://localhost:8080/rest/test`
6. Verify the resource is write accessible with the testuser.
`curl -i -u testuser:testpass -X PUT http://localhost:8080/rest/test/child`

# Interested parties


fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACRolesProvider.java
```diff
@@ -270,9 +270,14 @@
...
         if (hasVcardGroup) {
-            return triples.stream()
+            final List<String> values = new ArrayList();
+
+            final List<org.apache.jena.graph.Node>  objects = triples.stream()
                           .filter(triple -> triple.predicateMatches(VCARD_MEMBER_NODE))
-                          .map(Triple::getObject).flatMap(WebACRolesProvider::nodeToStringStream);
+                          .map(Triple::getObject).collect(Collectors.toList());
+            objects.stream().flatMap(WebACRolesProvider::nodeToStringStream).forEach(values::add);
+            objects.stream().flatMap(WebACRolesProvider::additionalAgentValues).forEach(values::add);
```
bbpennel: If i'm understanding correctly, the stream returned by getAgentMembers is now going to contain the original URI of the agent and the URI of the agent relative to the repository, is that correct? Are both versions needed? Or is just the relativized version needed? 
Sorry, I'm not that familiar with the webac module, so I'm trying to get a handle on if the duplication is necessary. I guess what I'm wondering is if what is needed here is a combination of `nodeToStringStream` and `additionalAgentValue` so that the URI would only be added once and to keep down the number of iterations.
Bit of a tangent, but are `GROUP_AGENT_BASE_URI_PROPERTY` or `USER_AGENT_BASE_URI_PROPERTY` expected to be changeable while the repo is running? Wondering if they should be retrieved once at start time and reused.

awoods: I believe @dbernstein is attempting to support both ACLs scenarios:
1. An ACL has Strings for user-ids (i.e. [additionalAgentValues](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACRolesProvider.java#L438))
```
vcard:hasMember "someuser"
```
2. An ACL has URIs for user-ids (i.e. [nodeToStringStream](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACRolesProvider.java#L285))
```
vcard:hasMember <http://example.com/someuser>
```

In both of the cases above, I believe we want the end result to be `someuser`. That being the case, it seems like the logic could be simplified to not include both:
```
someuser
http://example.com/someuser
```
..in the `values` List.

Specifically, if the `object` is a URI, we want to strip the URI prefix off (ala additionalAgentValues), we do not want to keep the URI (ala nodeToStringStream).

fcrepo-auth-webac/src/test/java/org/fcrepo/integration/auth/webac/WebACRecipesIT.java
```diff
@@ -1198,6 +1198,21 @@ public void testAgentGroupWithHashUris() throws Exception {
...
+        ingestTurtleResource("fedoraAdmin", "/acls/agent-group-list-with-member-uris.ttl",
+                             serverAddress + "/rest/agent-group-list-with-member-uris");
+        final String authorized = ingestObj("/rest/agent-group-with-vcard-member-as-uri");
+        ingestAcl("fedoraAdmin", "/acls/agent-group-with-vcard-member-as-uri.ttl", authorized + "/fcr:acl");
+        //check that test user is authorized to write
+        final HttpPut childPut = new HttpPut(authorized + "/child");
+        setAuth(childPut, "testuser");
+        assertEquals(HttpStatus.SC_CREATED, getStatus(childPut));
+
+        System.clearProperty(USER_AGENT_BASE_URI_PROPERTY);
```
awoods: Although the resetting of System Properties is handled by:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-auth-webac/src/test/java/org/fcrepo/integration/auth/webac/WebACRecipesIT.java#L76

Being explicit about it here is not too harmful.

fcrepo-auth-webac/src/test/resources/acls/agent-group-list.ttl
```diff
@@ -3,4 +3,4 @@
 <#authorized> a vcard:Group ;
           vcard:hasMember  "testuser" .
 <#unauthorized> a vcard:Group ;
-          vcard:hasMember  "anotheruser" .
+          vcard:hasMember  "anotheruser" .
```
awoods: This change can probably be reverted.

fcrepo-auth-webac/src/test/java/org/fcrepo/integration/auth/webac/WebACRecipesIT.java
dbernstein: Ah - I did't realize @Rule was taking care of that.  I'll remove it.

fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACRolesProvider.java
dbernstein: Seeing that we were using both for acl:agent I thought there might be some reason for that.  I can change that, no problem.

dbernstein: Re changeability of the system properties,  I believe they do need to be changeable for the integration tests.

PR #1463, opened 2018-10-26
Fcrepo-2923 - Allow modification of premis:hasSize property

bbpennel:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2923

Allows clients to delete or set premis:hasSize property.

# What's new?
* Removed premis:hasSize from set of server managed triples
* Added integration test that verifies the expected behavior of the four server generated triples for binaries: premis:hasSize, ebucore:filename, ebucore:hasMimeType, iana:describedby
* Removes duplicate testLdpContains test (it is already being tested in testLdpNamespace)

# How should this be tested?
New integration test.

Create a binary, then issue a PATCH or PUT request to replace the premis:hasSize property. This can be done with a request similar to:

```
echo 'INSERT { <> <http://www.loc.gov/premis/rdf/v1#hasSize> "99"^^<http://www.w3.org/2001/XMLSchema#long> } WHERE { }' | curl -XPATCH -ufedoraAdmin:fedoraAdmin -H "Content-Type: application/sparql-update" --data-binary "@-" "http://localhost:8080/rest/test_bin"
 ```

Attempt to retrieve the description and perform head requests against the binary to see that the new size is being used. 

# Additional Notes:
While clients can set the premis:hasSize property, it must be of type "http://www.w3.org/2001/XMLSchema#long" and there can only be one property of this predicate at a time (adding a new one will replace the original). If the client replaces the binary, the property will be overwritten with the size of the new file. If no premis:hasSize is present, then 0 will be returned as the `Content-Length` header.

iana:describedby is added during requests to retrieve non-rdf resource descriptions, so it cannot be truly removed by clients. Clients can, however, add additional iana:describedby relations and delete those. This is not a new behavior in this PR.

This change is intended to align behaviors with the documentation here https://wiki.duraspace.org/display/FEDORA5x/What+are+the+Server+Managed+Triples

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
@@ -32,6 +32,7 @@ import static javax.ws.rs.core.Response.Status.GONE;
 import static javax.ws.rs.core.Response.Status.NO_CONTENT;
 import static javax.ws.rs.core.Response.Status.OK;
 import static org.apache.commons.lang3.StringUtils.isNotEmpty;
+import static org.apache.jena.rdf.model.ModelFactory.createDefaultModel;
 import static org.fcrepo.http.commons.test.util.TestHelpers.parseTriples;
 import static org.fcrepo.kernel.api.FedoraTypes.FCR_METADATA;
 import static org.fcrepo.kernel.api.RdfLexicon.CONSTRAINED_BY;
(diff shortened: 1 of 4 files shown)


awoods:
@bbpennel : As a note, it looks like your PR message was truncated:
> This change is intended to align...

awoods:
Unlike what you note above in your PR comment, my local testing shows that the additional Datatype information (e.g. ""99"^^<http://www.w3.org/2001/XMLSchema#long>") is not necessary when setting `premis:hasSize`.
Am I missing something?

bbpennel:
@awoods the type restriction on premis:hasSize only appears to apply when setting the property on a non-rdf description. For a container you can put whatever you want in there. For a binary I get `org.modeshape.jcr.value.ValueFormatException: Error converting "what" from String to a Long` when providing a non-numeric type.

Finished that incomplete sentence and pushed a commit for that constant.

PR #1464, opened 2018-10-30
Requiring allowed list http domains end with slash

bbpennel:
External content allow list entry for http uri with a domain but no slash after it will log an error and be ignored.

# How should this be tested?

```
echo "http://example.com" > /tmp/allowed.txt
mvn jetty:run -Dfcrepo.external.content.allowed=/tmp/allowed.txt

# ERROR 15:58:32.025 (ExternalContentPathValidator) Invalid path http://example.com in configuration /Users/bbpennel/git/fcrepo4/fcrepo-webapp/allowed.txt, domain must end with a '/'

curl -v -X PUT -H "Link: <http://example.com/>; rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"; handling=\"proxy\"; type=\"text/plain\"" http://localhost:8080/rest/blocked_http -ufedoraAdmin:fedoraAdmin

# 400 Bad Request
# External content is disallowed by the server


rm /tmp/allowed.txt
echo "http://example.com/" > /tmp/allowed.txt

mvn jetty:run -Dfcrepo.external.content.allowed=/tmp/allowed.txt
# no error

curl -v -X PUT -H "Link: <http://example.com/>; rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"; handling=\"proxy\"; type=\"text/plain\"" http://localhost:8080/rest/http_bin -ufedoraAdmin:fedoraAdmin

# 201 created

curl -v -X PUT -H "Link: <http://example.com:8080/>; rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"; handling=\"proxy\"; type=\"text/plain\"" http://localhost:8080/rest/diff_port -ufedoraAdmin:fedoraAdmin

# 400 Bad Request
# Path did not match any allowed external content paths: http://example.com:8080/
```

# Interested parties
@bseeger @awoods

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ExternalContentPathValidator.java
@@ -51,6 +51,7 @@ public class ExternalContentPathValidator extends AutoReloadingConfiguration {
 
     private static final Pattern SCHEME_PATTERN = Pattern.compile("^(http|https|file):/.*");
 
+    // Pattern to check that an http uri contains a / after the domain if a domain is present
     private static final Pattern HTTP_DOMAIN_PATTERN = Pattern.compile("^(http|https)://([^/]+/.*|$)");
 
     private static final Pattern RELATIVE_MOD_PATTERN = Pattern.compile(".*(^|/)\\.\\.($|/).*");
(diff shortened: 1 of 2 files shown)


bseeger:
This looks good to me. One minor point around error messages.  If I have only one thing in the file and it's a domain with no trailing slash, the error upon creating external content is:  

`External content is disallowed by the server` - because it found nothing valid in the allowed file, so it is thinking external content is disallowed.  Which of course it is, but if I'm the system admin who set this up, this will be confusing because I thought I configured it.   But in that case simply checking the log files should show the issue, since there's an ERROR message logged there.

If there is another valid entry in the allowed file, the error is: 
`Path did not match any allowed external content paths: http://example.com:8080`

So, it's behaving correctly, but I wonder if we want to have clearer errors.   Or same error for both?   I think I am overthinking this and that this PR is fine as is.  Any other opinions about this?

bseeger:
After thinking about this some more, I'm fine with the errors the way they are.

PR #1470, opened 2018-11-01
Ensures that the properties of the memento are displayed.

dbernstein:
Resolves: https://jira.duraspace.org/browse/FCREPO-2926

**Improves the display of mementos**
* * *

**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2926, https://jira.duraspace.org/browse/FCREPO-2928

https://jira.duraspace.org/browse/FCREPO-2929
This PR ensures that mementos are displayed in a way that  mirrors the original resource, by making sure that the velocity template is using the original resource rather than the memento URI (which formerly was resolved from the URIInfo).

# What's new?
Now you should be able to view mementos like ordinary RDF resources in the HTML UI.  The properties are now rendered instead of appearing blank ( the issue called out by https://jira.duraspace.org/browse/FCREPO-2926).  Additionally any hash-uris associated with the original resource are displayed in the Other Resources section,  mirroring the way hash-uris are displayed for non-memento RDF sources.

# How should this be tested?
1.  Open the UI (http://localhost:8080/rest) in the browser.
2. create a container.
3. Add a hash-uri using the following sparql-update:
```
INSERT {
<#test>  <http://purl.org/dc/elements/1.1/description> "my description" 
} WHERE {}
```
4. create a version.
5. click on newly created version link and verify that the memento display is the same as the original resource exception that the memento's title is the memento uri rather than the original resource.

# Additional Notes:

Example:
* Does this change require documentation to be updated? 
* Does this change add any new dependencies? 
* Does this change require any other modifications to be made to the repository (ie. Regeneration activity, etc.)? 
* Could this change impact execution of existing code?

# Interested parties

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
@@ -263,6 +263,7 @@ public class ViewHelpers {
 
     /**
      * Same as above but takes a string.
+     * NB: This method is currently used in fcrepo-http-api/src/main/resources/views/default.vsl
      * @param subjectUri the URI
      * @return a node
      */
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
@@ -32,6 +32,7 @@ import static java.util.Collections.emptyMap;
 import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toMap;
 import static org.fcrepo.kernel.api.FedoraTypes.FCR_METADATA;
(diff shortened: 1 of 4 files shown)


fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/responses/ViewHelpers.java
```diff
@@ -239,6 +240,36 @@ public String getObjectsAsString(final Graph graph,
...
+            return subject;
+        }
+    }
+
+    /**
+     * Same as above but takes a string.
+     * @param subjectUri the URI
+     * @return a node
+     */
+    public Node getOriginalResource(final String subjectUri) {
```
awoods: This looks good and works well. 
One minor suggestion, since it appears that this method is not used (and may in the future be at risk of being deleted) please add a comment indicating that the method is used in `fcrepo-http-api/src/main/resources/views/default.vsl`.

PR #1471, opened 2018-11-02
Updates version to 5.1.0-SNAPSHOT for continued development.

dbernstein:
Updates the version of the master branch to make way for the release candidate branch.


dbernstein:
@whikloj :  good catch on those missed pom references.  I've updated the PR.

PR #1477, opened 2018-11-15
"View Versions" link in HTMLUI no longer returns 404 if no versions have been created

dbernstein:
**"View Versions" link in HTMLUI no longer returns 404 if no version is created yet for a resource.**
* * *

**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2934

Ensures that UI renders the versions list page even when there are no versions.  Previously the link would return 404 if there were no versions created.   Also includes an HTML UI integration test to verify this fix.

# How should this be tested?
1. fire up HTML UI 
2. create a resource.
3. click on "View Versions" link.  
4. Verify that  a non 404 response is returned.

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
@@ -222,7 +222,7 @@ public class StreamingBaseHtmlProvider implements MessageBodyWriter<RdfNamespace
                     // node will not exist while the end point is visible to the user.
                     return null;
                 } else {
-                    throw new RuntimeException(e);
+                    throw e;
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
```diff
@@ -213,7 +214,17 @@ private FedoraResource getResourceFromSubject(final String subjectUri) {
...
-            return translator().convert(translator().toDomain(values.get("path")));
+            try {
+                return translator().convert(translator().toDomain(values.get("path")));
+            } catch (final RuntimeException e) {
+                if (e.getCause() instanceof PathNotFoundException) {
+                    //there is at least one case (ie Time Map endpoints - aka /fcr:versions) where the underlying jcr
+                    // node will not exist while the end point is visible to the user.
+                    return null;
+                } else {
+                    throw new RuntimeException(e);
```
awoods: You can change this to:
```
throw e;
```

PR #1484, opened 2018-11-29
Allow fcrepo5 to build and run on Windows

birkland:
[FCREPO-2693](https://jira.duraspace.org/browse/FCREPO-2963)

* Fixes several tests that fail on Windows due to differences in file paths
* Conditionally disables one test due to the usage of a feature in `wiremock` that does not work on Windows
* Updates `AutoReloadingConfiguration` to be tolerant of exceptions thrown upon `Paths.get(string)`

# What's new?
The only change to runtime fcrepo code is in `AutoReloadingConfiguration`.  This PR changes its behavior such that if an exception is thrown when attempting to read a file path via `Paths.get(string)`, monitoring the file is simply disabled.  The previous behavior was a uncaught runtime exception that prevented the fcrepo webapp from loading.

# How should this be tested?

Just make sure tests still pass.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/http/api/ExternalContentPathValidatorTest.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/http/api/ExternalContentPathValidatorTest.java
@@ -232,7 +232,7 @@ public class ExternalContentPathValidatorTest {
 
     @Test
     public void testCaseInsensitive() throws Exception {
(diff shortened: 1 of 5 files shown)


birkland:
Use this one instead: https://github.com/fcrepo4/fcrepo4/pull/1485
  We're margining into 5.0.0-RC instead of master

PR #1498, opened 2018-12-19
cherrypick  fcrepo-2970 (allowing blank node versioning) into 5x-maintenance

dbernstein:


awoods:
Related to https://jira.duraspace.org/browse/FCREPO-2970

PR #1499, opened 2018-12-19
Check if metadata is available in commons.js

gvasold:
Added code which only adds the suffix fcr:metadata to a resource URL if metadata is available.
Background: For a possibly strange use case we have references to external resources which are not known to fedora, so fedora does not have any metadata about these resources. This led to a problem because checkIfNonRdfResource has code which tries to provide metadata instead of the binary content by adding the fcr:metadata suffix if a HEAD request does not have a Link-Header (see fcrepo4#1168)
For our non-fedora resources this leads to 404 errors. I worked around this by testing if adding the suffix brings up status code 200 in a second HEAD request.

# What's new?
Added 5 lines of JavaScript code which sends a HEAD request to the resource and only adds the fcr:metadata suffix if the status code is 200.

# Additional Notes:
This change does not change the original behaviour for resources known to Fedora, except that an additional HEAD request is made when the user clicks on a link.

### DIFF
--- a/fcrepo-webapp/src/main/webapp/static/js/common.js
+++ b/fcrepo-webapp/src/main/webapp/static/js/common.js
@@ -113,7 +113,12 @@
           // WARNING: Fragile code relying on magic suffix '/fcr:metadata' and absence of 'Link' header 
           // on external resource.
           if(!url.match(/.*fcr:(metadata|tx)/)){
-            newLocation = url + "/fcr:metadata";
+             http('HEAD', url + "/fcr/metadata", function(res) {
+		            if(res.status == 200) {
+			              newLocation = url + "/fcr:metadata";
+		            } else {
+		              newLocation = url;
+	            }
           }
           location.href = newLocation ;
           return;


fcrepo-webapp/src/main/webapp/static/js/common.js
```diff
@@ -113,7 +113,12 @@
           // WARNING: Fragile code relying on magic suffix '/fcr:metadata' and absence of 'Link' header 
           // on external resource.
           if(!url.match(/.*fcr:(metadata|tx)/)){
-            newLocation = url + "/fcr:metadata";
+             http('HEAD', url + "/fcr/metadata", function(res) {
+		            if(res.status == 200) {
+			              newLocation = url + "/fcr:metadata";
+		            } else {
+		              newLocation = url;
+	            }
```
whikloj: I think you are missing a `);` from the end of this line

fcrepo-webapp/src/main/webapp/static/js/common.js
```diff
@@ -113,7 +113,12 @@
           // WARNING: Fragile code relying on magic suffix '/fcr:metadata' and absence of 'Link' header 
           // on external resource.
           if(!url.match(/.*fcr:(metadata|tx)/)){
-            newLocation = url + "/fcr:metadata";
+             http('HEAD', url + "/fcr/metadata", function(res) {
```
whikloj: this should be `fcr:metadata`, you've got a typo in there.

whikloj:
Hey @gvasold. Thanks for this, just so I understand better. You have an external binary but you do not have any metadata for it? Can you provide how you create those items so I can test this out. Cheers.

gvasold:
After taking a closer look on it, I found out that the problem is very specific to our server setup, because we are hosting and referencing files located on the same host, but not in the repository. So for example a triple like

~~~
<> xsl:hasXsltToLATEX https://myserver.org/xsl/some.xsl
~~~

matches [this condition](https://github.com/fcrepo4/fcrepo4/blob/082cf8afdefecaed3ca97bb0b898c59e40dee29b/fcrepo-webapp/src/main/webapp/static/js/common.js#L341), which results in piping the click through  checkIfNonRdfResource() which adds the fcr:metadata suffix. Resources with other host names are not affected. 

As our setup is very special, I do not know if it makes sense to add this to Fedora as it includes an extra HEAD request for each click on a link pointing to the servers hostname, but if there's interest, I'd be happy provide a clean version of my fix (sorry for the typos, I've typed the code into the online editor of github, which was stupid).

whikloj:
@gvasold Hmmm it kinda does as the `/fcr:metadata` suffix is a Fedora-ism and should not be assumed to apply to all things. Now that I've got some more information I can try to create a couple tests to see how our javascript is handling them. Thanks.

bbpennel:
I am going to close this PR since it is over 5 years old

PR #1501, opened 2019-01-02
Include user-provided rdf:types a Link headers

awoods:
Resolves: https://jira.duraspace.org/browse/FCREPO-2968

Ensures that user-provided types on RDF and descriptions of nonRDF Sources are returned as `Link` headers in GET and HEAD requests.

# How should this be tested?
1. Create a resource (binary or container)
1. Add a type property
   ```
   INSERT {
    <> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://pcdm.org/models#File> .
   }
   WHERE {}
   ```
1. Ensure that the new type is contained as a `Link` header in GET and HEAD requests

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -560,19 +560,15 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
         }
 
         // Add user-provided types as Link headers... when a description exists
-        if (resource.getDescription() != null) {
-            for (final URI typeURI : resource.getDescription().getTypes()) {
+        final FedoraResource resourceDescription = resource.getDescription();
+        if (resourceDescription != null) {
+            for (final URI typeURI : resourceDescription.getTypes()) {
 
                 // Get namespace of type
                 final String type = typeURI.toString();
-                final String namespace;
-                if (type.contains("#")) {
-                    namespace = type.substring(0, type.indexOf('#') + 1);
-                } else {
-                    namespace = type.substring(0, type.lastIndexOf('/') + 1);
-                }
+                final String namespace = createURI(type).getNameSpace();
 
-                // Omit server-managed types
+                // Omit server-managed types, as they are added elsewhere
                 if (!isManagedNamespace.test(namespace)) {
                     servletResponse.addHeader(LINK, buildLink(type, "type"));
                 }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -558,6 +558,26 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
                 servletResponse.addHeader(LINK, buildLink(VERSIONING_TIMEMAP_TYPE, "type"));
             }
         }
+
+        // Add user-provided types as Link headers... when a description exists
+        if (resource.getDescription() != null) {
+            for (final URI typeURI : resource.getDescription().getTypes()) {
+
+                // Get namespace of type
+                final String type = typeURI.toString();
+                final String namespace;
+                if (type.contains("#")) {
+                    namespace = type.substring(0, type.indexOf('#') + 1);
+                } else {
+                    namespace = type.substring(0, type.lastIndexOf('/') + 1);
+                }
+
+                // Omit server-managed types
+                if (!isManagedNamespace.test(namespace)) {
+                    servletResponse.addHeader(LINK, buildLink(type, "type"));
+                }
+            }
+        }
     }
 
     /**
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
@@ -53,6 +53,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
+import java.net.URI;
 import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Collection;
@@ -456,7 +457,12 @@ public abstract class AbstractResourceIT {
 
     protected CloseableHttpResponse setProperty(final String pid, final String propertyUri, final String value)
             throws IOException {
-        return setProperty(pid, null, propertyUri, value);
+        return setProperty(pid, null, propertyUri, "\"" + value + "\"");
+    }
+
+    protected CloseableHttpResponse setProperty(final String pid, final String propertyUri, final URI value)
+            throws IOException {
+        return setProperty(pid, null, propertyUri, "<" + value.toString() + ">");
     }
 
     private CloseableHttpResponse setProperty(final String id, final String txId, final String propertyUri,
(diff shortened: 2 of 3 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -558,6 +558,26 @@ private void addResourceLinkHeaders(final FedoraResource resource, final boolean
...
+                final String type = typeURI.toString();
+                final String namespace;
+                if (type.contains("#")) {
+                    namespace = type.substring(0, type.indexOf('#') + 1);
+                } else {
+                    namespace = type.substring(0, type.lastIndexOf('/') + 1);
+                }
+
+                // Omit server-managed types
+                if (!isManagedNamespace.test(namespace)) {
```
birkland: The intent of this is unclear.  Is it to avoid duplicate link headers due to the fact that some managed namespaces are added elsewhere in the code (e.g. LDP types, memento types), or is the intent to assure that the types in the `fedora` ontology are not exposed, or both?

awoods: The intent of this check is to prevent duplicate `Link` headers, as the server-manged types are added elsewhere.

birkland:
For a feature such as this,  I think we ought to consider adding lightweight RFCs to the specification process.  I can easily see clients relying in this sort of behaviour, which is not contained in any spec.  

Writing up an RFC would at least:
 * Define a contract that client can claim it requires, and a repository can claim it provides
 * Possibly inform other implementations so that they can implement these sort of add-ons in similar ways
 * By providing a name and reference, help determine its popularity/usage among clients and servers (who may cite it as something they require or provide)
 * Possibly open the door to incorporating it into a future fcrepo spec, if it proves to be essential to the community.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -558,6 +558,26 @@ private void addResourceLinkHeaders(final FedoraResource resource, final boolean
                 servletResponse.addHeader(LINK, buildLink(VERSIONING_TIMEMAP_TYPE, "type"));
             }
         }
+
+        // Add user-provided types as Link headers... when a description exists
+        if (resource.getDescription() != null) {
```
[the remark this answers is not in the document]
awoods: @bbpennel : Out of curiosity, what tooling/process did you use to get the low-level timings?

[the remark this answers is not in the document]
bbpennel: @awoods `System.nanoTime()`, output statements and existing integration tests :) sorry, nothing magical.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
birkland: In that case, there may be a bug, as it has the side effect of removing types in the fedora namespace, as noted in the [original review comment](https://github.com/fcrepo4/fcrepo4/pull/1501#pullrequestreview-189032473):
> for example <> rdf:type <fedora:Resource> appears in RDF, but not as a link header. 

To show this, create a binary, and look at the triples in the /fcr:metadata resource.  You'll see something like:

        rdf:type                 fedora:NonRdfSourceDescription ;
        rdf:type                 fedora:Binary ;
        rdf:type                 <http://example.org/Bar> ;
        rdf:type                 <http://example.org#Foo> ;
        rdf:type                 fedora:Resource ;

Yet in the headers, the "fedora:Resource, fedora:Binary" are not present:

    Link: <http://mementoweb.org/ns#OriginalResource>; rel="type"
    Link: <http://mementoweb.org/ns#TimeGate>; rel="type"
    Link: <http://example.org/Bar>; rel="type"
    Link: <http://example.org/#Foo>; rel="type"

bbpennel: @birkland My impression is that the intent is to not display any of the server managed rdf types in the fedora namespace via the link headers (fedora:NonRdfSourceDescription is not present either). I don't know the reasons for that decision, except perhaps that they aren't included in the specification.

awoods: @birkland : It does not seem like the behavior you are observing would be related to updates in this PR. Are you witnessing the presence of `fedora:Resource` and `fedora:Binary` Link headers in `master`?

birkland: @bbpennel  @awoods I'm just trying to figure out what the desired relationship between between `rdf:type` expressed in RDF vs `Link` headers is supposed to be.  The jira ticket suggests it should be all rdf:type values:

> It would be helpful to clients if the header included ***all rdf:type values*** so that they could detect locally meaningful resource types

As demonstrated, his PR does _not_ include all rdf:type values present in the RDF.  If that is indeed the expected behaviour, then I think a clearer definition of the contract regarding exposed link headers needs to be articulated.   See also [my comment on writing this up as an RFC](https://github.com/fcrepo4/fcrepo4/pull/1501#issuecomment-451182902).  If another implementation wanted to do this, what sort of logic would they need to employ?

awoods: @birkland : I am reluctant to encourage the use of the `fedora:NonRdfSourceDescription`,    `fedora:Binary`, `fedora:*` RDF types in cases where the same information is represented by more standard ontologies... that we already use.

It would have been good to pull those out in 5.0.0... but we can wait for 6.0.0.

birkland: @awoods So is the contract this PR offers to clients one of the following?  

In addition to mandatory link rel=type headers required by the fedora specifications (LDP, memento, etc):
* A) link rel=type headers shall contain all rdf:types present in the body?
* B) link rel=type headers shall contain all non-server-managed rdf:types found in the body?
* C) link rel=type headers shall include all rdf:types present in the body, but we are making an exception for fcrepo5 because one of the properties are deprecated.  If Fedora5+ ever adds a new, non-deprecated, server-managed rdf:type, then we'll be bound to expose it in the headers.

awoods: @birkland : At this point, I think option `B` is the most correct.

birkland: @awoods OK.  Mind if I update the jira ticket with the following text:

The desired functionality is as follows.  In addition to mandatory link rel=type headers required by the fedora specifications (LDP, memento, etc):
* On LDP-RS (RDFSource) resources, Fedora shall expose link rel=type headers for all non-server-managed rdf:type triples found in its body whose subject is the LDP-RS
* On LDP-NR (NonRDFSource) resources, Fedora shall expose Link rel=type headers for all non-server-managed rdf:type triples found in the body of its descriptive LDP-RS whose subject is the LDP-NR
* On LDP-RS resources that describe an LDP-NR, Fedora shall expose Link rel=type headers for all non-server-managed rdf:type triples found in its body whose subject is either the LDP-RS, or the described LDP-NR

awoods: @birkland : please do. Thank you!

PR #1518, opened 2019-02-14
5.0.2 rc to 5.x-maintenance

yinlinchen:
merge 5.0.2-RC into 5.x-maintenance


dbernstein:
When I look at the "Files changed" it looks right.  However the ordering of the commits don't look quite right to me.  @awoods :  can you take a look and confirm that we're all good here?

PR #1522, opened 2019-04-24
Adds Support for X-State-Token in GET and HEAD requests.

dbernstein:
**Adds Support for X-State-Token in GET and HEAD requests.**
* * *

**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2976

This PR adds the X-State-Token header into the http responses for LDP-RS and LDP-NS objects including ACLs and LDPCVs.

# What's new?
The change is very simple:  the value of the Etag for a resource is being plugged into the X-State-Token header on GET and HEAD requests.

# How should this be tested?
Integration tests cover the functionality.  For manual tests, perform GETs and HEADs above mentioned classes of resources and you should see the  X-State-Token header in the responses.

# Additional Notes:

Example:
* Does this change require documentation to be updated? Yes
* Does this change require any other modifications to be made to the repository (ie. Regeneration activity, etc.)?  No
* Could this change impact execution of existing code? No

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -765,7 +765,13 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
         }
 
         if (!etag.getValue().isEmpty()) {
-            servletResponse.addHeader("ETag", etag.toString());
+            final String etagStrVal = etag.toString();
+            servletResponse.addHeader("ETag", etagStrVal);
+
+            //State Tokens, while not used for caching per se,  nevertheless belong
+            //here since we can conveniently reuse the value of the etag for
+            //our state token
+            servletResponse.addHeader("X-State-Token", etagStrVal);
         }
 
         if (date != null) {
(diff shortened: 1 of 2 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -765,7 +765,13 @@ protected void addCacheControlHeaders(final HttpServletResponse servletResponse,
         }
 
         if (!etag.getValue().isEmpty()) {
-            servletResponse.addHeader("ETag", etag.toString());
+            final String etagStrVal = etag.toString();
```
awoods: Have you inspected the value of `etagStrVal`? It is a weak-etag, no? i.e. `W/123456`

dbernstein: It is a weak etag.  Is that a problem? Any recommendations?

awoods: I believe the only reason we return weak etags is due to the lack of guarantees around the order of RDF triples.
It is probably a non-issue. It may seem a little less hokey if we don't include the `W/` in our state-token... but not a big deal.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/StateTokensIT.java
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.http.client.methods.HttpHead;
+import org.apache.http.client.methods.HttpPut;
+import org.junit.Test;
+
+/**
+ * @author dbernstein
+ */
+public class StateTokensIT extends AbstractResourceIT {
+
+    public static final String X_STATE_TOKEN_HEADER = "X-State-Token";
```
awoods: Can be `private`.

PR #1523, opened 2019-04-26
Adds Support for X-If-State-Token on PUTs and PATCHs.

dbernstein:
**Adds Support for X-If-State-Token on PUTs and PATCHs..**
* * *

**JIRA Tickets**: 
https://jira.duraspace.org/browse/FCREPO-2993
https://jira.duraspace.org/browse/FCREPO-2994

Ensures that PUTs and  PATCHs honor the X-If-State-Token header as defined by the Fedora specification.

# What's new?
I added a new method FedoraResource.getStateToken() in order to ensure that we distinguish between StateTokens and Etags.  Currently we are using weak etags.  However, that may change in Fedora 6.  Therefore we may want to handle the implementation differently in 6.

# How should this be tested?
Integration tests cover success and failure for both PUTs and PATCHs  for rdf, nonrdf and acl's.

Example:
* Does this change require documentation to be updated? Yes 
* Could this change impact execution of existing code? Unlikely.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -858,7 +858,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
             final String clientSuppliedStateToken = headers.getHeaderString("X-If-State-Token");
             if (clientSuppliedStateToken != null && !stateToken.equals(clientSuppliedStateToken)) {
                 throw new PreconditionException(format(
-                    "The client-supplied value ({0}) does not match the resource's current state token ({1}).",
+                    "The client-supplied value ({0}) does not match the current state token ({1}).",
                     clientSuppliedStateToken, stateToken), 412);
             }
         }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -18,6 +18,7 @@
 package org.fcrepo.http.api;
 
 import static com.google.common.base.Strings.nullToEmpty;
+import static java.text.MessageFormat.format;
 import static java.util.EnumSet.of;
 import static java.util.stream.Stream.concat;
 import static java.util.stream.Stream.empty;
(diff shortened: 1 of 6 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/StateTokensIT.java
```diff
@@ -144,4 +153,116 @@ public void testHeadHasStateTokenNonRDFSource() throws IOException {
...
+        }
+
+        //get state token
+        final String stateToken;
+        try (final CloseableHttpResponse response = execute(new HttpHead(location))) {
+            assertEquals(OK.getStatusCode(), getStatus(response));
+            stateToken = response.getFirstHeader(X_STATE_TOKEN_HEADER).getValue();
+        }
+
+        LOGGER.debug("stateToken={}", stateToken);
```
whikloj: This can probably be removed

dbernstein:
@whikloj : log lines removed.

dbernstein:
Just a note:  I also ran the fedora api testsuite against this PR: the remaining two failed tests passed as expected.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -847,6 +851,17 @@ void evaluateRequestPreconditions(final Request request,
...
                     : "Request failed due to unspecified failed precondition.", response.getStatus());
         }
+
+        final String method = request.getMethod();
+        if (method.equals(HttpPut.METHOD_NAME) || method.equals(HttpPatch.METHOD_NAME)) {
+            final String stateToken = resource.getStateToken();
+            final String clientSuppliedStateToken = headers.getHeaderString("X-If-State-Token");
+            if (clientSuppliedStateToken != null && !stateToken.equals(clientSuppliedStateToken)) {
+                throw new PreconditionException(format(
+                    "The client-supplied value ({0}) does not match the resource's current state token ({1}).",
```
awoods: [MessageFormat](https://docs.oracle.com/javase/8/docs/api/java/text/MessageFormat.html) has issues with single quotes. The following format produces:
```
The client-supplied value (junk) does not match the resource's current state token ({1}).
```
Instead of:
```
The client-supplied value (junk) does not match the resource's current state token (b644dca3c7af172cf56625bcd5877c247a213a92).
```

Suggestion: Change the message to (note, no singe quote `resource's`):
```
"The client-supplied value ({0}) does not match the current state token ({1})."
```

PR #1529, opened 2019-05-31
Ensures that the acl:default predicate behavior is inherited correctl…

dbernstein:
**Ensures that the acl:default predicate behavior is inherited correctly when referencing a child**
* * *

**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3000

Fixes the bug described in https://jira.duraspace.org/browse/FCREPO-3000 and adds an integration test to prove that it has been fixed. 

# How should this be tested?
See https://jira.duraspace.org/browse/FCREPO-3000 for manual test.

Example:
* Does this change require documentation to be updated? No
* Could this change impact execution of existing code? Possibly

# Interested parties
@whikloj @peichman-umd


dbernstein:
acl:default values that refer to children will not be allowed going forward.

PR #1535, opened 2019-07-15
Generate timemap paths in platform-independent way (#1516)

dbernstein:
**Generate timemap paths in platform-independent way**
* * *

**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-2986 

* Generate timemap paths in platform-independent way
* Special case for resources ending in slash
* Emulates the previous behavior where path termination with a slash is
irrelevant

# What's new?
This commit was included in the 5.0.2 release but did not get merged into master. Thus it is missing on 5.1.0-RC and is the likely cause of failures of the code to build on winows.

# How should this be tested?

To test, please build on windows and verify that it passes.

# Interested parties
@birkland


whikloj:
Should there be a PR against master with this change too?

PR #1538, opened 2019-07-26
towards connecting to the persistence api.

dbernstein:
Do not merge:  This PR is for discussion purposes only at this point.

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainerServiceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainerServiceImpl.java
@@ -1,0 +1,100 @@
+package org.fcrepo.kernel.impl;
+
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import static org.fcrepo.persistence.api.Options.Option.ARCHIVAL_GROUP;
+
+import org.fcrepo.kernel.api.FedoraSession;
+import org.fcrepo.kernel.api.models.Container;
+import org.fcrepo.kernel.api.services.ContainerService;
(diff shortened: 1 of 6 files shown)


awoods:
Is there a JIRA?

dbernstein:
@awoods :  all valid points:  I was thinking of this as a quick sketch based on the conversation @bseeger  and I were having during the Committer Hour. No Jira yet.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainerServiceImpl.java
```diff
@@ -0,0 +1,100 @@
...
+import org.fcrepo.kernel.api.services.ContainerService;
+import org.fcrepo.persistence.api.Options;
+import org.fcrepo.persistence.api.PersistentResource;
+import org.fcrepo.persistence.api.PersistentStorage;
+import org.fcrepo.persistence.api.PersistentStorageSession;
+
+/**
+ * @author dbernstein
+ */
+public class ContainerServiceImpl implements ContainerService {
```
birkland: One thing to think about is breaking free of the rdf == container paradigm that Fedora currently has.  When modeling in OCFL, for example, a container is analogous to a directory (with some associated RDF).  This isn't always desirable.  Sometimes you'd just want rdf as rdf/file; in LDP an RDFSource.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainerServiceImpl.java
```diff
@@ -0,0 +1,100 @@
...
+        if (container == null) {
+            return create(session, path, interactionModel);
+        } else {
+            return container;
+        }
+    }
+
+    private Container create(final FedoraSession session, final String path, final String interactionModel) {
+        final String psSessionId = resolvePeristentSessionId(session);
+        final PersistentStorageSession storageSession = persistentStorage.getSession(psSessionId);
```
birkland: What is the difference between a FedoraSession, and PersistentStorageSession?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainerServiceImpl.java
```diff
@@ -0,0 +1,100 @@
...
+ */
+public class ContainerServiceImpl implements ContainerService {
+
+    private PersistentStorage persistentStorage;
+
+    @Override
+    public Container findOrCreate(final FedoraSession session, final String path, final String interactionModel) {
+
+        //TODO validate interaction model
+        //if archival group, ensure parent is not archival group
```
birkland: So we'd need this API to say "I want to create an archival group", right?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainerServiceImpl.java
whikloj: It might be nice to not have to keep switching sessions for other sessions, if we can define them at a level that they can be used by all implementing classes.

dbernstein:
This code is obselete.

PR #1545, opened 2019-09-19
Added fcrepo-kernel-impl module

mohideen:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3046

* Other Relevant Links (Mailing list discussion, related pull requests, etc.)

# What does this Pull Request do?
Add a new module fcrepo-kernel-impl

# How should this be tested?

Verify the maven build includes the fcrepo-kernel-impl module

# Interested parties
@whikloj

PR #1546, opened 2019-09-19
 FCREPO-3046. Create Fedora Transaction Interfaces and classes

mohideen:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3046

- #1545 

Creates the Interfaces and simple implementation classes related to FedoraTransaction

# What's new?
Adds interfaces
- FedoraTransaction
- FedoraTransactionManager
Adds implementations
- FedoraTransactionImpl
- FedoraTransactionManagerImpl

# How should this be tested?
Verify that the build and the tests pass.

# Additional Notes:
This PR does not include the changes to delete FedoraSession. I will create a separate PR for that.

# Interested parties
@dbernstein

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraTransactionManager.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraTransactionManager.java
@@ -1,0 +1,42 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.kernel.api;
+
+/**
+ * The Fedora Transaction Manager abstraction
+ *
+ * @author mohideen
+ */
+public interface FedoraTransactionManager {
+
+    /**
+     * Create a new fedora transaction
+     * 
+     * @return {@link FedoraTransaction} The new fedora transaction
+     */
+    FedoraTransaction create();
+
+    /**
+     * Get an existing fedora transaction
+     * 
+     * @param transactionId the id of the transaction to be returned
+     * @return {@link FedoraTransaction} the fedora transaction associated with the provided id
+     */
+    FedoraTransaction get(String transactionId);
+
+}
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraTransactionImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraTransactionImpl.java
@@ -29,6 +29,9 @@ public class FedoraTransactionImpl implements FedoraTransaction {
     final String id;
 
     FedoraTransactionImpl(final String id) {
+        if (id == null || id.isEmpty()) {
+            throw new IllegalArgumentException("FedoraTransaction id should not be empty!");
+        }
         this.id = id;
     }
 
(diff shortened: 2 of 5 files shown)


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraTransactionManager.java
```diff
@@ -0,0 +1,42 @@
...
+     */
+    FedoraTransaction create();
+
+    /**
+     * Get an existing fedora transaction
+     * 
+     * @param transactionId the id of the transaction to be returned
+     * @return {@link FedoraTransaction} the fedora transaction associated with the provided id
+     */
+    FedoraTransaction get(String transactionId);
```
dbernstein: Perhaps this method should throw a TransactionNotFoundException if the transactionId is not recognized?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraTransactionImpl.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * The Fedora Transaction implementation
+ *
+ * @author mohideen
+ */
+public class FedoraTransactionImpl implements FedoraTransaction {
+
+    final String id;
+
+    FedoraTransactionImpl(final String id) {
```
dbernstein: Perhaps assert non-null value here.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraTransactionManagerImpl.java
```diff
@@ -0,0 +1,60 @@
...
+public class FedoraTransactionManagerImpl implements FedoraTransactionManager {
+
+    private final HashMap<String, FedoraTransaction> transactions;
+
+    FedoraTransactionManagerImpl() {
+        transactions = new HashMap();
+    }
+
+    @Override
+    public FedoraTransaction create() {
```
dbernstein: This method should be synchronized I believe.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraTransactionManager.java
dbernstein: Or better,  down in the implementation, it should throw a TransactionNotFoundException (which should be a RuntimeException) since the client won't be able to recover from an unrecognized transaction request.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraTransactionManagerImpl.java
```diff
@@ -0,0 +1,60 @@
...
+        transactions.put(txId, tx);
+        return tx;
+    }
+
+    @Override
+    public FedoraTransaction get(final String transactionId) {
+        if(transactions.containsKey(transactionId)) {
+            return transactions.get(transactionId);
+        } else {
+            throw new RuntimeException("No Transaction found with transactionId: " + transactionId);
```
dbernstein: It should throw a TransactionNotFoundException (which should be a RuntimeException) since the client won't be able to recover from an unrecognized transaction request.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/FedoraTransactionImpl.java
mohideen: @dbernstein Should it throw an Exception if the id is null or empty?

dbernstein: I think it is best to throw an  IllegalArgumentException

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/FedoraTransactionManagerTest.java
```diff
@@ -0,0 +1,60 @@
...
+        testTx = (FedoraTransactionImpl) testTxManager.create();
+        assertNotNull(testTx);
+    }
+
+    @Test
+    public void testGetTransaction() {
+        final FedoraTransactionImpl tx = (FedoraTransactionImpl) testTxManager.get(testTx.getId());
+        assertNotNull(tx);
+        assertEquals(testTx.getId(), tx.getId());
+    }
```
dbernstein: @mohideen :  would you also just add a test demonstrating the runtime exception when not found.

PR #1547, opened 2019-09-19
Add PersistentStorage and PersistentStorageTransaction interfaces

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3047

Adds two interfaces related to the depiction here:
https://wiki.duraspace.org/pages/viewpage.action?pageId=133366058

# What's new?

# How should this be tested?

Read and review.

# Additional Notes:

Example:

# Interested parties
@fcrepo4/committers


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorage.java
```diff
@@ -0,0 +1,78 @@
...
+    public FedoraResource read(final PersistentStorageTransaction psTxId, final String identifier);
+
+    /**
+     * Return a version of a resource from persistent storage
+     *
+     * @param identifier The identifier of the resource to retrieve.
+     * @param version The date/time of the version to retrieve.
+     * @return The version of the resource.
+     */
+    public FedoraResource read(final String identifier, final Instant version);
```
whikloj: I'll note that this does **not** allow for a transaction. I am unsure if it should, because we have said all actions would happen in some sort of transaction.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorage.java
```diff
@@ -0,0 +1,78 @@
...
+public interface PersistentStorage {
+
+    /**
+     * Create a new resource on the persistent storage.
+     *
+     * @param psTxId The persistent storage transaction for this action.
+     * @param resource The new fedora resource to persist to storage.
+     * @return the resource with any new information related to the persisting..?
+     */
+    public FedoraResource create(final PersistentStorageTransaction psTxId, final FedoraResource resource);
```
dbernstein: Should this method throw a PersistentStorageException?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorage.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Update an existing resource on persistent storage.
+     *
+     * @param psTxId The persistent storage transaction for this action.
+     * @param resource The changed fedora resource to persist to storage.
+     * @return the updated resource with any new information related to the persisting..?
+     */
+    public FedoraResource update(final PersistentStorageTransaction psTxId, final FedoraResource resource);
```
dbernstein: same as above

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorage.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Delete a resource from persistent storage.
+     *
+     * @param psTxId The persistent storage transaction for this action.
+     * @param resource The current fedora resource to delete.
+     * @return The tombstone for the removed resource.
+     */
+    public Tombstone delete(final PersistentStorageTransaction psTxId, final FedoraResource resource);
```
dbernstein: same as above

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorage.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Return a resource from persistent storage
+     *
+     * @param psTxId The persistent storage transaction for this action.
+     * @param identifier The identifier of the resource to retrieve.
+     * @return The resource.
+     */
+    public FedoraResource read(final PersistentStorageTransaction psTxId, final String identifier);
```
dbernstein: same as above - possibly NotFoundException exception?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorage.java
dbernstein: same as above

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorageTransaction.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @author whikloj
+ * @since 2019-09-19
+ */
+public interface PersistentStorageTransaction {
+
+    /**
+     * Prepares the transaction to be committed by ensuring it can be locked.
+     */
+    public void prepare();
```
dbernstein: I believe this method should throw a checked exception since the client would want to respond to possible conditions:  timeout waiting for locking mechanism, etag changed, etc.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorageTransaction.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Commits any changes in the current transaction to persistent storage.
+     */
+    public void commit();
```
dbernstein: I believe this method should throw a checked exception since the client would want to respond to failure with a rollback.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorage.java
bbpennel: i think it should, since resources created in a transaction are only visible within that transaction

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorageTransaction.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Rolls back any changes in the current transaction.
+     */
+    public void rollback();
```
dbernstein: again - we might want to take action in the case of a rollback failure. Checked exception?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorage.java
bbpennel: I also wanted to mention that when I was doing my flow diagram thing, I had FedoraTransactions being passed into these methods, while PersistentStorageTransaction was something used internally within the Persistence layer. Mainly mentioning that so we can discuss which approach is most fitting.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorage.java
birkland: Out of curiosity, why is the argument name `psTxId`, rather than `psTx`.  It threw me off a little bit, and made me think that the role of `psTxId` was just to identify a transaction, rather than provide access to it.

birkland: I'd say use runtime exceptions until it's clear what sort of checked exceptions would be needed.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorage.java
birkland: right, re: "only visible within that transaction".  It does raise the question as whether an implicit transaction is needed for a basic read, so maybe it can be considered nullable?  or optional?  or a different method?  Not sure.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorage.java
whikloj: @birkland sorry I was moving fast and copying. I'll update this variable name

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorage.java
whikloj: Based on the diagram it is the `PersistentStorageTransaction` that is passed to the `PersistentStorage`, unsure if there is value/deficit to passing the `FedoraTransaction`.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorage.java
whikloj: I am okay with `null`ing this for non-transaction requests. That does mean that somewhere higher up we are _not_ creating a transaction...which was not the plan so far. I do wonder if there will be a cumulative performance benefit to not creating transactions unless required.

whikloj: Also this is the `read` for a version of the resource. Versus the `read` further up which would probably be used when accessing a changed resource in a transaction.

However I guess if you are performing a **restore** you might link the `GET` of the Memento and the `PUT` of it over-top the HEAD of the resource together in a transaction.

In which case you would only need to be aware you were _inside_ a transaction if the version you want was created in the same transaction.

whikloj:
@dbernstein I know you said `checked` exceptions but we seem to have no checked exceptions (that I could find). We can certainly add some for the uses here. But to move quickly I just went with the nearest unchecked exception to describe the issue.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/PersistentStorage.java
dbernstein: Weren't we saying the creating a version within a transaction would not be allowed?  It seems to me that when reading committed versions,  a transaction is never needed.  I think that should be explicit in the api.

dbernstein: I think different methods for read a version versus reading the current state of the object within an uncommitted transaction is most clear (as it is currently defined)

bbpennel: Whether to support creating a version in a transaction may depend on whether or not ending a transaction creates a new version automatically. And that is likely going to be persistence impl specific.

Sorry, I missed that `version` param. I agree, the transaction should generally not matter when getting a version. The only edge case that comes to mind:
* the client deletes an object in a transaction, and then requests a past version of that object while passing along the transaction id (since it would probably be passing in the transaction id on every request). Should it get a 404/410, or should it get the past version?

I'm guessing that in modeshape you could create versions in a transaction, but I don't know if anyone really did that.

dbernstein:
@whikloj  is reworking the PR based on a meeting this afternoon.  We will most likely be abandoning this PR.  DO NOT MERGE.

whikloj:
This was replaced by #1548

PR #1550, opened 2019-09-23
Make ResourceFactory and remove containerService/binaryService

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3058

Removes old services and adds new factory interface.

Updates tests to remove references to containerService and/or binaryService.

# How should this be tested?

Review, build.

# Additional Notes:

* Could this change impact execution of existing code? heck yeah

# Interested parties
@fcrepo4/committers


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Interface to a factory to locate or create new resources.
+ *
+ * @author whikloj
+ * @since 2019-09-23
+ */
+public interface ResourceFactory {
+
+    /**
+     * Find or create a new container.
```
dbernstein: create-> initialize

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @param identifier The path or identifier for the resource.
+     * @param interactionModel The interaction model for the new/expected container.
+     * @return The container object.
+     */
+    public Container findOrInitContainer(final String identifier, final String interactionModel);
+
+    /**
+     * Find or create a new binary. Newly created binaries will have nothing backing them until they are populated by a
```
dbernstein: create -> initialize

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public FedoraBinary findOrInitBinary(final String identifier);
+
+    /**
+     * Find or create a binary description.
+     *
+     * @param identifier The path or identifier for the resource.
+     * @return The description object.
+     */
+    public NonRdfSourceDescription findOrInitBinaryDescription(final String identifier);
```
dbernstein: create -> initialize

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * Find or create a timemap.
+     *
+     * @param identifier The path or identifier for the resource.
+     * @return The timemap object.
+     */
+    public FedoraTimeMap findOrInitTimemap(final String identifier);
```
dbernstein: create -> initialize

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * Find or create a Webac ACL.
+     *
+     * @param identifier The path or identifier of the resource this ACL is linked to.
+     * @return The ACL object.
+     */
+    public FedoraWebacAcl findOrInitAcl(final String identifier);
```
dbernstein: create -> initialize

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
```diff
@@ -0,0 +1,75 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.kernel.api.models;
+
+/**
+ * Interface to a factory to locate or create new resources.
```
dbernstein: create -> initialize

dbernstein: Maybe emphasize here the fact the meaning of "initialize".

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * @author whikloj
+ * @since 2019-09-23
+ */
+public interface ResourceFactory {
```
bbpennel: what would you all think of including a `public FedoraResource getResource(identifier)` method here. It would throw an error if a resource wasn't found rather than initializing a new object. The object returned would be of the appropriate subclass, but it would be returned as a a FedoraResource, so callers would have to cast it if they cared about the difference. This would be used for GET and HEAD cases

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
```diff
@@ -0,0 +1,75 @@
...
+    public Container findOrInitContainer(final String identifier, final String interactionModel);
+
+    /**
+     * Find or create a new binary. Newly created binaries will have nothing backing them until they are populated by a
+     * service.
+     *
+     * @param identifier The path or identifier for the resource.
+     * @return The binary object.
+     */
+    public FedoraBinary findOrInitBinary(final String identifier);
```
bbpennel: In order to figure out which binary class to initialize, you would need to also pass in `handling`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
```diff
@@ -0,0 +1,90 @@
...
+public class ResourceFactoryImpl implements ResourceFactory {
+
+    private final PersistentStorageSession psSession;
+
+    /**
+     * Special constructor for testing.
+     *
+     * @param psFactory Persistent storage session factory.
+     */
+    protected ResourceFactoryImpl(final PersistentStorageSessionFactory psFactory) {
```
bbpennel: how are we expecting psSession to make it into here if this is a test method?

My sense is that we need to pick one of two directions:

1.  ResourceFactory should be a singleton and live for the entire life of the application. In this case it would be passed the current fedora transaction per method call, like how FedoraSession was passed into ContainerService before. This approach would be most appropriate if the factory needs to inject common dependencies in to the constructed resources.
2. ResourceFactory would be scoped to a single transaction. Right now it is scoping to a persistence transaction, but it seems more appropriate to me that its API would expect to work against a FedoraTransaction since it is at the kernel level. This case seems more appropriate if the persistence level is able to inject all of the necessary dependencies into the constructed resources.

This current implementation seems to be assuming that the resources are ready to go after `psSession.read`. Are there any non-persistence level details that might need to be added? At the moment I can mainly think of potentially needing access to the PersistentStorageSessionFactory in order to retrieve triples/content. I wonder if there would be any need for ways to look up other resources by identifier, or pull from indexes?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
```diff
@@ -0,0 +1,90 @@
...
+        this.psSession = psFactory.getReadOnlySession();
+    }
+
+    @Override
+    public Container findOrInitContainer(final String identifier, final String interactionModel) {
+        final FedoraResource container = findInPersistence(identifier);
+        if (!container.getTypes().stream().anyMatch(t -> interactionModel.equalsIgnoreCase(t.toString()))) {
```
bbpennel: getTypes would return all of the types assigned to a resource, wouldn't it? So there would be user assigned types. Maybe we need a `getInteractionModel(s)` method?
In terms of interaction models, are there any cases where two interaction models can be on the same resource?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
whikloj: This is the crux of the discussion I and @dbernstein were having, does this factory return the fully initialized and populated object?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
whikloj: I am perhaps not clear on what we are expecting the PersistentStorageSession to return.

I was thinking _if_ the resource was found then the storage would return the fully formed FedoraResource. If it doesn't return the fully formed object how does another service know where to access the bits to make it fully formed? Because I would assume those "bits" would also be in the persistent storage.

But then if you are `PATCH`ing, you would send this fully formed FedoraResource to some other service to apply new triples or binary or whatever to it.

As for the singleton vs transaction, I am fine with passing in a transaction (`FedoraTransaction` I believe) and generating a readonly session to the persistent storage to retrieve any objects *if* they exist.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
whikloj: I am fine with this and added something like this as `private` to the `ResourceFactoryImpl` stub.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
whikloj: I may be unclear on what the definition of an interaction model is. The LDP spec makes limited use of it as well.

We can assign both DirectContainer and RDFSource "types" to an object, and seemingly these are both interaction models but really I think the DirectContainer is the only interaction model.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
whikloj: Also to be clear, each of the calls (ie. `findOrInitContainer()`) should have a `FedoraTransaction` as the first argument?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
bbpennel: I don't think it needs to have all the incoming data from the client set at this point, and I take back this suggestion. I'll just copy my response from slack for the sake of having it here:

> Looking at the implementation of FedoraBinary, I’m now thinking it’d be fine not to include the handling parameter. It doesn’t initialize the implementation of the binary until its interacted with. So it would become a UrlBinary when its external binary url gets set.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
dbernstein: +1

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
dbernstein: +1

PR #1551, opened 2019-09-24
Adds DeleteResourceService interface as a replacement for FedoraResou…

dbernstein:
…rce.delete().

**Refactors delete() method out of FedoraResource and into a service interface.**
* * *

**JIRA Ticket**: Resolves: https://jira.duraspace.org/browse/FCREPO-3062

# How should this be tested?
No tests.  No functional change.  Just replaces one interface method with another.
# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/DeleteResourceService.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/DeleteResourceService.java
@@ -1,0 +1,36 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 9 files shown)


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/DeleteResourceService.java
```diff
@@ -0,0 +1,34 @@
...
+ * @author dbernstein
+ * @since 6.0.0
+ */
+public interface DeleteResourceService {
+  /**
+   * Delete the specified resource
+   *
+   * @param resource the Fedora resource to delete
+   */
+  void perform(final FedoraResource resource);
```
bbpennel: Probably need to pass in the `Transaction` in which the resource is being deleted

dbernstein: Or would the Transaction be passed to a ServiceFactory that would return services interfaces for this ( as well as update and replace properties services)?

bbpennel: That could work as well. They would be very short lived, not sure they'd ever need to be reused, so might make sense to pass in the resource at the same time as constructing. I don't think there'd be much cost either way, 

`serviceFactory.createDeleteResourceService(tx, res).perform();`

versus

`deleteResourceService.perform(tx, res)`

While the factory could provide extra dependencies for the service, spring could also provide that in the case of a singleton service. So if there isn't a lot of state info to hold in the service (that would make it not thread safe), I probably prefer the non-factory approach. The only thing that I can think of right now in favor of the factory is fewer things to inject

dbernstein: Okay, that makes sense to me.  I'll update the PR and create a JIRA to update the  UpdatePropertiesService and ReplacePropertiesService.

dbernstein: https://jira.duraspace.org/browse/FCREPO-3091

PR #1553, opened 2019-09-24
Remove redundent "Fedora" from class names.

peichman-umd:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3059

Removes redundant "Fedora" from class and interface names.

# What's new?
Just name changes.

# How should this be tested?
`mvn clean install` should do it.

# Additional Notes:
I left FedoraSession as-is since it is being removed anyway.

# Interested parties
@whikloj, @fcrepo4/committers


whikloj:
I am okay with this PR as well.

While I am not as concerned about the generic names mostly because I felt that sticking `Fedora` in-front just meant we were using it in this codebase. Which was implicit. 

But to make it easier for new developers and those of us aging devs with limited memory. If we feel some of these classes could use a more clear/concise name, this would be the time to do it.

As this changes a lot of learned code, it would be good to get other @fcrepo4/committers to have a look and have a say.

dbernstein:
I'm comfortable with the PR as is.

bbpennel:
Once the tests are passing, I think we should try to get this merged since it impacts so many other PRs.

PR #1556, opened 2019-09-26
Implement OCFLPersistentStorageSessionManager.

dbernstein:
**Implement OCFLPersistentStorageSessionManager**
* * *

**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3080

This PR builds on the minimal implementation of the OCFLPersistentStorageSessionFactory in the following ways: 
1. Renames PersistentStorageSessionFactory to PersistentStorageSessionManager to reflect more accurately its function.   This class is responsible for accessing the state of the collection of persistent storage sessions and providing access to them.

2. Adds state to OCFLPersistentSessionManager to support a minimal
implementation.  Also adds check for null sessionId.

# How should this be tested?
mvn clean install

# Interested parties
@whikloj


whikloj:
If this a going to be a singleton class should it have a `public` constructor?

dbernstein:
@whikloj :  If we let spring manage the creation of the  OCFLPersistentStorageSessionManager (as I believe we will since that would follow an established pattern it does need a public interface.

PR #1557, opened 2019-09-26
FCREPO-3086 - Separate server and client managed triples

bbpennel:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3086

This is somewhat exploratory, still trying to get a sense of how we are going to shuttle the pieces of the object back and forth.

* Refactors getTriples into two methods
* Adds a stubbed out FedoraResourceImpl with triples related methods populated
* Adds read methods for binary content, triples, and server managed properties
* Refactors workflow for getting an existing resource to start to flesh out how it might work without prepopulating the object.

# What's new?
FedoraResourceImpl and some implementation related to it

# How should this be tested?
There aren't any tests yet, although if this approach looks reasonable it might be good to add a test for ResourceFactoryImpl to this PR.


fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -460,11 +460,13 @@ private Response getBinaryContent(final String rangeValue, final FedoraResource
         }
 
     private RdfStream getTriples(final FedoraResource resource, final Set<? extends TripleCategory> x) {
-        return resource.getTriples(translator(), x);
+        // TODO Reimplement filtering of triples by categories
+        return resource.getTriples();
```
whikloj: I am just getting here with `POST` and I am wondering if we don't want to pass in a translator instance to convert `info:fedora/the/ldp/path` to `http://localhost:8080/fcrepo/rest/the/ldp/path` in the RDFStream?

bbpennel: I was thinking this might happen as another step in the RDF stream just prior to returning to clients, versus occurring when getting the base triples. This might allow us to transform the server managed and client managed properties at together.

fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/PersistentStorageSession.java
```diff
@@ -86,6 +88,39 @@ public FedoraResource read(final String identifier) throws PersistentItemNotFoun
...
+     */
+    public List<String> getTypes(final String identifier) throws PersistentItemNotFoundException;
+
+    /**
+     * Read the client managed triples for the provided resource, and store them to that resource.
+     *
+     * @param resource the resource.
+     * @throws PersistentItemNotFoundException If the identifier doesn't exist.
+     */
+    public void readTriples(final FedoraResource resource) throws PersistentItemNotFoundException;
```
dbernstein: I'm feeling a little uneasy with the pattern being introduced here.   I'm open to being persuaded otherwise but it seems cleaner to have separate methods that describe the various use cases we need for reading out FedoraResource or optionally a variable set of parameters in the read() method for specifying which _representation_ or _view_ of the resource is desired.   My concern (which may be unfounded) is that we are starting to blur the ideas of FedoraResource  as a conceptual entity and the notion of it as a  _view_ of it such that at any particular place in the code we may find it difficult to reason about a particular instance of FedoraResource's state.  I think this warrants a face-to-face team discussion.

bbpennel: Would separate methods for the various use cases mean something like this? (names meant to be descriptive, not realistic):
`readWithTriplesAndServerManagedProperties(id)`
`readWithTriples(id)` (for Prefer omitting SMTs)
`readWithServerManagerProperties(id)`
`readWithBinaryContentAndServerManagedProperties(id)`
`readWithHeaders(id)` (for HEAD request)

or the second idea with 
`read(id, TRIPLES, SMTS)`
`read(id, TRIPLES)`
`read(id, SMTS)`
`read(id, BINARY, SMTS)`
`read(id)` (headers automatic in all reads)

I guess it seemed simpler to me to lazy load the parts of the resource that were necessary when they were requested, versus figuring out all the needed data up front. It would also accommodate multiple reads of any of the streams without needing to make a new FedoraResource. For a HEAD request ideally it wouldn't need to load any of the RDF files, so `readHeaders()` would probably defer to an index.

Worth mentioning that this PR doesn't address how headers would be loaded without populating the SMT rdf stream. So there might need to be something like `readHeaders()` added in here as well, or some other client (either against a index or the Simple Search) for getting individual properties of a resource (which might be nice to have regardless).

We can talk about it next week

dbernstein: I prefer the second suggestion assuming that  you would be returning the populated FedoraResource rather than passing in an unpopulated instance.

FedoraResource read(id, FedoraResourceOption... option);

dbernstein: Also I wonder if the persistentStorageSession is really what we want.  Something like this might be a more clear for the CUD methods: 

Also I'm wondering if we should consider an alternative interface for the CUD methods: 

CreateResourceRequest req = new CreateResourceRequest(id, incomingtriples, headers); 

CreateResourceResponse resp =  ps.create(req);

I'm not sure that returning the FedoraResource on create() is a good idea since FedoraResource can be in any number of different states depending on what "view" of it is required. Perhaps it is better to let the client specify exactly what he/she wants on a read (as you did above).

bbpennel: In my most recent commit, I shifted things a little:
https://github.com/fcrepo4/fcrepo4/pull/1557/commits/5bc950ccb69a05d08b032086ca17f3b5e6c7d7bc
Now when a resource is read it would always start with all of its header information. The various streamed content (triples, server managed triples, binary content) would be loaded on demand. For a newly created or updated resource, `setTriples` would have been called, and the client provided rdf stream would be returned by the FedoraResource instead of pulling it from the persistence layer.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/FedoraResourceImpl.java
```diff
@@ -0,0 +1,262 @@
...
+import org.fcrepo.persistence.api.PersistentStorageSessionFactory;
+import org.fcrepo.persistence.api.exceptions.PersistentItemNotFoundException;
+
+/**
+ * Implementation of a Fedora resource, containing functionality common to the more concrete resource implementations.
+ *
+ * @author bbpennel
+ */
+public class FedoraResourceImpl implements FedoraResource {
+    private PersistentStorageSessionFactory pSessionFactory;
```
whikloj: Is this here to lazy load some data?

bbpennel: Yes, see the implementations of getTriples, getManagedProperties. Also, in FedoraBinary, getContent would lazy load.
https://github.com/fcrepo4/fcrepo4/pull/1557/files#diff-b3f3902fc824f0ad14ce5c56fe6afca8R208

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/FedoraResourceImpl.java
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    @Override
+    public RdfStream getTriples() {
+        if (triplesStream == null) {
+            try {
+                pSessionFactory.getSession(tx.getId())
+                        .readTriples(this);
+            } catch (final PersistentItemNotFoundException e) {
+                throw new ItemNotFoundException("Unable to retrieve triples for " + getId(), e);
```
whikloj: Maybe this should react differently if you have not persisted the FedoraResource yet? Like if you called create and then called getTriples() then it probably shouldn't check the storage. 
```
/**
 * Is this a brand new resource?
 */
private boolean isNew = true;

public RdfStream getTriples() {
    if (triplesStream == null && !isNew) {
            try {
                this.triplesStream = pSessionFactory.getSession(tx.getId()).readTriples(this);
            } catch (final PersistentItemNotFoundException e) {
                throw new ItemNotFoundException("Unable to retrieve triples for " + getId(), e);
            }
    }
    return triplesStream;
}
```

bbpennel: Good call, that makes sense, I can add that in soon.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
whikloj: What was your thinking for the old `RdfContexts`? It seems like instead of `getTriples(*, PROPERTIES)` now we will `getTriples().filter(PROPERTIES)`. Should we do something to RdfStream to setup some filters in there? Or are we going to cache these RdfStreams or are they _only_ containing the requested information. 

These are just questions, not comments on your work here. 👍

bbpennel: The `RdfContext`s sort of did a combination of filtering out and generating new triples. My thinking was that that logic would be moved into separate services which would do similar stuff. These services would be implemented/called in the places where they are needed rather than inside the FedoraResource class.

For instance, there could be a `InboundReferencesRdfService` which concats the inbound refs as rdf to the triples rdf stream.

It may also still be reasonable to keep the RdfContexts structure and just externalize the gluing together of all of them, but we're working against a pretty different model than the node and property API from modeshape, so I don't know if it'll end up being practical.

whikloj: So this `resource.getTriples()` just returns the user's RDF and nothing else? Every other type of triple (containment, membership, reference, server managed) will be add (or not) to the stream via these services?

bbpennel: Yeah pretty much. Server managed triples get added in at the HTTP layer via a service that calls getManagedProperties (and probably generates some other properties that are not stored) and concats that onto the original Triples stream. I'm not sure if its a good thing or not for the rdfStream returned by getTriples() to be modified or if it should always only return the triples that were returned by persistence.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/AbstractRdfSourceOperation.java
```diff
@@ -0,0 +1,41 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.kernel.impl.operations;
+
+import org.fcrepo.kernel.api.RdfStream;
+import org.fcrepo.kernel.api.operations.RdfSourceOperation;
+
+
+/**
```
dbernstein: Class description missing?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/AbstractResourceOperation.java
```diff
@@ -0,0 +1,70 @@
...
+ * limitations under the License.
+ */
+package org.fcrepo.kernel.impl.operations;
+
+import java.time.Instant;
+
+import org.fcrepo.kernel.api.operations.ResourceOperation;
+
+
+/**
```
dbernstein: Class description missing?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateRdfSourceOperationBuilder.java
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+package org.fcrepo.kernel.impl.operations;
+
+import org.fcrepo.kernel.api.RdfStream;
+import org.fcrepo.kernel.api.operations.RdfSourceOperation;
+import org.fcrepo.kernel.api.operations.RdfSourceOperationBuilder;
+
+
+/**
```
dbernstein: Class description?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/UpdateRdfSourceOperationBuilder.java
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+package org.fcrepo.kernel.impl.operations;
+
+import org.fcrepo.kernel.api.RdfStream;
+import org.fcrepo.kernel.api.operations.RdfSourceOperation;
+import org.fcrepo.kernel.api.operations.RdfSourceOperationBuilder;
+
+
+/**
```
dbernstein: Class description missing?

whikloj:
Whew, once I see the 23 new interfaces/classes I'm less convinced this is going to be easier to understand then mutating FedoraResource and we don't have any Delete requests here yet. 🤷‍♂ 

Do we think some of these could be combined? Do we need specific Create/Update for RDFSource and NonRdfSource objects?

bbpennel:
@whikloj Filling out the operation/factory/builder structure isn't much fun, but its mostly a lot of boiler plate. There will likely be a lot of shared code. I wasn't sure if we needed any delete requests or not. I changed the `delete` method to take an id instead of a resource in the hopes that was all that was needed.

I ended up with so many operations mainly because RDF and Non-RDF sources have different data needs, and I wanted to the persistence layer to be able to easily tell what kind of operation/object it was dealing with. The builder/factory structure attempts to corral clients into not adding invalid data for an object type or operation. Relaxing things some would probably reduce the number of classes, so that's an option if people feel like it'd be better.

dbernstein:
@whikloj and @bbpennel :  Strong thumbs up on having separate classes for each operation.   Inheriting from the abstract will limit the amount of boiler plate and I concur that making the pairing of parameters with operation explicit will make writing tests easier.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
```diff
@@ -21,27 +21,41 @@
...
      */
+    @Deprecated
     String getPath();
 
+    /**
+     * Get an object containing header information describing this resource.
+     *
+     * @return header information
+     */
+    ResourceHeaders getHeaders();
```
bseeger: I wonder if we want to ack these as Headers this deep in the system?  That's more the API layer.  But maybe that's okay?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
```diff
@@ -109,47 +83,66 @@ public FedoraResource getResource(final String identifier)
...
-        try {
-            final PersistentStorageSession psSession = getSession(transaction);
-            return psSession.read(identifier);
-        } catch (final PersistentItemNotFoundException e) {
-            throw new PathNotFoundException(e);
-        } catch (final PersistentStorageException e) {
-            // This is a big error, wrap as RepositoryRuntime and send it through.
-            throw new RepositoryRuntimeException(e);
-        }
+        return createResource(transaction, identifier);
```
bseeger: Why does getResource() call createResource()?  Is it creating a new one or simply creating a new object from information found at the persistence layer?  Thanks!

dbernstein: @bseeger :  my understanding of the shift is that FedoraResources are now being built and populated above the persistence layer.  So I think the latter is @bbpennel 's intention.

dbernstein:
@bbpennel : One other comment:  I wonder if it would be better to have a DeleteOperation for the sake of consistency.  I can go either way there,  but just thought I would throw it out there.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
bbpennel: The terminology may be a little confusing. This method is creating a FedoraResource object, rather than persisting its contents. I could rename it to `instantiateResource` or something similar if that would be clearer.

bbpennel:
@dbernstein i don't have a strong opinion about it, I'll go ahead and add a delete operation

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
bseeger: or `fetchResource`?

bseeger: `instantiateResource` would probably work better then fetch.

bbpennel: https://github.com/fcrepo4/fcrepo4/pull/1557/commits/f14c3aad76ba972d2e7ce1aad5fc347749209726 has got you covered

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
dbernstein: I think it's okay for now.  I say let's move forward with the PR.

dbernstein: @bseeger :  the reason I say this is that the FedoraResource mediates between the persistence layer and the http api.  So it really is not as deep as you may be thinking.

bbpennel: @dbernstein they are coming out of the persistence layer in this PR:
https://github.com/fcrepo4/fcrepo4/pull/1557/files#diff-92d44c0c3bc6cfbd223e78ccc6151bacR62

Although I could see that coming from a kernel layer index client that doesn't currently exist.

PR #1555, opened 2019-09-26
Add new identifier translator

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3088

https://docs.google.com/document/d/1IvZK_VInC8HhtMVs98jJ3czInpwX4Ro8MWJd8hb3W40/edit

# What does this Pull Request do?
Adds a new Converter to translate a URI to an internal Fedora ID.

In the most basic sense `doForward()` takes the URI, grabs the path without the base context. Then strips off any with `fcr:*` or `#something` and prepends `info:fedora/`

`doReverse` removes the `info:fedora/` prefix and makes what is left as the path of the URI template.

Due to the removal of `fcr:*`, etc this is not necessarily an strictly invertable conversion. 

For example:
`http://localhost:8080/fcrepo/rest/my_thing` -> `info:fedora/my_thing`
and
`info:fedora/my_thing` -> `http://localhost:8080/fcrepo/rest/my_thing`

But
`http://localhost:8080/fcrepo/rest/my_thing/fcr:acl` -> `info:fedora/my_thing`
and
`info:fedora/my_thing` -> `http://localhost:8080/fcrepo/rest/my_thing`

# How should this be tested?

Not a functioning repository, but there are unit tests.

# Additional Notes:

* Does this change require documentation to be updated? no
* Does this change add any new dependencies? no
* Does this change require any other modifications to be made to the repository (ie. Regeneration activity, etc.)? no
* Could this change impact execution of existing code? no

# Interested parties
@fcrepo4/committers

PR #1560, opened 2019-09-27
First crack at POST

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3075

This relies on the below PR, once that has been finalized I can refactor/rebase this to work
Relies on #1555 

First steps at altering the `createObject` method in FedoraLdp to support some of the new workflow.

Then dealing with the consequences mostly of replacing HttpResourceConverter with HttpIdentifierConverter.

# What's new?

Mostly continuing or starting a transaction, and revising `createFedoraResource` to use the `ResourceFactory`

# How should this be tested?

1. Review
1. `mvn clean install`
1. ....
1. Profit

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -47,6 +47,8 @@ import static org.apache.jena.riot.WebContent.ctSPARQLUpdate;
 import static org.apache.jena.riot.WebContent.ctTextCSV;
 import static org.apache.jena.riot.WebContent.ctTextPlain;
 import static org.apache.jena.riot.WebContent.matchContentType;
+import static org.fcrepo.http.api.HttpConstants.ATOMIC_ID;
+import static org.fcrepo.http.api.HttpConstants.ATOMIC_START;
 import static org.fcrepo.http.commons.domain.RDFMediaType.JSON_LD;
 import static org.fcrepo.http.commons.domain.RDFMediaType.N3_ALT2_WITH_CHARSET;
 import static org.fcrepo.http.commons.domain.RDFMediaType.N3_WITH_CHARSET;
@@ -166,7 +168,7 @@ public class FedoraLdp extends ContentExposingResource {
     }
 
     /**
-     * Create a new FedoraNodes instance for a given path
+     * Create a new FedoraLdp instance for a given path
      * @param externalPath the external path
      */
     @VisibleForTesting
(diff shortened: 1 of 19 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -607,9 +608,11 @@ public Response updateSparql(final InputStream requestBodyStream)
     public Response createObject(@HeaderParam(CONTENT_DISPOSITION) final ContentDisposition contentDisposition,
                                  @HeaderParam(CONTENT_TYPE) final MediaType requestContentType,
                                  @HeaderParam("Slug") final String slug,
-            final InputStream requestBodyStream,
+        final InputStream requestBodyStream,
                                  @HeaderParam(LINK) final List<String> rawLinks,
-                                 @HeaderParam("Digest") final String digest)
+                                 @HeaderParam("Digest") final String digest,
+                                 @HeaderParam(ATOMIC_START) final String AtomicStart,
```
dbernstein: I believe the detection of the Atomic-Id, Atomic-Start, and Atomic-Commit will all be handled at a higher level in the stack.  You should be able to access the current Transaction by calling FedoraBaseResource.session().getTransaction().  I have a PR in now that will add FedoraBaseResource.getTransaction() that effectively will do the same thing.  So all code related to resolving the transaction can go away in FedoraLdp.

whikloj: This is the Jersey endpoint that receives the request, no? Perhaps my understanding of JAX-RS is wrong but I thought this is where the request originates? Are you writing some sort of Interceptor?

bbpennel: This controller:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/Transactions.java#L48
has been used for starting and committing transactions in the past, so I suspect that is where it'll end up going forward. This is based off of the assumption that we be following Peter's current transaction API suggestions, which involves making a POST to `/rest/fcr:tx` to start the transaction:
https://github.com/fcrepo/fcrepo-specification-atomic-operations/pull/8/files#diff-eacf331f0ffc35d4b482f1d15a887d3bR120

So no header is needed to start a transaction. Only the Atomic-ID header would make it to FedoraLdp. Some of my earlier documentation from the sprint was using Atomic-Start based off of older versions of the spec before it was updated. Sorry for the confusion.

whikloj: Ohhhhhh okay. So I should remove the `Atomic-Start` header. Yes, I see that it has been removed, that was a mistake I was probably seeing the removed content of the atomic-batch-operations PR and still registering it.

PR #1561, opened 2019-09-27
Add getResource without transaction

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3090

Adds a method

# What's new?

# How should this be tested?

Review

# Interested parties
@fcrepo4/committers @mohideen

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
@@ -29,6 +29,16 @@ import org.fcrepo.kernel.api.exception.PathNotFoundException;
 public interface ResourceFactory {
 
     /**
+     * Get a resource from the persistence layer without using a transaction.
+     *
+     * @param identifier The path or identifier for the resource.
+     * @return The resource.
+     * @throws PathNotFoundException If the identifier cannot be found.
+     */
+    public FedoraResource getResource(final String identifier)
+            throws PathNotFoundException;
+
+    /**
      * Get a resource from the persistence layer.
      *
      * @param transaction The transaction this request is part of.
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
@@ -101,6 +101,12 @@ public class ResourceFactoryImpl implements ResourceFactory {
     }
 
     @Override
+    public FedoraResource getResource(final String identifier)
+            throws PathNotFoundException {
+        return getResource(null, identifier);
+    }
+
+    @Override
     public FedoraResource getResource(final Transaction transaction, final String identifier)
             throws PathNotFoundException {
         try {


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
```diff
@@ -28,6 +28,16 @@
...
 public interface ResourceFactory {
 
+    /**
+     * Get a resource from the persistence layer without using a transaction.
+     *
+     * @param identifier The path or identifier for the resource.
+     * @return The resource.
+     * @throws PathNotFoundException If the identifier cannot be found.
+     */
+    public FedoraResource getResource(final String identifier)
```
dbernstein: I believe you also need an _Instant version_ param, since we'll need this call for getting mementos. 

You could also just add an overloaded version of the above method with the added _Instant version_ param.   So I'm suggesting something like this: 

public FedoraResource getResource(final String identifier)
public FedoraResource getResource(final String identifier, Instant version)

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
```diff
@@ -100,6 +100,12 @@ public WebacAcl createAcl(final Transaction transaction, final String identifier
         return (WebacAcl) createResource(WebacAcl.class);
     }
 
+    @Override
+    public FedoraResource getResource(final String identifier)
+            throws PathNotFoundException {
+        return getResource(null, identifier);
```
dbernstein: I think it's cleaner for getResource(transaction, identifier) to throw an exception on null. Both methods will call into the PersistentStorageSession which also has separate methods for requesting transactional vs immutable state.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
dbernstein: above should be a change request.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
dbernstein: above should be a change request:

whikloj: We don't specify any exceptions being thrown in either the ResourceFactory or PersistentStorageManager.

The exception you are referencing is thrown by the OCFL implementation. The ResourceFactoryImpl is not aware of that specific implementation.

But regardless, this (ResourceFactory) is a separate API. It (IMHO) does not have to mimic the implementation that it is under it.

dbernstein: The exception I was thinking was a runtime IllegalArgumentException. 

I'm okay skipping the overloaded method since I don't believe we need support for it at this time (though we likely we need it later when we implement memento support).

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
dbernstein: I suppose this change isn't needed _at this time_ so we can skip.

whikloj:
Your choice, I think there are more moving pieces then I can keep track of
right now.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
whikloj: The question I have is why throw an Exception from the Persistence layer all the way to the HTTP layer? Wouldn't it make more sense to check that `Atomic-Id != null` at the top? 

Instead it seems to me (correct me if I'm wrong) you are proposing adding code to keep a clear path from HTTP -> Persistence to ensure the `null` gets passed down and then fails.

If an `Atomic-Id` derived from the header does not refer to a valid transaction, then you'd want to throw an Exception but again that should occur before you try to use the `Atomic-Id`. I'd say in some HTTP layer utility like [here](https://github.com/fcrepo4/fcrepo4/pull/1560/files#diff-3f238be750488e3730daa409a69174b6R116)

dbernstein: I think we're on the same page @whikloj.  I wasn't suggesting we throw an exception due to a missing transaction id in the persistence layer.

PR #1564, opened 2019-10-31
Updates PersistentStorageSession interface to reflect design discussi…

dbernstein:
…ons.
**JIRA Ticket**: No JIRA

A realization of my recommendations in https://docs.google.com/document/d/18rSFqqoxixmozZrgPKON59Ojpg5iOu4lsHDNMm23soY/edit#
which was a response to 
https://github.com/fcrepo4/fcrepo4/pull/1557/files/23e54a2030551fce54eec6a62a7d587ced04de80#r329231737
 

# How should this be tested?

Nothing to test - just skeletal code

# Interested parties

### DIFF
--- a/fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/CommitOption.java
+++ b/fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/CommitOption.java
@@ -1,0 +1,28 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.persistence.api;
+
+/**
+ * This option enum specifies how commits should be applied to the PersistentStorageSession.
+ *
+ * @author dbernstein
+ */
+public enum CommitOption {
+    NEW_VERSION,  //indicates that you want to commit changes to a new version
+    UNVERSIONED;  //indicates that you want to commit unversioned changes
+}
--- a/fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/CreateResourceRequest.java
+++ b/fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/CreateResourceRequest.java
@@ -1,0 +1,25 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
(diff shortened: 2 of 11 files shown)


fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/CreateResourceResponse.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.persistence.api;
+
+/**
+ * A class describes the result of creating a resource.
+ * @author dbernstein
+ */
+public class CreateResourceResponse {
```
whikloj: In the Google document you were returning the `FedoraResource` for all the various interactions, so I am wondering what these 3 new responses hold and how they are different from each other and from a GET request response?

fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/CreateResourceRequest.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.persistence.api;
+
+/**
+ * A class that holds data required for creating a resource.
+ * @author dbernstein
+ */
+public class CreateResourceRequest {
```
whikloj: Seems like the `CreateResourceRequest`, `DeleteResourceRequest` and `UpdateResourceRequest` could be a very thin wrapper around a base class as they will probably all need the information from the initial HTTP request. 

Which then leads me to ask if perhaps 3 separate classes is unnecessary and a `getOperation()` which could return `CREATE`, `UPDATE` or `DELETE` is sufficient. This class doesn't do any work it just shuttles information around, right?

whikloj:
Yeah, this is not what I thought we were discussing. In this code we seem to generate a "request" which we pass to the persistence layer where all the work happens and we get back a response which (I believe) we pass back to the end-user.

fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/CommitOption.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package org.fcrepo.persistence.api;
+
+/**
+ * This option enum specifies how commits should be applied to the PersistentStorageSession.
+ *
+ * @author dbernstein
+ */
+public enum CommitOption {
+    NEW_VERSION,  //indicates that you want to commit changes to a new version
```
bbpennel: My understanding was that `PersistentStorageSession.commit` committed changes for one to many objects, but it seems like whether changes were committed to a new version or unversioned would vary by object?

dbernstein: I would think that from the http layer's point of view they are either committing to the HEAD or they are wanting to create a new version.   The HTTP layer is not really aware of what OCFL objects are being touched, no?

fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/CreateResourceRequest.java
dbernstein: Right - I just thought it would be clearer to have a clearly defined set of parameters that go with each operation.  This PR is out of date, and Ben's latest changes supersede it but I think it still makes sense to have separate classes for each  operation (as is reflected in Ben's PR)

fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/CreateResourceResponse.java
dbernstein: I was thinking having a "response" object leaves open the door for returning other metadata related to the result - not just the FedoraResource.

dbernstein:
I've made some replies to the comments, but I'm going to close this PR for the sake of clarity since it has been superseded by @bbpennel 's latest updates to the other PR (https://github.com/fcrepo4/fcrepo4/pull/1557)

PR #1565, opened 2019-11-04
Implements DeleteResourceService

dbernstein:
**Implements DeleteResourceService**
* * *

**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3063

Implements DeleteResourceService and adds a unit test. One note:  since the http layer requires some checks on FedoraResource before passing it to the delete service,  I reverted back to passing the FedoraResource to the service rather than the fedora identifier.

# How should this be tested?

Covered by a unit test at this time.

# Interested parties


fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImplTest.java
```diff
@@ -0,0 +1,89 @@
...
+    @Mock
+    FedoraResource resource;
+
+    @InjectMocks
+    private DeleteResourceServiceImpl service;
+
+
+    @Before
+    public void setUp() {
+    }
```
whikloj: Is this necessary? Or if it is could you add 
```
// This method left intentionally blank.
```
please

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImplTest.java
```diff
@@ -0,0 +1,98 @@
...
+import org.fcrepo.kernel.api.Transaction;
+import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
+import org.fcrepo.kernel.api.models.FedoraResource;
+import org.fcrepo.kernel.api.operations.DeleteResourceOperationFactory;
+import org.fcrepo.kernel.impl.operations.DeleteResourceOperation;
+import org.fcrepo.kernel.impl.operations.DeleteResourceOperationBuilder;
+import org.fcrepo.persistence.api.PersistentStorageSession;
+import org.fcrepo.persistence.api.PersistentStorageSessionManager;
+import org.fcrepo.persistence.api.exceptions.PersistentStorageException;
+import org.junit.Before;
```
whikloj: I'm getting a checkstyle error because this is unused and needs to be removed.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
```diff
@@ -0,0 +1,56 @@
...
+public class DeleteResourceServiceImpl implements DeleteResourceService {
+
+    @Inject
+    private DeleteResourceOperationFactory factory;
+
+    @Inject
+    private PersistentStorageSessionManager psManager;
+
+    @Override
+    public void perform(final Transaction tx, final FedoraResource resource) {
```
whikloj: After our discussion this should receive the `final String fedoraId` from the HTTP layer and load the resource itself.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/api/ContainmentIndex.java
```diff
@@ -0,0 +1,40 @@
...
+ * @since 6.0.0
+ */
+public interface ContainmentIndex {
+
+    /**
+     * Return a stream of fedora identifiers contained by the specified fedora resource.
+     *
+     * @param tx The transaction.  If no transaction, null is okay.
+     * @param fedoraResource The containing fedora resource
+     * @return A stream of contained identifiers
```
whikloj: I know you haven't implemented this yet, but should we define the order here. Below you assume they are in greatest grandchild to child order (I think). Maybe we should state that here, so its clear when we implement.

Or the other way around, we can reverse it.

dbernstein: I am assuming that the index  returns identifiers that are directly contained by the specified object.  Thus the need for recursive calls to the index.

dbernstein: Maybe the index should allow for a `getAllDescents()` query?

dbernstein: My feeling is that we can optimize that later.

bseeger: Is this a `util`?  Or do you imagine this will store state?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/api/ContainmentIndex.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+package org.fcrepo.kernel.api;
+
+import java.util.stream.Stream;
+
+import org.fcrepo.kernel.api.models.FedoraResource;
+
+/**
+ * An interface defining the encapsulates containment relationships.
```
bseeger: The "encapsulates containment relationships" is kind of confusing.    Should it be `encapsulated`?  Or is this class a helper to determine what the containment relationships are for existing objects?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/api/ContainmentIndex.java
bseeger: (I'm wondering if this should be in a sub directory)

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+            });
+        }
+
+        //delete described resource if binary
+        if (fedoraResource instanceof Binary) {
+            delete(pSession, fedoraResource.getDescribedResource().getId());
+        }
+
+        //delete the acl if this is not the acl
+        if (!fedoraResource.isAcl()) {
```
bseeger: do we want to consider mementos as well, or is that for later?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/api/ContainmentIndex.java
dbernstein: This will remain as a placeholder until @peichman-umd moves the SearchIndex interface into place.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/api/ContainmentIndex.java
dbernstein: will fix.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
dbernstein: I think that's for later since we're not planning on deleting versions at the outset.

whikloj:
One last general question, does the service do the permission checking? I kind of assumed that when working in the `CreateResourceService` but I notice you didn't do that? Is that coming here later or being handled somewhere else?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+            LOGGER.debug(format("deleting of %s", fedoraResourceId));
+            final PersistentStorageSession pSession = this.psManager.getSession(tx.getId());
+            deleteRecursivelyFromBottom(tx, pSession, fedoraResource);
+        } catch (final PersistentStorageException ex) {
+            throw new RepositoryRuntimeException(format("failed to delete resource %s", fedoraResourceId), ex);
+        }
+    }
+
+
+    private void deleteRecursivelyFromBottom(final Transaction tx, final PersistentStorageSession pSession,
```
bbpennel: maybe `deleteDepthFirst`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class mediates delete operations between the kernel and persistent storage layers
+ *
+ * @author dbernstein
+ */
+public class DeleteResourceServiceImpl implements DeleteResourceService {
+
+    private Logger LOGGER = LoggerFactory.getLogger(DeleteResourceService.class);
```
bbpennel: `private static final Logger log`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+        final String fedoraId = fedoraResource.getId();
+
+        if (fedoraResource instanceof Container) {
+            final Stream<String> children = containmentIndex.getContainedBy(tx, fedoraResource);
+            children.forEach(childResourceId -> {
+                try {
+                    final FedoraResource res = resourceFactory.getResource(tx, childResourceId);
+                    deleteRecursivelyFromBottom(tx, pSession, res);
+                } catch (final PathNotFoundException ex) {
+                    LOGGER.warn(format("Path not found for %s: ignoring.", childResourceId));
```
bbpennel: I'm trying to think through what it would mean for the path to not be found. The index is out of sync? Or a concurrent operation already deleted the item? Trying to figure out the severity of this kind of failure.

Also, use the logger's formatting syntax, so `"Path not found for {}: ignoring.", childResourceId`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
bbpennel: If you delete a AG, does the whole OCFL object go away?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+
+        //delete described resource if binary
+        if (fedoraResource instanceof Binary) {
+            delete(pSession, fedoraResource.getDescribedResource().getId());
+        }
+
+        //delete the acl if this is not the acl
+        if (!fedoraResource.isAcl()) {
+            final FedoraResource acl = fedoraResource.getAcl();
+            if (fedoraResource.getAcl() != null) {
```
bbpennel: use the `acl` variable that was just declared instead of retrieving again. Same with the next line.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+                } catch (final PathNotFoundException ex) {
+                    LOGGER.warn(format("Path not found for %s: ignoring.", childResourceId));
+                } catch (final PersistentStorageException ex) {
+                    throw new RepositoryRuntimeException(format("failed to delete resource %s", fedoraId), ex);
+                }
+            });
+        }
+
+        //delete described resource if binary
+        if (fedoraResource instanceof Binary) {
```
bbpennel: this should probably be an `else if` of the previous if, since they can't both be true.

Also, I think this method would currently allow the binary description to be deleted independently, which we probably don't want to allow.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+
+        //delete the resource itself
+        delete(pSession, fedoraId);
+    }
+
+    private void delete(final PersistentStorageSession pSession, final String fedoraId)
+        throws PersistentStorageException {
+        LOGGER.debug(format("starting delete of %s", fedoraId));
```
bbpennel: use `{}` for formatting with the logger.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
```diff
@@ -0,0 +1,122 @@
...
+        final ResourceOperation deleteOp = deleteResourceFactory.deleteBuilder(fedoraId).build();
+        pSession.persist(deleteOp);
+        LOGGER.debug(format("deleted %s", fedoraId));
```
bbpennel: use `{}` for formatting with the logger.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImplTest.java
```diff
@@ -0,0 +1,101 @@
...
+    private ContainmentIndex containmentIndex;
+
+    @Mock
+    private PersistentStorageSessionManager psManager;
+
+    @Mock
+    private ResourceFactory resourceFactory;
+
+    @Mock
+    private DeleteResourceOperationFactory factory;
```
bbpennel: the factories and builders aren't going to have much life outside these services. I think it'd be reasonable to use concrete implementations of them in a unit test here

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImplTest.java
```diff
@@ -0,0 +1,101 @@
...
+    private DeleteResourceOperation operation;
+
+    @Mock
+    private Container container;
+
+
+    @InjectMocks
+    private DeleteResourceServiceImpl service;
+
+    private final String id = "test-resource";
```
bbpennel: constants should be in the form `private final static String ALL_CAPS`

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -343,6 +343,7 @@ private Response getMemento(final String datetimeHeader, final FedoraResource re
     @DELETE
     public Response deleteObject() {
         hasRestrictedPath(externalPath);
+
```
bbpennel: undo the whitespace change

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImplTest.java
```diff
@@ -0,0 +1,101 @@
...
+    private final String txId = "tx-1234";
+
+    @Test
+    public void testContainerDelete() throws Exception {
```
bbpennel: are more test cases going to be added later? We'll want to test deleting all the different types

whikloj:
Checked with @peichman-umd and this seems to be a servlet filter (my brain is much soup) so it would be done before this. So I am 👍

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
dbernstein: Re :
>  If you delete a AG, does the whole OCFL object go away?

Good question.  We will continue to support Tombstones no?  In that case, the OCFL Object would only go away once all the tombstone's were deleted.   

Also question:   

if `/my-ag`  is deleted, would we expect  `/my-ag/binary1/fcr:tombstone` to exist?
Then if we delete `/my-ag/fcr:tombstone` would we expect all tombstones of the children also to be deleted,  ie `/my-ag/binary1/fcr:tombstone`?

dbernstein: I'm going to add this question to https://wiki.duraspace.org/display/FF/Decisions+and+Open+Questions

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
dbernstein: Yes - not sure myself.  We could throw a runtime exception here.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
dbernstein: There is a check for the binary in the FedoraLdp.delete() method.  I could move it here.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
bbpennel: @dbernstein The current behavior in fcrepo5 is that a GET to the child resource of a deleted container tells you where the tombstone of the deleted parent is, rather than having tombstones for all the children.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
bbpennel: That would make sense to me

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImplTest.java
dbernstein: Yes.  though I could add some more now.

dbernstein:
> One last general question, does the service do the permission checking? I kind of assumed that when working in the `CreateResourceService` but I notice you didn't do that? Is that coming here later or being handled somewhere else?

I think you're right - it should be done.

dbernstein:
@whikloj :  I createdhttps://jira.duraspace.org/browse/FCREPO-3114  to a address a bug in Fedora 5.1.0 and  https://jira.duraspace.org/browse/FCREPO-3115  to ensure that we cover this use case in 6.0.0

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
```diff
@@ -0,0 +1,130 @@
...
+
+        if( fedoraResource instanceof NonRdfSourceDescription){
+            throw new RepositoryRuntimeException(
+                    format("A NonRdfSourceDescription cannot be deleted independently of the NonRDFSource:  %s",
+                            fedoraResourceId));
+        }
+
+
+        try {
+            log.debug(format("deleting of %s", fedoraResourceId));
```
bbpennel: switch to `{}` for formatting

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImplTest.java
bbpennel: This one could still be addressed

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImplTest.java
```diff
@@ -0,0 +1,161 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * DeleteResourceServiceTest
+ *
+ * @author dbernstein
+ */
+@RunWith(MockitoJUnitRunner.Strict.class)
+public class DeleteResourceServiceImplTest {
```
bbpennel: The only thing not being tested at the moment is recursive deletion of children. I'd understand if you'd prefer to put that off until you have final interface for getting children, but it's probably not going to be too different.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImplTest.java
dbernstein: Okay - done.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImplTest.java
dbernstein: Done.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
dbernstein: done.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+    private DeleteResourceOperationFactory deleteResourceFactory;
+
+    @Inject
+    private PersistentStorageSessionManager psManager;
+
+    @Override
+    public void perform(final Transaction tx, final FedoraResource fedoraResource) {
+        final String fedoraResourceId = fedoraResource.getId();
+
+        if( fedoraResource instanceof NonRdfSourceDescription){
```
bbpennel: the space missed :( should be `if (` and there should be a space after the closing `)`. Are you using a code formatting rules in your IDE? I know they exist for eclipse, not sure how intellij differs

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImplTest.java
```diff
@@ -0,0 +1,192 @@
...
+        verify(pSession).persist(captor.capture());
+        final DeleteResourceOperation containerOperation = captor.getValue();
+        assertEquals(resourceId, containerOperation.getResourceId());
+    }
+
+    @Test
+    public void testAclDelete() throws Exception {
+        when(acl.getId()).thenReturn(RESOURCE_ACL_ID);
+        when(acl.isAcl()).thenReturn(true);
+        final ArgumentCaptor<DeleteResourceOperation> aclCaptor = ArgumentCaptor.forClass(DeleteResourceOperation.class);
```
bbpennel: I can't remember if these were here before, but I'd suggest using the `@Captor` annotation since this is in all the tests. See:
https://www.baeldung.com/mockito-annotations#captor-annotation

dbernstein: done.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/DeleteResourceServiceImpl.java
dbernstein: done.

PR #1570, opened 2019-11-05
Adds ReplacePropertiesServiceImpl

bseeger:
First pass at adding ReplacePropertiesService Implementation.  
This adds the service and updates the pieces that were using some of the code that got pulled into this (from ContentExposingResource). 

https://jira.duraspace.org/browse/FCREPO-3060


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/UpdatePropertiesService.java
```diff
@@ -40,4 +40,5 @@
   void updateProperties(final FedoraResource fedoraResource,
                         final String sparqlUpdateStatement,
                         final RdfStream originalTriples) throws MalformedRdfException, AccessDeniedException;
+
```
bbpennel: should probably revert this since there aren't any actual changes to this file

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/ReplacePropertiesService.java
```diff
@@ -31,12 +31,14 @@
    * Replace the properties of this object with the properties from the given
    * model
    *
+   * @param tx the Transaction
    * @param fedoraResource the Fedora resource to update
-   * @param inputModel the input model
-   * @param originalTriples the original triples
+   * @param requestBodyStream the body of the request
+   * @param contentType the original triples
```
bbpennel: update this param description to match the field

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/ReplacePropertiesServiceImpl.java
```diff
@@ -0,0 +1,217 @@
...
+     * @throws MalformedRdfException in case rdf json cannot be parsed
+     */
+    private Model parseBodyAsModel(final InputStream requestBodyStream,
+            final String contentType, final FedoraResource resource) throws MalformedRdfException {
+        final Lang format = contentTypeToLang(contentType);
+
+        final Model inputModel;
+        try {
+            inputModel = createDefaultModel();
+            inputModel.read(requestBodyStream, "www.example.com", format.getName().toUpperCase());
```
bbpennel: don't think `"www.example.com"` is what we want here, i think we want the base to be the uri of the resource. Although this might be a problem since the the RDF will probably have the HTTP URI for the resource and the kernel won't know how to convert that to an internal fedora uri... I wonder if model parsing needs to happen in the http layer because of this.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/ReplacePropertiesServiceImplTest.java
```diff
@@ -0,0 +1,117 @@
...
+    private Transaction tx;
+
+    @Mock
+    private PersistentStorageSession pSession;
+
+    @Mock
+    private PersistentStorageSessionManager psManager;
+
+    @Mock
+    private RdfSourceOperationFactory factory;
```
bbpennel: this unit test could probably use real instances of the factories/operations, maybe draw the line at the persistence classes?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/ReplacePropertiesServiceImpl.java
whikloj: I moved this same method down a layer so we'll need to merge our uses.

As for the other thing, I am of the mind that in the HTTP layer we parse the body and do URI replacement for any URI in the domain of the repository. Then pass it along so at this level we might have `info:fedora/some/object` as the URI.
ie.
```
 protected Model parseBodyAsModel(final InputStream requestBodyStream,
                                   final String contentType, final String fedoraId) throws MalformedRdfException {
```

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/ReplacePropertiesServiceImplTest.java
whikloj: The implementation classes for these interfaces are just stubs returning `null`, so unless we want to expand the scope we probably need this work done in a separate ticket and then pulled into the tests.
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/RdfSourceOperationFactoryImpl.java

bbpennel: I think that is in scope for this ticket, the ticket is for the workflow of replacing properties rather than for the service specifically

bseeger:
@bbpennel @whikloj @dbernstein  - this is ready for review.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/ReplacePropertiesService.java
dbernstein: This has been fixed.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/UpdatePropertiesService.java
dbernstein: This has been fixed.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/ReplacePropertiesServiceImpl.java
dbernstein: Reworked.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/ReplacePropertiesServiceImplTest.java
dbernstein: Looks like RdfSourceOperationFactoryImpl was implemented, but it was not pulled into this test.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/UpdateRdfSourceOperationBuilder.java
```diff
@@ -19,33 +19,72 @@
...
+            final Model model = ModelFactory.createDefaultModel();
+            final Lang lang = contentTypeToLang(mimetype);
+            model.read(contentStream, this.resourceId, lang.getName().toUpperCase());
+            stream = fromModel(ResourceFactory.createResource(this.resourceId).asNode(), model);
+        } else {
+            stream = null;
+        }
+
+        tripleStream = stream;
+        return this;
```
whikloj: Just noting we both have a function (CreateRdfSourceOperationBuilder and here) that does this, perhaps we need an AbstractRdfSourceOperationBuilder to extend from?

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/operations/CreateRdfSourceOperationBuilderTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private RdfSourceOperationBuilder builder;
+
+    private Model model;
+
+    private String turtleGraph;
+
+    private RdfStream stream;
+
+    private final String resourceId = "info:fedora/test-subject";
```
bbpennel: a bunch of these value should be named as constants, so CAPS and static.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/AbstractServiceTest.java
```diff
@@ -0,0 +1,237 @@
...
+import java.io.InputStream;
+import java.util.List;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@RunWith(MockitoJUnitRunner.Silent.class)
+public class AbstractServiceTest {
+
+    private final List<String> stringTypesNotValid = Stream.of(MEMENTO_TYPE, RESOURCE.toString(),
```
bbpennel: use `Arrays.asList` instead of passing through a stream for this and the next variable. These should also be constants.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/AbstractServiceTest.java
```diff
@@ -0,0 +1,237 @@
...
+            FEDORA_RESOURCE.toString()).collect(Collectors.toList());
+
+    private final List<String> stringTypesValid = Stream.of(MEMENTO_TYPE, CONTAINER.toString(),
+            BASIC_CONTAINER.toString()).collect(Collectors.toList());
+
+    private final String defaultInteractionModel = DEFAULT_INTERACTION_MODEL.toString();
+
+    private final AbstractService service = new AbstractService();
```
bbpennel: `AbstractService` should be... `abstract` :) These tests are valuable, but they should be run against concrete implementations. You could move these test methods to test classes for specific services where they are applicable (for instance, interaction model stuff only really applies to creating resources). If there are any that apply to all services, then you could make this into a parent class for the other test classes, in which case you'll have to have an instantiation of the service available to these tests.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/AbstractServiceTest.java
```diff
@@ -0,0 +1,237 @@
...
+                "<http://example.org/some/location/image.tiff>; " +
+                        "rel=\"http://fedora.info/definitions/fcrepo#ExternalContent\"; " +
+                        "handling=\"proxy\"; type=\"image/tiff\"", "<http://example.org/some/otherlocation>; rel='acl'")
+                .collect(Collectors.toList());
+        service.checkAclLinkHeader(links);
+    }
+
+    @Test(expected = RequestWithAclLinkHeaderException.class)
+    public void testCheckAclLinkHeaderFailNoQ() throws Exception {
+        final List<String> links = Stream.of("<" + NON_RDF_SOURCE.toString() + ">; rel=\"type\"",
```
bbpennel: Please use Arrays.asList for all of these link construction operations

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImplTest.java
```diff
@@ -0,0 +1,264 @@
...
+    @Mock
+    private RdfSourceOperation operation;
+
+    @Mock
+    private ExternalContent extContent;
+
+    @InjectMocks
+    private CreateResourceServiceImpl createResourceService;
+
+    private final static String txId = "tx1234";
```
bbpennel: constant

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImplTest.java
```diff
@@ -0,0 +1,264 @@
...
+
+    private final static Collection<String> basicContainerTypes;
+
+    private final Model model = ModelFactory.createDefaultModel();
+
+    private final Collection<String> digests = Stream.of("urn:sha1:12345abced")
+            .collect(Collectors.toCollection(HashSet::new));
+
+    static {
+        nonRdfSourceTypes = Collections.singleton(NON_RDF_SOURCE.toString());
```
bbpennel: Should probably immediately assign these values rather than going through a static block

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImplTest.java
```diff
@@ -0,0 +1,264 @@
...
+
+    private final static String txId = "tx1234";
+
+    private final static Collection<String> nonRdfSourceTypes;
+
+    private final static Collection<String> basicContainerTypes;
+
+    private final Model model = ModelFactory.createDefaultModel();
+
+    private final Collection<String> digests = Stream.of("urn:sha1:12345abced")
```
bbpennel: Since the implementation doesn't matter here, I'd suggest replacing this with `Collections.singleton`

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImplTest.java
```diff
@@ -0,0 +1,264 @@
...
+    public void testSlugIsNullRdf() throws Exception {
+        final String fedoraId = UUID.randomUUID().toString();
+        when(psSession.getHeaders(fedoraId, null)).thenReturn(resourceHeaders);
+        when(resourceHeaders.getTypes()).thenReturn(basicContainerTypes);
+        createResourceService.perform(txId, fedoraId, null, null, null, model);
+        verify(psSession).persist(ArgumentMatchers.any(ResourceOperation.class));
+    }
+
+    @Test
+    public void testWithSlugExistsRdf() throws Exception {
```
bbpennel: Might make sense to verify that the resource operation persisted doesn't have an internal id with the slug in it. You can declare a ArgumentCaptor for the test class and use it to get the ResourceOperation persisted back. 
```
    @Captor
    private ArgumentCaptor<ResourceOperation> operationCaptor;
```

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImplTest.java
```diff
@@ -0,0 +1,264 @@
...
+        final String fedoraId = UUID.randomUUID().toString();
+        when(psSession.getHeaders(fedoraId, null)).thenReturn(resourceHeaders);
+        when(psSession.getHeaders(fedoraId + "/" + "testSlug", null)).thenReturn(resourceHeaders);
+        when(resourceHeaders.getTypes()).thenReturn(basicContainerTypes);
+        createResourceService.perform(txId, fedoraId, "testSlug", null, null, model);
+        verify(psSession).persist(ArgumentMatchers.any(ResourceOperation.class));
+    }
+
+    @Test
+    public void testWithSlugDoesntExistsRdf() throws Exception {
```
bbpennel: Should probably make sure the id created contains the slug

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImplTest.java
```diff
@@ -0,0 +1,264 @@
...
+        createResourceService.perform(txId, fedoraId, null, null, null, digests,
+                null, null);
+        verify(psSession).persist(ArgumentMatchers.any(ResourceOperation.class));
+    }
+
+    @Test
+    public void testWithSlugExistsBinary() throws Exception {
+        final String fedoraId = UUID.randomUUID().toString();
+        when(psSession.getHeaders(fedoraId, null)).thenReturn(resourceHeaders);
+        when(psSession.getHeaders(fedoraId + "/" + "testSlug", null)).thenReturn(resourceHeaders);
```
bbpennel: Might make sense to use the helper method added in this PR for joining these strings

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImplTest.java
```diff
@@ -0,0 +1,264 @@
...
+        final String fedoraId = UUID.randomUUID().toString();
+        when(psSession.getHeaders(fedoraId, null)).thenReturn(resourceHeaders);
+        when(resourceHeaders.getTypes()).thenReturn(basicContainerTypes);
+        createResourceService.perform(txId, fedoraId, null, null, null, digests,
+                null, null);
+        verify(psSession).persist(ArgumentMatchers.any(ResourceOperation.class));
+    }
+
+    @Test
+    public void testWithSlugExistsBinary() throws Exception {
```
bbpennel: Check the id of the created binary

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImplTest.java
```diff
@@ -0,0 +1,264 @@
...
+    private UniqueValueSupplier minter;
+
+    @Mock
+    private PersistentStorageSession psSession;
+
+    @Mock
+    private ResourceHeaders resourceHeaders;
+
+    @Mock
+    private RdfSourceOperationBuilder builder;
```
bbpennel: builder and operation are not being used or injected, you can remove them i think

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/ReplacePropertiesServiceImplTest.java
```diff
@@ -0,0 +1,125 @@
...
+    private Transaction tx;
+
+    @Mock
+    private PersistentStorageSession pSession;
+
+    @Mock
+    private PersistentStorageSessionManager psManager;
+
+    @Mock
+    FedoraResource resource;
```
bbpennel: scope

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/ReplacePropertiesServiceImplTest.java
```diff
@@ -0,0 +1,125 @@
...
+    public void setup() {
+        when(tx.getId()).thenReturn(TX_ID);
+        when(psManager.getSession(anyString())).thenReturn(pSession);
+        when(resource.getId()).thenReturn(FEDORA_ID);
+        when(factory.updateBuilder(eq(FEDORA_ID))).thenReturn(builder);
+        when(builder.build()).thenReturn(operation);
+    }
+
+    @Test
+    @Ignore
```
bbpennel: the only test in here is ignored. When would we try to implement it and/or other tests?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -402,105 +407,66 @@ public Response createOrReplaceObjectRdf(
...
 
-        final FedoraResource resource;
+        final String contentType = extContent == null ? requestContentType.toString() : extContent.getContentType();
 
-        final String path = toPath(translator(), externalPath);
+        final String interactionModel = checkInteractionModel(links);
 
-        final AcquiredLock lock = lockManager.lockForWrite(path, transaction, nodeService);
+        // TODO: check if it exists or if we're creating it.
+        if ("We exist in persistent storage already".isEmpty()) {
```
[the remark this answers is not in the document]
whikloj: Yeah, I wasn't sure where we were going to get this information (index or persistence). Once we have a service defined I can replace this.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ExternalContent.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.io.InputStream;
+
+/**
+ * Interface for the ExternalContent information class.
+ * @author bseeger
+ */
+public interface ExternalContent {
+
+    String PROXY = "proxy";
```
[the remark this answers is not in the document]
whikloj: 🤷‍♂ They were in there entirely own file before, but as they are part of ExternalContent I thought they could stay here. An enum might be a good idea to make checks more consistent? Not sure.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
```diff
@@ -0,0 +1,274 @@
...
+        }
+    }
+
+    /**
+     * Check that we don't try to provide an ACL Link header.
+     * @param links list of the link headers provided.
+     * @throws RequestWithAclLinkHeaderException If we provide an rel="acl" link header.
+     */
+    void checkAclLinkHeader(final List<String> links) throws RequestWithAclLinkHeaderException {
+        final Predicate matcher = Pattern.compile("rel=[\"']?acl[\"']?").asPredicate();
```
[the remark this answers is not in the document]
whikloj: I guess this check should happen in the HTTP layer, I'll leave it as is with a TODO to move it to FedoraLdp (I think)

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
```diff
@@ -0,0 +1,274 @@
...
+
+    /**
+     * This method does two things:
+     * - Throws an exception if an authorization has both accessTo and accessToClass
+     * - Adds a default accessTo target if an authorization has neither accessTo nor accessToClass
+     *
+     * @param fedoraId the fedora Id
+     * @param inputModel to be checked and updated
+     */
+    protected void ensureValidACLAuthorization(final String fedoraId, final Model inputModel) {
```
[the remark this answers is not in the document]
whikloj: Would it make sense to put this test in the HttpRdfService now as we are parsing the provided RDF there anyways?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
```diff
@@ -0,0 +1,274 @@
...
+        return new DefaultRdfStream(asNode(fedoraId), serverManagedProperties.stream());
+    }
+
+    /**
+     * Populate server managed properties.
+     * Override in subclasses to add additional triples, always call super() first;
+     */
+    void populateServerManagedTriples(final String fedoraId) {
+        final ZonedDateTime now = ZonedDateTime.now();
+        serverManagedProperties.add(new Triple(
```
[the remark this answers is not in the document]
whikloj: You want me to remove all this?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/UpdateRdfSourceOperationBuilder.java
bseeger: Oh, interesting. Okay. I just pushed a change to this PR to add an AbstractRdfSourceOperationBuilder to refactor this.  I'm looking at the rest of your comments before I touch more things.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ExternalContent.java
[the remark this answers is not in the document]
bbpennel: Hmm, maybe its not worth the trouble unless we find ourselves using these in a lot of places. It would be kind of nice/safer to be able to pass an enum around instead of a string for the handling though

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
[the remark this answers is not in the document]
bbpennel: Hmm, my understanding is that this is very specific to creating/updating WebACs? HttpRdfService is pretty well scoped to translating URIs in RDF bodies at the moment, so I think it's in a good place. To me at least it would make sense to have a service dedicated to updating/validating ACLs, even if it hits the persistence layer like any other RDF resource

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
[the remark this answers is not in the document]
bbpennel: Nah, its okay, we can resolve this once the SMT stuff gets updated

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -936,10 +912,10 @@ protected Response createUpdateResponse(final FedoraResource resource, final boo
         }
     }
 
-    protected static MediaType getSimpleContentType(final MediaType requestContentType) {
+    protected static String getSimpleContentType(final MediaType requestContentType) {
```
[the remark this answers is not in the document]
whikloj: MediaType is an HTTP construct and in many places we to `mediaType.toString()`. It seemed easier to just deal in strings unless we really needed to know that the type was `image` and the subType was `jpeg`.

Because left over code that expected MediaType I casted to try to keep those places running but they could probably be removed.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -408,99 +393,61 @@ public Response createOrReplaceObjectRdf(
...
-                    }
-
-                    if (requestContentType == null && emptyRequest) {
-                        throw new ClientErrorException("Resource Already Exists", CONFLICT);
-                    }
-                    throw new NotSupportedException("Invalid Content Type " + requestContentType);
-                }
-            } catch (final Exception e) {
-                checkForInsufficientStorageException(e, e);
+            if ("Resource Exists".isEmpty()) {
```
whikloj: I'm just realize that this pseudo code sounds wrong. Not sure what I was thinking but it meant the object does exist.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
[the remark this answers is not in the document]
bbpennel: Okay, I didn't think there were any other calls to this method other than the ones that go through MediaType.valueOf, but maybe more are expected?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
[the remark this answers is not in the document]
bbpennel: Line 127 and 128 still give me warnings `Predicate is a raw type. References to generic type Predicate<T> should be parameterized`. Changing from Predicate to var will resolve it.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImpl.java
```diff
@@ -71,12 +76,17 @@
...
                         final InputStream requestBody, final ExternalContent externalContent) {
         final PersistentStorageSession pSession = this.psManager.getSession(txId);
-        final String fullPath = doInternalPerform(pSession, fedoraId, slug, linkHeaders);
-        final Collection<URI> uriDigests = digest.stream().map(URI::create).collect(Collectors.toCollection(HashSet::new));
+        checkAclLinkHeader(linkHeaders);
+        checkParent(pSession, (isContained ? fedoraId : findExistingAncestor(fedoraId)));
+
+        final String fullPath = isContained ? getResourcePath(pSession, fedoraId, slug) : fedoraId;
+
+        final Collection<URI> uriDigests = (digest == null ?
```
bbpennel: this conditional is backwards

bbpennel: Also, you can use `Collections.emptySet()` instead of instantiating a new hashset in the null case, unless we are expecting to modify the set of digests (its possible we might need to add in the system generated checksum later, but we may still not want to change the user supplied set of digests)

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImpl.java
```diff
@@ -0,0 +1,229 @@
...
+    @Inject
+    private UniqueValueSupplier minter;
+
+    @Override
+    public void perform(final String txId, final String fedoraId, final String slug, final boolean isContained,
+                        final String contentType, final List<String> linkHeaders, final Collection<String> digest,
+                        final InputStream requestBody, final ExternalContent externalContent) {
+        final PersistentStorageSession pSession = this.psManager.getSession(txId);
+        checkAclLinkHeader(linkHeaders);
+        checkParent(pSession, (isContained ? fedoraId : findExistingAncestor(fedoraId)));
```
bbpennel: could you add some short comments explaining the isContained behaviors? There's a lot happening on these two lines, I was having a somedifficulty following the cases and the expected outcomes. Putting the result of this ternary into its own variable might help a little too.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImplTest.java
```diff
@@ -117,87 +133,146 @@ public void setUp() {
...
+        final String persistedId = operationCaptor.getValue().getResourceId();
+        assertNotEquals(fedoraId, persistedId);
+        assertNotEquals(childId, persistedId);
+        assertTrue(persistedId.startsWith(fedoraId));
+    }
+
+    @Test
+    public void testWithSlugDoesntExistsRdf() throws Exception {
+        final String fedoraId = UUID.randomUUID().toString();
+        final String childId = fedoraId + File.separator + "testSlug";
```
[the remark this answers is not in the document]
whikloj: Ok so to clarify are we specifying that our IDs will always use a backslash? We didn't say that before, but it does make sense.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImpl.java
whikloj: So `isContained` is a way to tell if the `fedoraId` is the parent of the eventual container or the final path. 
Basically 
```
if (httpRequestMethod.equals("POST")) {
   isContained = true;
else if (httpRequestMethod.equals("PUT")) {
  isContained = false;
}
```
Because when we get into the service we need to either treat `fedoraId` as the parent of the resource we are creating (POST, so check that its not a binary and that it does exist) or that `fedoraId` is the final path of the resource we are creating (PUT, so find the a containment parent from the path and check that it exists).

This might be changed once the containment index is resolved. But I wanted to make clear the difference between making a resource called `Slug` under `fedoraId` and making `fedoraId` (possibly under something).

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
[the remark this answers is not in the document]
whikloj: Once this PR is merged, I can have a look and see if this function is even needed anymore.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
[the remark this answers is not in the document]
bbpennel: I've created a follow up ticket to address this https://jira.lyrasis.org/browse/FCREPO-3134

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImpl.java
```diff
@@ -0,0 +1,213 @@
...
+        if (isPairTree) {
+            throw new CannotCreateResourceException("Objects cannot be created under pairtree nodes");
+        }
+
+        String finalSlug;
+        if (slug == null) {
+            finalSlug = this.minter.get();
+        } else {
+            try {
+                pSession.getHeaders(fedoraId + "/" + slug, null);
```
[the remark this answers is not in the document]
bbpennel: think this still needs to be addressed, or you can resolve this is you don't want to address it. Wasn't sure how to interpret the 👍

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/functions/FedoraUtils.java
```diff
@@ -15,19 +15,25 @@
...
-/**
- * @author bseeger
- */
-
-public final class ExternalContent {
-
-    public static final String PROXY = "proxy";
-
-    public static final String REDIRECT = "redirect";
+public final class FedoraUtils {
```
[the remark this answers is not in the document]
bbpennel: I still think this class should be renamed, `FedoraUtils` could apply to anything in fedora

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
```diff
@@ -0,0 +1,290 @@
...
+import org.fcrepo.kernel.api.exception.ACLAuthorizationConstraintViolationException;
+import org.fcrepo.kernel.api.exception.MalformedRdfException;
+import org.fcrepo.kernel.api.exception.RequestWithAclLinkHeaderException;
+import org.fcrepo.kernel.api.exception.ServerManagedPropertyException;
+import org.fcrepo.kernel.api.exception.ServerManagedTypeException;
+import org.fcrepo.kernel.api.rdf.DefaultRdfStream;
+
+import org.slf4j.Logger;
+
+public class AbstractService {
```
bbpennel: While we're here, lets make this class `abstract`

and add javadoc

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
```diff
@@ -0,0 +1,290 @@
...
+    /**
+     * Looks through an RdfStream for rdf:types in the LDP namespace or server managed predicates.
+     *
+     * @param model The RDF model.
+     */
+    protected void checkForSmtsLdpTypes(final Model model) {
+        final StmtIterator it = model.listStatements();
+        while (it.hasNext()) {
+            final Statement st = it.next();
+            if ((st.getPredicate().hasURI(RDF_TYPE) && st.getObject().isURIResource() &&
```
bbpennel: Sorry, I just noticed this. There is an existing RDF.type constant that you can use, and do an equality comparison here since it is a Property.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImpl.java
bseeger: fixing in next commit

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
bseeger: Will do

whikloj: If you make it abstract you'll need to define an abstract method (I think). Maybe this should be `ServiceCommons` or something 🤷‍♂

bbpennel: hmm, i don't think you have to do anything special. Doing it as a utility is a possibility, but it'd require some refactoring.

PR #1574, opened 2019-11-07
First pass at CreateResourceService and POST

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3105

Adds a CreateResourceService and partial implementation to handle a POST type create.

Also creates a utility class (could actually be a parent class if we want) with some helper functions.

# How should this be tested?

Review it...does it make sense?

# Interested parties
@fcrepo4/committers


bseeger:
I think you need to build this out as well:  https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateRdfSourceOperationBuilder.java

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/functions/resourceServiceFunctions.java
```diff
@@ -0,0 +1,90 @@
...
+
+import org.fcrepo.kernel.api.exception.RequestWithAclLinkHeaderException;
+import org.fcrepo.kernel.api.exception.ServerManagedTypeException;
+
+import javax.ws.rs.core.Link;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class resourceServiceFunctions {
```
bseeger: Should resource be lowercase?  Is that a java thing?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImpl.java
```diff
@@ -0,0 +1,142 @@
...
+
+import javax.inject.Inject;
+import javax.ws.rs.core.Link;
+
+import java.io.InputStream;
+import java.net.URI;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class CreateResourceServiceImpl implements CreateResourceService {
```
bbpennel: this is a very HTTP POST oriented approach, which is probably fine. If we ever have a need for creating resources outside of the http layer then we would probably need smaller services. It would be good to have a consistent approach across services as to whether or RDF and non-RDF sources happen in the same service.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/functions/resourceServiceFunctions.java
whikloj: Mistake on my part, this class is gone

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImpl.java
```diff
@@ -0,0 +1,142 @@
...
+        checkAclLinkHeader(linkHeaders);
+        final PersistentStorageSession pSession = this.psManager.getSession(txId);
+        final ResourceHeaders parent;
+        try {
+            parent = pSession.getHeaders(fedoraId, null);
+        } catch (PersistentItemNotFoundException exc) {
+            throw new ItemNotFoundException(String.format("Item %s was not found", fedoraId), exc);
+        }
+
+        // TODO: Verify permissions Write or Append permissions on parent.
```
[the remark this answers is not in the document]
whikloj: You are correct sir.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImpl.java
```diff
@@ -0,0 +1,142 @@
...
+
+        final boolean isParentBinary = parent.getTypes().stream().anyMatch(t -> t.equalsIgnoreCase(NON_RDF_SOURCE.toString()));
+        if (isParentBinary) {
+            // Binary is not a container, can't have children.
+            throw new CannotCreateResourceException("NonRdfSource resources cannot contain other resources");
+        }
+        // TODO: Will this type still be needed?
+        final boolean isPairTree = parent.getTypes().stream().anyMatch(t -> t.equalsIgnoreCase(FEDORA_PAIRTREE));
```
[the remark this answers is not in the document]
whikloj: I guess we haven't, I pulled this out of `FedoraLdp`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImpl.java
```diff
@@ -0,0 +1,142 @@
...
+
+        final ResourceOperation createOp;
+
+        if (interactionModel.equalsIgnoreCase(NON_RDF_SOURCE.toString())) {
+            // TODO: Implement the NonRdfSourceOperationFactory
+            createOp = null;
+        } else {
+            // TODO: Verify no server managed triples are in the requestBody.
+            createOp = rdfSourceOperationFactory.createBuilder(fullPath, interactionModel)
+                    .triples(requestBody, contentType).build();
```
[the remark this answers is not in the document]
whikloj: In my mind the requestBody here would already have been translated. No validation just converting external URIs to internal IDs. So we were not passing the converters around.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImpl.java
whikloj: This is what I meant by knowing the operation, if I pass in POST or PUT then I could make changes. But that would make this much larger. 

It might make more sense to have an actual `CreateResourceServiceWrapper` that takes the HTTP method and all the arguments and passes off to `CreateResourceServiceImpl` or `CreateOrReplaceResourceServiceImpl`

whikloj: Or I guess just calling the appropriate method in `FedoraLdp` makes more sense 🤦‍♂

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/functions/resourceServiceFunctions.java
```diff
@@ -0,0 +1,90 @@
...
+     * @param linkTypes Link headers with rel="type"
+     * @param isRdfContentType Is the Content-type a known RDF type?
+     * @param contentPresent Is there content present on the request body?
+     * @param isExternalContent Is there Link headers that define external content?
+     * @return The determined or default interaction model.
+     */
+    public static String determineInteractionModel(final List<String> linkTypes,
+                                                   final boolean isRdfContentType, final boolean contentPresent,
+                                                   final boolean isExternalContent) {
+        final String interactionModel = linkTypes == null ? null : linkTypes.stream().filter(INTERACTION_MODELS_FULL::contains).findFirst()
```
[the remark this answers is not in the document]
whikloj: Guess I could check and see what happens now.

whikloj:
This has be subsumed by commits in #1570

PR #1575, opened 2019-11-08
Adds HttpRdfService to help translate uris to fedora ids in the rdf.

bseeger:
HttpRdfService will translate from resource URI to internal resource ID

This is part of https://jira.duraspace.org/browse/FCREPO-3060 and is used by any of the code that needs the URIs to be translated into internal fedora IDs. 

**JIRA Ticket**: part of https://jira.duraspace.org/browse/FCREPO-3060

Adds a service at the HTTP Layer that will look at the InputStream rdf and return it as a Model or RdfStream, having translated the URIs into internal fedora ids in the data. 

# Interested parties
@whikloj @bbpennel

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -978,8 +978,8 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
                                              final MediaType contentType,
                                              final RdfStream resourceTriples) throws MalformedRdfException {
 
-        final Model inputModel = httpRdfService.bodyToInternalModel(resource, requestBodyStream, contentType,
-                                                                    httpIdentifierConverter);
+        final Model inputModel = httpRdfService.bodyToInternalModel(resource.getId(), requestBodyStream, contentType);
+
 
         ensureValidMemberRelation(inputModel);
 
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -44,7 +44,6 @@ import static javax.ws.rs.core.Variant.mediaTypes;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.apache.http.HttpStatus.SC_BAD_REQUEST;
 import static org.apache.jena.graph.NodeFactory.createURI;
-import static org.apache.jena.rdf.model.ModelFactory.createDefaultModel;
 import static org.apache.jena.rdf.model.ResourceFactory.createProperty;
 import static org.apache.jena.rdf.model.ResourceFactory.createResource;
 import static org.apache.jena.rdf.model.ResourceFactory.createStatement;
@@ -89,7 +88,6 @@ import static org.fcrepo.kernel.api.RequiredRdfContext.SERVER_MANAGED;
 import static org.fcrepo.kernel.api.services.VersionService.MEMENTO_RFC_1123_FORMATTER;
 import static org.slf4j.LoggerFactory.getLogger;
 
-import com.fasterxml.jackson.core.JsonParseException;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Splitter;
 import java.io.IOException;
@@ -126,7 +124,6 @@ import javax.ws.rs.core.Response;
 
 import org.apache.http.client.methods.HttpPatch;
 import org.apache.http.client.methods.HttpPut;
-import org.apache.jena.atlas.RuntimeIOException;
 import org.apache.jena.graph.Graph;
 import org.apache.jena.graph.Node;
 import org.apache.jena.graph.Triple;
@@ -134,9 +131,9 @@ import org.apache.jena.rdf.model.Model;
 import org.apache.jena.rdf.model.Property;
 import org.apache.jena.rdf.model.RDFNode;
 import org.apache.jena.rdf.model.Statement;
-import org.apache.jena.riot.Lang;
-import org.apache.jena.riot.RiotException;
+import org.fcrepo.http.api.services.HttpRdfService;
 import org.fcrepo.http.commons.api.HttpHeaderInjector;
+import org.fcrepo.http.commons.api.rdf.HttpIdentifierConverter;
 import org.fcrepo.http.commons.api.rdf.HttpTripleUtil;
 import org.fcrepo.http.commons.domain.MultiPrefer;
 import org.fcrepo.http.commons.domain.PreferTag;
@@ -241,6 +238,12 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
 
     @Inject
     protected UpdatePropertiesService updatePropertiesService;
+
+    @Inject
+    protected HttpIdentifierConverter httpIdentifierConverter;
+
+    @Inject
+    protected HttpRdfService httpRdfService;
 
     private static final Predicate<Triple> IS_MANAGED_TYPE = t -> t.getPredicate().equals(type.asNode()) &&
             isManagedNamespace.test(t.getObject().getNameSpace());
@@ -974,7 +977,9 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
                                              final InputStream requestBodyStream,
                                              final MediaType contentType,
                                              final RdfStream resourceTriples) throws MalformedRdfException {
-        final Model inputModel = parseBodyAsModel(requestBodyStream, contentType, resource);
+
+        final Model inputModel = httpRdfService.bodyToInternalModel(resource, requestBodyStream, contentType,
+                                                                    httpIdentifierConverter);
 
         ensureValidMemberRelation(inputModel);
 
@@ -983,34 +988,6 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
         replacePropertiesService.replaceProperties(resource, inputModel, resourceTriples);
     }
 
-    /**
-     * Parse the request body as a Model.
-     *
-     * @param requestBodyStream rdf request body
-     * @param contentType content type of body
-     * @param resource the fedora resource
-     * @return Model containing triples from request body
-     * @throws MalformedRdfException in case rdf json cannot be parsed
-     */
-    private Model parseBodyAsModel(final InputStream requestBodyStream,
-            final MediaType contentType, final FedoraResource resource) throws MalformedRdfException {
-        final Lang format = contentTypeToLang(contentType.toString());
-
-        final Model inputModel;
-        try {
-            inputModel = createDefaultModel();
-            inputModel.read(requestBodyStream, getUri(resource).toString(), format.getName().toUpperCase());
-            return inputModel;
-        } catch (final RiotException e) {
-            throw new BadRequestException("RDF was not parsable: " + e.getMessage(), e);
-
-        } catch (final RuntimeIOException e) {
-            if (e.getCause() instanceof JsonParseException) {
-                throw new MalformedRdfException(e.getCause());
-            }
-            throw new RepositoryRuntimeException(e);
-        }
-    }
 
     /**
      * This method throws an exception if the arg model contains a triple with 'ldp:hasMemberRelation' as a predicate
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
@@ -62,19 +62,19 @@ public class HttpRdfService {
     /**
      * Parse the request body to a Model, with the URI to Fedora ID translations done.
      *
-     * @param resourceId the ID of the Fedora resource
+     * @param extResourceId the external ID of the Fedora resource
      * @param stream the input stream containing the RDF
      * @param contentType the media type of the RDF
      * @return RdfStream containing triples from request body, with Fedora IDs in them
      * @throws MalformedRdfException in case rdf json cannot be parsed
      * @throws BadRequestException in the case where the RDF syntax is bad
      */
-    public Model bodyToInternalModel(final String resourceId, final InputStream stream,
+    public Model bodyToInternalModel(final String extResourceId, final InputStream stream,
                                      final MediaType contentType)
                                      throws RepositoryRuntimeException, BadRequestException {
 
-        final Model model = parseBodyAsModel(stream, contentType, resourceId);
-        final List<Statement> insertStatements = new ArrayList<Statement>();
+        final Model model = parseBodyAsModel(stream, contentType, extResourceId);
+        final List<Statement> insertStatements = new ArrayList<>();
         final StmtIterator stmtIterator = model.listStatements();
 
         while (stmtIterator.hasNext()) {
@@ -86,8 +86,9 @@ public class HttpRdfService {
             RDFNode obj = stmt.getObject();
             if (stmt.getObject().isResource()) {
                 final String objString = stmt.getObject().asResource().getURI();
-                obj = model.getResource(idTranslator.inExternalDomain(objString) ?
-                    idTranslator.toInternalId(objString) : objString);
+                if (idTranslator.inExternalDomain(objString)) {
+                    obj = model.getResource(idTranslator.toInternalId(objString));
+                }
             }
 
             if (!subj.equals(originalSubj) || !obj.equals(stmt.getObject())) {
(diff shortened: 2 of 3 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
```diff
@@ -0,0 +1,142 @@
...
+     * @throws MalformedRdfException in case rdf json cannot be parsed
+     * @throws BadRequestException in the case where the RDF syntax is bad
+     */
+    public Model bodyToInternalModel(final FedoraResource resource, final InputStream stream,
+                                     final MediaType contentType, final HttpIdentifierConverter idTranslator)
+                                     throws RepositoryRuntimeException, BadRequestException {
+
+        final Model model = parseBodyAsModel(stream, contentType, resource);
+
+        final StmtIterator stmtIterator = model.listStatements();
```
bbpennel: I was testing this some, and I'm finding that it's not behaving the way I remember it working before. It appears to iterate through the newly added items in the main look. So I think we're going to have to add the new statements to a list and insert them after the loop like you suggested yesterday. 

There were a couple other issues (only need to add/remove when there is a change, the object might not be a literal, etc) and I ended up finding I had to update it to figure out what feedback to give, so sorry if this is a bit much but here is a suggested updated implementation:
```
        final Model model = parseBodyAsModel(stream, contentType, resource);

        final var insertStatements = new ArrayList<Statement>();
        final StmtIterator stmtIterator = model.listStatements();
        while (stmtIterator.hasNext()) {
            final Statement stmt = stmtIterator.nextStatement();

            final String originalSubj = stmt.getSubject().getURI();
            final String subj = idTranslator.convert(originalSubj);

            var obj = stmt.getObject();
            if (stmt.getObject().isResource()) {
                final String objString = stmt.getObject().asResource().getURI();
                obj = model.getResource(idTranslator.convert(objString));
            }

            if (!subj.equals(originalSubj) || !obj.equals(stmt.getObject())) {
                insertStatements.add(new StatementImpl(model.getResource(subj),
                        stmt.getPredicate(),
                        obj));

                stmtIterator.remove();
            }
        }

        model.add(insertStatements);

        LOGGER.debug("Model: {}", model);
        return model;
```

fcrepo-http-api/src/test/java/org/fcrepo/http/api/services/HttpRdfServiceTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Mock
+    FedoraResource resource;
+
+    @InjectMocks
+    private HttpRdfService httpRdfService;
+
+    private final String resourceUri = "www.example.com/fedora/rest/resource1";
+    private final String systemId = "info:fedora/resource1";
+    private final String rdfString = "@prefix dc: <http://purl.org/dc/elements/1.1/> . \n" +
```
bbpennel: you probably already know this, but you'll want to test some triples that have objects that are URIs, including some that are in domain and some that aren't.

Also, all the `www.example.com` things should begin with `http://` so that are valid absolute uris, otherwise they will be evaluated as relative

bseeger: Yes - that's the second pass I want to make: ensure that the triples that need to get changed actually change.  I'll add the `http://` to URIs.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
```diff
@@ -0,0 +1,142 @@
...
+ * A service that will translate the resourceURI to fedora ID in the Rdf InputStream
+ *
+ * @author bseeger
+ * @author bbpennel
+ * @since 2019-11-07
+ */
+
+public class HttpRdfService {
+
+    private static final Logger LOGGER = getLogger(HttpRdfService.class);
```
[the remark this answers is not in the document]
bseeger: I agree.  I created this so we can talk about it some more as a group: https://jira.duraspace.org/browse/FCREPO-3122

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
bseeger: Thanks, Ben.  Putting this in.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
```diff
@@ -0,0 +1,142 @@
...
+    public static Model parseBodyAsModel(final InputStream requestBodyStream,
+                                         final MediaType contentType,
+                                         final FedoraResource resource) throws BadRequestException,
+                                         RepositoryRuntimeException {
+        final Lang format = contentTypeToLang(contentType.toString());
+
+        final Model inputModel;
+        try {
+            inputModel = createDefaultModel();
+            inputModel.read(requestBodyStream, resource.getId(), format.getName().toUpperCase());
```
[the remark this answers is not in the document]
bseeger: aka the externalPath var that's handed around in ContentExposingResource?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
```diff
@@ -0,0 +1,157 @@
...
+        while (stmtIterator.hasNext()) {
+            final Statement stmt = stmtIterator.nextStatement();
+            final String originalSubj = stmt.getSubject().getURI();
+            final String subj = idTranslator.inExternalDomain(originalSubj) ?
+                idTranslator.toInternalId(originalSubj) : originalSubj;
+
+            RDFNode obj = stmt.getObject();
+            if (stmt.getObject().isResource()) {
+                final String objString = stmt.getObject().asResource().getURI();
+                obj = model.getResource(idTranslator.inExternalDomain(objString) ?
```
bbpennel: if written without the ternary conditional you could probably save the system from having to do any work in the case of out of domain URIs:
```
if (idTranslator.inExternalDomain(objString)) {
  obj = model.getResource(objString);
}

File: fcrepo-http-api/src/test/java/org/fcrepo/http/api/services/HttpRdfServiceTest.java (hunk 2 above)
Comment (bseeger) [reply fcrepo-http-api/src/test/java/org/fcrepo/http/api/services/HttpRdfServiceTest.java]: I have tests for this now.

File: fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java [hunk 6]
```diff
@@ -0,0 +1,157 @@
...
+
+    private static final Logger log = getLogger(HttpRdfService.class);
+
+    @Inject
+    private HttpIdentifierConverter idTranslator;
+
+    /**
+     * Parse the request body to a Model, with the URI to Fedora ID translations done.
+     *
+     * @param resourceId the ID of the Fedora resource
```
bbpennel: might want to change the variable name to reflect that it must be an external uri representation of the resource identifier in all these methods, or at least the javadoc

fcrepo-http-api/src/test/java/org/fcrepo/http/api/services/HttpRdfServiceTest.java
```diff
@@ -0,0 +1,114 @@
...
+        verifyTriples(stream);
+    }
+
+    private void verifyTriples(final Model model)  {
+        verifyTriples(fromModel(model.getResource(fedoraId).asNode(),model));
+    }
+
+    private void verifyTriples(final RdfStream rdfStream) {
+
+        final List<Triple> triples = rdfStream.map(triple -> {
```
bbpennel: This is a bit of an unusual use of the streaming API (there's actually a `peek` method that you can use for doing stuff to items in the stream without consuming them). It might be easier/more readable to transform the stream into a `Model`. I think Jared may have been doing so in one of the tests he was working on. Then you can use nice methods like `model.contains` to check for entire triples at once

fcrepo-http-api/src/test/java/org/fcrepo/http/api/services/HttpRdfServiceTest.java
```diff
@@ -0,0 +1,114 @@
...
+            assertFalse(triple.getSubject().getURI().equals(fedoraUri));
+            assertFalse(triple.getObject().isURI() && triple.getObject().toString().equals(fedoraUri));
+            return triple;
+        })
+        .filter(triple -> triple.getObject().isURI() && triple.getObject().getURI().equals(externalUri))
+        .collect(Collectors.toList());
+
+        assertTrue(triples.size() == 1);
```
bbpennel: should use assertEquals for equality tests

fcrepo-http-api/src/test/java/org/fcrepo/http/api/services/HttpRdfServiceTest.java
bbpennel: oh i just realized you convert from a model to a stream. Might be easier to skip that step

fcrepo-http-api/src/test/java/org/fcrepo/http/api/services/HttpRdfServiceTest.java
```diff
@@ -0,0 +1,114 @@
...
+        when(idTranslator.toInternalId(fedoraUri)).thenReturn(fedoraId);
+        when(idTranslator.inExternalDomain(fedoraUri)).thenReturn(true);
+        when(idTranslator.inExternalDomain(externalUri)).thenReturn(false);
+
+        when(resource.getId()).thenReturn(fedoraId);
+
+        final Model model = httpRdfService.bodyToInternalModel(fedoraUri, requestBodyStream,
+            contentType);
+
+        assertTrue(model.listStatements().toList().size() > 0);
```
bbpennel: there's actually a `model.isEmpty()` method that might serve this purpose?

fcrepo-http-api/src/test/java/org/fcrepo/http/api/services/HttpRdfServiceTest.java
```diff
@@ -0,0 +1,114 @@
...
+    @Mock
+    HttpIdentifierConverter idTranslator;
+
+    @Mock
+    FedoraResource resource;
+
+    @InjectMocks
+    private HttpRdfService httpRdfService;
+
+    private final String fedoraUri = "http://www.example.com/fedora/rest/resource1";
```
bbpennel: These should probably all be constants, so static and ANGRY CAPS

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
bseeger: I have to translate it as well, but I think I get what you're saying here and will change it.

birkland:
I haven't had time to look too deeply in the code, but have a question:  

If somebody `POST`s an RDFSource with the triple `<http://example.org/fcrepo/rest/foo> <whatever:rel> "hello"`, when they do a `GET`, they'll get that triple back.

If somebody POSTs an RDFSource with the triple `<info:fedora/my:foo> <whatever:rel> "hello"`, will they get back  `<info:fedora/my:foo> <whatever:rel> "hello"`?

If the answer is "yes", I think everything is all good.  If the answer is "no, they'll get back `<http://example.org/fcrepo/rest/my:foo> <whatever:rel> "hello"`, I think that's a problem for fedora 3 folks who still want to deal with RDF that has `info:fedora` subjects/objects.

whikloj:
@birkland You will get back the full URI, so would it be okay to reject requests that POST with the internal identifier prefix already inside the RDF?

ie. POST `<http://myserver.com/fcrepo/rest/object1> <some:pred> "Whatever"`
translates to `<info:fedora/object1> <some:pred> "Whatever"`

POST `<info:fedora/object1> <some:pred> "Whatever"`
400 Bad Request - you can't use our prefix in your RDF.

whikloj:
Maybe the easiest is to switch our internal prefix to something new.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
whikloj: Also it doesn't appear like we are translating predicates, that should probably be done as well?

fcrepo-http-api/src/test/java/org/fcrepo/http/api/services/HttpRdfServiceTest.java
whikloj: I think @bseeger is doing this because the actual `bodyToInternalStream` is already outputting a stream so this test will be needed regardless, unless we want to convert the stream to a model in the test. I think it was @dbernstein in his Persister PR #1577 that did the `model.contains`

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
bbpennel: We haven't traditionally done translation of predicates, I think it part because they aren't references to Fedora resources. I'm okay with saving the system that extra work until we have a use case for it.

bbpennel:
@birkland like jared said, it would be converted. If this is an important sticking point to Fedora3 migrators, as Jared suggested, we could use a different prefix for Fedora6 resources for disambiguation purposes, `info:fedora6/`, etc. 

It also seems like we need to come to a decision as to whether or not subject/object references to Fedora objects are special, and the degree to which Fedora can take control of the RDF submitted to it.

birkland:
OK.  Just raising a flag.  The impression I got from fcrepo3 folks is that they don't want to have to buy into http URIs, and want to just want to be able to use what they had.  (i.e. if they had a triple store that indexed info URIs before, they'll have a triple store that indexes the same info URIs on fcrepo6).   But please check with the fcrepo3 stakeholders.

fcrepo-http-api/src/test/java/org/fcrepo/http/api/services/HttpRdfServiceTest.java
bseeger: Thanks!!  @whikloj - That's exactly why I did it.  I'll check out @dbernstein's PR.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -242,6 +239,12 @@
     @Inject
     protected UpdatePropertiesService updatePropertiesService;
 
+    @Inject
+    protected HttpIdentifierConverter httpIdentifierConverter;
```
whikloj: I'm not sure you can inject this, the reason the old one was generated via a function call is because we need to pass the current path into it. 
So if we called `translator()` in `FedoraLdp` the path is `@Path("/{path: .*}")` whereas if we call it in  `FedoraVersioning` the path is `@Path("/{path: .*}/fcr:versions")`.

fcrepo-http-api/src/test/java/org/fcrepo/http/api/services/HttpRdfServiceTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+        assertTrue(stream.toString().length() > 0);
+        verifyTriples(stream);
+    }
+
+    private void verifyTriples(final Model model)  {
+
+        // make sure it changed the fedora uri to id, but didn't touch the non fedora domain
+        assertTrue(model.contains(model.createResource(FEDORA_ID),
+            model.createProperty(DCTerms.isPartOf.toString()),
```
bbpennel: DCTerms.isPartOf should already be a property, so don't need to convert it

bseeger: Ah, good point! Fixing

fcrepo-http-api/src/test/java/org/fcrepo/http/api/services/HttpRdfServiceTest.java
```diff
@@ -0,0 +1,126 @@
...
+            model.createResource(NON_FEDORA_URI)));
+
+        // make sure there are no triples that have the subject of the fedora uri
+        assertFalse(model.containsResource(model.createResource(FEDORA_URI)));
+
+        // make sure it translated the fedora uri to fedora id for this triple
+        assertTrue(model.contains(model.createResource(FEDORA_ID),
+            model.createProperty(DCTerms.isPartOf.toString()),
```
bbpennel: same as above

fcrepo-http-api/src/test/java/org/fcrepo/http/api/services/HttpRdfServiceTest.java
bseeger: fixing in both spots, that is

fcrepo-http-api/src/test/java/org/fcrepo/http/api/services/HttpRdfServiceTest.java
```diff
@@ -0,0 +1,126 @@
...
+            CONTENT_TYPE);
+
+        assertTrue(stream.toString().length() > 0);
+        verifyTriples(stream);
+    }
+
+    private void verifyTriples(final Model model)  {
+
+        // make sure it changed the fedora uri to id, but didn't touch the non fedora domain
+        assertTrue(model.contains(model.createResource(FEDORA_ID),
```
bbpennel: while i'm pestering you, I guess it might be nice to initialize `var fedoraIdResc = model.createResource(FEDORA_URI);` and reuse it

PR #1576, opened 2019-11-11
FCREPO-3095 - Implement OCFL Object Session

bbpennel:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3095

Implements the core functionality involved in committing changes to an OCFL object.

# What's new?
* Implements all methods exception `prepare`
* Add unit tests for object session
* Adds additional methods to the object session interface: deleteObject, close
* Adds throws expectations

# How should this be tested?

There is a unit test for now.

# Additional Notes:
This PR does not address locking beyond synchronization to prevent simultaneous modifications to the same object session.

Also, if an object is deleted in a session, it will cause the OCFL object to be deleted at commit time. The same session can also create a new OCFL object with the same identifier.

# Interested parties
@dbernstein, @pwinckles, @fcrepo4/committers

### DIFF
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
@@ -33,6 +33,7 @@ import java.util.HashSet;
 import java.util.Set;
 import java.util.function.Consumer;
 
+import org.apache.commons.io.FileUtils;
 import org.fcrepo.persistence.api.exceptions.PersistentItemNotFoundException;
 import org.fcrepo.persistence.api.exceptions.PersistentStorageException;
 import org.fcrepo.persistence.ocfl.api.CommitOption;
@@ -59,6 +60,8 @@ public class DefaultOCFLObjectSession implements OCFLObjectSession {
 
     private Set<String> deletePaths;
 
+    private boolean objectDeleted;
+
     private MutableOcflRepository ocflRepository;
 
     /**
@@ -74,6 +77,7 @@ public class DefaultOCFLObjectSession implements OCFLObjectSession {
         this.stagingPath = stagingPath;
         this.ocflRepository = ocflRepository;
         this.deletePaths = new HashSet<>();
+        this.objectDeleted = false;
     }
 
     /**
@@ -107,6 +111,7 @@ public class DefaultOCFLObjectSession implements OCFLObjectSession {
         final var stagedPath = resolveStagedPath(subpath);
         final var hasStagedChanges = hasStagedChanges(stagedPath);
 
+        // If the subpath exists in the staging path for this session, then delete from there
         if (hasStagedChanges) {
             // delete the file from the staging path
             try {
@@ -120,13 +125,41 @@ public class DefaultOCFLObjectSession implements OCFLObjectSession {
         if (!newInSession(subpath)) {
             deletePaths.add(subpath);
         } else if (!hasStagedChanges) {
-            // Doesn't exist in staging or head version, so file doesn't exist
-            throw new PersistentItemNotFoundException(format("Could not find %s within object %s",
-                    subpath, objectIdentifier));
+            // File is neither in the staged or exists in the head version, so file cannot be found for deletion
+            if (objectDeleted && isStagingEmpty()) {
+                // File can't be found because the object no longer exists
+                throw new PersistentItemNotFoundException(format(
+                        "Could not delete from object %s, it does not exist", objectIdentifier));
+            } else {
+                throw new PersistentItemNotFoundException(format("Could not find %s within object %s",
+                        subpath, objectIdentifier));
+            }
+        }
+    }
+
+    @Override
+    public void deleteObject() throws PersistentStorageException {
+        objectDeleted = true;
+        // Reset state of the object
+        if (!isStagingEmpty()) {
+            try {
+                FileUtils.cleanDirectory(stagingPath.toFile());
+            } catch (final IOException e) {
+                throw new PersistentStorageException("Unable to cleanup staging path", e);
+            }
+            deletePaths = new HashSet<>();
+        } else if (!ocflRepository.containsObject(objectIdentifier)) {
+            // fail if attempting to delete object that does not exist and has no staged changes
+            throw new PersistentItemNotFoundException(format(
+                    "Cannot delete object %s, it does not exist", objectIdentifier));
         }
     }
 
     private boolean newInSession(final String subpath) {
+        // If the object was deleted in this session, then content can only be new
+        if (objectDeleted) {
+            return true;
+        }
         // If the object isn't created yet, then there is no history for the subpath
         if (!ocflRepository.containsObject(objectIdentifier)) {
             return true;
@@ -151,14 +184,13 @@ public class DefaultOCFLObjectSession implements OCFLObjectSession {
                 throw new PersistentItemNotFoundException(format("Could not find %s within object %s",
                         subpath, objectIdentifier));
             }
-        } else {
+        } else if (!objectDeleted) {
             // Fall back to the head version
             return readVersion(subpath, ObjectVersionId.head(objectIdentifier));
         }
-    }
 
-    private boolean hasStagedChanges(final Path path) {
-        return path.toFile().exists();
+        throw new PersistentItemNotFoundException(format("Could not find %s within object %s",
+                subpath, objectIdentifier));
     }
 
     /**
@@ -166,6 +198,12 @@ public class DefaultOCFLObjectSession implements OCFLObjectSession {
      */
     @Override
     public InputStream read(final String subpath, final String version) throws PersistentItemNotFoundException {
+        // If the object was deleted, then only uncommitted staged files can be available
+        if (objectDeleted) {
+            throw new PersistentItemNotFoundException(format("Could not find %s within object %s",
+                    subpath, objectIdentifier));
+        }
+
         return readVersion(subpath, ObjectVersionId.version(objectIdentifier, version));
     }
 
@@ -198,10 +236,20 @@ public class DefaultOCFLObjectSession implements OCFLObjectSession {
      * {@inheritDoc}
      */
     @Override
-    public String commit(final CommitOption commitOption) {
+    public String commit(final CommitOption commitOption) throws PersistentStorageException {
         if (commitOption == null) {
             throw new IllegalArgumentException("Invalid commit option provided: " + commitOption);
         }
+        // Perform requested deletion of the object
+        if (objectDeleted) {
+            deleteExistingObject();
+            // no new state provided for the object, this is just committing the delete
+            if (isStagingEmpty()) {
+                return null;
+            }
+            // new state exists, object is being recreated after delete
+        }
+
         // Determine if a new object needs to be created
         if (isNewObject()) {
             return commitNewObject(commitOption);
@@ -210,8 +258,17 @@ public class DefaultOCFLObjectSession implements OCFLObjectSession {
         }
     }
 
-    private String commitNewObject(final CommitOption commitOption) {
+    private void deleteExistingObject() throws PersistentStorageException {
+        ocflRepository.purgeObject(objectIdentifier);
+    }
+
+    private String commitNewObject(final CommitOption commitOption) throws PersistentStorageException {
         final var commitInfo = new CommitInfo().setMessage("initial commit");
+
+        // Prevent creation of empty OCFL objects
+        if (isStagingEmpty()) {
+            throw new PersistentStorageException("Cannot create empty OCFL object " + objectIdentifier);
+        }
 
         if (NEW_VERSION.equals(commitOption)) {
             // perform commit to new version
@@ -232,7 +289,9 @@ public class DefaultOCFLObjectSession implements OCFLObjectSession {
     private String commitUpdates(final CommitOption commitOption) {
         // Updater which pushes all updated files and then performs queued deletes
         final Consumer<OcflObjectUpdater> commitChangeUpdater = updater -> {
-            updater.addPath(stagingPath, "", MOVE_SOURCE, OVERWRITE);
+            if (!isStagingEmpty()) {
+                updater.addPath(stagingPath, "", MOVE_SOURCE, OVERWRITE);
+            }
             deletePaths.forEach(updater::removeFile);
         };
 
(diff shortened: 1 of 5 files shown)


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
```diff
@@ -89,85 +115,267 @@ public void write(final String subpath, final InputStream stream) {
...
     /**
      * {@inheritDoc}
      */
     @Override
-    public InputStream read(final String subpath) {
-        if (hasStagedChanges(subpath)) {
-            // TODO read the staged version of the file
-        } else {
-            // TODO read the head version of the file from the ocfl object
+    public InputStream read(final String subpath) throws PersistentItemNotFoundException {
```
pwinckles: Do the `read` methods need session checks?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
```diff
@@ -89,85 +115,267 @@ public void write(final String subpath, final InputStream stream) {
...
     private boolean isNewObject() {
-        return false;
+        return !ocflRepository.containsObject(objectIdentifier);
     }
 
     private Path resolveStagedPath(final String subpath) {
         return stagingPath.resolve(subpath);
     }
+
+    private synchronized void assertSessionOpen() throws PersistentSessionClosedException {
```
pwinckles: I'm not sure this needs to be `synchronized`. Why is it a problem if it's called concurrently?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
```diff
@@ -89,85 +114,267 @@ public void write(final String subpath, final InputStream stream) {
...
+    public synchronized void prepare() {
+        // TODO check for conflicts and lock the object
     }
 
     /**
      * {@inheritDoc}
      */
     @Override
-    public void commit(final CommitOption commitOption) {
+    public synchronized String commit(final CommitOption commitOption) throws PersistentStorageException {
```
pwinckles: Is there any concurrency control above this class? If not, is it a problem if any of the methods on this class are executed concurrently? `synchronized` only helps prevent concurrent execution of the same method. If that doesn't matter here, that's fine, but, if it does matter, you'd probably need to use a lock.

peichman-umd:
Looks good to me so far.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSessionTest.java
```diff
@@ -0,0 +1,654 @@
...
+        assertNoMutableHead(OBJ_ID);
+        assertFileInHeadVersion(OBJ_ID, FILE1_SUBPATH, FILE_CONTENT1);
+        assertFileInHeadVersion(OBJ_ID, FILE2_SUBPATH, FILE_CONTENT2);
+        // Initial version is empty since created with mutable head
+        assertFileNotInVersion(OBJ_ID, "v1", FILE1_SUBPATH);
+        assertFileNotInVersion(OBJ_ID, "v1", FILE2_SUBPATH);
+    }
+
+    @Test(expected = PersistentSessionClosedException.class)
+    public void write_CommittedSession() throws Exception {
```
dbernstein: Just curious about the underscores in the method names...

dbernstein: I'm assuming those are just for clarity?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
bbpennel: @pwinckles My understanding (I had to remind myself about the specifics earlier when working on this) is that a synchronized method is the equivalent of wrapping the body of the method with `synchronized (this) { }`, which would cause blocking on calls to any other synchronized methods. 

My intent was to prevent simultaneous write operations to the same object/staging path, but to allow simultaneous reads to continue to happen. Perhaps that's not realistic though, since a file could get deleted/overwritten mid-read. If that's the case, then I should probably synchronize the read methods as well, or more precisely lock based on stagingPath and/or ocflRepository usage. Probably the latter. I was hoping to keep from having synchronize or locking blocks all over the place, but probably not avoidable.

I don't think a higher level will manage synchronized access to the whole class.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
```diff
@@ -61,25 +81,31 @@ public DefaultOCFLObjectSession(final String objectIdentifier, final Path stagin
...
      */
     @Override
-    public void write(final String subpath, final InputStream stream) {
-        // TODO write contents to the subpath within the staging path
+    public synchronized void write(final String subpath, final InputStream stream) throws PersistentStorageException {
+        assertSessionOpen();
 
         // Determine the staging path for the incoming content
-        resolveStagedPath(subpath);
+        final var stagedPath = resolveStagedPath(subpath);
```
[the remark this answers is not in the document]
bbpennel: I think it should have been generated and sanitized at a higher level, but we will need to make sure that happens in integration testing.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSessionTest.java
bbpennel: I can switch away from this naming scheme if it'd be preferred for consistency. Its to break up the method being tested from state and expectations. There don't seem to be particularly well defined best practices for naming unit tests in java and they can be pretty hard to read. I felt like this was more readable when I came across it while ago while looking for best practices.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
pwinckles: You're absolutely right about `synchronized`. Sorry!

Yes, there is a danger that a resource could be deleted while it's being read. However, using a lock to protect reads would be quite tricky because you're using InputStreams. If you want to guarantee a read of a resource in Fedora staging you'd have to hold the read lock until the stream is closed.

bbpennel: Huh, I got a kind of surprisingly result when testing to see what would happen if I deleted the file after opening an input stream. It continued to read the contents:
```
        final File dme = tempFolder.newFile("deleteme.txt");
        FileUtils.writeStringToFile(dme, "whatever");

        final InputStream inne = new FileInputStream(dme);
        inne.read(new byte[4]);
        dme.delete();
        System.out.println(dme.exists());
        System.out.print(IOUtils.toString(inne));
```
outputs:
```
false
ever
```
Moving the file produced the same result. So maybe I don't need to worry about the file going away? I get the same result with a 10000 character string. Either way you make a good point about the locking not doing much since i'm passing back an inputstream that would be consumed after the lock was released. My inclination is to leave the read methods unsynchronized. It might be a good idea to introduce a concurrency test at some point, although i'm not sure how to structure it at the moment for this class

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
```diff
@@ -89,85 +115,267 @@ public void write(final String subpath, final InputStream stream) {
...
+    private InputStream readVersion(final String subpath, final ObjectVersionId version) throws PersistentItemNotFoundException {
+        try {
+            // read the head version of the file from the ocfl object
+            final var file = ocflRepository.getObject(version)
+                    .getFile(subpath);
+            if (file == null) {
+                throw new PersistentItemNotFoundException(format("Could not find %s within object %s version %s",
+                        subpath, objectIdentifier, version.getVersionId()));
+            }
+            return file.getStream();
```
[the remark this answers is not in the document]
bbpennel: I will disable it for the time being, but that is potentially valuable to know in case its useful for performing fixity checks later. I am guessing it uses whatever algorithm's digest is in the inventory? I would suggest that the `on()` method return the stream rather than void though

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
```diff
@@ -89,85 +115,267 @@ public void write(final String subpath, final InputStream stream) {
...
+            }
+            deletePaths.forEach(updater::removeFile);
+        };
+
+        if (NEW_VERSION.equals(commitOption)) {
+            // check if a mutable head exists for this object
+            if (ocflRepository.hasStagedChanges(objectIdentifier)) {
+                // Persist the current changes to the mutable head, and then commit the head as a version
+                ocflRepository.stageChanges(ObjectVersionId.head(objectIdentifier),
+                        new CommitInfo(),
```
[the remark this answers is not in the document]
bbpennel: oh i didn't think of that, I'll switch

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
pwinckles: Interesting... Does the same thing happen when you do that to a file created outside the test? I wonder if Java is fudging anything. Like saying something is deleted when it's only queued for deletion or buffering file contents in memory.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
bbpennel: I was thinking it wouldn't do any harm to read from a closed session since it'd fall back to the head version, but its probably best to force consistency by disallowing session reuse. I'll change it

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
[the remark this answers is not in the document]
pwinckles: Yeah, I can do that. I'll add a new method and let you know when it's there. `on` is a method from the underlying MessageDigestInputStream.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
bbpennel: just tried creating the file outside of java and then doing the same operations (minus the write portion). Same result, says the file is deleted, outputs file contents, and the file is gone when i look afterwards. Maybe it reads the whole thing? or maybe it has a reference to the file, and the OS doesn't truly delete its contents until all the references are cleared.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
```diff
@@ -61,25 +81,31 @@ public DefaultOCFLObjectSession(final String objectIdentifier, final Path stagin
...
-        resolveStagedPath(subpath);
+        final var stagedPath = resolveStagedPath(subpath);
+        final var parentPath = stagedPath.getParent();
+
+        try {
+            // Fill in any missing parent directories
+            Files.createDirectories(parentPath);
+            // write contents to subpath within the staging path
+            Files.copy(stream, stagedPath, StandardCopyOption.REPLACE_EXISTING);
+            stream.close();
```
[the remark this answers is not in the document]
bbpennel: Nope. I've moved the close to a finally block

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
pwinckles: Cool!

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSessionTest.java
dbernstein: I think if it improves readability we should keep it.  As Emerson said "A foolish consistency is the hobgoblin of little minds."

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
```diff
@@ -61,25 +84,36 @@ public DefaultOCFLObjectSession(final String objectIdentifier, final Path stagin
...
+            Files.createDirectories(parentPath);
+            // write contents to subpath within the staging path
+            Files.copy(stream, stagedPath, StandardCopyOption.REPLACE_EXISTING);
+        } catch (final IOException e) {
+            throw new PersistentStorageException("Unable to persist content to " + stagedPath, e);
+        } finally {
+            try {
+                stream.close();
+            } catch (final IOException e) {
+                log.error("Failed to close inputstream while writing {}", subpath, e);
```
whikloj: Would this not swallow (while logging) the error and possibly leave the stream open? Could that end up being a memory leak?

whikloj: Also noting my lack of experience tracing memory leaks in the first place.

bbpennel: that could be the result, but we also can't close the stream, so I think we're kind of stuck with it.

PR #1578, opened 2019-11-12
Refactor to a more understandable converter syntax

whikloj:
**JIRA Ticket**: none

Stops using the Converter interface and creates more developer friendly method names to make it clear what you are doing with the converter.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpIdentifierConverter.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpIdentifierConverter.java
@@ -99,6 +99,19 @@ public class HttpIdentifierConverter {
     }
 
     /**
+     * Test if the provided external URI is in the domain of this repository.
+     *
+     * If it is not in the domain we can't convert it.
+     *
+     * @param httpUri the external URI.
+     * @return true if it is in domain.
+     */
+    public boolean inExternalDomain(final String httpUri) {
+        LOGGER.debug("Checking if http URI {} is in domain", httpUri);
+        return getPath(httpUri) != null;
+    }
+
+    /**
      * Convert an internal identifier to an external URI.
      *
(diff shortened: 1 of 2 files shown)


fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpIdentifierConverter.java
```diff
@@ -99,8 +98,13 @@ protected String doForward(final String httpUri) {
...
-    @Override
-    protected String doBackward(final String fedoraId) {
+    /**
+     * Convert an internal identifier to an external URI.
+     *
+     * @param fedoraId the internal identifier.
+     * @return the external URI.
+     */
+    protected String toExternalId(final String fedoraId) {
         LOGGER.debug(String.format("Translating Fedora ID %s to Http URI", fedoraId));
```
bbpennel: I know its not part of this PR, but it'd still be best to switch to using `{}` notation for this log statement and the other ones in this file (its a minor performance issue)

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/api/rdf/HttpIdentifierConverter.java
```diff
@@ -80,12 +78,13 @@ public HttpIdentifierConverter(final UriBuilder uriBuilder) {
...
-
-    @Override
-    protected String doForward(final String httpUri) {
+    /**
+     * Convert an external URI to an internal ID.
+     *
+     * @param httpUri the external URI.
+     * @return the internal identifier.
+     */
+    protected String toInternalId(final String httpUri) {
```
bbpennel: I think this has to be public, otherwise you won't be able to call it from outside the `http/api/rdf` package. Same with `toExternalId`

awoods:
Thanks for this, @whikloj !

PR #1577, opened 2019-11-11
Implements CreateRDFSourcePersister.persist().

dbernstein:
**Implements CreateRDFSourcePersister.persist()**
* * *

**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3094

# What does this Pull Request do?
It implements the CreateRDFSourcePersister.persist() method which mediates the interaction between the OCFLObjectSession and the ResourceOperation. It includes a unit test.

# Interested parties
Tag (@ mention) interested parties or, if unsure, @fcrepo4/committers

PR #1581, opened 2019-11-13
Allow persisters to handle more than one operation

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3098

Allows a persister to handle more than one operation type if it makes sense.

ie. When Updating RDF at the persistence layer, would we not just be writing it again? So it seems the work done by @dbernstein in #1577 would be re-used for the Update process.

# How should this be tested?

Tests are included.

# Additional Notes:

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractPersister.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractPersister.java
@@ -22,6 +22,8 @@ import org.fcrepo.kernel.api.operations.ResourceOperationType;
 import org.fcrepo.persistence.ocfl.api.Persister;
 
 import java.lang.reflect.ParameterizedType;
+import java.util.HashSet;
+import java.util.Set;
 
 /**
  * A base abstract persister class
@@ -31,10 +33,19 @@ import java.lang.reflect.ParameterizedType;
  */
 public abstract class AbstractPersister<T extends ResourceOperation> implements Persister<T> {
 
-    private ResourceOperationType resourceOperationType;
+    private final Set<ResourceOperationType> resourceOperationType;
+
+    AbstractPersister(final Set<ResourceOperationType> resourceOperationType) {
+        this.resourceOperationType = resourceOperationType;
+    }
 
     AbstractPersister(final ResourceOperationType resourceOperationType) {
-        this.resourceOperationType = resourceOperationType;
+        this();
+        this.resourceOperationType.add(resourceOperationType);
+    }
+
+    AbstractPersister() {
+        this.resourceOperationType = new HashSet<>();
     }
 
     @Override
@@ -46,10 +57,10 @@ public abstract class AbstractPersister<T extends ResourceOperation> implements 
         final var interfaces = operation.getClass().getInterfaces();
 
         //ensure that at least one of them match.
-        for (var i : interfaces) {
+        for (final var i : interfaces) {
             if (clazz.equals(i)) {
                 //return true if the operation types match.
-                return this.resourceOperationType.equals(operation.getType());
+                return this.resourceOperationType.contains(operation.getType());
             }
         }
         return false;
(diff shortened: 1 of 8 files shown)


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/CreateUpdateRDFSourcePersister.java
```diff
@@ -17,40 +17,51 @@
...
+import java.util.Set;
 
 /**
  * This class implements the persistence of a new RDFSource
  *
  * @author dbernstein
  * @since 6.0.0
  */
-public class CreateRDFSourcePersister extends AbstractPersister<RdfSourceOperation> {
+public class CreateUpdateRDFSourcePersister extends AbstractPersister<RdfSourceOperation> {
```
bbpennel: Would `RDFSourcePersister` be too minimal? I'm not sure if there are going to be other RDF source persisters... Seems like deleting would potentially cover all object types (although there would have to be special logic according to what the interaction model was).

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/CreateUpdateRDFSourcePersisterTest.java
```diff
@@ -17,6 +17,20 @@
  */
 package org.fcrepo.persistence.ocfl.impl;
 
+import static org.apache.jena.graph.NodeFactory.createLiteral;
```
bbpennel: Wonder if there's an option in intellij to not have it rearrange imports all the time?

whikloj: I adjusted my imports to try and match @dbernstein's setup, but I always have `static` imports at the top. I can manually revert these.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/CreateUpdateRDFSourcePersister.java
```diff
@@ -17,40 +17,51 @@
...
 
-    private static final Logger log = LoggerFactory.getLogger(CreateRDFSourcePersister.class);
+    private static final Logger log = LoggerFactory.getLogger(CreateUpdateRDFSourcePersister.class);
+
+    private static final Set<ResourceOperationType> OPERATION_TYPES = new HashSet<>();
```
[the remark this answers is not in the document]
whikloj: `var` seems to be only for local variables, so it is not allowed here. But I will use the `asList()` idea.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractPersister.java
```diff
@@ -17,39 +17,60 @@
...
-
         //retrieve the interfaces of the operation class
         final var interfaces = operation.getClass().getInterfaces();
 
         //ensure that at least one of them match.
-        for (var i : interfaces) {
-            if (clazz.equals(i)) {
-                //return true if the operation types match.
-                return this.resourceOperationType.equals(operation.getType());
+        for (final var i : interfaces) {
```
bbpennel: could this method be reduced to
```
@Override
    public boolean handle(final ResourceOperation operation) {
        //ensure that at least one of them match.
        for (final var c : this.resourceOperations) {
            if (c.isInstance(operation)) {
                //return true if the operation types match.
                return this.resourceOperationType.contains(operation.getType());
            }
        }
        return false;
    }
```

whikloj: yes, yes it can

PR #1583, opened 2019-11-13
Implement OCFLPersistentStorageSession.getTriples()

dbernstein:
**Implements Implement OCFLPersistentStorageSession.getTriples() and OCFLPersistentStorageSession.getManagedProperties() for unversioned content.**
* * *

**JIRA Ticket**:  https://jira.duraspace.org/browse/FCREPO-3119

This PR implements the OCFLPersistentStorageSession.getTriples() and OCFLPersistentStorageSession.getManagedProperties()  methods when the version parameter is null.  Included is a unit test that writes resources to disk and then reads them out  before and after the commit.
  
# What's new?
PersistentStorageSession.commit() now has a CommitOption param. 

# How should this be tested?
Covered by a unit test.

# Additional Notes:

Folks should be aware of the change to the PersistentStorageSession.commit() method as we'll need to translate the users intention to create a new version vs commit to unversioned at the close of a transaction (in the case that autoversioning is not enabled).
# Interested parties

### DIFF
--- a/fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/PersistentStorageSession.java
+++ b/fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/PersistentStorageSession.java
@@ -23,8 +23,6 @@ import java.time.Instant;
 import org.fcrepo.kernel.api.RdfStream;
 import org.fcrepo.kernel.api.models.ResourceHeaders;
 import org.fcrepo.kernel.api.operations.ResourceOperation;
-import org.fcrepo.persistence.api.exceptions.PersistentItemNotFoundException;
-import org.fcrepo.persistence.api.exceptions.PersistentSessionClosedException;
 import org.fcrepo.persistence.api.exceptions.PersistentStorageException;
 
 /**
@@ -58,8 +56,7 @@ public interface PersistentStorageSession {
      * @param version instant identifying the version of the resource to read from.
      *      If null, then the head version is used.
      * @return header information
-     * @throws PersistentItemNotFoundException If the identifier doesn't exist.
-     * @throws PersistentSessionClosedException If the session is closed.
+     * @throws PersistentStorageException  Either a PersistentItemNotFoundException or PersistentSessionClosedException
      */
     public ResourceHeaders getHeaders(final String identifier, final Instant version)
             throws PersistentStorageException;
@@ -71,10 +68,10 @@ public interface PersistentStorageSession {
      * @param version instant identifying the version of the resource to read from. If null, then the head version is
      *        used.
      * @return the triples as an RdfStream.
-     * @throws PersistentItemNotFoundException If the identifier doesn't exist.
+     * @throws PersistentStorageException  Either a PersistentItemNotFoundException or PersistentSessionClosedException
      */
     public RdfStream getTriples(final String identifier, final Instant version)
-            throws PersistentItemNotFoundException, PersistentStorageException;
+            throws PersistentStorageException;
 
     /**
      * Get the server managed properties for this provided resource.
@@ -83,12 +80,10 @@ public interface PersistentStorageSession {
      * @param version instant identifying the version of the resource to read from. If null, then the head version is
      *        used.
      * @return the server managed properties as an RdfStream.
-     * @throws PersistentItemNotFoundException If the identifier doesn't exist.
-     * @throws PersistentSessionClosedException If the session is closed.
-     * @throws PersistentStorageException Other issues.
+     * @throws PersistentStorageException  Either a PersistentItemNotFoundException or PersistentSessionClosedException
      */
     public RdfStream getManagedProperties(final String identifier, final Instant version)
-            throws PersistentItemNotFoundException, PersistentSessionClosedException, PersistentStorageException;
+            throws PersistentStorageException;
 
     /**
      * Get the persisted binary content for the provided resource.
@@ -97,10 +92,10 @@ public interface PersistentStorageSession {
      * @param version instant identifying the version of the resource to read from. If null, then the head version is
      *        used.
      * @return the binary content.
-     * @throws PersistentItemNotFoundException If the identifier doesn't exist.
+     * @throws PersistentStorageException  Either a PersistentItemNotFoundException or PersistentSessionClosedException
      */
     public InputStream getBinaryContent(final String identifier, final Instant version)
-            throws PersistentItemNotFoundException;
+            throws PersistentStorageException;
 
     /**
      * Commits any changes in the current sesssion to persistent storage.
--- a/fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/PersistentStorageSession.java
+++ b/fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/PersistentStorageSession.java
@@ -24,6 +24,7 @@ import org.fcrepo.kernel.api.RdfStream;
 import org.fcrepo.kernel.api.models.ResourceHeaders;
 import org.fcrepo.kernel.api.operations.ResourceOperation;
 import org.fcrepo.persistence.api.exceptions.PersistentItemNotFoundException;
+import org.fcrepo.persistence.api.exceptions.PersistentSessionClosedException;
 import org.fcrepo.persistence.api.exceptions.PersistentStorageException;
 
 /**
@@ -58,9 +59,10 @@ public interface PersistentStorageSession {
      *      If null, then the head version is used.
      * @return header information
      * @throws PersistentItemNotFoundException If the identifier doesn't exist.
+     * @throws PersistentSessionClosedException If the session is closed.
      */
     public ResourceHeaders getHeaders(final String identifier, final Instant version)
-            throws PersistentItemNotFoundException;
+            throws PersistentStorageException;
 
     /**
      * Get the client managed triples for the provided resource.
@@ -72,7 +74,7 @@ public interface PersistentStorageSession {
      * @throws PersistentItemNotFoundException If the identifier doesn't exist.
      */
     public RdfStream getTriples(final String identifier, final Instant version)
-            throws PersistentItemNotFoundException;
+            throws PersistentItemNotFoundException, PersistentStorageException;
 
     /**
      * Get the server managed properties for this provided resource.
@@ -82,9 +84,11 @@ public interface PersistentStorageSession {
      *        used.
      * @return the server managed properties as an RdfStream.
      * @throws PersistentItemNotFoundException If the identifier doesn't exist.
+     * @throws PersistentSessionClosedException If the session is closed.
+     * @throws PersistentStorageException Other issues.
      */
     public RdfStream getManagedProperties(final String identifier, final Instant version)
-            throws PersistentItemNotFoundException;
+            throws PersistentItemNotFoundException, PersistentSessionClosedException, PersistentStorageException;
 
     /**
      * Get the persisted binary content for the provided resource.
@@ -100,10 +104,10 @@ public interface PersistentStorageSession {
 
     /**
      * Commits any changes in the current sesssion to persistent storage.
-     *
+     * @param option The commit option indicating the
      * @throws PersistentStorageException Error during commit.
      */
-    public void commit() throws PersistentStorageException;
+    public void commit(final CommitOption option) throws PersistentStorageException;
 
     /**
      * Rolls back any changes in the current session.
--- a/fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/PersistentStorageSession.java
+++ b/fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/PersistentStorageSession.java
@@ -99,10 +99,9 @@ public interface PersistentStorageSession {
 
     /**
      * Commits any changes in the current sesssion to persistent storage.
-     * @param option The commit option indicating the
      * @throws PersistentStorageException Error during commit.
      */
-    public void commit(final CommitOption option) throws PersistentStorageException;
+    public void commit() throws PersistentStorageException;
 
     /**
      * Rolls back any changes in the current session.
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentSessionManager.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentSessionManager.java
@@ -17,12 +17,13 @@
  */
 package org.fcrepo.persistence.ocfl;
 
+import org.fcrepo.persistence.api.PersistentStorageSession;
+import org.fcrepo.persistence.api.PersistentStorageSessionManager;
+import org.fcrepo.persistence.ocfl.api.OCFLObjectSessionFactory;
+
 import javax.inject.Inject;
 import java.util.HashMap;
 import java.util.Map;
-
-import org.fcrepo.persistence.api.PersistentStorageSession;
-import org.fcrepo.persistence.api.PersistentStorageSessionManager;
 
 /**
  * OCFL implementation of PersistentStorageSessionManager
@@ -36,6 +37,9 @@ public class OCFLPersistentSessionManager implements PersistentStorageSessionMan
     private OCFLPersistentStorageSession readOnlySession;
 
     private Map<String,OCFLPersistentStorageSession> sessionMap;
+
+    @Inject
+    private OCFLObjectSessionFactory objectSessionFactory;
 
     @Inject
     private FedoraToOCFLObjectIndex fedoraOcflIndex;
@@ -60,7 +64,9 @@ public class OCFLPersistentSessionManager implements PersistentStorageSessionMan
             return session;
         }
 
-        final OCFLPersistentStorageSession newSession = new OCFLPersistentStorageSession(sessionId, fedoraOcflIndex);
+        final OCFLPersistentStorageSession newSession = new OCFLPersistentStorageSession(sessionId,
+                                                                                         fedoraOcflIndex,
+                                                                                         objectSessionFactory);
 
         sessionMap.put(sessionId, newSession);
 
@@ -70,7 +76,7 @@ public class OCFLPersistentSessionManager implements PersistentStorageSessionMan
     @Override
     public PersistentStorageSession getReadOnlySession() {
         if(this.readOnlySession == null) {
-            this.readOnlySession = new OCFLPersistentStorageSession(fedoraOcflIndex);
+            this.readOnlySession = new OCFLPersistentStorageSession(fedoraOcflIndex, objectSessionFactory);
         }
 
         return this.readOnlySession;
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentSessionManager.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentSessionManager.java
@@ -34,9 +34,9 @@ import java.util.Map;
  */
 public class OCFLPersistentSessionManager implements PersistentStorageSessionManager {
 
-    private OCFLPersistentStorageSession readOnlySession;
+    private PersistentStorageSession readOnlySession;
 
-    private Map<String,OCFLPersistentStorageSession> sessionMap;
+    private Map<String, PersistentStorageSession> sessionMap;
 
     @Inject
     private OCFLObjectSessionFactory objectSessionFactory;
@@ -54,32 +54,28 @@ public class OCFLPersistentSessionManager implements PersistentStorageSessionMan
     @Override
     public PersistentStorageSession getSession(final String sessionId) {
 
-        if(sessionId == null) {
+        if (sessionId == null) {
             throw new IllegalArgumentException("session id must be non-null");
         }
 
-        final OCFLPersistentStorageSession session = sessionMap.get(sessionId);
+        final PersistentStorageSession session = sessionMap.get(sessionId);
 
-        if(session != null) {
+        if (session != null) {
             return session;
         }
 
-        final OCFLPersistentStorageSession newSession = new OCFLPersistentStorageSession(sessionId,
-                                                                                         fedoraOcflIndex,
-                                                                                         objectSessionFactory);
-
+        final PersistentStorageSession newSession = new OCFLPersistentStorageSession(sessionId,
+                fedoraOcflIndex,
+                objectSessionFactory);
         sessionMap.put(sessionId, newSession);
-
         return newSession;
     }
 
     @Override
     public PersistentStorageSession getReadOnlySession() {
-        if(this.readOnlySession == null) {
+        if (this.readOnlySession == null) {
             this.readOnlySession = new OCFLPersistentStorageSession(fedoraOcflIndex, objectSessionFactory);
         }
-
         return this.readOnlySession;
     }
-
 }
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
@@ -230,7 +230,7 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
     }
 
     @Override
-    public synchronized void commit(final CommitOption option) throws PersistentStorageException {
+    public synchronized void commit() throws PersistentStorageException {
         ensureCommitNotStarted();
         if (isReadOnly()) {
             // No changes to commit.
@@ -265,6 +265,7 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
 
             //perform commit
             for (OCFLObjectSession objectSession : sessions) {
+                final CommitOption option = objectSession.getDefaultCommitOption();
                 objectSession.commit(option);
                 sessionsToRollback.add(new CommittedSession(objectSession, option));
             }
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
@@ -319,6 +319,8 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
                 if (cs.option == NEW_VERSION) {
                     //TODO rollback to previous OCFL version
                     //add any failure messages here.
+                    rollbackFailures.add(format("rollback of previously committed versions is not yet supported",
+                            cs.session));
                 } else {
                     rollbackFailures.add(format("%s was already committed to the unversioned head", cs.session));
                 }
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
@@ -79,7 +79,7 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
 
     private final static List<Persister> PERSISTER_LIST = new ArrayList<>();
 
-    private List<CommittedSession> sessionsToRollback = null;
+    private List<CommittedSession> sessionsToRollback = new ArrayList<>();
 
     private State state = State.COMMIT_NOT_STARTED;
 
@@ -105,7 +105,7 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
     private enum State {
         COMMIT_NOT_STARTED,
         COMMIT_STARTED,
-        PREPARE_FAILRED,
+        PREPARE_FAILED,
         COMMITTED,
         COMMIT_FAILED,
         ROLLING_BACK,
@@ -176,7 +176,7 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
 
     private void ensureCommitNotStarted() throws PersistentSessionClosedException {
         if (!state.equals(State.COMMIT_NOT_STARTED)) {
-            throw new PersistentSessionClosedException("The session cannot be committed: current state: " + state);
+            throw new PersistentSessionClosedException("The session cannot be committed in the  " + state + " state");
         }
     }
 
@@ -256,7 +256,7 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
 
             LOGGER.debug("Prepare succeeded.");
         } catch (final Exception e) {
-            this.state = State.PREPARE_FAILRED;
+            this.state = State.PREPARE_FAILED;
             throw new PersistentStorageException("Commit failed due to : " + e.getMessage(), e);
         }
 
@@ -268,7 +268,6 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
                 objectSession.commit(option);
                 sessionsToRollback.add(new CommittedSession(objectSession, option));
             }
-
 
             state = State.COMMITTED;
             LOGGER.info("Successfully committed {}", this);
@@ -285,12 +284,23 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
             return;
         }
 
-        if (!state.equals(State.COMMIT_FAILED) && !state.equals(State.PREPARE_FAILRED)) {
+        if (!(state.equals(State.COMMIT_FAILED) || state.equals(State.PREPARE_FAILED) ||
+                state.equals(State.COMMIT_NOT_STARTED))) {
             throw new PersistentStorageException("This session cannot be rolled back in this state: " + state);
         }
 
+        final boolean commitWasStarted = this.state != State.COMMIT_NOT_STARTED;
+
         this.state = State.ROLLING_BACK;
         LOGGER.info("rolling back...");
+
+        if (!commitWasStarted) {
+            //if the commit had not been started at the time this method was invoked
+            //we must ensure that all persist operations are complete before we close any
+            //ocfl object sessions. If the commit had been started then this synchronization step
+            //will have already occurred and is thus unnecessary.
+            this.phaser.arriveAndAwaitAdvance();
+        }
 
         //close any uncommitted sessions
         final List<OCFLObjectSession> committedSessions = this.sessionsToRollback.stream().map(c -> c.session).collect(Collectors.toList());
@@ -300,29 +310,31 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
             obj.close();
         }
 
-        // rollback committed sessions
-        //for each committed session, rollback if possible
-        final List<String> rollbackFailures = new ArrayList<>(this.sessionsToRollback.size());
-        for (CommittedSession cs : this.sessionsToRollback) {
-            if (cs.option == NEW_VERSION) {
-                //TODO rollback to previous OCFL version
-                //add any failure messages here.
-            } else {
-                rollbackFailures.add(format("%s was already committed to the unversioned head", cs.session));
+        if (commitWasStarted) {
+
+            // rollback committed sessions
+            //for each committed session, rollback if possible
+            final List<String> rollbackFailures = new ArrayList<>(this.sessionsToRollback.size());
+            for (CommittedSession cs : this.sessionsToRollback) {
+                if (cs.option == NEW_VERSION) {
+                    //TODO rollback to previous OCFL version
+                    //add any failure messages here.
+                } else {
+                    rollbackFailures.add(format("%s was already committed to the unversioned head", cs.session));
+                }
+            }
+            //throw an exception if any sessions could not be rolled back.
+            if (rollbackFailures.size() > 0) {
+                state = State.ROLLBACK_FAILED;
+                final StringBuilder builder = new StringBuilder();
+                builder.append("Unable to rollback successfully due to the following reasons: \n");
+                for (String failures : rollbackFailures) {
+                    builder.append("        " + failures + "\n");
+                }
+
+                throw new PersistentStorageException(builder.toString());
             }
         }
-        //throw an exception if any sessions could not be rolled back.
-        if (rollbackFailures.size() > 0) {
-            state = State.ROLLBACK_FAILED;
-            final StringBuilder builder = new StringBuilder();
-            builder.append("Unable to rollback successfully due to the following reasons: \n");
-            for (String failures : rollbackFailures) {
-                builder.append("\t\t" + failures + "\n");
-            }
-
-            throw new PersistentStorageException(builder.toString());
-        }
-
         this.state = State.ROLLED_BACK;
         LOGGER.info("rolled back successfully.");
 
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
@@ -21,18 +21,18 @@ import static java.lang.String.format;
 
 import java.io.InputStream;
 import java.time.Instant;
-import java.util.HashMap;
-import java.util.LinkedList;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
 import org.fcrepo.kernel.api.RdfStream;
 import org.fcrepo.kernel.api.models.ResourceHeaders;
 import org.fcrepo.kernel.api.operations.ResourceOperation;
+import org.fcrepo.persistence.api.CommitOption;
 import org.fcrepo.persistence.api.PersistentStorageSession;
 import org.fcrepo.persistence.api.exceptions.PersistentItemNotFoundException;
+import org.fcrepo.persistence.api.exceptions.PersistentSessionClosedException;
 import org.fcrepo.persistence.api.exceptions.PersistentStorageException;
-import org.fcrepo.persistence.api.CommitOption;
 import org.fcrepo.persistence.ocfl.api.OCFLObjectSession;
 import org.fcrepo.persistence.ocfl.api.OCFLObjectSessionFactory;
 import org.fcrepo.persistence.ocfl.api.Persister;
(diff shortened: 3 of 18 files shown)


fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
```diff
@@ -0,0 +1,151 @@
...
+        this.objectSessionFactory = new DefaultOCFLObjectSessionFactory(stagingDir, repoDir, workDir);
+        session = createSession(index, objectSessionFactory);
+
+    }
+
+    private OCFLPersistentStorageSession createSession(final FedoraToOCFLObjectIndex index,
+                                                       final OCFLObjectSessionFactory objectSessionFactory){
+       return new OCFLPersistentStorageSession(new Random().nextLong() + "", index, objectSessionFactory);
+    }
+
```
bbpennel: Two extra tests it might make sense to add now:
* test getting triples from a binary description
* test for operation that doesn't map to a persister

fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/PersistentStorageSession.java
```diff
@@ -82,9 +84,11 @@ public RdfStream getTriples(final String identifier, final Instant version)
      *        used.
      * @return the server managed properties as an RdfStream.
      * @throws PersistentItemNotFoundException If the identifier doesn't exist.
+     * @throws PersistentSessionClosedException If the session is closed.
+     * @throws PersistentStorageException Other issues.
      */
     public RdfStream getManagedProperties(final String identifier, final Instant version)
-            throws PersistentItemNotFoundException;
+            throws PersistentItemNotFoundException, PersistentSessionClosedException, PersistentStorageException;
```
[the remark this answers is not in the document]
dbernstein: What do you suggest?  Shall I catch the PersistentItemNotFoundException and throw the rest as a RepositoryRuntimeException?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPeristentStorageUtils.java
```diff
@@ -89,17 +93,76 @@ public static String relativizeSubpath(final String ancestorResourceId, final St
...
+
+            final var is = new ByteArrayInputStream(os.toByteArray());
+            session.write(subpath + getRDFFileExtension(), is);
             log.debug("wrote {} to {}", subpath, session);
         } catch (final IOException ex) {
             throw new PersistentStorageException(format("failed to write subpath %s in %s", subpath, session), ex);
         }
     }
+
+    private static InputStream readFile(final String version,  final OCFLObjectSession objSession, final String filename)
```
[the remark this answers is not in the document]
dbernstein: Sure.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPeristentStorageUtils.java
```diff
@@ -89,17 +93,76 @@ public static String relativizeSubpath(final String ancestorResourceId, final St
...
+     * @param identifier The resource identifier
+     * @param version The version.  If null, the head state will be returned.
+     * @param objSession The OCFL object session
+     * @param filePath The path to the desired file.
+     * @return
+     * @throws PersistentStorageException
+     */
+    public static RdfStream getRdfStream(final String identifier, final Instant version,
+                                         final OCFLObjectSession objSession,
+                                         final String filePath) throws PersistentStorageException {
```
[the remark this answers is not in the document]
dbernstein: Will do.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPeristentStorageUtils.java
```diff
@@ -89,17 +93,76 @@ public static String relativizeSubpath(final String ancestorResourceId, final St
...
+
+        final String versionNumber = resolveVersionId(objSession, version);
+        try (final InputStream is = readFile(null, objSession, filePath)) {
+            final Model model = createDefaultModel();
+            RDFDataMgr.read(model, is, DEFAULT_RDF_FORMAT.getLang());
+            return new DefaultRdfStream(createURI(identifier),
```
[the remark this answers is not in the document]
dbernstein: Great suggestion.  I didn't know that was there.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentSessionManager.java
```diff
@@ -60,7 +64,9 @@ public PersistentStorageSession getSession(final String sessionId) {
             return session;
         }
 
-        final OCFLPersistentStorageSession newSession = new OCFLPersistentStorageSession(sessionId, fedoraOcflIndex);
+        final OCFLPersistentStorageSession newSession = new OCFLPersistentStorageSession(sessionId,
+                                                                                         fedoraOcflIndex,
+                                                                                         objectSessionFactory);
 
         sessionMap.put(sessionId, newSession);
```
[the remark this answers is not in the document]
dbernstein: Yes, good idea,  will do.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentSessionManager.java
```diff
@@ -60,7 +64,9 @@ public PersistentStorageSession getSession(final String sessionId) {
             return session;
         }
 
-        final OCFLPersistentStorageSession newSession = new OCFLPersistentStorageSession(sessionId, fedoraOcflIndex);
+        final OCFLPersistentStorageSession newSession = new OCFLPersistentStorageSession(sessionId,
```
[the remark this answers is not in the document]
dbernstein: I'll be sure to reformat.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -67,25 +73,32 @@
...
      * @param fedoraOcflIndex the index
+     * @param objectSessionFactory the session factory
      */
-    protected OCFLPersistentStorageSession(final String sessionId, final FedoraToOCFLObjectIndex fedoraOcflIndex) {
+    protected OCFLPersistentStorageSession(final String sessionId, final FedoraToOCFLObjectIndex fedoraOcflIndex,
+                                           final OCFLObjectSessionFactory objectSessionFactory) {
         this.sessionId = sessionId;
         this.fedoraOcflIndex = fedoraOcflIndex;
+        this.objectSessionFactory = objectSessionFactory;
         this.sessionMap = new HashMap<>();
```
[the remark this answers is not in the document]
dbernstein: +1

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -97,33 +110,38 @@ public String getId() {
     public void persist(final ResourceOperation operation) throws PersistentStorageException {
         actionNeedsWrite();
 
-        //resolve the mapping between the fedora resource identifier and the associated OCFL object.
-        final FedoraOCFLMapping mapping = fedoraOcflIndex.getMapping(operation.getResourceId());
+        try {
+            mutex.lock();
```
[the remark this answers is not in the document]
dbernstein: Cool - I'll try that.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -97,33 +110,38 @@ public String getId() {
...
-        //resolve the persister based on the operation
-        final Persister persister = PERSISTER_LIST.stream().filter(p -> p.handle(operation)).findFirst().orElse(null);
+            //get the session.
+            final OCFLObjectSession objSession = findOrCreateSession(mapping.getOcflObjectId());
 
-        if (persister == null) {
-            throw new UnsupportedOperationException(format("The %s is not yet supported", operation.getClass()));
-        }
+            //resolve the persister based on the operation
+            final Persister persister = PERSISTER_LIST.stream().filter(p -> p.handle(operation)).findFirst().orElse(null);
```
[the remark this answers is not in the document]
dbernstein: +1

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -48,6 +53,8 @@
  */
 public class OCFLPersistentStorageSession implements PersistentStorageSession {
 
+    private static final String FEDORA_METADATA_SUFFIX = "/fcr:metadata";
```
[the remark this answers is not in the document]
dbernstein: +1

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -153,12 +189,37 @@ public InputStream getBinaryContent(final String identifier, final Instant versi
...
             return;
         }
-        // commit changes.
+
+        try {
+            mutex.lock();
+            //prepare session for commit
+            final Collection<OCFLObjectSession> sessions = this.sessionMap.values();
+
+            for(OCFLObjectSession objectSession : sessions){
```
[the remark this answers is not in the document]
dbernstein: +1

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -153,12 +189,37 @@ public InputStream getBinaryContent(final String identifier, final Instant versi
     }
 
     @Override
-    public void commit() throws PersistentStorageException {
+    public void commit(final CommitOption option) throws PersistentStorageException {
```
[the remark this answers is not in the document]
dbernstein: Do we want to implement per object versioning down the road?  I thought we were going to punt in the short term.  Either way,  I would suggest we create a separate ticket and address later. Unless you think that will cause more pain than it is worth.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -153,12 +189,37 @@ public InputStream getBinaryContent(final String identifier, final Instant versi
...
+            }
+
+            //perform commit
+            for(OCFLObjectSession objectSession : sessions){
+                objectSession.commit(option);
+            }
+
+            //close each session
+            for(OCFLObjectSession objectSession : sessions){
+                objectSession.close();
```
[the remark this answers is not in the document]
dbernstein: Okay - good to know.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -153,12 +189,37 @@ public InputStream getBinaryContent(final String identifier, final Instant versi
...
+            }
+
+            //purge object sessions
+            sessionMap.clear();
```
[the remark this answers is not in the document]
dbernstein: I wasn't sure about session reuse in the case of the OCFLPeristentStorageSession.  I guess we were thinking it would be associated with a transaction so I suppose not.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -153,12 +189,37 @@ public InputStream getBinaryContent(final String identifier, final Instant versi
...
+
+        } finally {
+            mutex.unlock();
```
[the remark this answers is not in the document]
dbernstein: The rollback scenario  for failing during prepare() as well as during object commits is fairly straightforward when committing new versions.  When committing to the mutable head,  I'm not sure how we rollback if a commit fails.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSessionFactory.java
```diff
@@ -0,0 +1,86 @@
...
+    public DefaultOCFLObjectSessionFactory(final File ocflStagingDir, final File ocflStorageRootDir,
+                                           final File ocflWorkDir) {
+        ocflStagingDir.mkdirs();
+        ocflStorageRootDir.mkdirs();
+        ocflWorkDir.mkdirs();
+
+        this.ocflStagingRoot = ocflStagingRoot;
+        this.ocflRepository = new OcflRepositoryBuilder().buildMutable(
+                new FileSystemOcflStorage(ocflStorageRootDir.toPath(),
+                        new ObjectIdPathMapperBuilder().buildFlatMapper()),
```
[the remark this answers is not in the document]
dbernstein: Sure.   Good catch.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
dbernstein: +1

fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/PersistentStorageSession.java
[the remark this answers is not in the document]
bbpennel: Could just throw PersistentStorageException. All the others are subclasses of it I believe. Just means the javadoc can't have separate `@throws` tags I believe. Can probably still call them out, just not with `@throws`

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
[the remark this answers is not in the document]
bbpennel: If we are punting, it would probably be easier to punt by having the interface line up with the eventual goal. So that could mean leaving this parameter out here and then passing a default commit option into all of the object sessions. So everything might commit to a new version or mutable head for now. Otherwise, the clients calling this will either have to try to figure out what to pass, in which we'll remove later, or they'll all have to pick a default, in which case it'd be easier to centralize that here.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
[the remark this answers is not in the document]
bbpennel: I don't think you have to address rollback in this PR, since its likely to be involved even in the "new version" scenario. I just wasn't sure where it was going to be trigger. Like would this method call the rollback itself on failure, or would whatever client was calling the commit catch the exception and trigger the rollback.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPeristentStorageUtils.java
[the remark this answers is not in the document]
whikloj: https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/DefaultRdfStream.java#L73

fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/PersistentStorageSession.java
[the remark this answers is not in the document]
dbernstein: 👍

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPeristentStorageUtils.java
```diff
@@ -89,17 +93,76 @@ public static String relativizeSubpath(final String ancestorResourceId, final St
...
+     * @param filePath The path to the desired file.
+     * @return
+     * @throws PersistentStorageException
+     */
+    public static RdfStream getRdfStream(final String identifier, final Instant version,
+                                         final OCFLObjectSession objSession,
+                                         final String filePath) throws PersistentStorageException {
+
+        final String versionNumber = resolveVersionId(objSession, version);
+        try (final InputStream is = readFile(null, objSession, filePath)) {
```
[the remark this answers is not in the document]
dbernstein: Yes.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -97,33 +110,38 @@ public String getId() {
...
 
+            if (persister == null) {
+                throw new UnsupportedOperationException(format("The %s is not yet supported", operation.getClass()));
+            }
 
-        //perform the operation
-        persister.persist(objSession, operation, mapping);
-    }
+            //perform the operation
+            persister.persist(objSession, operation, mapping);
```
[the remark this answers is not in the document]
dbernstein: The nice thing about it in the current form is that casting is not required.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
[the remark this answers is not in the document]
dbernstein: How will users of OCFLPersistentStorageSession indicate that they want to override the default behavior in a particular  invocation?  Namely, imagine the default commit behavior is UNVERSIONED, but the user desires to mint a version?  Do we need a separate method? 
commitVersion()?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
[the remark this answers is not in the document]
dbernstein: I'm going to take a swing at a partial rollback implementation.  I'm thinking if the commit fails, an exception will be thrown.  Then the user will call rollback().   Here's a JIRA for completing the rollback implementation: https://jira.duraspace.org/browse/FCREPO-3130

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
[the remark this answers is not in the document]
dbernstein: > bbpennel:waffle: 2:52 PM
@dbernstein for an object that does not have autoversioning turned on, you would issue a request to create a memento, which would probably go through a CreateVersionPersister or something like that, which would call objectSession.commit(NEW_VERSION);
Its a bit awkward since we aren’t allowing creating mementos as part of a transactions since it needs to immediately commit the object session, which is a bit unusual
i don’t know that there’s a use case or API means of telling something with autoversioning to not version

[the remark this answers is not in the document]
dbernstein: Hmmm.  looking back at this issue with committing versions I'm not sure I like it.
I'm going to hold off on changing that  for now in my PR.  I think it needs more discussion.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -152,12 +225,52 @@ public InputStream getBinaryContent(final String identifier, final Instant versi
...
+
+        try {
+            this.sessionsToRollback = new ArrayList<>(sessions.size());
+
+            //perform commit
+            for (OCFLObjectSession objectSession : sessions) {
+                objectSession.commit(option);
+                sessionsToRollback.add(new CommittedSession(objectSession, option));
+            }
+
```
bbpennel: i feel like andrew would want me to suggest removing this extra line

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -166,7 +279,48 @@ public void rollback() throws PersistentStorageException {
             // No changes to rollback
             return;
         }
-        // rollback changes
+
+        if (!state.equals(State.COMMIT_FAILED) && !state.equals(State.PREPARE_FAILRED)) {
```
bbpennel: Seems like you could "rollback" or at least abort a session in the `COMMIT_NOT_STARTED` state, it would only need to close sessions rather than rollback. Or are you thinking that a user requested transaction cancel would go through a different method?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -166,7 +279,48 @@ public void rollback() throws PersistentStorageException {
...
+            throw new PersistentStorageException("This session cannot be rolled back in this state: " + state);
+        }
+
+        this.state = State.ROLLING_BACK;
```
bbpennel: Hmm, I wonder if this should wait for all the `persist` calls to finish. That'll mainly only matter if this method is going to handle user requested transaction cancels. For post-commit failure situations, it probably isn't necessary since the commit will have already blocked to wait for them, so it'd be a programming error to reach this method without passing through commit first.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -166,7 +279,48 @@ public void rollback() throws PersistentStorageException {
...
+                rollbackFailures.add(format("%s was already committed to the unversioned head", cs.session));
+            }
+        }
+        //throw an exception if any sessions could not be rolled back.
+        if (rollbackFailures.size() > 0) {
+            state = State.ROLLBACK_FAILED;
+            final StringBuilder builder = new StringBuilder();
+            builder.append("Unable to rollback successfully due to the following reasons: \n");
+            for (String failures : rollbackFailures) {
+                builder.append("\t\t" + failures + "\n");
```
bbpennel: Wonder if we should use spaces? I don't see any other uses of tabs in the project, but i also don't know if we have any list structured exception bodies like this.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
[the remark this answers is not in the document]
bbpennel: So looks like the answer is that whatever called `commit` will need to call `rollback`. I am guessing the flow will be that if any exception is thrown during commit, rollback should be called? Just checking to make sure the `State` of the session doesn't need to be exposed.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
bbpennel: Since things have changed, we should probably test:
* persist fails if commit has happened
* getTriples fails if commit has started
* commit fails if already committed
* rollback test for uncommitted session
* rollback test for session with commits to mutable heads
* we should probably test for read only session behaviors at some point (persist fail, etc)

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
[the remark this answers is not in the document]
dbernstein: That's how I'm thinking about it.  No need to expose State.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
dbernstein: No - I think that makse sense. I'll change it.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
dbernstein: Yes. I think this makes sense.  Good catch.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
dbernstein: 👍

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
dbernstein: Will do all except the last: these behaviors are covered already by tests in OCFLPersistentSessionManagerTest.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -56,35 +72,66 @@
...
+        CommittedSession(final OCFLObjectSession session, final CommitOption option) {
+            this.session = session;
+            this.option = option;
+        }
+    }
+
+    private enum State {
+        COMMIT_NOT_STARTED,
+        COMMIT_STARTED,
+        PREPARE_FAILRED,
```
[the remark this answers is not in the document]
bbpennel: Please change to `PREPARE_FAILED`

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
```diff
@@ -195,4 +223,243 @@ public void getTriplesOfBinaryDescription() throws Exception {
...
+        //perform the create rdf operation
+        session.persist(rdfSourceOperation);
+
+        //commit to OCFL
+        session.commit(CommitOption.UNVERSIONED);
+
+        //this should fail
+        try {
+            session.persist(rdfSourceOperation);
```
bbpennel: Do `@Test(expected = PersistentStorageException.class)` instead of the catch

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
```diff
@@ -195,4 +223,243 @@ public void getTriplesOfBinaryDescription() throws Exception {
...
+
+        //perform the create rdf operation
+        session.persist(rdfSourceOperation);
+
+        //commit to OCFL
+        session.commit(CommitOption.UNVERSIONED);
+
+        //this should fail
+        try {
+            session.getTriples(resourceId, null);
```
bbpennel: Do `@Test(expected = PersistentStorageException.class)` instead of the catch

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
```diff
@@ -195,4 +223,243 @@ public void getTriplesOfBinaryDescription() throws Exception {
...
+
+        //perform the create rdf operation
+        session.persist(rdfSourceOperation);
+
+        //commit to OCFL
+        session.commit(CommitOption.UNVERSIONED);
+
+        //this should fail
+        try {
+            session.commit(CommitOption.UNVERSIONED);
```
bbpennel: Do `@Test(expected = PersistentStorageException.class)` instead of the catch

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
```diff
@@ -195,4 +223,243 @@ public void getTriplesOfBinaryDescription() throws Exception {
...
+        //perform the create rdf operation
+        session1.persist(rdfSourceOperation2);
+
+        //get triples should now fail because the session is effectively closed.
+        try {
+            session1.commit(CommitOption.UNVERSIONED);
+            fail("session1.commit(...) invocation should fail.");
+        } catch (PersistentStorageException ex) {
+            //attempted rollback should also fail:
+            try {
```
bbpennel: this doesn't need to be inside the catch block

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
```diff
@@ -195,4 +223,243 @@ public void getTriplesOfBinaryDescription() throws Exception {
...
+        });
+
+        final PersistentStorageSession session1 = createSession(index, mockSessionFactory);
+        //persist the two operations
+        session1.persist(rdfSourceOperation);
+
+        new Thread(() -> {
+            try {
+                session1.commit(CommitOption.UNVERSIONED);
+            } catch (PersistentStorageException e) {
```
bbpennel: If this commit fails that would be a problem wouldn't it? Maybe call `fail()` in the catch block or throw a runtime exception

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
```diff
@@ -195,4 +223,243 @@ public void getTriplesOfBinaryDescription() throws Exception {
...
+        }).start();
+
+        //get triples should now fail because the commit has started.
+        try {
+            //sleep for a fraction of a second to ensure
+            //commit thread runs first.
+            Thread.sleep(500);
+            session1.getTriples(resourceId, null);
+            fail("session1.getTriples(...) invocation should fail.");
+        } catch (Exception ex) {
```
bbpennel: I'd suggest adding an `(expected = PersistentStorageException.class)` to the test method and change the test method to `throws Exception`. At the moment this would pass if any exception were thrown, but we only want some exceptions to pass.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
```diff
@@ -195,4 +223,243 @@ public void getTriplesOfBinaryDescription() throws Exception {
...
+
+    @Test
+    public void getTriplesFailsIfCommitHasAlreadyStarted() throws PersistentStorageException {
+        when(mapping.getOcflObjectId()).thenReturn(OCFL_OBJECT_ID);
+        when(mapping.getParentFedoraResourceId()).thenReturn(parentId);
+        when(index.getMapping(resourceId)).thenReturn(mapping);
+
+        final Node resourceUri = createURI(resourceId);
+
+        when(rdfSourceOperation.getTriples()).thenReturn(new DefaultRdfStream(resourceUri, Stream.empty()));
```
bbpennel: These four statements are in most of the tests, sometimes more than once. Maybe you can break it out into a well named helper method that takes a rdfSourceOperation to help with readability/length?

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
```diff
@@ -195,4 +223,243 @@ public void getTriplesOfBinaryDescription() throws Exception {
...
+            } catch (final PersistentStorageException e) {
+                assertTrue("exception does not contain expected text",
+                        e.getMessage().contains("already committed to the unversioned "));
+            }
+        }
+    }
+
+    @Test
+    public void getTriplesFailsIfCommitHasAlreadyStarted() throws PersistentStorageException {
+        when(mapping.getOcflObjectId()).thenReturn(OCFL_OBJECT_ID);
```
bbpennel: these three lines are in almost every test. Could you break them out into a helper method with an informative name to help with readability/length?

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
```diff
@@ -195,4 +223,243 @@ public void getTriplesOfBinaryDescription() throws Exception {
...
+        try {
+            session.getTriples(resourceId, null);
+            fail("second session.getTriples(...) invocation should have failed.");
+        } catch (PersistentStorageException ex) {
+            //expected failure
+        }
+    }
+
+    @Test
+    public void commitFailsIfAlreadyCommitted() throws PersistentStorageException {
```
bbpennel: I don't think its a problem, but Mohamed's PR takes the approach committing a committed transaction does nothing. Same with rolling back a rolled back transaction. Just mentioning that in case you'd prefer consistency, but its probably fine for the persistence layer to be less forgiving.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
bbpennel: Okay. Also realizing we aren't testing that rollback fails if the state is not in one of the 3 happy states. Would be good to verify that

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
dbernstein: I didn't use `expected` because if for some reason the previous persist() or commit() invocations throw a PersistentStorageException,  we wouldn't want the test to pass.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
dbernstein: same as above.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
dbernstein: same as above.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
dbernstein: Are you sure?

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
dbernstein: Okay.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
dbernstein: Same as similar.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
dbernstein: +1

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
dbernstein: +1

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
dbernstein: We can revisit if it becomes a problem.  I think it's okay for now.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
bbpennel: Okay, that makes sense

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
bbpennel: In this case how about catching and ignoring PersistentStorageException, but letting other exceptions fail the test?

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSessionTest.java
bbpennel: Pretty sure, if no exception is thrown then it will fail first, and if any exception other than the expected one is thrown the the test will also fail. But its mainly cosmetic, so it doesn't matter too much

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentSessionManagerTest.java
```diff
@@ -64,6 +63,9 @@
     @Mock
     private FedoraOCFLMapping mapping;
 
+    @Mock
+    private OCFLObjectSessionFactory objectSessionFactory;
+
```
bbpennel: Sorry, I just noticed that `testResourcePath` is not being used by this test, so it should be removed.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSessionFactory.java
```diff
@@ -0,0 +1,86 @@
...
+     * @param ocflStorageRootDir The OCFL storage root directory
+     * @param ocflWorkDir        The OCFL work directory
+     */
+    public DefaultOCFLObjectSessionFactory(final File ocflStagingDir, final File ocflStorageRootDir,
+                                           final File ocflWorkDir) {
+        ocflStagingDir.mkdirs();
+        ocflStorageRootDir.mkdirs();
+        ocflWorkDir.mkdirs();
+
+        this.ocflStagingRoot = ocflStagingRoot;
```
bbpennel: this line is setting the variable to itself since the constructor isn't receiving an ocflStagingRoot. This is probably a bug since ocflStagingRoot is referenced further down in the class

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentSessionManagerTest.java
bbpennel: Also, this test is producing empty directories with names like "-3242157528316277694" and "8742058730346907586" in the root of `fcrepo-persistence-ocfl` each time I run it. It may be related to the ocflStagingRoot not getting set in `DefaultOCFLObjectSessionFactory`.

We may need to add a test of `DefaultOCFLObjectSessionFactory` that goes all the way down to persisting something so that we can verify that its creating sessions with appropriate paths.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSessionFactory.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * A default implemenntation of the {@link org.fcrepo.persistence.ocfl.api.OCFLObjectSessionFactory} interface.
+ *
+ * @author dbernstein
+ * @since 6.0.0
+ */
+public class DefaultOCFLObjectSessionFactory implements OCFLObjectSessionFactory {
+
+    private static final File DEFAULT_STAGING_DIR = new File(getProperty("fcrepo.ocfl.staging.dir",
+            getProperty("java.io.tmpdir") + File.separator + "fcrepo-ocfl-staging"));
```
bbpennel: There is a constant available in apache commons for the default tmpdir property if you'd like to use it http://commons.apache.org/proper/commons-lang/javadocs/api-release/org/apache/commons/lang3/SystemUtils.html#JAVA_IO_TMPDIR

I'm not sure if it would cause any problems if the tmpdir ended with a `/`. It might be a little safer to use `Paths.get(getProperty("java.io.tmpdir"), "fcrepo-ocfl-staging")`

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/OCFLPersistentSessionManagerTest.java
dbernstein: Variable removed. I'm no longer seeing the empty directories.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSessionFactory.java
dbernstein: Fixed.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSessionFactory.java
dbernstein: +1

PR #1586, opened 2019-11-15
FCREPO-3128. Complete TransactionImpl implementation.

mohideen:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3128

Completes the implementation of the TransactionImpl

# What's new?
- Changed `getExpires` method on Transaction Interface to return `Instant` instead of `Optional<Instant>`

# How should this be tested?
Verify that the build and tests pass.

# Additional Notes:
Currently, the only time a Transaction get removed from the TransactionManager's list is when there is a request request to get a transaction which has expired. Do we need to remove transactions on commit/rollback or should we move the finished (commited/rolledback/expired) transaction to a different list. This discussion probably can wait until we move to a database backed transaction store.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/Transaction.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/Transaction.java
@@ -19,7 +19,6 @@ package org.fcrepo.kernel.api;
 
 import java.time.Duration;
 import java.time.Instant;
-import java.util.Optional;
 
 /**
  * The Fedora Transaction abstraction
@@ -71,6 +70,11 @@ public interface Transaction {
     public void expire();
 
     /**
+     * Has the transaction expired?
+     */
+    public boolean hasExpired();
+
+    /**
     * Update the expiry by the provided amount
     * @param amountToAdd the amount of time to add
     * @return the new expiration date
@@ -81,7 +85,7 @@ public interface Transaction {
      * Get the date this session expires
      * @return expiration date, if one exists
      */
-    Optional<Instant> getExpires();
+    Instant getExpires();
 
     /**
      * Refresh the transaction to extend its expiration window.
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
@@ -17,11 +17,18 @@
  */
 package org.fcrepo.kernel.impl;
 
+
+import static java.time.Duration.ofMinutes;
+
 import java.time.Duration;
 import java.time.Instant;
-import java.util.Optional;
 
 import org.fcrepo.kernel.api.Transaction;
+import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
+import org.fcrepo.persistence.api.PersistentStorageSession;
+import org.fcrepo.persistence.api.exceptions.PersistentStorageException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * The Fedora Transaction implementation
@@ -30,28 +37,62 @@ import org.fcrepo.kernel.api.Transaction;
  */
 public class TransactionImpl implements Transaction {
 
+    private final static Logger log = LoggerFactory.getLogger(TransactionImpl.class);
+
     final String id;
+
+    final TransactionManagerImpl txManager;
+
+    final Duration DEFAULT_TIMEOUT = ofMinutes(3);
 
     boolean shortLived = true;
 
-    TransactionImpl(final String id) {
+    Instant expiration;
+
+    boolean expired = false;
+
+    boolean rolledback = false;
+
+    boolean commited = false;
+
+    protected TransactionImpl(final String id, final TransactionManagerImpl txManager) {
         if (id == null || id.isEmpty()) {
             throw new IllegalArgumentException("Transaction id should not be empty!");
         }
         this.id = id;
+        this.txManager = txManager;
+        this.expiration = Instant.now().plus(timeout());
     }
 
     @Override
     public void commit() {
-        // Prepare Persistence Transactions
-        // Commit Persistence Transactions
+        failIfExpired();
+        failIfRolledback();
+        if(this.commited) {
+            return;
+        }
+        try {
+            log.debug("Commiting transaction {}", id);
+            this.getPersistentSession().commit();
+            this.commited = true;
+        } catch (final PersistentStorageException ex) {
+            throw new RepositoryRuntimeException("failed to commit transaction " + id, ex);
+        }
     }
 
     @Override
     public void rollback() {
-        // Rollback Persistence Transactions
-
-        // Delete Transaction from TransactionManager state?
+        failIfCommited();
+        if(this.rolledback) {
+            return;
+        }
+        try {
+            log.debug("Rolling back transaction {}", id);
+            this.getPersistentSession().rollback();
+            this.rolledback = true;
+        } catch (final PersistentStorageException ex) {
+            throw new RepositoryRuntimeException("failed to rollback transaction " + id, ex);
+        }
     }
 
     @Override
@@ -59,11 +100,6 @@ public class TransactionImpl implements Transaction {
         return id;
     }
 
-    /**
-     * Set transaction short-lived state.
-     * 
-     * @param shortLived boolean true (short-lived - default) or false (not short-lived)
-     */
     @Override
     public void setShortLived(final boolean shortLived) {
         this.shortLived = shortLived;
@@ -76,20 +112,31 @@ public class TransactionImpl implements Transaction {
 
     @Override
     public void expire() {
-        // TODO Auto-generated method stub
+        this.expiration = Instant.now();
+        this.expired = true;
+    }
 
+    @Override
+    public boolean hasExpired() {
+        if (this.expired) {
+            return true;
+        }
+        this.expired = this.expiration.isBefore(Instant.now());
+        return this.expired;
     }
 
     @Override
     public Instant updateExpiry(final Duration amountToAdd) {
-        // TODO Auto-generated method stub
-        return null;
+        failIfExpired();
+        failIfCommited();
+        failIfRolledback();
+        this.expiration = this.expiration.plus(amountToAdd);
+        return this.expiration;
     }
 
     @Override
-    public Optional<Instant> getExpires() {
-        // TODO Auto-generated method stub
-        return null;
+    public Instant getExpires() {
+        return this.expiration;
     }
 
     @Override
@@ -101,7 +148,34 @@ public class TransactionImpl implements Transaction {
 
     @Override
     public void refresh() {
-        // TODO Auto-generated method stub
+        updateExpiry(timeout());
     }
 
+    private Duration timeout() {
+        // TODO Get the user configured timeout?
+        // Otherwise, use the default timeout
+        return DEFAULT_TIMEOUT;
+    }
+
+    private PersistentStorageSession getPersistentSession() {
+        return this.txManager.getPersistentStorageSessionManager().getSession(this.id);
+    }
+
+    private void failIfExpired() {
+        if(hasExpired()) {
+            throw new RuntimeException("Transaction with transactionId: " + id + " expired!");
+        }
+    }
+
+    private void failIfCommited() {
+        if(this.commited) {
+            throw new RuntimeException("Transaction with transactionId: " + id + " is already committed!");
+        }
+    }
+
+    private void failIfRolledback() {
+        if(this.rolledback) {
+            throw new RuntimeException("Transaction with transactionId: " + id + " is already rolledback!");
+        }
+    }
 }
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
@@ -25,6 +25,7 @@ import java.time.Instant;
 
 import org.fcrepo.kernel.api.Transaction;
 import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
+import org.fcrepo.kernel.api.exception.TransactionRuntimeException;
 import org.fcrepo.persistence.api.PersistentStorageSession;
 import org.fcrepo.persistence.api.exceptions.PersistentStorageException;
 import org.slf4j.Logger;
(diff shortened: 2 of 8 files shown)


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/Transaction.java
```diff
@@ -81,7 +85,7 @@
      * Get the date this session expires
      * @return expiration date, if one exists
      */
-    Optional<Instant> getExpires();
+    Instant getExpires();
```
mohideen: Do we want to support transactions that will not expire? or Was there any other reason for using an `Optional` here?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
```diff
@@ -76,20 +112,31 @@ public boolean isShortLived() {
...
+        }
+        this.expired = this.expiration.isBefore(Instant.now());
+        return this.expired;
     }
 
     @Override
     public Instant updateExpiry(final Duration amountToAdd) {
-        // TODO Auto-generated method stub
-        return null;
+        failIfExpired();
```
mohideen: Should we allow the expiration to be update if the transaction has already expired?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
```diff
@@ -30,40 +37,69 @@
  */
 public class TransactionImpl implements Transaction {
 
+    private final static Logger log = LoggerFactory.getLogger(TransactionImpl.class);
+
     final String id;
```
bbpennel: Hmm, I think all the class variables should be declared as private unless they need to be protected

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
```diff
@@ -30,40 +37,69 @@
...
+        this.expiration = Instant.now().plus(timeout());
     }
 
     @Override
     public void commit() {
-        // Prepare Persistence Transactions
-        // Commit Persistence Transactions
+        failIfExpired();
+        failIfRolledback();
+        if(this.commited) {
```
bbpennel: formatting

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
```diff
@@ -30,40 +37,69 @@
...
         if (id == null || id.isEmpty()) {
             throw new IllegalArgumentException("Transaction id should not be empty!");
         }
         this.id = id;
+        this.txManager = txManager;
+        this.expiration = Instant.now().plus(timeout());
     }
 
     @Override
     public void commit() {
```
bbpennel: Transaction instances will be shared across requests, won't they? Se we probably need some kind of locking or synchronization scheme so that we don't run into race conditions with simultaneous commit/rollback/expire/update calls. I'm not sure if we need anything more sophisticated than making most of the public methods synchronized. The persistent session classes are taking care of a lot of the more exact locking

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
```diff
@@ -30,40 +37,69 @@
...
+        failIfRolledback();
+        if(this.commited) {
+            return;
+        }
+        try {
+            log.debug("Commiting transaction {}", id);
+            this.getPersistentSession().commit();
+            this.commited = true;
+        } catch (final PersistentStorageException ex) {
+            throw new RepositoryRuntimeException("failed to commit transaction " + id, ex);
```
bbpennel: If a commit fails, should the transaction still be usable? From the work that Danny has been doing, it seems like we are going to want to trigger an automatic rollback on failure, which would indicate that the Transaction should no longer be usable for requests. This might be resolvable by simply calling `rollback()` in this catch block before throwing the error.

I have the feeling that we'll want more precise exceptions being thrown at some point so we can tell conflicts from server errors, but I don't think there's anything you can do about it at this point.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
```diff
@@ -30,40 +37,69 @@
...
+        }
     }
 
     @Override
     public void rollback() {
-        // Rollback Persistence Transactions
-
-        // Delete Transaction from TransactionManager state?
+        failIfCommited();
+        if(this.rolledback) {
```
bbpennel: space after `if`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
```diff
@@ -30,40 +37,69 @@
...
+            return;
+        }
+        try {
+            log.debug("Rolling back transaction {}", id);
+            this.getPersistentSession().rollback();
+            this.rolledback = true;
```
bbpennel: whether or not rollback succeeds, I think we still want to mark the transaction as closed/rolled back to prevent further changes. So maybe move this up a line?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
```diff
@@ -101,7 +148,34 @@ public void commitIfShortLived() {
...
+        // Otherwise, use the default timeout
+        return DEFAULT_TIMEOUT;
+    }
+
+    private PersistentStorageSession getPersistentSession() {
+        return this.txManager.getPersistentStorageSessionManager().getSession(this.id);
+    }
+
+    private void failIfExpired() {
+        if(hasExpired()) {
```
bbpennel: please add a space after `if` for all these methods

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
```diff
@@ -101,7 +148,34 @@ public void commitIfShortLived() {
...
+            throw new RuntimeException("Transaction with transactionId: " + id + " expired!");
```
bbpennel: We should at least throw RepositoryRuntimeExceptions for these methods. It would be better to throw more specific exceptions so we can set appropriate response codes. Are there exceptions leftover from the previous transaction implementation?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionManagerImpl.java
```diff
@@ -43,18 +50,28 @@ public synchronized Transaction create() {
...
         transactions.put(txId, tx);
         return tx;
     }
 
     @Override
     public Transaction get(final String transactionId) {
         if(transactions.containsKey(transactionId)) {
-            return transactions.get(transactionId);
+            final Transaction transaction = transactions.get(transactionId);
+            if(transaction.hasExpired()) {
```
bbpennel: please format with spaces after `if`'s in this file

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionManagerImpl.java
```diff
@@ -43,18 +50,28 @@ public synchronized Transaction create() {
...
+                transaction.rollback();
+                transactions.remove(transactionId);
+                throw new RuntimeException("Transaction with transactionId: " + transactionId +
```
bbpennel: It'd be good to throw a more specific exception

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/TransactionImplTest.java
```diff
@@ -0,0 +1,173 @@
...
+        assertEquals(false, testTx.isShortLived());
+    }
+
+    @Test
+    public void testCommit() {
+        testTx.commit();
+        try {
+            verify(psSession).commit();
+        } catch (PersistentStorageException e) {
+            e.printStackTrace();
```
bbpennel: I'd suggest adding `throws Exception` to the test method instead to keep the test cleaner. The same holds true for all the other tests like this

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/TransactionImplTest.java
```diff
@@ -0,0 +1,173 @@
...
+            verify(psSession, never()).rollback();
+        } catch (PersistentStorageException e) {
+            e.printStackTrace();
+        }
+    }
+
+    @Test
+    public void testExpire() {
+        testTx.expire();
+        assertEquals(true, testTx.hasExpired());
```
bbpennel: use `assertTrue`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
```diff
@@ -30,40 +37,69 @@
  */
 public class TransactionImpl implements Transaction {
 
+    private final static Logger log = LoggerFactory.getLogger(TransactionImpl.class);
+
     final String id;
 
+    final TransactionManagerImpl txManager;
+
+    final Duration DEFAULT_TIMEOUT = ofMinutes(3);
```
bbpennel: this should be `static final`

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/TransactionImplTest.java
```diff
@@ -0,0 +1,173 @@
...
+    @Mock
+    private TransactionManagerImpl txManager;
+
+    @Mock
+    private PersistentStorageSessionManager pssManager;
+
+    @Mock
+    private PersistentStorageSession psSession;
+
+    @Before
```
bbpennel: Add tests:
* commit already committed transaction
* automatic rollback when commit fails
* rollback already rolledback tx

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/TransactionImplTest.java
```diff
@@ -0,0 +1,173 @@
...
+            e.printStackTrace();
+        }
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void testCommitExpired() {
+        testTx.expire();
+        testTx.commit();
+        try {
+            verify(psSession).commit();
```
bbpennel: this verify is never called since an exception is thrown before it. You can wrap the commit in a try and put the verify in a finally block

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/TransactionImplTest.java
```diff
@@ -0,0 +1,173 @@
...
+            e.printStackTrace();
+        }
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void testCommitRolledbackTx() {
+        testTx.rollback();
+        testTx.commit();
+        try {
+            verify(psSession, never()).commit();
```
bbpennel: not called, see above

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/TransactionImplTest.java
```diff
@@ -0,0 +1,173 @@
...
+            e.printStackTrace();
+        }
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void testRollbackCommited() {
+        testTx.commit();
+        testTx.rollback();
+        try {
+            verify(psSession, never()).rollback();
```
bbpennel: not called, see above

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/TransactionImplTest.java
```diff
@@ -0,0 +1,173 @@
...
+    @Test
+    public void testRefresh() {
+        final Instant previousExpiry = testTx.getExpires();
+        testTx.refresh();
+        assertTrue(testTx.getExpires().isAfter(previousExpiry));
+    }
+
+    @Test
+    public void testNewTransactionNotExpired() {
+        assertTrue(testTx.getExpires().compareTo(Instant.now()) > 0);
```
bbpennel: For readability/consistency, could this be
`assertFalse(testTx.getExpires().isAfter(previousExpiry));`
instead?

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/TransactionManagerImplTest.java
```diff
@@ -60,6 +60,12 @@ public void testGetTransaction() {
...
     @Test(expected = RuntimeException.class)
     public void testGetTransactionWithInvalidID() {
-        final TransactionImpl tx = (TransactionImpl) testTxManager.get("invalid-id");
+        testTxManager.get("invalid-id");
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void testGetExpiredTransaction() {
+        testTx.expire();
+        testTxManager.get(testTx.getId());
```
bbpennel: Hmm, there's actually a bug here. The runtime exception you're getting is an NPE caused by like 91 here:
https://github.com/fcrepo4/fcrepo4/pull/1586/files#diff-1704d54771cc04a4b70dddfed086aba7R91

The persistent session is null, so it can't call rollback. It'd be helpful to verify that rollback gets called on the persistent session in this test. Also, expecting a `RepositoryRuntimeException` instead of RuntimeException would catch this particular bug (although a more specific exception would be better).

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/TransactionManagerImplTest.java
```diff
@@ -60,6 +60,12 @@ public void testGetTransaction() {
 
     @Test(expected = RuntimeException.class)
     public void testGetTransactionWithInvalidID() {
-        final TransactionImpl tx = (TransactionImpl) testTxManager.get("invalid-id");
+        testTxManager.get("invalid-id");
```
bbpennel: Lets at least get this test expecting a RepositoryRuntimeException instead of RuntimeException. Might have to change the class for that.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
bbpennel: no i don't think so. We should probably be rolling back expired transactions on a timer over in the transaction manager rather than exclusively waiting for a request against it

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionManagerImpl.java
bbpennel: Still need to update this formatting. There are also a few other formatting issues in this file, like there should be a space between `while` and its opening `(` at line 53

whikloj:
I'm good with this once the formatting issues are resolved. 👍

PR #1587, opened 2019-11-15
Allow one Create/Update persister

whikloj:
**JIRA Ticket**: https://jira.duraspace.org/browse/FCREPO-3096 & https://jira.duraspace.org/browse/FCREPO-3097

 Combines the Create and Update persisters for both Rdf and NonRdf into one. As they are just writing some user data and some system data.

# How should this be tested?

Review, internal unit tests.

# Interested parties
@fcrepo4/committers


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/CreateUpdatePersister.java
```diff
@@ -43,30 +44,54 @@
  * @author dbernstein
  * @since 6.0.0
  */
-public class RDFSourcePersister extends AbstractPersister {
+public class CreateUpdatePersister extends AbstractPersister {
```
bbpennel: While I don't think its doing much harm, I'm not sure i see a strong reason to merge RDF and Non-RDF persisters together here. Its a short implementation, but it seems like the core action taken is different based on rdf vs non-rdf. I guess I'm not going to push too hard for reverting the merge, but it might be a little less in the "single responsibility" department.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/NonRdfSourcePersisterTest.java
```diff
@@ -0,0 +1,120 @@
...
+
+        persister.persist(session, nonRdfSourceOperation, mapping);
+
+        // verify user content
+        verify(session).write(eq("child"), userContentCaptor.capture());
+        final InputStream userContent = userContentCaptor.getValue();
+        assertEquals(inputContent, IOUtils.toString(userContent, StandardCharsets.UTF_8));
+
+        //verify server triples
+        verify(session).write(eq(INTERNAL_FEDORA_DIRECTORY + "/child" + DEFAULT_RDF_EXTENSION), serverTriplesIsCaptor.capture());
```
bbpennel: this might fail in windows, might want to use either `File.separator` or `Path.get`

PR #1596, opened 2019-12-08
Changes to get closer to running

whikloj:
**JIRA Ticket**: n/a

In trying to get the system to respond to a GET request to the repo root I found some small issues.
1. We pass in the path part only, so the HttpIdentifierConverter needs to handle that scenario as well as a full URI.
2. We need a ResourceFactory Injected into the HTTP layer to get resources, that or we need to start radically altering how to "get" a resource up from the base.
3. Some of the `private static` fields had Inject annotations on them and it seems like this is not allowed. So I made them non-static.

After all this I am getting to [here in OCFLPersistentStorageSession](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java#L204) but it fails as its `null` because we don't have an implementation of the `FedoraToOCFLObjectIndex` interface to inject yet.

I'm not clear what this is and how its different from the FedoraOCFLMapping, but this is where the request fails.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/FedoraOCFLMapping.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/FedoraOCFLMapping.java
@@ -17,14 +17,11 @@
  */
 package org.fcrepo.persistence.ocfl.impl;
 
-import org.springframework.stereotype.Component;
-
 /**
  * A mapping that links the parent fedora resource to its corresponding OCFL object.
  *
  * @author dbernstein
  */
-@Component
 public class FedoraOCFLMapping {
     private String parentFedoraResourceId;
     private String ocflObjectId;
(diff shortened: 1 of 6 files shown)


dbernstein:
@whikloj :  the FedoraToOCFLObjectIndex maintains the set of FedoraOCFLMappings.  I've create a JIRA for this:  https://jira.lyrasis.org/browse/FCREPO-3155

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/FedoraOCFLMapping.java
```diff
@@ -17,11 +17,14 @@
...
 package org.fcrepo.persistence.ocfl.impl;
 
+import org.springframework.stereotype.Component;
+
 /**
  * A mapping that links the parent fedora resource to its corresponding OCFL object.
  *
  * @author dbernstein
  */
+@Component
```
dbernstein: This @Component doesn't belong here.  It should go on the implementation of FedoraToOCFLObjectIndex when it exists.

PR #1597, opened 2019-12-09
Adds simple in-memory implementation for FedoraToOCFLObjectIndex.

dbernstein:
Resolves: https://jira.lyrasis.org/browse/FCREPO-3155

**Simple in-memory implementation for FedoraToOCFLObjectIndex**
* * *

**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3155

Creates a simple in-memory implementation of  FedoraToOCFLObjectIndex and injects it into the OCFLPersistentStorageSession.

# How should this be tested?

Covered by unit tests.

# Additional Notes:

# Interested parties

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateNonRdfSourceOperation.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateNonRdfSourceOperation.java
@@ -18,13 +18,11 @@
 package org.fcrepo.kernel.impl.operations;
 
 import static org.fcrepo.kernel.api.RdfLexicon.NON_RDF_SOURCE;
-import static org.fcrepo.kernel.api.operations.ResourceOperationType.CREATE;
 
 import java.io.InputStream;
 import java.net.URI;
 
 import org.fcrepo.kernel.api.operations.CreateResourceOperation;
-import org.fcrepo.kernel.api.operations.ResourceOperationType;
 
 /**
  * Operation for creating a new non-rdf source
@@ -79,8 +77,4 @@ public class CreateNonRdfSourceOperation extends AbstractNonRdfSourceOperation i
         this.parentId = parentId;
     }
 
-    @Override
-    public ResourceOperationType getType() {
-        return CREATE;
-    }
 }
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateNonRdfSourceOperation.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateNonRdfSourceOperation.java
@@ -18,11 +18,13 @@
 package org.fcrepo.kernel.impl.operations;
 
 import static org.fcrepo.kernel.api.RdfLexicon.NON_RDF_SOURCE;
+import static org.fcrepo.kernel.api.operations.ResourceOperationType.CREATE;
 
 import java.io.InputStream;
 import java.net.URI;
 
 import org.fcrepo.kernel.api.operations.CreateResourceOperation;
+import org.fcrepo.kernel.api.operations.ResourceOperationType;
 
 /**
  * Operation for creating a new non-rdf source
@@ -75,5 +77,10 @@ public class CreateNonRdfSourceOperation extends AbstractNonRdfSourceOperation i
      */
     public void setParentId(final String parentId) {
         this.parentId = parentId;
+    }
+
+    @Override
+    public ResourceOperationType getType() {
+        return CREATE;
     }
 }
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateNonRdfSourceOperation.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateNonRdfSourceOperation.java
@@ -61,6 +61,11 @@ public class CreateNonRdfSourceOperation extends AbstractNonRdfSourceOperation i
     }
 
     @Override
+    public boolean isArchivalGroup() {
+        return false;
+    }
+
+    @Override
     public String getParentId() {
         return parentId;
     }
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateRdfSourceOperation.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateRdfSourceOperation.java
@@ -19,6 +19,9 @@ package org.fcrepo.kernel.impl.operations;
 
 import org.fcrepo.kernel.api.RdfStream;
 import org.fcrepo.kernel.api.operations.CreateResourceOperation;
+import org.fcrepo.kernel.api.operations.ResourceOperationType;
+
+import static org.fcrepo.kernel.api.operations.ResourceOperationType.CREATE;
 
 /**
  * Operation to create an RDF source.
@@ -67,5 +70,10 @@ public class CreateRdfSourceOperation extends AbstractRdfSourceOperation impleme
      */
     public void setParentId(final String parentId) {
         this.parentId = parentId;
+    }
+
+    @Override
+    public ResourceOperationType getType() {
+        return CREATE;
     }
 }
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateRdfSourceOperation.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateRdfSourceOperation.java
@@ -19,9 +19,6 @@ package org.fcrepo.kernel.impl.operations;
 
 import org.fcrepo.kernel.api.RdfStream;
 import org.fcrepo.kernel.api.operations.CreateResourceOperation;
-import org.fcrepo.kernel.api.operations.ResourceOperationType;
-
-import static org.fcrepo.kernel.api.operations.ResourceOperationType.CREATE;
 
 /**
  * Operation to create an RDF source.
@@ -72,8 +69,4 @@ public class CreateRdfSourceOperation extends AbstractRdfSourceOperation impleme
         this.parentId = parentId;
     }
 
-    @Override
-    public ResourceOperationType getType() {
-        return CREATE;
-    }
 }
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateRdfSourceOperation.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateRdfSourceOperation.java
@@ -53,6 +53,11 @@ public class CreateRdfSourceOperation extends AbstractRdfSourceOperation impleme
     }
 
     @Override
+    public boolean isArchivalGroup() {
+        return false;
+    }
+
+    @Override
     public String getParentId() {
         return parentId;
     }
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
@@ -18,6 +18,8 @@
 package org.fcrepo.persistence.ocfl;
 
 import static java.lang.String.format;
+import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_ID_PREFIX;
+import org.fcrepo.kernel.api.operations.CreateResourceOperation;
 import static org.fcrepo.persistence.ocfl.OCFLPersistentStorageUtils.resolveVersionId;
 
 import java.io.InputStream;
@@ -34,6 +36,7 @@ import org.fcrepo.persistence.api.PersistentStorageSession;
 import org.fcrepo.persistence.api.exceptions.PersistentItemNotFoundException;
 import org.fcrepo.persistence.api.exceptions.PersistentSessionClosedException;
 import org.fcrepo.persistence.api.exceptions.PersistentStorageException;
+import org.fcrepo.persistence.ocfl.api.FedoraToOCFLObjectIndex;
 import org.fcrepo.persistence.ocfl.api.OCFLObjectSession;
 import org.fcrepo.persistence.ocfl.api.OCFLObjectSessionFactory;
 import org.fcrepo.persistence.ocfl.api.Persister;
@@ -163,8 +166,18 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
             }
 
             //resolve the mapping between the fedora resource identifier and the associated OCFL object.
-            final FedoraOCFLMapping mapping = fedoraOcflIndex.getMapping(operation.getResourceId());
+            final String resourceId = operation.getResourceId();
+            final FedoraOCFLMapping mapping = fedoraOcflIndex.getMapping(resourceId);
 
+            //if no mapping exists, create one
+            if(mapping == null) {
+                if(operation instanceof CreateResourceOperation) {
+                    final String parentId = ((CreateResourceOperation)operation).getParentId();
+                    fedoraOcflIndex.addMapping(resourceId,  parentId, mintOCFLObjectId(parentId));
+               } else {
+                    throw new PersistentStorageException("Unable to resolve parent identifier for " + resourceId);
+                }
+            }
             //get the session.
             final OCFLObjectSession objSession = findOrCreateSession(mapping.getOcflObjectId());
 
@@ -175,6 +188,15 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
             phaser.arriveAndDeregister();
         }
 
+    }
+
+    private String mintOCFLObjectId ( final String fedoraIdentifier) {
+        //TODO make OBJECT minting more configurable.
+        String bareFedoraIdentifier = fedoraIdentifier;
+        if(fedoraIdentifier.indexOf(FEDORA_ID_PREFIX) == 0){
+            bareFedoraIdentifier = fedoraIdentifier.substring(FEDORA_ID_PREFIX.length());
+        }
+        return bareFedoraIdentifier.replace("/", "_");
     }
 
     private void ensureCommitNotStarted() throws PersistentSessionClosedException {
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
@@ -19,7 +19,6 @@ package org.fcrepo.persistence.ocfl;
 
 import static java.lang.String.format;
 import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_ID_PREFIX;
-import org.fcrepo.kernel.api.operations.CreateResourceOperation;
 import static org.fcrepo.persistence.ocfl.OCFLPersistentStorageUtils.resolveVersionId;
 
 import java.io.InputStream;
@@ -29,13 +28,16 @@ import java.util.List;
 import java.util.Map;
 
 import org.fcrepo.kernel.api.RdfStream;
+import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
 import org.fcrepo.kernel.api.models.ResourceHeaders;
+import org.fcrepo.kernel.api.operations.CreateResourceOperation;
 import org.fcrepo.kernel.api.operations.ResourceOperation;
 import org.fcrepo.persistence.api.CommitOption;
 import org.fcrepo.persistence.api.PersistentStorageSession;
 import org.fcrepo.persistence.api.exceptions.PersistentItemNotFoundException;
 import org.fcrepo.persistence.api.exceptions.PersistentSessionClosedException;
 import org.fcrepo.persistence.api.exceptions.PersistentStorageException;
+import org.fcrepo.persistence.ocfl.api.FedoraOCFLMappingNotFoundException;
 import org.fcrepo.persistence.ocfl.api.FedoraToOCFLObjectIndex;
 import org.fcrepo.persistence.ocfl.api.OCFLObjectSession;
 import org.fcrepo.persistence.ocfl.api.OCFLObjectSessionFactory;
@@ -88,6 +90,8 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
     private List<CommittedSession> sessionsToRollback = new ArrayList<>();
 
     private State state = State.COMMIT_NOT_STARTED;
+
+    private static final String DEFAULT_REPOSITORY_ROOT_OCFL_OBJECT_ID = "_fedora_repository_root";
 
     static {
         PERSISTER_LIST.add(new RDFSourcePersister());
@@ -167,17 +171,10 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
 
             //resolve the mapping between the fedora resource identifier and the associated OCFL object.
             final String resourceId = operation.getResourceId();
-            final FedoraOCFLMapping mapping = fedoraOcflIndex.getMapping(resourceId);
 
-            //if no mapping exists, create one
-            if(mapping == null) {
-                if(operation instanceof CreateResourceOperation) {
-                    final String parentId = ((CreateResourceOperation)operation).getParentId();
-                    fedoraOcflIndex.addMapping(resourceId,  parentId, mintOCFLObjectId(parentId));
-               } else {
-                    throw new PersistentStorageException("Unable to resolve parent identifier for " + resourceId);
-                }
-            }
+            //retrieve the mapping
+            final FedoraOCFLMapping mapping = findOrCreateFedoraOCFLMapping(operation, resourceId);
+
             //get the session.
             final OCFLObjectSession objSession = findOrCreateSession(mapping.getOcflObjectId());
 
@@ -190,13 +187,49 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
 
     }
 
-    private String mintOCFLObjectId ( final String fedoraIdentifier) {
-        //TODO make OBJECT minting more configurable.
+    private FedoraOCFLMapping findOrCreateFedoraOCFLMapping(final ResourceOperation operation, final String resourceId)
+            throws PersistentStorageException {
+        FedoraOCFLMapping mapping;
+
+        try {
+            mapping = fedoraOcflIndex.getMapping(resourceId);
+        } catch (FedoraOCFLMappingNotFoundException e) {
+            //if no mapping exists, create one
+            if (operation instanceof CreateResourceOperation) {
+                final String parentId = ((CreateResourceOperation) operation).getParentId();
+                fedoraOcflIndex.addMapping(resourceId, parentId, mintOCFLObjectId(parentId));
+                try {
+                    mapping = fedoraOcflIndex.getMapping(resourceId);
+                } catch (FedoraOCFLMappingNotFoundException ex) {
+                    throw new PersistentStorageException("Failed to retrieve new mapping for " + resourceId);
+                }
+            } else {
+                throw new PersistentStorageException("Unable to resolve parent identifier for " + resourceId);
+            }
+        }
+
+        return mapping;
+    }
+
+    private String mintOCFLObjectId(final String fedoraIdentifier) {
+        //TODO make OCFL Object Id minting more configurable.
         String bareFedoraIdentifier = fedoraIdentifier;
-        if(fedoraIdentifier.indexOf(FEDORA_ID_PREFIX) == 0){
+        if (fedoraIdentifier.indexOf(FEDORA_ID_PREFIX) == 0) {
             bareFedoraIdentifier = fedoraIdentifier.substring(FEDORA_ID_PREFIX.length());
         }
-        return bareFedoraIdentifier.replace("/", "_");
+
+        //ensure no accidental collisions with the root ocfl identifier
+        if (bareFedoraIdentifier.equals(DEFAULT_REPOSITORY_ROOT_OCFL_OBJECT_ID)) {
+            throw new RepositoryRuntimeException(bareFedoraIdentifier + " in a reserved identifier");
+        }
+
+        bareFedoraIdentifier = bareFedoraIdentifier.replace("/", "_");
+
+        if (bareFedoraIdentifier.length() == 0) {
+            bareFedoraIdentifier = DEFAULT_REPOSITORY_ROOT_OCFL_OBJECT_ID;
+        }
+
+        return bareFedoraIdentifier;
     }
 
     private void ensureCommitNotStarted() throws PersistentSessionClosedException {
@@ -223,7 +256,7 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
 
         ensureCommitNotStarted();
 
-        final FedoraOCFLMapping mapping = fedoraOcflIndex.getMapping(identifier);
+        final FedoraOCFLMapping mapping = getFedoraOCFLMapping(identifier);
         final OCFLObjectSession objSession = findOrCreateSession(mapping.getOcflObjectId());
         final String fedoraSubpath = relativizeSubpath(mapping.getParentFedoraResourceId(), identifier);
         final String ocflSubpath = resolveOCFLSubpath(fedoraSubpath);
@@ -240,12 +273,20 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
         return deserializeHeaders(headerStream);
     }
 
+    private FedoraOCFLMapping getFedoraOCFLMapping(final String identifier) throws PersistentStorageException {
+        try {
+            return fedoraOcflIndex.getMapping(identifier);
+        } catch (FedoraOCFLMappingNotFoundException e) {
+            throw new PersistentStorageException(e.getMessage());
+        }
+    }
+
     @Override
     public RdfStream getTriples(final String identifier, final Instant version)
             throws PersistentStorageException {
         ensureCommitNotStarted();
 
-        final FedoraOCFLMapping mapping = fedoraOcflIndex.getMapping(identifier);
+        final FedoraOCFLMapping mapping = getFedoraOCFLMapping(identifier);
         final OCFLObjectSession objSession = findOrCreateSession(mapping.getOcflObjectId());
         final String fedoraSubpath = relativizeSubpath(mapping.getParentFedoraResourceId(), identifier);
         final String ocflSubpath = resolveOCFLSubpath(fedoraSubpath);
@@ -255,12 +296,13 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
 
     /**
      * Returns a list of immutable versions associated with the specified fedora identifier
+     *
      * @param fedoraIdentifier The fedora identifier
      * @return The list of instants that map to the underlying versions
      * @throws PersistentStorageException Due the underlying resource not existing or is otherwise unreadable.
      */
-     List<Instant> listVersions(final String fedoraIdentifier) throws PersistentStorageException {
-        final FedoraOCFLMapping mapping = fedoraOcflIndex.getMapping(fedoraIdentifier);
+    List<Instant> listVersions(final String fedoraIdentifier) throws PersistentStorageException {
+        final FedoraOCFLMapping mapping = getFedoraOCFLMapping(fedoraIdentifier);
         final OCFLObjectSession objSession = findOrCreateSession(mapping.getOcflObjectId());
         return OCFLPersistentStorageUtils.listVersions(objSession);
     }
@@ -269,7 +311,7 @@ public class OCFLPersistentStorageSession implements PersistentStorageSession {
     public RdfStream getManagedProperties(final String identifier, final Instant version)
             throws PersistentStorageException {
         ensureCommitNotStarted();
-        final FedoraOCFLMapping mapping = fedoraOcflIndex.getMapping(identifier);
+        final FedoraOCFLMapping mapping = getFedoraOCFLMapping(identifier);
         final OCFLObjectSession objSession = findOrCreateSession(mapping.getOcflObjectId());
         final String fedoraSubpath = relativizeSubpath(mapping.getParentFedoraResourceId(), identifier);
         final String ocflSubpath = resolveOCFLSubpath(fedoraSubpath);
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
@@ -29,23 +29,25 @@ public interface FedoraToOCFLObjectIndex {
      * Retrieve identification information for the OCFL object which either contains, or is identified by,
      * the provided fedora resource id. In other words the method will find the closest resource that is persisted
      * as an OCFL object and returns its identifiers.
-     *
+     * <p>
      * If you pass fedora identifier that is not part of an archival group such as
      * "my/fedora/binary/fcr:metadata"  the parent fedora resource returned in the mapping will be "my/fedora/binary".
-     *
+     * <p>
      * Contrast this  with an Archival Group example:  if you pass in "my/archival-group/binary/fcr:metadata" the
      * parent returned in the mapping would be "my/archival-group".
      *
      * @param fedoraResourceIdentifier the fedora resource identifier
      * @return the mapping
+     * @throws FedoraOCFLMappingNotFoundException when no mapping exists for the specified identifier.
      */
-    FedoraOCFLMapping getMapping(final String fedoraResourceIdentifier);
+    FedoraOCFLMapping getMapping(final String fedoraResourceIdentifier) throws FedoraOCFLMappingNotFoundException;
 
     /**
      * Adds a mapping to the index
+     *
      * @param fedoraResourceIdentifier The fedora resource
-     * @param parentFedoraResourceId The parent of the fedora resource
-     * @param ocflObjectId The ocfl object id
+     * @param parentFedoraResourceId   The parent of the fedora resource
+     * @param ocflObjectId             The ocfl object id
      */
     public void addMapping(final String fedoraResourceIdentifier, final String parentFedoraResourceId,
                            final String ocflObjectId);
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
@@ -29,10 +29,10 @@ public interface FedoraToOCFLObjectIndex {
      * Retrieve identification information for the OCFL object which either contains, or is identified by,
      * the provided fedora resource id. In other words the method will find the closest resource that is persisted
      * as an OCFL object and returns its identifiers.
-     * <p>
+     *
      * If you pass fedora identifier that is not part of an archival group such as
      * "my/fedora/binary/fcr:metadata"  the parent fedora resource returned in the mapping will be "my/fedora/binary".
-     * <p>
+     *
      * Contrast this  with an Archival Group example:  if you pass in "my/archival-group/binary/fcr:metadata" the
      * parent returned in the mapping would be "my/archival-group".
      *
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
@@ -48,8 +48,9 @@ public interface FedoraToOCFLObjectIndex {
      * @param fedoraResourceIdentifier The fedora resource
      * @param parentFedoraResourceId   The parent of the fedora resource
      * @param ocflObjectId             The ocfl object id
+     * @return  The newly created mapping
      */
-    public void addMapping(final String fedoraResourceIdentifier, final String parentFedoraResourceId,
+    public FedoraOCFLMapping addMapping(final String fedoraResourceIdentifier, final String parentFedoraResourceId,
                            final String ocflObjectId);
 }
 
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.fcrepo.persistence.ocfl;
+package org.fcrepo.persistence.ocfl.api;
 
 import org.fcrepo.persistence.ocfl.impl.FedoraOCFLMapping;
 
@@ -40,5 +40,14 @@ public interface FedoraToOCFLObjectIndex {
      * @return the mapping
      */
     FedoraOCFLMapping getMapping(final String fedoraResourceIdentifier);
+
+    /**
+     * Adds a mapping to the index
+     * @param fedoraResourceIdentifier The fedora resource
+     * @param parentFedoraResourceId The parent of the fedora resource
+     * @param ocflObjectId The ocfl object id
+     */
+    public void addMapping(final String fedoraResourceIdentifier, final String parentFedoraResourceId,
+                           final String ocflObjectId);
 }
 
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
@@ -46,11 +46,11 @@ public interface FedoraToOCFLObjectIndex {
      * Adds a mapping to the index
      *
      * @param fedoraResourceIdentifier The fedora resource
-     * @param parentFedoraResourceId   The parent of the fedora resource
+     * @param fedoraRootObjectId   The fedora resource that maps directly to the ocfl object
      * @param ocflObjectId             The ocfl object id
      * @return  The newly created mapping
      */
-    public FedoraOCFLMapping addMapping(final String fedoraResourceIdentifier, final String parentFedoraResourceId,
+    public FedoraOCFLMapping addMapping(final String fedoraResourceIdentifier, final String fedoraRootObjectId,
                            final String ocflObjectId);
 }
 
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractNonRdfSourcePersister.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractNonRdfSourcePersister.java
@@ -27,7 +27,7 @@ import static org.fcrepo.persistence.common.ResourceHeaderUtils.populateBinaryHe
 import static org.fcrepo.persistence.common.ResourceHeaderUtils.populateExternalBinaryHeaders;
 import static org.fcrepo.persistence.common.ResourceHeaderUtils.touchCreationHeaders;
 import static org.fcrepo.persistence.common.ResourceHeaderUtils.touchModificationHeaders;
-import static org.fcrepo.persistence.ocfl.OCFLPersistentStorageUtils.relativizeSubpath;
+import static org.fcrepo.persistence.ocfl.impl.OCFLPersistentStorageUtils.relativizeSubpath;
 
 import java.util.HashSet;
 import java.util.Set;
@@ -40,7 +40,9 @@ import org.fcrepo.kernel.api.operations.ResourceOperationType;
 import org.fcrepo.persistence.api.WriteOutcome;
 import org.fcrepo.persistence.api.exceptions.PersistentStorageException;
 import org.fcrepo.persistence.common.ResourceHeadersImpl;
+import org.fcrepo.persistence.ocfl.api.FedoraToOCFLObjectIndex;
 import org.fcrepo.persistence.ocfl.api.OCFLObjectSession;
+import org.fcrepo.persistence.ocfl.api.OCFLObjectSessionFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -50,22 +52,24 @@ import org.slf4j.LoggerFactory;
  * @author whikloj
  * @since 6.0.0
  */
-public class NonRdfSourcePersister extends AbstractPersister {
+abstract class AbstractNonRdfSourcePersister extends AbstractPersister {
 
-    private static final Logger log = LoggerFactory.getLogger(NonRdfSourcePersister.class);
+    private static final Logger log = LoggerFactory.getLogger(AbstractNonRdfSourcePersister.class);
 
     private static final Set<ResourceOperationType> OPERATION_ACTIONS = new HashSet<>(asList(CREATE, UPDATE));
 
     /**
      * Constructor
      */
-    public NonRdfSourcePersister() {
-        super(NonRdfSourceOperation.class, OPERATION_ACTIONS);
+    AbstractNonRdfSourcePersister(final Class<? extends ResourceOperation> resourceOperation,
+                                  final ResourceOperationType resourceOperationType,
+                                  final OCFLObjectSessionFactory objectFactory,
+                                  final FedoraToOCFLObjectIndex index) {
+        super(resourceOperation, resourceOperationType, objectFactory, index);
     }
 
-    @Override
-    public void persist(final OCFLObjectSession session, final ResourceOperation operation,
-            final FedoraOCFLMapping mapping) throws PersistentStorageException {
+    protected void persistNonRDFSource(final ResourceOperation operation, final FedoraOCFLMapping mapping,
+                                       final OCFLObjectSession objectSession) throws PersistentStorageException {
         log.debug("persisting ({}) to {}", operation.getResourceId(), mapping.getOcflObjectId());
         final String subpath = relativizeSubpath(mapping.getParentFedoraResourceId(), operation.getResourceId());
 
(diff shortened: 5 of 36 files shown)


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -163,8 +166,18 @@ public void persist(final ResourceOperation operation) throws PersistentStorageE
...
 
             //resolve the mapping between the fedora resource identifier and the associated OCFL object.
-            final FedoraOCFLMapping mapping = fedoraOcflIndex.getMapping(operation.getResourceId());
-
+            final String resourceId = operation.getResourceId();
+            final FedoraOCFLMapping mapping = fedoraOcflIndex.getMapping(resourceId);
+
+            //if no mapping exists, create one
+            if(mapping == null) {
+                if(operation instanceof CreateResourceOperation) {
```
whikloj: Nit picky but the lack of space between `if` and `(` on these two lines makes my brain hurt.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -18,6 +18,8 @@
 package org.fcrepo.persistence.ocfl;
 
 import static java.lang.String.format;
+import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_ID_PREFIX;
+import org.fcrepo.kernel.api.operations.CreateResourceOperation;
```
whikloj: I really think we need to pick how we are organizing imports.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
```diff
@@ -40,5 +40,14 @@
...
     FedoraOCFLMapping getMapping(final String fedoraResourceIdentifier);
+
+    /**
+     * Adds a mapping to the index
+     * @param fedoraResourceIdentifier The fedora resource
+     * @param parentFedoraResourceId The parent of the fedora resource
+     * @param ocflObjectId The ocfl object id
+     */
+    public void addMapping(final String fedoraResourceIdentifier, final String parentFedoraResourceId,
+                           final String ocflObjectId);
```
whikloj: How do we add a mapping for the root of an OCFL object? Is the `fedoraResourceIdentifier == parentFedoraResourceId`?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/FedoraToOCFLObjectIndexImpl.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public FedoraOCFLMapping getMapping(final String fedoraResourceIdentifier) {
+        return fedoraOCFLMappingMap.get(fedoraResourceIdentifier);
+    }
+
+    @Override
+    public void addMapping(final String fedoraResourceIdentifier, final String parentFedoraResourceId, final String ocflObjectId) {
+        FedoraOCFLMapping mapping = fedoraOCFLMappingMap.get(parentFedoraResourceId);
+
+        if(mapping == null){
```
whikloj: nit pick `if (mapping == null) {`

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -177,6 +190,15 @@ public void persist(final ResourceOperation operation) throws PersistentStorageE
 
     }
 
+    private String mintOCFLObjectId ( final String fedoraIdentifier) {
+        //TODO make OBJECT minting more configurable.
+        String bareFedoraIdentifier = fedoraIdentifier;
+        if(fedoraIdentifier.indexOf(FEDORA_ID_PREFIX) == 0){
```
whikloj: This isn't path minting right, this is transforming a Fedora ID into an OCFL ID?

whikloj: Also, nit picky space `if (fed...`

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/FedoraToOCFLObjectIndexImpl.java
```diff
@@ -0,0 +1,53 @@
...
+ * @author dbernstein
+ * @since 6.0.0
+ */
+@Component
+public class FedoraToOCFLObjectIndexImpl implements FedoraToOCFLObjectIndex {
+    private Map<String,FedoraOCFLMapping> fedoraOCFLMappingMap = Collections.synchronizedMap(new HashMap<>());
+
+    @Override
+    public FedoraOCFLMapping getMapping(final String fedoraResourceIdentifier) {
+        return fedoraOCFLMappingMap.get(fedoraResourceIdentifier);
```
whikloj: The interface is not clear on what this should return, but all the code around this seems to assume it will return a `FedoraOCFLMapping` object. Currently this returns `null` if not found, so should probably throw an exception. Also it finds nothing because the root of the repository is not a real thing, we probably need to pre-load an entry for `info:fedora/`

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
dbernstein: In some cases it may be a simple transaction.  In other cases it may be a minting.  It depends on what strategy the user decides to use: c.f. https://jira.lyrasis.org/browse/FCREPO-3156

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
dbernstein: yes

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -177,6 +187,51 @@ public void persist(final ResourceOperation operation) throws PersistentStorageE
...
 
+    private FedoraOCFLMapping findOrCreateFedoraOCFLMapping(final ResourceOperation operation, final String resourceId)
+            throws PersistentStorageException {
+        FedoraOCFLMapping mapping;
+
+        try {
+            mapping = fedoraOcflIndex.getMapping(resourceId);
+        } catch (FedoraOCFLMappingNotFoundException e) {
+            //if no mapping exists, create one
+            if (operation instanceof CreateResourceOperation) {
```
bbpennel: I'm not sure at the moment how else to arrange things, but it feels a little odd to me to have management of the index being so directly controlled by this session. There will be other places that add to the index, at a minimum during a rebuild scenario. It might make sense for the index to know how to populate itself based off what's on disk, so a call to `getMapping` would seek an existing record first in the index and then on disk, before failing. I know this is first pass implementation, just wondering if more of the guts should be internalized.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/FedoraToOCFLObjectIndexImplTest.java
```diff
@@ -0,0 +1,51 @@
...
+    private static final String RESOURCE_ID_2 = "info:fedora/parent/child1";
+
+    private static final String PARENT_RESOURCE_ID = "info:fedora/parent";
+    private static final String OCFL_ID = "ocfl-id";
+
+    @Test
+    public void test() throws Exception {
+        final FedoraToOCFLObjectIndexImpl index = new FedoraToOCFLObjectIndexImpl();
+
+        index.addMapping(RESOURCE_ID_1, PARENT_RESOURCE_ID, OCFL_ID);
```
bbpennel: might be good to test two different resources with different ocfl_ids, etc? right now resc1 and 2 have the same id and same details.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
bbpennel: Could you add a test to verify the root resource case?

Also, I'm not sure at the moment where the root's parent gets set to itself? I wonder if it'd make sense for the parent of the root to be null instead, otherwise it might come back as its own child in some queries, unless i'm misunderstanding.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/FedoraToOCFLObjectIndexImplTest.java
```diff
@@ -0,0 +1,51 @@
...
+package org.fcrepo.persistence.ocfl.impl;
+
+import static org.junit.Assert.assertEquals;
+import org.junit.Test;
+
+/**
+ * @author dbernstein
+ * @since 6.0.0
+ */
+public class FedoraToOCFLObjectIndexImplTest {
```
bbpennel: might be good to add a test case to verify what happens on `getMapping` when the mapping doesn't exist.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
```diff
@@ -177,6 +187,51 @@ public void persist(final ResourceOperation operation) throws PersistentStorageE
...
+            }
+        }
+
+        return mapping;
+    }
+
+    private String mintOCFLObjectId(final String fedoraIdentifier) {
+        //TODO make OCFL Object Id minting more configurable.
+        String bareFedoraIdentifier = fedoraIdentifier;
+        if (fedoraIdentifier.indexOf(FEDORA_ID_PREFIX) == 0) {
```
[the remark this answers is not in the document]
dbernstein: I'll fix.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/FedoraToOCFLObjectIndexImplTest.java
dbernstein: oops - good catch.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/FedoraToOCFLObjectIndexImplTest.java
dbernstein: +1

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
dbernstein: I'll take a look and see what makes sense.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
dbernstein: I think the confusion may be around terminology. It's not really the "parent" it is the fedora resource that corresponds directly to the OCFL object. I'm not sure what to call that.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/OCFLPersistentStorageSession.java
dbernstein: I agree that OCFLPersistentStorageSession should not be responsible for populating the index when the resources already exist on disk and that yes, maybe some kind of lazy loading would be a good idea ( and also we could support preloading the index).   Still I believe we need a way to populate the index when new objects are created no?

bbpennel: I think its reasonable for the index to have both `getMapping` and `getOrCreateMapping` where the latter attempts to lazy load the value into the index by whatever means it feels is appropriate.

so `getOrCreateMapping` would:

1. Check to see if the resource id existed in the index
2. if not, then check to see if the resource id was persisted as a OCFL object. So given `atomic/nested_ag` it would try to see if the ocfl root contained an object with that id. This covers the "new root resource" case.
3. if not, then it would remove the last segment of the identifier to get the parent and check for it in the index. If it finds an entry for that parent, then the parent's root resource would become the current resource's root resource. So given `my_ag/folder/binary.txt`, it would look up `my_ag/folder`, then use its root resource which is `my_ag`

The parent should not be new at the same time as the new resource, so it should be indexed (unless the index got wiped, in which cause it should get repopulated at startup of the server).

I think the current approach might be problematic since the operation's parent id is the resource the contains the new resource, which is not necessarily the root resource (in the case of being contained by an atomistic resource).

dbernstein: In step 2,  how would the OCFL Id be resolved in order to make that check?

bbpennel: Is the OCFL id a function of the fedora id, or is it totally independent? If its totally independent, then it can only be found with a separate index of OCFL object ids or a crawl of all OCFL objects in the ocfl root.

dbernstein: I think the answer is both:  it can be deterministic or independent depending on the object object minter. I believe we know we'll need to support independent ids.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/CreateNonRdfSourcePersisterTest.java
```diff
@@ -98,12 +103,12 @@
 
     private static final Long EXTERNAL_CONTENT_SIZE = 526632l;
 
-    private final NonRdfSourcePersister persister = new NonRdfSourcePersister();
+    private CreateNonRdfSourcePersister persister;
 
     @Before
     public void setUp() throws Exception {
```
bbpennel: Maybe add a test for what happens if you try to create a resource that already exists?

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/CreateRDFSourcePersisterTest.java
```diff
@@ -67,18 +40,39 @@
...
+import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.withSettings;
+
 /**
  * @author dbernstein
  * @since 6.0.0
  */
 @RunWith(MockitoJUnitRunner.class)
-public class RDFSourcePersisterTest {
+public class CreateRDFSourcePersisterTest {
```
bbpennel: Maybe add a test for what happens if you try to create a resource that already exists? And a test to make sure you can't create something with the same id as the repo root

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/DeleteResourcePersisterTest.java
```diff
@@ -48,31 +54,52 @@
...
+        operation = mock(RdfSourceOperation.class);
+        persister = new DeleteResourcePersister(this.sessionFactory, this.index);
+
+        when(psSession.findOrCreateSession(anyString())).thenReturn(session);
+        when(index.getMapping(anyString())).thenReturn(mapping);
+    }
+
 
     @Test
     public void testDeleteSubPath() throws Exception {
```
bbpennel: add test for deleting object that doesn't exist?

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/UpdateNonRdfSourcePersisterTest.java
```diff
@@ -0,0 +1,190 @@
...
+
+        persister = new UpdateNonRdfSourcePersister(objectSessionFactory, index);
+
+        when(psSession.findOrCreateSession(anyString())).thenReturn(session);
+        when(index.getMapping(anyString())).thenReturn(mapping);
+
+    }
+
+    @Test
+    public void testNonRdfExistingResource() throws Exception {
```
bbpennel: test for updating a resource that doesn't exist

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/UpdateRDFSourcePersisterTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+        when(session.write(anyString(), any(InputStream.class))).thenReturn(writeOutcome);
+        when(psSession.findOrCreateSession(anyString())).thenReturn(session);
+        when(index.getMapping(anyString())).thenReturn(mapping);
+
+        persister = new UpdateRDFSourcePersister(this.sessionFactory, this.index );
+    }
+
+    @Test
+    public void testPersistExistingResource() throws Exception {
```
bbpennel: test for updating a resource that doesn't exist

bbpennel: I know that the AbstractPersisterTest tests `handle` in a generic sense, but it would probably be good to verify that the persister impls handle the expected types of operations, since that is unique to each persister.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/UpdateRDFSourcePersisterTest.java
```diff
@@ -0,0 +1,188 @@
...
+@RunWith(MockitoJUnitRunner.class)
+public class UpdateRDFSourcePersisterTest {
+
+    private static final String RESOURCE_ID = "info:fedora/parent/child";
+
+    private static final String ROOT_RESOURCE_ID = "info:fedora/parent";
+
+    private static final String USER_PRINCIPAL = "fedoraUser";
+
+    private static final Instant CREATED_DATE = Instant.parse("2019-11-22T12:10:04.000000Z");
```
bbpennel: CREATED_DATE and MODIFIED_DATE are not used

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/UpdateNonRdfSourcePersisterTest.java
```diff
@@ -0,0 +1,190 @@
...
+
+    private static final String ROOT_RESOURCE_ID = "info:fedora/parent";
+
+    private static final String USER_PRINCIPAL = "fedoraUser";
+
+    private static final String CONTENT_BODY = "this is some example content";
+
+    private static final Long LOCAL_CONTENT_SIZE = Long.valueOf(CONTENT_BODY.length());
+
+    private static final String EXTERNAL_URL = "http://example.com/file.txt";
```
bbpennel: none of these external variables are being used. Perhaps an external binary test should be added?

bbpennel: This is also making me realize that the persister probably needs to be able to handle the case of switching from an internal to an external binary, which would entail deleting the binary (or is it writing a 0 byte file? I forget at the moment). Switching from an external to internal probably already works. Not sure if you want to handle that in this PR or make a followup.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractNonRdfSourcePersister.java
```diff
@@ -50,24 +52,26 @@
  * @author whikloj
  * @since 6.0.0
  */
-public class NonRdfSourcePersister extends AbstractPersister {
+abstract class AbstractNonRdfSourcePersister extends AbstractPersister {
 
-    private static final Logger log = LoggerFactory.getLogger(NonRdfSourcePersister.class);
+    private static final Logger log = LoggerFactory.getLogger(AbstractNonRdfSourcePersister.class);
 
     private static final Set<ResourceOperationType> OPERATION_ACTIONS = new HashSet<>(asList(CREATE, UPDATE));
```
bbpennel: this is private and not used anywhere, i think you can safely remove it.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/UpdateNonRdfSourcePersisterTest.java
dbernstein: I think we want this in a separate PR / JIRA.  @bbpennel  would you mind creating a jira for this one?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/CreateRDFSourcePersister.java
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public CreateRDFSourcePersister(final OCFLObjectSessionFactory objectFactory,
+                                    final FedoraToOCFLObjectIndex index) {
+        super(RdfSourceOperation.class, CREATE, objectFactory, index);
+    }
+
+    @Override
+    public void persist(final OCFLPersistentStorageSession session, final ResourceOperation operation)
+            throws PersistentStorageException {
+        final FedoraOCFLMapping mapping = findOrCreateFedoraOCFLMapping(operation, session);
```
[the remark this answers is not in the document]
dbernstein: That's a valid point - I will change.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/UpdateRDFSourcePersisterTest.java
dbernstein: This persister should never be called when the resource does not exist.  That case will be handled by CreateRDFSourcePersister.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/UpdateNonRdfSourcePersisterTest.java
dbernstein: This persister should never be called when the resource does not exist. That case will be handled by CreateNonRDFSourcePersister.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/CreateRDFSourcePersisterTest.java
dbernstein: I believe we'll be checking at a higher level to make sure that you can create a resource that  already exists. Also,  when the fedora is initialized it will use this code path to create a repo root if none exists.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/UpdateNonRdfSourcePersisterTest.java
dbernstein: I believe we'll be checking at a higher level for this.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/UpdateRDFSourcePersisterTest.java
dbernstein: We'll check at a higher level for updating resources that don't exist.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/CreateRDFSourcePersisterTest.java
bbpennel: I guess its a bit edgy, but concurrent requests could attempt to create the same resource (simultaneous PUTs to the same resource being the most likely), unless there is going to be a locking mechanism further up. It can be nice to be explicit about expectations, but we can move ahead without the check if you feel confident it should be handled elsewhere.

I'd still recommend having a test to verify that you can't create an RDF or Non-RDF resource with the same ocfl id as the root since you have an explicit check for that https://github.com/fcrepo4/fcrepo4/blob/c0108e45a8632253a55555e519457c1d6ebdc1e3/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractPersister.java#L135

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/UpdateNonRdfSourcePersisterTest.java
bbpennel: https://jira.lyrasis.org/browse/FCREPO-3158

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/CreateNonRdfSourcePersister.java
```diff
@@ -0,0 +1,58 @@
...
+    CreateNonRdfSourcePersister(final OCFLObjectSessionFactory objectFactory,
+                                final FedoraToOCFLObjectIndex index) {
+        super(NonRdfSourceOperation.class, CREATE, objectFactory, index);
+    }
+
+    @Override
+    public void persist(final OCFLPersistentStorageSession session, final ResourceOperation operation) throws PersistentStorageException {
+        final String ocflId = resolveRootObjectId((CreateResourceOperation) operation, session);
+        final OCFLObjectSession ocflObjectSession = session.findOrCreateSession(ocflId);
+        final FedoraOCFLMapping mapping = findOrCreateFedoraOCFLMapping(operation, session);
```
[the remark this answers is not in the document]
dbernstein: I will make all the persisters package-private with protected constructors.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractPersister.java
```diff
@@ -84,13 +78,70 @@ protected static ResourceHeaders readHeaders(final OCFLObjectSession objSession,
...
+
+    protected FedoraOCFLMapping getMapping(final String resourceId) throws PersistentStorageException {
+        try {
+            return this.index.getMapping(resourceId);
+        } catch (FedoraOCFLMappingNotFoundException e){
+            throw new PersistentStorageException(e.getMessage());
+        }
+    }
+
+    protected FedoraOCFLMapping findOrCreateFedoraOCFLMapping(final ResourceOperation operation,
```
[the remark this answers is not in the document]
dbernstein: Yes that makes sense. Will do.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/CreateRDFSourcePersister.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * This class implements the persistence of a new RDFSource
+ *
+ * @author dbernstein
+ * @since 6.0.0
+ */
+public class CreateRDFSourcePersister extends AbstractRDFSourcePersister {
+
+    private static final Logger log = LoggerFactory.getLogger(CreateRDFSourcePersister.class);
```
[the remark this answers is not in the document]
dbernstein: will do.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractPersister.java
```diff
@@ -40,30 +43,21 @@
...
-
-    private final Set<Class<? extends ResourceOperation>> resourceOperations;
-
-    AbstractPersister(final Set<Class<? extends ResourceOperation>> resourceOperations, final Set<ResourceOperationType> resourceOperationType) {
-        this.resourceOperations = resourceOperations;
-        this.resourceOperationType = resourceOperationType;
-    }
+    private final Class<? extends ResourceOperation> resourceOperation;
+    private final ResourceOperationType resourceOperationType;
+    private final OCFLObjectSessionFactory objectFactory;
```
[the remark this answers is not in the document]
dbernstein: yes.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateNonRdfSourceOperation.java
```diff
@@ -71,4 +78,9 @@ public String getParentId() {
     public void setParentId(final String parentId) {
         this.parentId = parentId;
     }
+
+    @Override
+    public ResourceOperationType getType() {
+        return CREATE;
```
whikloj: Why are we overriding this with the same output? Maybe we don't need the `default` implementation [here](https://github.com/fcrepo4/fcrepo4/pull/1597/files#diff-98af73889d05ae57565ffea92694d7b6R44-R46)?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateRdfSourceOperation.java
```diff
@@ -63,4 +71,9 @@ public String getParentId() {
     public void setParentId(final String parentId) {
         this.parentId = parentId;
     }
+
+    @Override
+    public ResourceOperationType getType() {
+        return CREATE;
```
whikloj: Same question about overriding the default implementation?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
```diff
@@ -29,16 +29,28 @@
      * Retrieve identification information for the OCFL object which either contains, or is identified by,
      * the provided fedora resource id. In other words the method will find the closest resource that is persisted
      * as an OCFL object and returns its identifiers.
-     *
+     * <p>
```
whikloj: This is an IntelliJ thing, which you can disable.
`Preferences -> Editor -> Code Style -> Java -> JavaDoc -> Other -> Generate "<p>" on empty lines`

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractPersister.java
```diff
@@ -38,32 +40,18 @@
...
-
-    private final Set<ResourceOperationType> resourceOperationType;
-
-    private final Set<Class<? extends ResourceOperation>> resourceOperations;
-
-    AbstractPersister(final Set<Class<? extends ResourceOperation>> resourceOperations, final Set<ResourceOperationType> resourceOperationType) {
-        this.resourceOperations = resourceOperations;
-        this.resourceOperationType = resourceOperationType;
-    }
+abstract class AbstractPersister implements Persister {
```
whikloj: This should probably be explicit with `public` unless it should be `protected`?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractNonRdfSourcePersister.java
```diff
@@ -50,24 +46,25 @@
...
+                                  final FedoraToOCFLObjectIndex index) {
+        super(resourceOperation, resourceOperationType, index);
     }
 
-    @Override
-    public void persist(final OCFLObjectSession session, final ResourceOperation operation,
-            final FedoraOCFLMapping mapping) throws PersistentStorageException {
-        log.debug("persisting ({}) to {}", operation.getResourceId(), mapping.getOcflObjectId());
-        final String subpath = relativizeSubpath(mapping.getParentFedoraResourceId(), operation.getResourceId());
+    protected void persistNonRDFSource(final ResourceOperation operation,
```
whikloj: Needs javadoc

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractPersister.java
```diff
@@ -84,13 +72,45 @@ protected static ResourceHeaders readHeaders(final OCFLObjectSession objSession,
...
-        //ensure that at least one of them match.
-        for (final var c : this.resourceOperations) {
-            if (c.isInstance(operation)) {
-                //return true if the operation types match.
-                return this.resourceOperationType.contains(operation.getType());
-            }
+            return resourceOperation.isInstance(operation) && resourceOperationType.equals(operation.getType());
+    }
+
+    protected FedoraOCFLMapping getMapping(final String resourceId) throws PersistentStorageException {
```
whikloj: Needs javadoc

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractPersister.java
```diff
@@ -84,13 +72,45 @@ protected static ResourceHeaders readHeaders(final OCFLObjectSession objSession,
...
+        try {
+            return this.index.getMapping(resourceId);
+        } catch (FedoraOCFLMappingNotFoundException e){
+            throw new PersistentStorageException(e.getMessage());
         }
-        return false;
     }
+
+    protected FedoraOCFLMapping findMapping(final ResourceOperation operation,
```
whikloj: Needs javadoc

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractPersister.java
```diff
@@ -84,13 +72,45 @@ protected static ResourceHeaders readHeaders(final OCFLObjectSession objSession,
...
+        final String resourceId = operation.getResourceId();
+
+        try {
+            return index.getMapping(resourceId);
+        } catch (FedoraOCFLMappingNotFoundException e) {
+            throw new PersistentStorageException("Unable to resolve parent identifier for " + resourceId);
+        }
+    }
+
+    protected String resolveRootObjectId(final CreateResourceOperation operation,
```
whikloj: Needs javadoc

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractRDFSourcePersister.java
```diff
@@ -38,35 +25,45 @@
...
+    protected AbstractRDFSourcePersister(final Class<? extends ResourceOperation> resourceOperation,
+                               final ResourceOperationType resourceOperationType,
+                               final FedoraToOCFLObjectIndex index) {
+        super(resourceOperation, resourceOperationType, index);
     }
 
-    @Override
-    public void persist(final OCFLObjectSession session, final ResourceOperation operation,
-            final FedoraOCFLMapping mapping) throws PersistentStorageException {
+    protected void persistRDF(final OCFLObjectSession session, final ResourceOperation operation,
```
whikloj: Needs javadoc

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageSession.java
```diff
@@ -129,6 +125,14 @@ protected OCFLPersistentStorageSession(final String sessionId, final FedoraToOCF
         this.fedoraOcflIndex = fedoraOcflIndex;
         this.objectSessionFactory = objectSessionFactory;
         this.sessionMap = new ConcurrentHashMap<>();
+
+        //load the persister list if empty
+        persisterList.add(new CreateRDFSourcePersister(this.fedoraOcflIndex));
+        persisterList.add(new UpdateRDFSourcePersister(this.fedoraOcflIndex));
+        persisterList.add(new CreateNonRdfSourcePersister(this.fedoraOcflIndex));
+        persisterList.add(new UpdateNonRdfSourcePersister(this.fedoraOcflIndex));
+        persisterList.add(new DeleteResourcePersister(this.fedoraOcflIndex));
```
whikloj: I'm not sure if this makes difference, just wondering why we moved away from a static list of persisters to an instance list?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtils.java
```diff
@@ -74,26 +79,52 @@ private OCFLPersistentStorageUtils() {
...
+    public static String relativizeSubpath(final String rootObjectId, final String resourceId) {
+        if (resourceId.equals(rootObjectId)) {
+            return resourceId.substring(FEDORA_ID_PREFIX.length());
+        } else if (resourceId.startsWith(rootObjectId)) {
+
+            final var rawSubpath = resourceId.substring(rootObjectId.length() + 1);
+            //grab the last path segment of the rootObjectId
+            final var lastPathSegment = rootObjectId.substring(rootObjectId.lastIndexOf("/") + 1);
+
+            if (rawSubpath.endsWith(FCR_ACL) || rawSubpath.endsWith(FCR_METADATA)) {
```
bbpennel: Kind of wondering, could this logic be accomplished with a string replace, replacing `FCR_ACL` => `-description`, etc?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractPersister.java
whikloj: I don't think we have a policy. I was just trying to be able to understand this code if I have to work with it. As long as you and @dbernstein understand how this all works then I'm fine.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtils.java
whikloj: I'm wondering how this works with things like
```
info:fedora/some/binary/fcr:metadata/fcr:versions
```
or are we stripping off the versions endpoint before here?

But yes I think something like this could work (if I understand the purpose)
```
if (rawSubpath.endsWith(FCR_ACL)) {
  return lastPathSegment + rawSubpath.replace("/" + FCR_ACL, "-acl");
} else if (rawSubpath.endsWith(FCR_METADATA)) {
  return lastPathSegment + rawSubpath.replace("/" + FCR_METADATA, "-description");
} else {
  return lastPathSegment + "/" +  rawSubpath;
}
```

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtils.java
```diff
@@ -74,26 +79,52 @@ private OCFLPersistentStorageUtils() {
...
+                }
+
+                String translatedSuffix = null;
+                if (suffix.equals(FCR_METADATA)) {
+                    translatedSuffix = "-description";
+                } else {
+                    translatedSuffix = "-acl";
+                }
+
+                return lastPathSegment + (beforeSlash != null ? beforeSlash : "") + translatedSuffix;
```
whikloj: You don't add a slash between `lastPathSegment` and `beforeSlash` here but you add one between `lastPathSegment` and `rawSubpath` below?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractNonRdfSourcePersister.java
```diff
@@ -50,24 +52,26 @@
  * @author whikloj
  * @since 6.0.0
  */
-public class NonRdfSourcePersister extends AbstractPersister {
+abstract class AbstractNonRdfSourcePersister extends AbstractPersister {
```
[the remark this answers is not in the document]
dbernstein: class can't be protected.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageSession.java
```diff
@@ -184,7 +177,7 @@ private void ensureCommitNotStarted() throws PersistentSessionClosedException {
     }
 
 
-    private OCFLObjectSession findOrCreateSession(final String ocflId) {
+    OCFLObjectSession findOrCreateSession(final String ocflId) {
```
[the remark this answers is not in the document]
dbernstein: It's package protected, thus I believe it is correct as is.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageSession.java
```diff
@@ -218,27 +211,36 @@ public ResourceHeaders getHeaders(final String identifier, final Instant version
...
     /**
      * Returns a list of immutable versions associated with the specified fedora identifier
+     *
      * @param fedoraIdentifier The fedora identifier
      * @return The list of instants that map to the underlying versions
      * @throws PersistentStorageException Due the underlying resource not existing or is otherwise unreadable.
      */
-     List<Instant> listVersions(final String fedoraIdentifier) throws PersistentStorageException {
-        final FedoraOCFLMapping mapping = fedoraOcflIndex.getMapping(fedoraIdentifier);
+    List<Instant> listVersions(final String fedoraIdentifier) throws PersistentStorageException {
```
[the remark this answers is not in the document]
dbernstein: same as above.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/UpdateNonRdfSourcePersister.java
```diff
@@ -0,0 +1,55 @@
...
+
+import static org.fcrepo.kernel.api.operations.ResourceOperationType.UPDATE;
+
+/**
+ * This class implements the persistence of a NonRDFSource
+ *
+ * @author whikloj
+ * @since 6.0.0
+ */
+class UpdateNonRdfSourcePersister extends AbstractNonRdfSourcePersister {
```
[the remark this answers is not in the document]
dbernstein: It is correct as is, I believe. It doesn't need to be public.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/UpdateNonRdfSourcePersister.java
```diff
@@ -0,0 +1,55 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(UpdateNonRdfSourcePersister.class);
+
+    /**
+     * Constructor
+     */
+    UpdateNonRdfSourcePersister(final OCFLObjectSessionFactory objectFactory,
```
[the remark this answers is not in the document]
dbernstein: same as above.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/CreateNonRdfSourcePersisterTest.java
dbernstein: Checking for exisitng resources should be handled at a higher level.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/CreateRDFSourcePersisterTest.java
dbernstein: okay.  will do.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/api/FedoraToOCFLObjectIndex.java
dbernstein: 👍

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateNonRdfSourceOperation.java
dbernstein: tests were failing without it.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractPersister.java
dbernstein: It should not be visible outside the package.  `protected` does not apply to  non-inner classes

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateNonRdfSourceOperation.java
dbernstein: good point.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/operations/CreateRdfSourceOperation.java
dbernstein: valid point.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtils.java
```diff
@@ -88,32 +88,18 @@ private OCFLPersistentStorageUtils() {
...
-                    translatedSuffix = "-description";
-                } else {
-                    translatedSuffix = "-acl";
-                }
-
-                return lastPathSegment + (beforeSlash != null ? beforeSlash : "") + translatedSuffix;
+            if (subpath.endsWith(FCR_ACL)) {
+                return subpath.replace( "/" + FCR_ACL, "-acl");
+            } else if (rawSubpath.endsWith(FCR_METADATA)) {
+                return subpath.replace( "/" + FCR_METADATA, "-description");
```
bbpennel: Hopefully I'm not misinterpreting this, but it seems like if the root object id is something like `info:fedora/123/abc`, the block at line 90 will return `123/abc`. But if requesting resource `info:fedora/123/abc/sub/path/file.txt` inside the OCFL object, then the second block would return `abc/sub/path/file.txt` because of the `lastPathSegment` bit. Seems like we'd want a consistent representation of the root object id returned? Namely `123/abc/sub/path/file.txt`

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtils.java
```diff
@@ -88,32 +88,18 @@ private OCFLPersistentStorageUtils() {
...
             } else {
-                return lastPathSegment + "/" +  rawSubpath;
+                return subpath;
```
bbpennel: This might be totally legit, but i'm having a hard time thinking of what cases would have a root object id followed by something other than a `/`? I'm not sure hash urls would be resolved here.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/FedoraToOCFLObjectIndexImplTest.java
bbpennel: I don't think this has been resolved yet

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/FedoraToOCFLObjectIndexImplTest.java
```diff
@@ -48,4 +50,12 @@ public void test() throws Exception {
         assertEquals(ROOT_RESOURCE_ID, mapping1.getRootObjectIdentifier());
         assertEquals(OCFL_ID, mapping1.getOcflObjectId());
     }
+
+    @Test(expected = FedoraOCFLMappingNotFoundException.class)
+    public void testNotExists() throws Exception {
+        final FedoraToOCFLObjectIndexImpl index = new FedoraToOCFLObjectIndexImpl();
+
+
+        final FedoraOCFLMapping mapping1 = index.getMapping(RESOURCE_ID_1);
```
bbpennel: you don't need to store the result to a variable here (causes warning for me)

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtilsTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Test class for {@link OCFLPersistentStorageUtils}
+ *
+ * @author dbernstein
+ */
+public class OCFLPersistentStorageUtilsTest {
+
+    @Test
+    public void testRelativizeSubpath() {
```
bbpennel: Add a test where `rootObjectId` is the same as `resourceId`.
And maybe one where there is stuff after the root id other than a `/` if that's a valid case?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtils.java
bbpennel: Good question about `fcr:versions`. It probably gets stripped off before this point since it goes through a different controller and has to be resolved into requesting a timemap rather than a file on disk.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractNonRdfSourcePersister.java
```diff
@@ -56,22 +50,21 @@
...
+        super(resourceOperation, resourceOperationType, index);
     }
 
-    protected void persistNonRDFSource(final ResourceOperation operation, final FedoraOCFLMapping mapping,
-                                       final OCFLObjectSession objectSession) throws PersistentStorageException {
-        log.debug("persisting ({}) to {}", operation.getResourceId(), mapping.getOcflObjectId());
-        final String subpath = relativizeSubpath(mapping.getRootObjectIdentifier(), operation.getResourceId());
+    protected void persistNonRDFSource(final ResourceOperation operation,
+                                       final OCFLObjectSession objectSession, final String rootIdentifier,
+                                       final String ocflId) throws PersistentStorageException {
```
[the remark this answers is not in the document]
bbpennel: This still should be addressed

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractPersister.java
bbpennel: only having two people able to understand the code doesn't seem like a good goal to me, so i'm in favor of adding documentation

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractPersister.java
```diff
@@ -40,30 +43,21 @@
...
 
-    private final Set<ResourceOperationType> resourceOperationType;
-
-    private final Set<Class<? extends ResourceOperation>> resourceOperations;
-
-    AbstractPersister(final Set<Class<? extends ResourceOperation>> resourceOperations, final Set<ResourceOperationType> resourceOperationType) {
-        this.resourceOperations = resourceOperations;
-        this.resourceOperationType = resourceOperationType;
-    }
+    private final Class<? extends ResourceOperation> resourceOperation;
```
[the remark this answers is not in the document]
bbpennel: don't think this has been addressed ^

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/FedoraToOCFLObjectIndexImplTest.java
dbernstein: will remove.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtils.java
dbernstein: Yes - I agree.  I believe this should be fixed with the last update.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtilsTest.java
dbernstein: Test added for root equals resource.
I'm not sure I understand the second issue you're raising.  Are you talking about something other than nested subpaths?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtils.java
dbernstein: I put that in there just to ensure that an illegal argument exception  is thrown when 
root = info:fedora/test/a 
resource = info:fedora/test/ab

dbernstein: I put the check for the  trailing slash there to ensure that an error is thrown when 

root =         info:fedora/test/a
resource = info:fedora/test/ab

but not when

resource = info:fedora/test/a/b

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtils.java
dbernstein: Yes - that's my understanding too.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtils.java
dbernstein: This has been addressed.

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtilsTest.java
bbpennel: The second case is related to the second half of the conditional here:
https://github.com/fcrepo4/fcrepo4/blob/fcfea26ded89335cfc1802a1994b1792eac96d9a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtils.java#L97

I think i misread the conditional yesterday though, on second look i'm realizing it is disallowing the case where there is no slash after the root id, which you already have a test for.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtils.java
```diff
@@ -81,17 +81,22 @@ private OCFLPersistentStorageUtils() {
...
      * Returns the relative subpath of the resourceId based on the root object's resource id.
      *
-     * @param rootObjectId The fedora root object identifier
-     * @param resourceId   The identifier of the resource whose subpath you wish to resolve.
+     * @param rootFedoraObjectId The fedora root object identifier
+     * @param fedoraResourceId   The identifier of the resource whose subpath you wish to resolve.
      * @return The resolved subpath
      */
-    public static String relativizeSubpath(final String rootObjectId, final String resourceId) {
+    public static String relativizeSubpath(final String rootFedoraObjectId, final String fedoraResourceId) {
```
bbpennel: Normally I'd expect a relative subpath to not include the parent, so given:
rootid = `info:fedora/my/root`
resource id = `info:fedora/my/root/some/more/stuff`
I'd expect the relative path to be `some/more/stuff`, but at the moment returns `root/some/more/stuff`. The former is how `Paths.relativize` works, although I understand not everything has to follow this model.

However, the result of this method appears to be passed directly into object session methods. If i'm understanding correctly, the session methods would be working with a subpath within the ocfl object that includes the last segment of the id of the root object, which I don't think is the structure we intended.

In other words, on disk I think we'd get 
`ocfl_root/my_root/v001/data/content/root/some/more/stuff`
whereas I would expect 
`ocfl_root/my_root/v001/data/content/some/more/stuff` 
since we can't assume that imported OCFL is going to put the contents of an object into a data subdirectory named after the ocfl object. I haven't tried running all the parts together to verify this though, this is just based off of my reading.

Also, `OCFLPersistentStorageUtils.resolveOCFLSubpath` currently only strips off the fcr:metadata suffix, but that already happens in `relativizeSubpath`. All calls of `resolveOCFLSubpath` receive the result of `relativizeSubpath`, so maybe we can drop `resolveOCFLSubpath` now?

PR #1606, opened 2020-01-20
Stage object sessions in their own subdirectories.

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3174

Currently all storage requests generate operations that need to be checked for changes. These all share the exact same directory so when we do the check of operations the first operation persists all files to its OCFL object. This causes an error.

# How should this be tested?

It still doesn't work but I was `POST`ing a new resource and would get a error related to rolling back the transaction. With this change you should get an error related to adding cache control headers. (See next PR for work on that)

# Interested parties
@fcrepo4/committers


whikloj:
That is a good idea, I'll see if I can whip one up.

PR #1607, opened 2020-01-20
Return identifier from CreateResourceService

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3175

We need to return the identifier from the kernel/service level to the HTTP level to send out a response. This PR starts returning the identifier from the service layer.

# How should this be tested?

This will still throw errors on a POST, but its incremental and tests have been updated.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImplTest.java
+++ b/fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImplTest.java
@@ -158,16 +158,14 @@ public class CreateResourceServiceImplTest {
     public void testNoParentRdf_Post() throws Exception {
         final String fedoraId = UUID.randomUUID().toString();
         when(psSession.getHeaders(fedoraId, null)).thenThrow(PersistentItemNotFoundException.class);
-        createResourceService.perform(TX_ID, USER_PRINCIPAL, fedoraId, null, true, null,
-                model);
+        createResourceService.perform(TX_ID, USER_PRINCIPAL, fedoraId, null, true, null, model);
     }
 
     @Test
     public void testNoParentRdf_Put() throws Exception {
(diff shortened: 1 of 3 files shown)


fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImplTest.java
```diff
@@ -158,52 +158,60 @@ public void setUp() {
     public void testNoParentRdf_Post() throws Exception {
         final String fedoraId = UUID.randomUUID().toString();
         when(psSession.getHeaders(fedoraId, null)).thenThrow(PersistentItemNotFoundException.class);
-        createResourceService.perform(TX_ID, USER_PRINCIPAL, fedoraId, null, true, null, model);
```
bbpennel: think some accidental parameter squishing happened here, should be a space after `true,`

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImplTest.java
```diff
@@ -443,90 +469,90 @@ public void testWithSlugDoesntExistsExternal() throws Exception {
     public void testSendingValidInteractionModel() {
         // If you provide a valid interaction model, you should always get it back.
         final String expected = BASIC_CONTAINER.toString();
-        final String model1 = createResourceService.determineInteractionModel(STRING_TYPES_VALID, false, false,
-                false);
+        final String model1 = createResourceService.determineInteractionModel(STRING_TYPES_VALID, false,
+                false, false);
```
bbpennel: looks like your IDE went to town around here, but i think its fine.

whikloj:
@bbpennel I'm sorry I tried, but my Intellij experience might just be making it worse.

PR #1608, opened 2020-01-22
doesResourceExist function

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3137

Adds a doesResourceExist method to the ResourceFactory

# How should this be tested?

Just review, I have not added this to any other code yet.

# Interested parties
@fcrepo4/committers


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/GetResourceServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+    @Override
+    public FedoraResource getResource(final Transaction transaction, final String fedoraId, final Instant version) {
+        final PersistentStorageSession psSession;
+        if (transaction == null) {
+            psSession = psManager.getReadOnlySession();
+        } else {
+            psSession = psManager.getSession(transaction.getId());
+        }
+        try {
+            final ResourceHeaders headers = psSession.getHeaders(fedoraId, version);
```
bbpennel: ResourceFactory has `getResource` which includes logic for determining what interaction model a resource uses and initializing the correct FedoraResource class based off it:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java#L84
So I think you can call that rather than going through `getHeaders`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/GetResourceServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+            final FedoraResource resource;
+            if (headers.getInteractionModel().equals(NON_RDF_SOURCE.toString())) {
+                resource = new BinaryImpl(fedoraId, transaction, psManager, rsFactory);
+            } else {
+                resource = new ContainerImpl(fedoraId, transaction, psManager, rsFactory);
+            }
+            // Commit session so it doesn't hang around.
+            psSession.commit();
```
bbpennel: I don't think we want to commit the session if the transaction existed before this service was called, otherwise it might prematurely commit changes. Doesn't seem like we have a `commitIfShortLived` equivalent for `PersistentStorageSession`, so might have to rely on checking if `transaction` was null or not.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/GetResourceServiceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+        final PersistentStorageSession psSession;
+        if (transaction == null) {
+            psSession = psManager.getReadOnlySession();
+        } else {
+            psSession = psManager.getSession(transaction.getId());
+        }
+        try {
+            psSession.getHeaders(fedoraId, version);
+            // Commit session so it doesn't hang around.
+            psSession.commit();
```
bbpennel: Same thing as in the other method

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/GetResourceServiceImpl.java
bbpennel: Also, we should probably commit in a finally block, otherwise we might leak sessions when exceptions occur

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/GetResourceServiceImplTest.java
```diff
@@ -0,0 +1,137 @@
...
+    public void setUp() {
+        when(psManager.getReadOnlySession()).thenReturn(psSession);
+        when(psManager.getSession(ArgumentMatchers.any())).thenReturn(psSession);
+        when(transaction.getId()).thenReturn(UUID.randomUUID().toString());
+        setField(service, "psManager", psManager);
+        setField(service, "rsFactory", rsFactory);
+    }
+
+    @Test
+    public void doesResourceExists() throws Exception {
```
bbpennel: Finding some of these tests names a bit hard to read grammatically. Bit redundant, maybe `doesResourceExistExists` or something along those lines. Similarly, `doesResourceExistDoesntExist`, etc, so that the method being tested is consistently in there. Or maybe some kind of delimiter between `doesResource_Exists` to split what's being tested from the state being tested.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/GetResourceServiceImplTest.java
```diff
@@ -0,0 +1,137 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentMatchers;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.Silent.class)
+public class GetResourceServiceImplTest {
```
bbpennel: Might be worth adding tests to verify the expected session behaviors depending on if a transaction is provided or not.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/GetResourceServiceImpl.java
whikloj: The commit here is on a persistent storage session we just created. Would that not mean nothing else has occurred in this session and should not cause trouble? I'm a little unclear about how the various sessions work so i could be missing something.

whikloj:
@bbpennel Moved the single function into ResourceFactory, so you'll probably want to have another look at this.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/GetResourceServiceImplTest.java
whikloj: Added in https://github.com/fcrepo4/fcrepo4/pull/1608/commits/40e294ecc2a3cfdc9fd41b4eb9ccb5cce81451d0

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/GetResourceServiceImpl.java
whikloj: I have moved the commit to a finally block, but as it also can throw an exception I have had to squash and log that exception. Not sure if that is an acceptable trade-off.

bbpennel: Since it retrieves a persistent session using the transaction when present, it could grab an ongoing modification session. What you have here looks good to me, it failing to close the temporary session doesn't really impact that rest of the operation, so logging is fine.

bbpennel:
Its a little hard to judge whether or not GetResourceService would have been helpful without these being used anywhere, but i think this is good now.

PR #1611, opened 2020-01-27
Ensures that POSTs without explicit Content-Type headers use turtle b…

dbernstein:
…y default.

Resolves:  https://jira.lyrasis.org/browse/FCREPO-3188

* * *

**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3188

* Ensures that POSTs without an explicit Content-Type header do not fail. Rather, they assume text/turtle.
* Formerly @Ignore'd integration tests now work.  In the process of reenabling the integration tests, I also fixed a bug related to link header parsing.

# How should this be tested?

`
# create the resource
curl -v -XPOST -H "Slug: blah" -ufedoraAdmin:fedoraAdmin -H "Link: <http://www.w3.org/ns/ldp#BasicContainer>;rel=\"type\"" http://localhost:8080/rest

#Verify that the operation succeeds and a subsequent GET returns the resource. content type text/turtle
`
curl -I -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/blah
`

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -47,6 +47,9 @@ import static org.fcrepo.http.commons.domain.RDFMediaType.TEXT_HTML_WITH_CHARSET
 import static org.fcrepo.http.commons.domain.RDFMediaType.TEXT_PLAIN_WITH_CHARSET;
 import static org.fcrepo.http.commons.domain.RDFMediaType.TURTLE_WITH_CHARSET;
 import static org.fcrepo.http.commons.domain.RDFMediaType.TURTLE_X;
+import static org.fcrepo.http.commons.domain.RDFMediaType.TURTLE_TYPE;
+import static org.fcrepo.http.commons.domain.RDFMediaType.APPLICATION_OCTET_STREAM_TYPE;
+
 import static org.fcrepo.kernel.api.RdfLexicon.INTERACTION_MODEL_RESOURCES;
 import static org.fcrepo.kernel.api.RdfLexicon.VERSIONED_RESOURCE;
 import static org.fcrepo.kernel.api.services.VersionService.MEMENTO_RFC_1123_FORMATTER;
@@ -89,7 +92,6 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.jena.rdf.model.Model;
 import org.apache.jena.rdf.model.Resource;
 import org.fcrepo.http.commons.domain.PATCH;
-import org.fcrepo.http.commons.domain.RDFMediaType;
 import org.fcrepo.kernel.api.FedoraTypes;
 import org.fcrepo.kernel.api.RdfStream;
 import org.fcrepo.kernel.api.exception.AccessDeniedException;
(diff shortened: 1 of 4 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -436,15 +441,17 @@ public Response createOrReplaceObjectRdf(
                         .toString(),
                     model);
             } else {
-                createResourceService.perform(transaction.getId(), getUserPrincipal(), fedoraId, null, false, links,
+                fedoraId = createResourceService.perform(transaction.getId(), getUserPrincipal(), fedoraId, null,
```
bbpennel: Should probably have this on two lines rather than going to three, or one line per parameter.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -135,6 +136,7 @@
     private static final String WANT_DIGEST = "Want-Digest";
 
     private static final String DIGEST = "Digest";
+    public static final MediaType DEFAULT_CONTENT_TYPE = RDFMediaType.TURTLE_TYPE;
```
bbpennel: While this is important, I think this is only the default content type for rdf sources. For non-rdf sources it should be `application/octet-stream` I would think?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -548,41 +555,39 @@ public Response createObject(@HeaderParam(CONTENT_DISPOSITION) final ContentDisp
...
+            newFedoraId = createResourceService.perform(transaction.getId(), getUserPrincipal(), fedoraId, slug,
+                    true, links,
+                    model);
         }
-        // TODO: How to generate a response.
-        LOGGER.debug("Finished creating resource with path: {}", externalPath());
 
-        final FedoraResource resource;
+        LOGGER.debug("Finished creating resource: {}", newFedoraId);
+        transaction.commit();
```
bbpennel: I see you ran into this commit order issue too. I'll just add that it should be `commitIfShortLived` so that it won't end a multi-transaction commit prematurely. That holds true for all the `commit` in this class actually.

dbernstein: will fix.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
dbernstein: will fix

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
dbernstein: will fix.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
dbernstein: This is fixed.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
dbernstein: done

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
dbernstein: done.

bbpennel:
The command `curl -v -XPOST -H "Slug: blah" -ufedoraAdmin:fedoraAdmin -H "Link: <http://www.w3.org/ns/ldp#BasicContainer>; rel='type'" http://localhost:8080/rest` worked as expected

PR #1612, opened 2020-01-27
FCREPO-3179 - POST to Create Binaries and descriptions

bbpennel:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3179

Binaries and their descriptions should not be created via POST.

# What's new?
* Implementation and creation of descriptions (Non-RDF source descriptions)
* Filling out and cleanup of Binary classes.
  * getProxyUrl and getRedirectUrl were merged since they are now identical
* Some fixes regarding committing transactions for POST
* Avoid NPE if no content disposition provided during binary creation.

# How should this be tested?
There are updated and reenabled tests. There are not many ways to manually test this until GET is implemented for binaries.

# Additional Notes:
Note: digests are not evaluated yet.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/NonRdfSourceDescriptionImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/NonRdfSourceDescriptionImpl.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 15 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/NonRdfSourceDescriptionImpl.java
```diff
@@ -0,0 +1,52 @@
...
+    public NonRdfSourceDescriptionImpl(final String id,
+            final Transaction tx,
+            final PersistentStorageSessionManager pSessionManager,
+            final ResourceFactory resourceFactory) {
+        super(id, tx, pSessionManager, resourceFactory);
+    }
+
+    @Override
+    public FedoraResource getDescribedResource() {
+        // TODO must return the described binary
```
dbernstein: Maybe create a JIRA if it doesn't already exist so we don't lose track of it?

dbernstein:
@bbpennel :  there's just one comment regarding a JIRA to be created and of course the rebase.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/NonRdfSourceDescriptionImpl.java
bbpennel: I created https://jira.lyrasis.org/browse/FCREPO-3190 which I think will include this

PR #1613, opened 2020-01-28
FCREPO-3079 - GET binary

bbpennel:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3079

Enables retrieval of binaries at the persistence and http layers.

# What's new?
* Implements getBinaryContent at persistence layer
* Reorganization of @GET method to skip unnecessary creation of rdf stream when getting binary content, and to separate binary from rdf retrieval, similar to how the HEAD method is organized.
  * this included removing binary retrieval from the `getContent` method, and removing the `Range` parameter from various API methods that were only receiving it to pass it to `getContent` (where it was ignored since they didn't retrieve binaries)
* Fix issue with content size of binaries when content disposition is present but no content size specified
* Removes some more unused binary
* Reenables some binary related integration tests
  * many still can't be enabled because they depend on `PUT`

# How should this be tested?
Various tests, and should be able to POST to create a binary with one click, and then perform a GET on it to retrieve the file.

# Additional Notes:
Still doesn't support digest related features, and still many binary retrieval integration tests disabled.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageSession.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageSession.java
@@ -20,6 +20,7 @@ package org.fcrepo.persistence.ocfl.impl;
 import static java.lang.String.format;
 import static org.fcrepo.persistence.ocfl.impl.OCFLPersistentStorageUtils.getSidecarSubpath;
 import static org.fcrepo.persistence.ocfl.impl.OCFLPersistentStorageUtils.resolveVersionId;
+import static org.fcrepo.persistence.ocfl.impl.OCFLPersistentStorageUtils.getBinaryStream;
 
 import java.io.InputStream;
 import java.time.Instant;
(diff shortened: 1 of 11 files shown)


fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageSessionTest.java
```diff
@@ -591,4 +597,97 @@ public void getTriplesFromPreviousVersion() throws Exception {
...
+        // perform the create non-rdf source operation
+        session.persist(binOperation);
+
+        // create a new session and verify the returned rdf stream.
+        final var result = IOUtils.toString(session.getBinaryContent(RESOURCE_ID, null), UTF_8);
+
+        assertEquals(BINARY_CONTENT, result);
+    }
+
+    private NonRdfSourceOperation mockNonRdfSourceOperation(final String content,
```
whikloj: Nit-pick, maybe the name should specify that you are mocking a Create NonRdfSource Operation. Eventually we might need one for Update as well.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageSession.java
```diff
@@ -253,9 +254,16 @@ public RdfStream getTriples(final String identifier, final Instant version)
 
     @Override
     public InputStream getBinaryContent(final String identifier, final Instant version)
```
whikloj: Should this not take a transaction? What if the binary is created and then `GET`ted inside a single long-running transaction?

bbpennel: The class this method is in is effectively the transaction at the persistence layer, so it should be resolving to the state of the long running transaction that the GET is taking place in

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageSessionTest.java
bbpennel: updated

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageSession.java
whikloj: This went 90% over my head, but I think I got enough to be okay.

PR #1614, opened 2020-01-29
FCREPO-3177 - Return 303 on GET for non-existant resource

robyj:
- Added unique PathNotFoundExceptionMapper and mock test for that file

Resolves: https://jira.lyrasis.org/browse/FCREPO-3177

* * *

https://jira.lyrasis.org/browse/FCREPO-3177

Instead of generating a 500 server error when trying to GET a non existant resource, the server
gives you a 404 error

# What's new?
Created new ExceptionMapper for PathNotFoundException and mockito test to test it

curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/blah

# Interested parties
@dbernstein 
@fcrepo4/committers


fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapper.java
```diff
@@ -0,0 +1,62 @@
...
+
+    Boolean showStackTrace = true;
+
+    private static final Logger LOGGER =
+        getLogger(PathNotFoundExceptionMapper.class);
+
+    @Override
+    public Response toResponse(final PathNotFoundException e) {
+
+        LOGGER.error("Exception intercepted by PathNotFoundExceptionMapper: {}\n", e.getMessage());
```
bbpennel: Don't think we need to keep this log statement since the next line logs the error at debug level. Maybe reduce it down to `LOGGER.debug("Exception intercepted by PathNotFoundExceptionMapper")` so that if users want debug mode they would see that it was caught by this handler, but otherwise i don't think its significant

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapper.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Catch PathNotFoundException
+ *
+ * @author robyj
+ */
+@Provider
+public class PathNotFoundExceptionMapper implements
+        ExceptionMapper<PathNotFoundException>, ExceptionDebugLogging {
+
+    Boolean showStackTrace = true;
```
bbpennel: this variable doesn't appear to be used, i think you can remove it and its setter

fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapperTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import javax.ws.rs.core.Response;
+
+import org.fcrepo.kernel.api.exception.PathNotFoundException;
+
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * <p>PathNotFoundExceptionMapperTest class.</p>
```
bbpennel: Remove the `<p>` tags. I think there is a way to get intellij to not generate these, you might want to ask in slack.

fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapperTest.java
```diff
@@ -0,0 +1,59 @@
...
+    public void setUp() {
+        testObj = new PathNotFoundExceptionMapper();
+    }
+
+    @Test
+    public void testToResponse() {
+        final PathNotFoundException input = new PathNotFoundException("xyz");
+        Response actual = testObj.toResponse(input);
+        assertEquals(NOT_FOUND.getStatusCode(), actual.getStatus());
+        assertTrue(actual.getEntity() != "xyz");
```
bbpennel: I'd suggest doing a positive check here. The entity is "Error: xyz", you could convert the entity to a string and either do assertEquals to the whole value, or do `.contains("xyz")` to check that just the expected error message is present.

fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapperTest.java
```diff
@@ -0,0 +1,59 @@
...
+        testObj.showStackTrace = false;
```
bbpennel: I don't think this line or the ones below it are testing anything new, and can likely be removed. `showStackTrace` doesn't currently do anything, and the response is the same in the first and second run.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapper.java
robyj: agreed

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapper.java
robyj: agreed. i'm removing the line plus the function that uses it

fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapperTest.java
robyj: tags are being removed

fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapperTest.java
robyj: I've changed it to assertEquals(actual.getEntity(),"Error: xyz")

fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapperTest.java
robyj: Mostly agreed. I've left the test assertNotEquals(INTERNAL_SERVER_ERROR) just so nothing in the future causes a server to send 500 instead of 404 (or 303 :p)

PR #1616, opened 2020-01-30
Added unique PathNotFoundExceptionMapper and mock test for that file

robyj:
Resolves https://jira.lyrasis.org/browse/FCREPO-3177

* * *

https://jira.lyrasis.org/browse/FCREPO-3177

Solves the issue where using GET to get info on a non-existant item returns a 500 server error code, it should return a 404 error code. This code adds a specific exception handler for the PathNotFoundException that returns a 404 server error code whereas before it fell through to a WildcardException that would return a 500 error code. There is also a mockito test unit for the new exception handler

# How should this be tested?
curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/blah
 

# Interested parties

### DIFF
--- a/fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapperTest.java
+++ b/fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapperTest.java
@@ -17,10 +17,8 @@
  */
 package org.fcrepo.http.commons.exceptionhandlers;
 
-import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
 import static javax.ws.rs.core.Response.Status.NOT_FOUND;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotEquals;
 
 import javax.ws.rs.core.Response;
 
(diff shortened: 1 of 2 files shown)


fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapperTest.java
```diff
@@ -0,0 +1,55 @@
...
+    public void setUp() {
+        testObj = new PathNotFoundExceptionMapper();
+    }
+
+    @Test
+    public void testToResponse() {
+        final PathNotFoundException input = new PathNotFoundException("xyz");
+        final Response actual = testObj.toResponse(input);
+        assertEquals(NOT_FOUND.getStatusCode(), actual.getStatus());
+        assertEquals(actual.getEntity(),"Error: xyz");
```
bbpennel: space after ,

fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapperTest.java
```diff
@@ -0,0 +1,55 @@
...
+        assertNotEquals(INTERNAL_SERVER_ERROR.getStatusCode(), actual.getStatus());
```
bbpennel: Line 51 is checking that `actual.getStatus()` equals 404, so i don't think its possible for this line to ever receive 500, since line 51 would have failed in that case.

fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapperTest.java
robyj: completed

fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/PathNotFoundExceptionMapperTest.java
robyj: the check against INTERNAL_SERVER_ERROR and the imports used for that check have been removed

bbpennel:
I think we might want to improve the response message at some point (right now it contains the internal id and a persistence error, which is way different than the html 404 page we used to get), but i think this is good

PR #1618, opened 2020-01-31
FCREPO-3148 - Transmission Fixity checks of binaries

bbpennel:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3148

Implements transmission fixity checks for binaries during write

# What's new?
* Adds a utility for performing multiple digests while reading an input stream
* Adds support for sha512 and sha512/256 digests
* Enables related ITs
* Changes InvalidChecksumException to a runtime exception since it does not generally need to be caught and handle by fedora, aside from the exception/response handler for it.

# How should this be tested?
Running the tests, and can performing commands like:
```
$ curl -ufedoraAdmin:fedoraAdmin -XPUT -H"Content-type: text/xml" -H"digest: sha=a9d6457ac8a0306dc46f86c1dca8c0feedaedf38,sha512/256=66ea1db10971c6bc8201909dcb3703c97075b22494397fb0f6caacbb958017d4" --data-binary "@pom.xml" http://localhost:8080/rest/testpom3

http://localhost:8080/rest/testpom3

$ curl -ufedoraAdmin:fedoraAdmin -XPUT -H"Content-type: text/xml" -H"digest: sha=a9d6457ac8a0306dc46f86c1dca8c0feedaedf38,sha512/256=bada1db197fb0f6caacbb958017d4" --data-binary "@pom.xml" http://localhost:8080/rest/testpom3

Checksum mismatch, computed SHA-512/256 digest 66ea1db10971c6bc8201909dcb3703c97075b22494397fb0f6caacbb958017d4 did not match expected value bada1db10971c6bc8201909dcb3703c97075b22494397fb0f6caacbb958017d4
```

# Additional Notes:
None.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-persistence-common/src/main/java/org/fcrepo/persistence/common/MultiDigestInputStreamWrapper.java
+++ b/fcrepo-persistence-common/src/main/java/org/fcrepo/persistence/common/MultiDigestInputStreamWrapper.java
@@ -1,0 +1,109 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.persistence.common;
+
+import static java.lang.String.format;
+import static org.apache.commons.codec.binary.Hex.encodeHexString;
+import static org.apache.commons.lang3.StringUtils.substringAfterLast;
+import static org.fcrepo.kernel.api.utils.ContentDigest.getAlgorithm;
+
+import java.io.InputStream;
+import java.net.URI;
+import java.security.DigestInputStream;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.fcrepo.kernel.api.exception.InvalidChecksumException;
+import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
+
+/**
+ * Wrapper for an InputStream that allows for the computation and evaluation
+ * multiple digests at once
+ *
+ * @author bbpennel
+ */
+public class MultiDigestInputStreamWrapper {
+
+    private InputStream sourceStream;
+
+    private Map<String, String> algToDigest;
+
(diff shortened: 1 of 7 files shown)


fcrepo-persistence-common/src/main/java/org/fcrepo/persistence/common/MultiDigestInputStreamWrapper.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.fcrepo.kernel.api.exception.InvalidChecksumException;
+import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
+
+/**
+ * Wrapper for an InputStream that allows for the computation and evaluation
+ * multiple digests at once
```
dbernstein: multiple -> of multiple

fcrepo-persistence-common/src/main/java/org/fcrepo/persistence/common/MultiDigestInputStreamWrapper.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    /**
+     * After consuming the inputstream, verify that all of the computed digests
+     * matched the provided digests.
+     *
+     * @throws InvalidChecksumException thrown if any of the digests did not match
+     */
+    public void checkFixity() throws InvalidChecksumException {
+        for (final var entry: algToDigestStream.entrySet()) {
```
whikloj: It seems like `checkFixity` relies on the code having called `getInputStream` first. If you call `checkFixity` without calling `getInputStream` `algToDigestStream` would have no entries so this doesn't check anything.

I think.

bbpennel: that's correct, I tried to document this requirement in the javadoc for this method with "After consuming the inputstream"

whikloj: Maybe move the digest wrapping stuff to a private method that can be called if `algToDigestStream.size() != algToDigest.size()` ? Otherwise just use the existing `algToDigestStream`

bbpennel: If you call checkFixity before getInputStream is called then you'd get a checksum mismatch because the digests it'd get back would be for a 0 byte stream, since nothing had been read so far. 

I could add an explicit failure if getInputStream hadn't been called yet, mainly as a warning to developers. It still wouldn't guarantee that the stream had been read though. I guess it could check that the stream was closed? Not 100% sure that's reliable, may depend on the type of inputstream...

bbpennel: Okay, I added a check in to explicitly make sure getInputStream was called first

fcrepo-persistence-common/src/main/java/org/fcrepo/persistence/common/MultiDigestInputStreamWrapper.java
bbpennel: fixed

PR #1620, opened 2020-01-31
Cleanup OCFL after tests

bbpennel:
**JIRA Ticket**: N/A

Adds a utility which sets up an OCFL base directory and related system properties in a location that gets cleaned up when shutdown occurs.

# What's new?
Also switches OCFL directory constants over to directories so that the values can more reliably be programmatically changed, as in the case of tests.

# How should this be tested?
Running the tests and verifying that no OCFL junk is left over (used to go into a java temp dir)

# Additional Notes:

# Interested parties
@awoods

### DIFF
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/TemporaryOCFLRepositoryHelper.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/TemporaryOCFLRepositoryHelper.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.persistence.ocfl;
+
+import static org.fcrepo.persistence.ocfl.impl.OCFLConstants.OCFL_STAGING_PROPERTY;
+import static org.fcrepo.persistence.ocfl.impl.OCFLConstants.OCFL_STORAGE_ROOT_PROPERTY;
+import static org.fcrepo.persistence.ocfl.impl.OCFLConstants.OCFL_WORK_PROPERTY;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import org.apache.commons.io.FileUtils;
+
+/**
+ * Utility which sets up a temporary base directory for OCFL, establishes
+ * system properties to make use of it, and cleans up on shutdown.
+ *
+ * @author bbpennel
+ */
+public class TemporaryOCFLRepositoryHelper {
+
+    /**
+     * Initialize the helper
+     *
+     * @throws IOException thrown if the base directory cannot be created
+     */
+    public TemporaryOCFLRepositoryHelper() throws IOException {
+        final Path ocflBase = Files.createDirectory(Paths.get("target/ocfl_base"));
+        final Path ocflStagingDir = ocflBase.resolve("staging");
+        final Path ocflStorageRootDir = ocflBase.resolve("storage_root");
+        final Path ocflWorkDir = ocflBase.resolve("work");
+
+        System.setProperty(OCFL_STAGING_PROPERTY, ocflStagingDir.toString());
+        System.setProperty(OCFL_STORAGE_ROOT_PROPERTY, ocflStorageRootDir.toString());
+        System.setProperty(OCFL_WORK_PROPERTY, ocflWorkDir.toString());
+
+        Runtime.getRuntime().addShutdownHook(new Thread(() ->
+                FileUtils.deleteQuietly(ocflBase.toFile())));
+    }
+}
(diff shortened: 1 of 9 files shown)


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/TemporaryOCFLRepositoryHelper.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.apache.commons.io.FileUtils;
+
+/**
+ * Utility which sets up a temporary base directory for OCFL, establishes
+ * system properties to make use of it, and cleans up on shutdown.
+ *
+ * @author bbpennel
+ */
+public class TemporaryOCFLRepositoryHelper {
```
awoods: This class is only used in tests. It would be ideal if we could move it into `src/text/java/...` ...which might require having the modules that use it in tests to include a "test-jar", such as:
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-commons/pom.xml#L281-L303
and
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-integration-rdf/pom.xml#L62-L68

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -4102,6 +4102,9 @@ public void testConcurrentPutsWithPairtrees() throws InterruptedException, IOExc
     }
 
     @Test
+    // TODO Test flaps in travis, causing more PUTs to succeed some of the time, possibly
+    // because the etag is timestamp based. Should also make sure the correct failure status returned
+    @Ignore
```
awoods: This seems unrelated to the current work.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/TemporaryOCFLRepositoryHelper.java
bbpennel: I had considered doing that, but wasn't sure I wanted to add fcrepo-persistence-ocfl as a test dependency everywhere. Guess I should get over that though, I'll add it.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
bbpennel: It is, but the test is failing all builds. It passed in the original commit that enabled it though, and passes locally. I could open a separate PR if you prefer.

awoods: I do not like creating unnecessary work... but if this commit gets squashed with the work you are doing on build directories, it may be an odd mix when analyzing in the future.

If we are going to turn off the test, should we have a JIRA (linked in the @Ignore annotation) as a reminder to turn it back on?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/TemporaryOCFLRepositoryHelper.java
awoods: Would it be possible to set the system-properties in the surefire/pom.xml configurations?
```
        System.setProperty(OCFL_STAGING_PROPERTY, ocflStagingDir.toString());
        System.setProperty(OCFL_STORAGE_ROOT_PROPERTY, ocflStorageRootDir.toString());
        System.setProperty(OCFL_WORK_PROPERTY, ocflWorkDir.toString());
```

..or some such thing?

bbpennel: Hmm, it is possible, but it doesn't seem to pick up the properties when i run the tests in my IDE, only when I run them via maven. Sounds like eclipse may simply not support it:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=388683
Not sure if intellij is more observant. I might be inclined to stick with the utility since I almost always run tests via my IDE, but what do you think? I can remove the cleanup on shutdown to allow for inspection.

bbpennel: If you'd like to test, this branch has surefire properties instead of the util:
https://github.com/fcrepo4/fcrepo4/tree/ocfl-temp
If you run any tests in FedoraLdpIT it will log the OCFL paths near the beginning of the run.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
bbpennel: I've moved it over to https://github.com/fcrepo4/fcrepo4/pull/1623 and added a note about it to https://jira.lyrasis.org/browse/FCREPO-2023

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/TemporaryOCFLRepositoryHelper.java
bbpennel: An alternative is to tell people to setup debug configs in their IDEs with the system properties, I think that would work with relative paths. Any preference?

awoods:
Where does this stand? I don't see any additional commits based on the previous comments.

bbpennel:
@awoods I left a few questions and additional info on the comment thread above. But it looks like we already have config for setting these paths when running tests via maven:
https://github.com/fcrepo4/fcrepo4/blob/8d24414a2143285387b30c44d244b62270628c2b/pom.xml#L504-L506
So the only situation that needs to be addressed is when running the tests outside of maven. That leaves us with two options, either add a utility like this to set the properties or tell developers to config their IDEs to set these properties during tests.

awoods:
@bbpennel : If possible, I would rather not build IDE support into the code base. If you have found a way to run the tests in your IDE, documenting that approach would be ideal.
Two potential documentation targets:
- https://wiki.lyrasis.org/display/FF/The+Basics or
- https://wiki.lyrasis.org/display/FF/Testing

bbpennel:
I'll go ahead and close this

awoods:
@bbpennel : were you able to find a way to run the tests in your IDE?

PR #1631, opened 2020-02-18
Changed the name of the object metadata files

pwinckles:
**JIRA Ticket**: [FCREPO-3195](https://jira.lyrasis.org/browse/FCREPO-3195) && [FCREPO-3210](https://jira.lyrasis.org/browse/FCREPO-3210)

Fixes the problem @awoods reported with rebuilds not working correctly due to the difference between the Fedora object id and the OCFL object id. The rebuild logic now does not make any assumptions about the naming of sidecar files. It iterates over each sidecar in an ocfl object, if it discovers the object is an archival group, it updates the index mappings accordingly.

# How should this be tested?

Follow @awoods steps as described in [FCREPO-3210](https://jira.lyrasis.org/browse/FCREPO-3210)

# Interested parties
@awoods @bbpennel @dbernstein

### DIFF
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/FedoraToOCFLObjectIndexUtilImpl.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/FedoraToOCFLObjectIndexUtilImpl.java
@@ -28,10 +28,11 @@ import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
 import javax.inject.Inject;
+import java.util.ArrayList;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static java.lang.String.format;
 import static org.fcrepo.persistence.common.ResourceHeaderSerializationUtils.deserializeHeaders;
-import static org.fcrepo.persistence.ocfl.impl.OCFLPersistentStorageUtils.getSidecarSubpath;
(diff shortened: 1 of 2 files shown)


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtils.java
```diff
@@ -125,10 +131,8 @@ public static  String resolveOCFLSubpath(final String rootFedoraObjectId, final
         final var lastPathSegment = rootObjectId.substring(rootObjectId.lastIndexOf("/") + 1);
 
         String subPath;
-        if (FEDORA_ID_PREFIX.equals(rootFedoraObjectId) && "".equals(fedoraSubpath)) {
-            subPath = DEFAULT_REPOSITORY_ROOT_OCFL_OBJECT_ID;
```
awoods: I am not seeing how this case is now being accounted for?

pwinckles: The repository root object is just treated like every other object. There's no need for its metadata files to be named differently. At least, I was unable to identify a reason for them to be treated differently based on my analysis of the code, but perhaps I missed something.

awoods: Besides this question, this pull-request should be good to go. It fixes the broken rebuild scenario.
It would be good to know from others if `root` requires special handling.

PR #1635, opened 2020-02-21
Translate internal IDs to external URIs for GET

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3161

Translates internal Fedora IDs back into external URIs when GETting a resource.

# What's new?
* Adds new translation function to the HttpRdfService.
* Modifies the `getContent` service to remove the RdfStream argument (as it was always an empty stream anyways)
* Modifies the `getResource` function in the StreamingHtmlProvider to properly get the resource.

# How should this be tested?

Before the PR
1. PUT a resource.
1. GET the resource
1. Notice the address has the form "info:fedora/XXXX"

Build this PR
1. PUT a resource.
1. GET the resource
1. Notice the address has the form "http://localhost:8080/(fcrepo)/rest/XXXX"

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -21,7 +21,6 @@ import static com.google.common.base.Strings.nullToEmpty;
 import static java.net.URI.create;
 import static java.text.MessageFormat.format;
 import static java.util.stream.Collectors.toSet;
-import static java.util.stream.Stream.concat;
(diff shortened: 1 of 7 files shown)


awoods:
It is odd that we were creating and passing along empty RdfStreams in FedoraAcl, FedoraLdp and FedoraVersioning... but there does not appear to be a need to keep that around.

whikloj:
Yes, I'm not sure why it started. Perhaps we thought we might pre-populate the stream but never did and then the pattern just started to get repeated.

PR #1637, opened 2020-03-02
Add rdf:type ldp:RDFSource to Rdf source resources

whikloj:

**Note**: this PR only partially resolves the above ticket.

* Fixes the StreamingBaseHtmlProvider to use the correct `ldp:RDFSource` and `ldp:NonRDFSource` types for template matching. 
* Adds the `ldp:RDFSource` type to all `Container`s
* Re-instates one of the `FedoraHtmlIT` tests

# How should this be tested?

Before the PR an RDF source (like a `ldp:BasicContainer`) would have only that type, after this PR it will also have the `ldp:RDFSource` type.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ContainerImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ContainerImpl.java
@@ -17,11 +17,24 @@
  */
 package org.fcrepo.kernel.impl.models;
 
+import static org.apache.jena.graph.NodeFactory.createURI;
+import static org.fcrepo.kernel.api.RdfLexicon.RDF_SOURCE;
+
+import java.util.stream.Stream;
+
+import org.apache.jena.graph.Triple;
+import org.apache.jena.vocabulary.RDF;
+import org.fcrepo.kernel.api.RdfStream;
 import org.fcrepo.kernel.api.Transaction;
+import org.fcrepo.kernel.api.exception.ItemNotFoundException;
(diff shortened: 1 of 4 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ContainerImpl.java
```diff
@@ -48,4 +61,19 @@ public ContainerImpl(final String id, final Transaction tx, final PersistentStor
     public FedoraResource getDescribedResource() {
         return this;
     }
+
+    @Override
+    public RdfStream getTriples() {
+        try {
```
awoods: Instead, could you use the extended class's `getTriples()` method here by changing this method to something like:
```
final Stream<Triple> extraTriples = Stream.of(
                new Triple(createURI(id), RDF.Init.type().asNode(), RDF_SOURCE.asNode()));
return new DefaultRdfStream(createURI(id), Stream.concat(super.getTriples(), extraTriples));
```

whikloj: No because the method returns a `RdfStream` and not `Stream<Triple>`. I had thought of that. I could get the `super.getTriples()` and then extract the `Stream<Triple>`. Not sure if that is easier though.

whikloj: Oh maybe I can, I had some trouble with returning Streams of triples but perhaps I was mistaken about the cause. Doing a test build.

PR #1638, opened 2020-03-03
Add Repository Root type to Repository Root

whikloj:

**Note**: This PR does not completely resolve the above ticket

Adds the `http://fedora.info/definitions/v4/repository#RepositoryRoot` type to the repository root when it is initialized.
Also makes that type a static constant.

# How should this be tested?

Before the PR, your root object had only the `ldp:BasicContainer` type.
After the PR, it also has the above mentioned `RepositoryRoot` type.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/RepositoryInitializer.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/RepositoryInitializer.java
@@ -63,13 +63,6 @@ public class RepositoryInitializer {
     @Inject
     private FedoraToOCFLObjectIndexUtil fedoraToOCFLObjectIndexUtil;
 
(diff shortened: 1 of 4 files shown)


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/RepositoryInitializer.java
```diff
@@ -54,6 +63,13 @@
     @Inject
     private FedoraToOCFLObjectIndexUtil fedoraToOCFLObjectIndexUtil;
 
+    private static Stream<Triple> REPOSITORY_ROOT_TRIPLES = Stream.of(
```
awoods: Since repository initialization is a one-time event, we may not want to retain these two static class constants. 
Suggestion: move them into the `initialize()` method as local variables.

PR #1642, opened 2020-03-06
Handle binary HTML test

whikloj:

Last PR for the above ticket

Creates a `getDescribedResource` method for `NonRdfSourceDescription`s so they can describe the correct resource.

Modifies the `managedPropertyService` to use the `getDescribedResource` to generate correct triples.

Cleans up some redundant logic in `StreamingBaseHtmlProvider`

Removes an un-used Predicate from `ContentExposingResource`, possibly might need to be re-added to a service later.

Uncomments the last test in `FedoraHtmlIT`

This test required the addition of a `createObject()` because it previously tried to 
```
PUT http://localhost:8080/1234-5678/file
```
But we currently  try to load `1234-5678` as a resource to be the parent, this will need to be fixed with the implementation of ghost nodes and can be tested by any number of tests in FedoraLdpIT that use the `createDatastream()` method.

For example:
* https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java#L496
* https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java#L626
* https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java#L762
* https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java#L1124
* as well as many more

# How should this be tested?

Tests pass, but also when you GET the `/fcr:metadata` for a binary the RDF should have the subject of the binary and NOT have a `fedora:NonRdfSourceDescription` type.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
@@ -28,7 +28,6 @@ import static org.fcrepo.http.commons.domain.RDFMediaType.TEXT_HTML_WITH_CHARSET
 import static org.fcrepo.http.commons.session.TransactionProvider.ATOMIC_ID_HEADER;
 import static org.fcrepo.kernel.api.RdfLexicon.NON_RDF_SOURCE;
 import static org.fcrepo.kernel.api.RdfLexicon.RDF_SOURCE;
-import static org.fcrepo.kernel.api.RdfLexicon.REPOSITORY_NAMESPACE;
 import static org.fcrepo.kernel.api.RdfCollectors.toModel;
 import static org.fcrepo.kernel.api.RdfLexicon.REPOSITORY_ROOT;
 import static org.slf4j.LoggerFactory.getLogger;
@@ -176,7 +175,7 @@ public class StreamingBaseHtmlProvider implements MessageBodyWriter<RdfNamespace
             .forEach(key -> templatesMapBuilder.put(key, velocity.getTemplate(getTemplateLocation(key))));
 
         templatesMap = templatesMapBuilder
-            .put(REPOSITORY_NAMESPACE + "RepositoryRoot", velocity.getTemplate(getTemplateLocation("root")))
+            .put(REPOSITORY_ROOT.toString(), velocity.getTemplate(getTemplateLocation("root")))
             .put(NON_RDF_SOURCE.toString(), velocity.getTemplate(getTemplateLocation("binary")))
             .put(RDF_SOURCE.toString(), velocity.getTemplate(getTemplateLocation("resource"))).build();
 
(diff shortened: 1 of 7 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
```diff
@@ -164,10 +164,6 @@ void init() throws IOException {
 
         templatesMap = templatesMapBuilder
             .put(REPOSITORY_NAMESPACE + "RepositoryRoot", velocity.getTemplate(getTemplateLocation("root")))
-            .put(REPOSITORY_NAMESPACE + "Binary", velocity.getTemplate(getTemplateLocation("binary")))
```
awoods: I think we should probably keep these for rendering these types (except, 'Pairtree').
Is there a reason why you believe they should go away?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
```diff
@@ -257,9 +253,6 @@ private Template getTemplate(final Model rdf, final Node subject,
             .map(x -> ((HtmlTemplate) x).value()).filter(templatesMap::containsKey).findFirst()
             .orElseGet(() -> {
                 final List<String> types = multiValueURI(rdf.getResource(subject.getURI()), type);
-                if (types.contains(REPOSITORY_NAMESPACE + "RepositoryRoot")) {
-                    return REPOSITORY_NAMESPACE + "RepositoryRoot";
```
awoods: I suspect this was put in-place to prioritize the "RepositoryRoot" template... did your observations indicate otherwise?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/NonRdfSourceDescriptionImpl.java
```diff
@@ -47,6 +58,20 @@ public NonRdfSourceDescriptionImpl(final String id,
     @Override
     public FedoraResource getDescribedResource() {
         // TODO must return the described binary
```
awoods: This comment can probably now be removed.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
whikloj: Seemingly a `fedora:Binary` will always also have a `ldp:NonRDFSource` type and therefore will get the template using that type. Same with `fedora:Container` and `ldp:RDFSource`. So having these seemed redundant.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
whikloj: I did not do a lot of research into the reason for this. But `RepositoryRoot` is the first entry in the `templateMap` and so if this `types.contains()` is going to match then it should also match in the following `types.stream()...` and the `findFirst()` should return the first match, which should be `RepositoryRoot`.

whikloj: Now the test required me to re-add this. Remove it and it works, rebase and it fails. 🤷‍♂

PR #1645, opened 2020-03-11
First pass at containment index

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3164

Adds some more methods to the API.
Implements a commit and rollback method.
Adds some more tests.
Alters the database setup structure.

# How should this be tested?

Right now this is just via the tests, so it should build and the tests should handle expectations.

Next phase is wiring it into the kernel.

# Interested parties
@fcrepo4/committers @pwinckles

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
@@ -17,14 +17,19 @@
  */
 package org.fcrepo.kernel.impl;
 
+import static org.slf4j.LoggerFactory.getLogger;
+
 import org.fcrepo.kernel.api.ContainmentIndex;
 import org.fcrepo.kernel.api.Transaction;
+import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
 import org.fcrepo.kernel.api.models.FedoraResource;
-import org.springframework.jdbc.BadSqlGrammarException;
+import org.slf4j.Logger;
 import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import org.springframework.stereotype.Component;
-import org.springframework.transaction.annotation.Transactional;
+import org.springframework.transaction.PlatformTransactionManager;
+import org.springframework.transaction.TransactionStatus;
+import org.springframework.transaction.support.DefaultTransactionDefinition;
 
 import javax.annotation.PostConstruct;
 import javax.inject.Inject;
@@ -38,10 +43,15 @@ import java.util.stream.Stream;
 @Component
 public class ContainmentIndexImpl implements ContainmentIndex {
 
+    private static final Logger LOGGER = getLogger(ContainmentIndexImpl.class);
+
     @Inject
     private DataSource dataSource;
 
     private NamedParameterJdbcTemplate jdbcTemplate;
+
+    @Inject
+    private PlatformTransactionManager platformTransactionManager;
 
     private static final String RESOURCES_TABLE = "resources";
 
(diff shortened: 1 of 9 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
```diff
@@ -239,6 +274,30 @@ public void removeContainedBy(final Transaction tx, final FedoraResource parent,
...
+    @Override
+    public void commitTransaction(final Transaction tx) {
+        if (tx != null) {
+            final String txId = tx.getId();
+            final NamedParameterJdbcTemplate jdbcTemplate = getNamedParameterJdbcTemplate();
+            final MapSqlParameterSource parameterSource = new MapSqlParameterSource();
+            parameterSource.addValue("transactionId", txId);
+            jdbcTemplate.update(COMMIT_ADD_RECORDS, parameterSource);
+            jdbcTemplate.update(COMMIT_DELETE_RECORDS, parameterSource);
+            jdbcTemplate.update(COMMIT_CLEANUP, parameterSource);
```
pwinckles: These updates should probably be made within a transaction.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
```diff
@@ -87,48 +118,80 @@
...
+            TRANSACTION_ID_COLUMN + " = :transactionId";
 
     /**
      * Connect to the database
      */
     @PostConstruct
-    private void setup() throws SQLException {
-        conn = dataSource.getConnection();
+    private void setup() {
+        final NamedParameterJdbcTemplate jdbcTemplate = getNamedParameterJdbcTemplate();
```
pwinckles: JdbcTemplates are thread safe. This class really should only ever have a single instance: https://docs.spring.io/spring/docs/current/spring-framework-reference/data-access.html#jdbc-JdbcTemplate-idioms

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
```diff
@@ -87,48 +118,80 @@
...
+        final MapSqlParameterSource parameterSource = new MapSqlParameterSource();
         // create the tables that don't already exist
-        if (tableIsMissing(RESOURCES_TABLE)) {
-            conn.createStatement().execute(RESOURCES_TABLE_DDL);
+        try {
+            jdbcTemplate.update(RESOURCES_TABLE_DDL, parameterSource);
+        } catch (BadSqlGrammarException e) {
```
pwinckles: Couldn't there be other reasons a BadSqlGrammerException is thrown that don't indicate the table already exists? Does the table already exists exception have an error code? This may also vary between db implementations. Can you use a CREATE ... IF NOT EXISTS construct?

whikloj: So there is a wrapped exception, but it is from the database layer which in the test case is h2. If you wire a postgresql or mysql I imagine it will be different.
I can certainly add the h2 exceptions here and once we start wiring in more robust database engines we can extend the `catch`?

whikloj: Alternatively I can re-add the `isTableMissing` function?

birkland: `CREATE TABLE IF NOT EXISTS` seems most straightforward from my perspective.  Then there won't be any "happy" exceptions to explicitly ignore.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
```diff
@@ -275,12 +288,13 @@ public void removeContainedBy(final Transaction tx, final FedoraResource parent,
     }
 
     @Override
+    @Transactional
```
pwinckles: Is Spring's TransactionManager already enabled?

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/transaction/annotation/EnableTransactionManagement.html

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
```diff
@@ -144,91 +209,65 @@ private NamedParameterJdbcTemplate getNamedParameterJdbcTemplate() {
...
+
+    @Override
+    public void addContainedBy(final String txID, final String parentID, final String childID) {
         final MapSqlParameterSource parameterSource = new MapSqlParameterSource();
-        parameterSource.addValue("parent", parent.getId());
-        parameterSource.addValue("child", child.getId());
-        if (txId != null) {
-            parameterSource.addValue("transactionId", txId);
-            jdbcTemplate.update(INSERT_CHILD_IN_TRANSACTION, parameterSource);
+        parameterSource.addValue("parent", parentID);
```
awoods: There are multiple places throughout this class where literals are used ("child", "parent", transactionId). Should probably use the static constants defined at the top of the class.

whikloj: So technically the `child` here matches to the `:child` part of the SQL statement in.
```
"SELECT * FROM " + RESOURCES_TABLE + " WHERE " + FEDORA_ID_COLUMN + " = :child"
```

So we don't want to use the constants, but perhaps using a more descriptive name to differentiate between columns and arguments?

whikloj: To clarify, the above was a question. I'm not going to mess around with all the various SQL statements unless its required.

awoods: I was not thinking of the SQL statements, but rather the String literals used in calls to `parameterSource.addValue()`. It seems more maintainable if "parent", "child" and "transactionId" were replaced by constants. 
Searching for "parameterSource.addValue(" shows 12 instances.

whikloj: To recap the Slack discussion, the replacing of literals with constants would work for the literals `parent` and `transactionId` due to them (fortuitously) being the same as the constants, however `child` would have to remain as a literal or be modified or have a new constant defined.

I'm not against this work, but I also don't see it as valuable and would prefer to push it off for now.

awoods: Punting here is fine.

PR #1650, opened 2020-03-23
Add datetime type to created/lastmodified date, add additional contai…

whikloj:
…ner types

**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3223

This PR does not fulfill all of the above ticket.

Adds a bunch of other `rdf:type` to a Container. Also adds the appropriate xsd:datetime type to the created date and last modified date server managed triples.

# How should this be tested?

Build and create a RDF source, then GET it.
Before you will get the interaction model (ie. `ldp:DirectContainer`) and `ldp:RdfSource` types
Now you will also `ldp:Container`, `fedora:Container` and `fedora:Resource` types.
Also the `fedora:created` and `fedora:lastmodified` will now have the `xsd:dateTime` type 

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImpl.java
@@ -41,6 +41,7 @@ import javax.ws.rs.core.Link;
 import org.apache.jena.rdf.model.Model;
 import org.fcrepo.kernel.api.RdfStream;
 import org.fcrepo.kernel.api.exception.CannotCreateResourceException;
+import org.fcrepo.kernel.api.exception.InteractionModelViolationException;
 import org.fcrepo.kernel.api.exception.ItemNotFoundException;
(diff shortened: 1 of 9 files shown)


whikloj:
Amended to fix the response when trying to add a resource inside a binary.

whikloj:
I think this might also resolve https://jira.lyrasis.org/browse/FCREPO-3168

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImpl.java
```diff
@@ -194,8 +195,8 @@ private void checkParent(final String txId, final PersistentStorageSession pSess
             final ResourceHeaders parent;
             try {
                 // Make sure the parent exists.
+                // TODO: object existence can be from the index, but we don't have interaction model. Should we add it?
```
awoods: That seems like a reasonable optimization.

whikloj: https://jira.lyrasis.org/browse/FCREPO-3258

PR #1655, opened 2020-04-04
Make Repository rebuilder handle binaries

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3268

Assigns the binary as the parent of the description when sending information to the persistence layer and then uses the parent and the interaction model to ensure the binary description gets re-added to the index correctly.

# How should this be tested?

Before PR.

1. Add a binary to Fedora. 
     ```
     > curl -ufedoraAdmin:fedoraAdmin -i -H"Content-type: text/plain" -H"Slug: binary1" --data-binary "Some text content" -XPOST http://localhost:8080/rest
     ```
1. `GET` the binary and its metadata, validate it works.
    ```
    > curl -i -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/binary1
    > curl -i -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/binary1/fcr:metadata
     ```
1. Shutdown Fedora and delete the index file, `fedoraToOcflIndex.tsv` in the OCFL work directory.
1. Restart Fedora
1. Try to get the binary and metadata, see errors
1. Shutdown Fedora, pull in this PR, compile
1. Delete the index file again
1. Restart Fedora
1. Try to get the binary and metadata, see correct results.

# Interested parties
@fcrepo4/committers


awoods:
Windows blues :( https://travis-ci.com/github/fcrepo4/fcrepo4/jobs/314228669

whikloj:
Fails at the RebuildIT and never recovers all tests before it succeed then it fails and all tests after it fail. I tried to pull a windows docker image and test this but that fails. I don't have any other ideas so I'll leave this here for people with more Windows development experience.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraHtmlIT.java
```diff
@@ -104,7 +104,7 @@ public void testGetRootTemplate() throws IOException {
         method.addHeader(ACCEPT, "text/html");
         try (final CloseableHttpResponse response = execute(method)) {
             final String html = EntityUtils.toString(response.getEntity());
-            assertTrue(contains(html, "class=\"fcrepo_root\""));
+            assertTrue(html, contains(html, "class=\"fcrepo_root\""));
```
bbpennel, line 107: not that i mind this, but was this intentionally left in?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/FedoraToOCFLObjectIndexUtilImpl.java
```diff
@@ -80,8 +83,11 @@ public void rebuild() {
                                 fedoraIds.add(headers.getId());
                                 if (headers.isArchivalGroup()) {
                                     rootId.set(headers.getId());
+                                } else if (Objects.equals(headers.getInteractionModel(),
+                                        PERSISTENCE_FEDORA_NON_RDF_SOURCE_DESCRIPTION)) {
+                                    rootId.set(headers.getParent());
```
bbpennel, line 88: I'm trying to understand if this is the case and I can't test at the moment, so I may be totally off, but it seems like if you have an archival group which contains a binary, then everything else in the AG would also have the binary as its root instead of the AG? Do we only want to set the root for a description to the binary if its not in an AG, or do we always want to do this?

dbernstein: I think that's right @bbpennel.  And we also want to account for fcr:acl  as well.

fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/PersistenceConstants.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Class for persistence constants.
+ */
+public final class PersistenceConstants {
+
+    /**
+     * Bringing from kernel level.
+     */
+    public static final String PERSISTENCE_FEDORA_NON_RDF_SOURCE_DESCRIPTION = FEDORA_NON_RDF_SOURCE_DESCRIPTION_URI;
```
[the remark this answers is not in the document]
whikloj: It did seem to work for me, I'll try again. Maybe when I rebased on the latest changes I missed something.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraHtmlIT.java
whikloj: This was part of @awoods PR on this branch. Now I'm not sure so much.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/FedoraToOCFLObjectIndexUtilImpl.java
whikloj: Maybe changing this to
```
} else if (Objects.equals(headers.getInteractionModel(),
                                        FEDORA_NON_RDF_SOURCE_DESCRIPTION_URI) &&
               rootId.get() == null) {
```
So it only sets it if the archival group hasn't set it, and if the archival group comes later it would overwrite.

fcrepo-persistence-api/src/main/java/org/fcrepo/persistence/api/PersistenceConstants.java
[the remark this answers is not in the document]
whikloj: I'm really not sure anymore, but it is not needed and I have removed it. 🤷

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/FedoraToOCFLObjectIndexUtilImpl.java
whikloj: Perhaps this is my lack of knowledge of how this was designed. In the case of an AG we know the root, but in the case of something else what (in the JSON) identifies it as the root?

bbpennel: hmm, i don't think anything in the json identifies it as the root. We had talked about potentially using a special name for the root object, although i'm forgetting what it was at the moment. Seems like once we find the root, we don't need to hunt for roots anymore, so maybe only do any of `if (headers.isArchivalGroup()) {` block if the root id is still null?

whikloj: Right, my question is what is the root ID for a non-archival group? I'd prefer not to check for each Fedora type, so if we can codify this in the JSON all the better.

Before this PR the root for the binary and binary description was `null` because the code was only setting a root for AGs or if there was a single file it became its own root.

We need a way for multiple files in a single object (binary description, ACL, etc) but not an AG to have the root defined. Maybe on top of `"archivalGroup":(true|false)` we need a `"objectRoot":(true|false)`?

dbernstein: For a non-archival group,  the root is itself. 
http://localhost:8080/rest/binary ->   root= info:fedora/binary  resource=info:fedora/binary
http://localhost:8080/rest/binary ->   root=  info:fedora/binary resource= info:fedora/binary/fcr:metadata

dbernstein: Perhaps I'm missing something?

whikloj: What I am wondering is is there a way to determine the correct root Id for resources without checking if 
```
if (
  Objects.equals(obj.getInteractionModel(), FEDORA_NON_RDF_SOURCE_DESCRIPTION_URI) ||
  obj.getId().endsWith(FCR_ACL))
) {
```

Otherwise what I am doing will (as in your PR) have to be expanded to cover a bunch of different cases. If we can generalize that relationship between supporting resources (like binary descriptions and ACLs) and the resources they support, then we don't need to check the type of object in the rebuild process.

dbernstein: I think I see what you're saying.   If we add "objectRoot": (true|false) to the ResourceHeaders.json, then we don't have to deal with inspecting the interaction model.

PR #1657, opened 2020-04-07
fix concurrency bugs

pwinckles:
Fixes a few concurrency bugs in Fedora's ocfl code.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentSessionManager.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentSessionManager.java
@@ -37,7 +37,7 @@ import java.util.concurrent.ConcurrentHashMap;
 @Component
 public class OCFLPersistentSessionManager implements PersistentStorageSessionManager {
 
-    private PersistentStorageSession readOnlySession;
+    private volatile PersistentStorageSession readOnlySession;
 
     private Map<String, PersistentStorageSession> sessionMap;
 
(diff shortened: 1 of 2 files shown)


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentSessionManager.java
```diff
@@ -61,23 +61,20 @@ public PersistentStorageSession getSession(final String sessionId) {
...
                 fedoraOcflIndex,
-                objectSessionFactory);
-        sessionMap.put(sessionId, newSession);
-        return newSession;
+                objectSessionFactory));
     }
 
     @Override
     public PersistentStorageSession getReadOnlySession() {
         if (this.readOnlySession == null) {
```
whikloj, line 72: Is this line needed as it is repeated inside the `synchronized (this) {` block? I feel like it was just an oversight?

pwinckles: No synchronization is required.

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent(K,java.util.function.Function)

>  The entire method invocation is performed atomically, so the function is applied at most once per key. Some attempted update operations on this map by other threads may be blocked while computation is in progress, so the computation should be short and simple, and must not attempt to update any other mappings of this map.

pwinckles: Whoops, I got confused about which line you were referring to and thought you meant L64.

pwinckles: Do you mean L72?

If so, then yes, both conditionals are needed. https://en.wikipedia.org/wiki/Double-checked_locking

whikloj: Ah okay, thanks for the explanation.

PR #1662, opened 2020-04-14
Periodic Transactions garbage collection

bbpennel:
**JIRA Ticket**: none

Fills in periodic cleanup of transactions.

Transactions are automatically rolled back if they expire.
Closed transactions will be remain available for retrieving status until their expiration time has passed. For transactions that expire automatically, they will stick around until the next garbage collection.

# What's new?
Transaction cleanup and tests.

# How should this be tested?

You should be able to create a transaction, cancel it, check its status to see that it tells you it was rolled back, and then check its status again after 3 minutes to see that it now gives a 404.

# Additional Notes:
Session cleanup timer currently just reuses the session timeout environment variable.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionManagerImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionManagerImpl.java
@@ -19,7 +19,8 @@ package org.fcrepo.kernel.impl;
 
 import static java.util.UUID.randomUUID;
 
-import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import javax.inject.Inject;
 
@@ -30,6 +31,7 @@ import org.fcrepo.kernel.api.exception.TransactionClosedException;
 import org.fcrepo.kernel.api.exception.TransactionNotFoundException;
 import org.fcrepo.kernel.api.observer.EventAccumulator;
 import org.fcrepo.persistence.api.PersistentStorageSessionManager;
+import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Component;
 
 /**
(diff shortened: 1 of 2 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionManagerImpl.java
```diff
@@ -52,10 +54,32 @@
...
+     */
+    @Scheduled(fixedDelayString = "#{systemProperties['fcrepo.session.timeout'] ?: 180000}")
+    public void cleanupClosedTransactions() {
+        final var txIt = transactions.entrySet().iterator();
+        while (txIt.hasNext()) {
+            final var txEntry = txIt.next();
+            final var tx = txEntry.getValue();
+
+            // Cleanup if transaction is closed and past its expiration time
+            if (tx.isCommitted() || tx.isRolledBack()) {
```
whikloj, line 71: Don't we remove the transaction as soon as it is committed or rolled back?

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/TransactionManagerImplTest.java
```diff
@@ -88,8 +97,108 @@ public void testGetTransactionWithInvalidID() {
...
+        } catch(final TransactionClosedException e) {
+            //expected
+        }
+
+        assertNotNull("Continuing transaction must be present",
+                testTxManager.get(continuingTx.getId()));
+
+        testTxManager.cleanupClosedTransactions();
+
+        // Verify that the closed transactions are stick around since they haven't expired yet
```
whikloj, line 139: I see now that this we are hanging on to transaction regardless of state until the clean up period. Just wondering if there is any value to having those transactions that have been actively committed or rolled back?

bbpennel: If there are multiple clients using the same transaction it might be helpful to be able to check the status, which we aren't disallowing at the moment. Figured it'd be good to have a consistent window for being able to look up what happened to a transaction

PR #1663, opened 2020-04-16
refactor event api, removing info map

pwinckles:

Refactors Event, replacing the info map with fields and removing OptionalValues.

# What's new?

Nothing new.

# How should this be tested?

Mutate a resource and [see events](https://wiki.lyrasis.org/display/FEDORA51/How+to+Inspect+Event+Messages+Generated+by+Fedora).

# Interested parties

@fcrepo4/committers


fcrepo-jms/src/main/java/org/fcrepo/jms/AbstractJMSPublisher.java
```diff
@@ -73,6 +74,7 @@
      * @throws JMSException if JMS exception occurred
      */
     @Subscribe
+    @AllowConcurrentEvents
```
awoods, line 77: What affect does this annotation have?

pwinckles: https://github.com/google/guava/blob/master/guava/src/com/google/common/eventbus/AllowConcurrentEvents.java

It allows the code to be called from multiple threads.

awoods: Yes, the name suggested as much. What was happening before? Is this basically a small performance optimization?

pwinckles: https://github.com/google/guava/blob/master/guava/src/com/google/common/eventbus/EventBus.java#L71

Previously, it would only process one event at a time. I assume this was an oversight and a misconfiguration. I believe the event code is thread-safe, or am I missing something?

awoods: Yes, this method should be thread-safe. "Oversight" is probably the reason that the method was not given this annotation before. 
I was just wondering if you had any other reason for making the change... it does not sound like it.

pwinckles: Nope! I just saw that it wasn't annotated and added it. :)

PR #1664, opened 2020-04-16
Use RDF extension when looking for files

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/projects/FCREPO/issues/FCREPO-3278

When we write RDF we append the `.nt` extension to the id, but when we delete we just look for a file with the id name.

When trying to delete a child in an AG I got a 500 because it was trying to find a file called `test2`, but the actual file was `test2.nt`. 

This refactors the Delete operations to track whether we are deleting a RDFSource or NonRDFSource object.

# How should this be tested?

Before the PR:
1. Create an Archival Group
     ```
     curl -i -ufedoraAdmin:fedoraAdmin -H"Slug: parent" -H'Link: <http://fedora.info/definitions/v4/repository#ArchivalGroup>;rel="type"' -H"Content-type: text/turtle" -XPOST http://localhost:8080/rest
     ```
1. Create a child resource in the AG.
     ```
     curl -i -ufedoraAdmin:fedoraAdmin -H"Slug: child" -H"Content-type: text/turtle" -XPOST http://localhost:8080/rest/parent
     ```
1. Delete the child resource
    ```
     curl -i -ufedoraAdmin:fedoraAdmin -XDELETE http://localhost:8080/rest/parent/child
    ```
1. See 500 and explosion
1. Pull in this PR 
1. Repeat steps
1. Joy.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DeleteResourcePersister.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DeleteResourcePersister.java
@@ -17,9 +17,13 @@
  */
 package org.fcrepo.persistence.ocfl.impl;
 
+import static org.fcrepo.kernel.api.RdfLexicon.NON_RDF_SOURCE;
 import static org.fcrepo.kernel.api.operations.ResourceOperationType.DELETE;
 import static org.fcrepo.persistence.ocfl.impl.OCFLPersistentStorageUtils.relativizeSubpath;
+import static org.fcrepo.persistence.ocfl.impl.OCFLPersistentStorageUtils.resolveExtensions;
 import static org.fcrepo.persistence.ocfl.impl.OCFLPersistentStorageUtils.resolveOCFLSubpath;
+
+import java.util.Objects;
 
 import org.fcrepo.kernel.api.operations.ResourceOperation;
 import org.fcrepo.persistence.api.exceptions.PersistentStorageException;
@@ -53,7 +57,11 @@ class DeleteResourcePersister extends AbstractPersister {
         } else {
             final var relativeSubPath = relativizeSubpath(fedoraResourceRoot, operation.getResourceId());
             final var ocflSubPath = resolveOCFLSubpath(fedoraResourceRoot, relativeSubPath);
-            objectSession.delete(ocflSubPath);
+            final var headers = readHeaders(objectSession, ocflSubPath);
+            final var filePath = resolveExtensions(ocflSubPath,
+                    !Objects.equals(NON_RDF_SOURCE.toString(), headers.getInteractionModel())
+            );
+            objectSession.delete(filePath);
         }
     }
 }
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/OcflPersistenceIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/OcflPersistenceIT.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.integration.http.api;
+
+import static org.slf4j.LoggerFactory.getLogger;
+import static javax.ws.rs.core.Response.Status.CREATED;
+import static javax.ws.rs.core.Response.Status.NOT_FOUND;
+import static javax.ws.rs.core.Response.Status.NO_CONTENT;
+import static javax.ws.rs.core.Response.Status.OK;
+
+import static org.junit.Assert.assertEquals;
+
+import org.apache.http.client.methods.CloseableHttpResponse;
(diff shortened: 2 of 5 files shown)


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DeleteResourcePersister.java
```diff
@@ -46,13 +48,15 @@ public void persist(final OCFLPersistentStorageSession session, final ResourceOp
...
         final var objectSession = session.findOrCreateSession(mapping.getOcflObjectId());
+        final var isRdf = ((DeleteResourceOperation) operation).isRdf();
         log.debug("Deleting {} from {}", resourceId, mapping.getOcflObjectId());
         if (fedoraResourceRoot.equals(resourceId)) {
             // We are at the root of the object.
             objectSession.deleteObject();
         } else {
             final var relativeSubPath = relativizeSubpath(fedoraResourceRoot, operation.getResourceId());
-            final var ocflSubPath = resolveOCFLSubpath(fedoraResourceRoot, relativeSubPath);
+            final var ocflSubPath = resolveOCFLSubpath(fedoraResourceRoot, relativeSubPath) +
```
pwinckles, line 58: This seems like something that should be done within the Utils class so that it can be reused other places if needed.

whikloj: Like in the `resolveOCFLSubpath` method?

pwinckles: Yeah, that would seem to make sense. I probably won't get around to it for a few weeks, but I'd like to refactor that whole pathing mess. It's very confusing.

whikloj: Except that the `resolveOCFLSubpath` really gets the subpath, which is used to resolve the sidecar files so adding `.nt` there would cause problems.

This is a difference we are dealing with in Delete because we use one persister to delete both RDFSource and NonRDFSource whereas to create or update we have two separate persisters (the RDF one adds this `.nt`).

I can duplicate the persister, but I think just having this in the DeleteResourcePersister might be ok.

pwinckles: How about a `resolveRdfOcflSubpath()` method? Not a big deal either way and not a blocker for the PR.

whikloj: except that the subpath technically does not include the `.nt` we only add that here https://github.com/fcrepo4/fcrepo4/blob/cd97a2e7a869f9598b64b8fc61fa95af6db5e83b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtils.java#L193

I'm a little out of my depth with how this stuff is operating, if you want me to make a method I will but it will not be used for all instances of "subpath" but instead only be called to replace the above line (in `writeRDF`), [this line in `getTriples()`](https://github.com/fcrepo4/fcrepo4/blob/cd97a2e7a869f9598b64b8fc61fa95af6db5e83b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageSession.java#L223) and in the delete persister to refer to the exact file.

pwinckles: It's fine. Just leave it.

whikloj:
@pwinckles I wasn't sure if the `resolveOCFLSubpath` method was used to build the path to the entire object so I went with a separate method to handle file extensions. If this is more of what you want I can add the commit here. 
Other comparison -> https://github.com/fcrepo4/fcrepo4/compare/master...whikloj:FCREPO-3278-2
Specific commit -> https://github.com/fcrepo4/fcrepo4/commit/0f51dabe3727234c390d6d673b01aac118bc7c14

pwinckles:
@whikloj, yeah that does look a little cleaner. 👍

whikloj:
@pwinckles added

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DeleteResourcePersister.java
```diff
@@ -46,13 +48,14 @@ public void persist(final OCFLPersistentStorageSession session, final ResourceOp
         final var fedoraResourceRoot = mapping.getRootObjectIdentifier();
         final var resourceId = operation.getResourceId();
         final var objectSession = session.findOrCreateSession(mapping.getOcflObjectId());
+        final var isRdf = ((DeleteResourceOperation) operation).isRdf();
```
bbpennel, line 51: Would it be an option to call `readHeaders` here in the persister to determine that the object being deleted is a binary or not, rather than passing it down through the Operation stuff?

whikloj: Maybe...the OCFL layer is still a bit of a tangle moving stuff around too me. Is that preferable to passing it down?

whikloj: Ugh, ok I'm going to start over with a fresh branch instead of trying to revert all the changes here. I'll force push over top later.

dbernstein: I don't have a _strong_ opinion on it,  but it does seem cleaner not to pass the isRdf flag.

bbpennel: Sorry to blow things up on you like that. Just seems preferable to have the data coming from as close as possible if its an option, fewer things to lineup or places for things to go wrong.

PR #1666, opened 2020-04-24
Change delete to remove data files and set headers to deleted.

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3033

Changes how deletes work from an immediate purge to just removing the data files and setting the resources as deleted in their header files. This allows the ResourceFactory to determine if the requested resource was deleted and provide a tombstone link.

# How should this be tested?

Before the PR all deleted objects would return a 404, now they should return a 410 GONE and include a Link header to a tombstone location. For NonRdfSourceDescription, TimeMaps and Mementos the link header refers to the original resource + `fcr:tombstone`

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-persistence-common/src/main/java/org/fcrepo/persistence/common/ResourceHeadersImpl.java
+++ b/fcrepo-persistence-common/src/main/java/org/fcrepo/persistence/common/ResourceHeadersImpl.java
@@ -62,6 +62,8 @@ public class ResourceHeadersImpl implements ResourceHeaders {
 
     private boolean objectRoot;
 
+    private boolean deleted;
+
     @Override
     public String getId() {
         return id;
@@ -257,5 +259,18 @@ public class ResourceHeadersImpl implements ResourceHeaders {
         } else {
             return objectRoot;
         }
+    }
+
+    /**
+     * Set deleted status flag.
+     * @param deleted true if deleted (a tombstone).
+     */
+    public void setDeleted(final boolean deleted) {
+        this.deleted = deleted;
+    }
+
+    @Override
+    public boolean isDeleted() {
+        return deleted;
     }
 }
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DeleteResourcePersister.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DeleteResourcePersister.java
@@ -19,15 +19,20 @@ package org.fcrepo.persistence.ocfl.impl;
 
 import static org.fcrepo.kernel.api.RdfLexicon.NON_RDF_SOURCE;
 import static org.fcrepo.kernel.api.operations.ResourceOperationType.DELETE;
+import static org.fcrepo.persistence.ocfl.impl.OCFLPersistentStorageUtils.isSidecarSubpath;
 import static org.fcrepo.persistence.ocfl.impl.OCFLPersistentStorageUtils.relativizeSubpath;
 import static org.fcrepo.persistence.ocfl.impl.OCFLPersistentStorageUtils.resolveExtensions;
 import static org.fcrepo.persistence.ocfl.impl.OCFLPersistentStorageUtils.resolveOCFLSubpath;
 
+import java.time.Instant;
 import java.util.Objects;
 
 import org.fcrepo.kernel.api.operations.ResourceOperation;
 import org.fcrepo.persistence.api.exceptions.PersistentStorageException;
+import org.fcrepo.persistence.api.exceptions.PersistentStorageRuntimeException;
+import org.fcrepo.persistence.common.ResourceHeadersImpl;
 import org.fcrepo.persistence.ocfl.api.FedoraToOCFLObjectIndex;
+import org.fcrepo.persistence.ocfl.api.OCFLObjectSession;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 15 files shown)


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DeleteResourcePersister.java
```diff
@@ -52,16 +57,68 @@ public void persist(final OCFLPersistentStorageSession session, final ResourceOp
...
+    }
+
+    /**
+     * Simple utility to delete a path's files and mark them as deleted in the headers file.
+     * @param path Path to delete
+     * @param session Session to delete the path in.
+     */
+    private void deletePath(final String path, final OCFLObjectSession session) throws PersistentStorageException{
+        // readHeaders and writeHeaders need the subpath where as delete needs the file name. So remove any extensions.
+        final var no_extension = (path.contains(".") ? path.substring(0, path.indexOf(".")) : path);
```
pwinckles, line 90: Is it possible for a path to have a `.` that is unrelated to a file extension?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DeleteResourcePersister.java
```diff
@@ -52,16 +57,68 @@ public void persist(final OCFLPersistentStorageSession session, final ResourceOp
...
-                    !Objects.equals(NON_RDF_SOURCE.toString(), headers.getInteractionModel())
-            );
-            objectSession.delete(filePath);
+            final var headers = (ResourceHeadersImpl) readHeaders(objectSession, ocflSubPath);
+            final boolean isRdf = !Objects.equals(NON_RDF_SOURCE.toString(), headers.getInteractionModel());
+            final var filePath = resolveExtensions(ocflSubPath, isRdf);
+            deletePath(filePath, objectSession, headers);
+            if (!isRdf) {
+                // Delete the description too.
+                final var descPath = resolveExtensions(ocflSubPath + "-description", true);
```
pwinckles, line 77: Would acls need to be deleted here too? I know we aren't writing any yet. Just asking out of curiosity.

fcrepo-persistence-common/src/main/java/org/fcrepo/persistence/common/ResourceHeadersImpl.java
```diff
@@ -62,6 +62,8 @@
 
     private boolean objectRoot;
 
+    private boolean deleted;
```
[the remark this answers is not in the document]
whikloj: As long as you set a sensible default in the ResourceHeadersImpl then I think we are good. For instance, in this case if you stand up a F6 create an object. Then stop F6, build this PR and stand it up again the headers for those existing objects (not having the key) should be `"delete":false`. As for removing properties, that would probably require us to list them as an ignored property.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DeleteResourcePersister.java
```diff
@@ -52,16 +57,68 @@ public void persist(final OCFLPersistentStorageSession session, final ResourceOp
...
         if (fedoraResourceRoot.equals(resourceId)) {
-            // We are at the root of the object.
-            objectSession.deleteObject();
+            // We are at the root of the object, so delete all the data files.
+            try {
+                objectSession.listHeadSubpaths().filter(p -> !isSidecarSubpath(p)).forEach(
+                        p -> deletePathWrapped(p, objectSession));
+            } catch (final PersistentStorageRuntimeException exc) {
+                // Rethrow the exception as a checked exception
+                throw new PersistentStorageException(exc.getCause().getMessage());
```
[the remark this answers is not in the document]
whikloj: fair point

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DeleteResourcePersister.java
whikloj: unsure, we could have something that removes only specific extensions?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DeleteResourcePersister.java
whikloj: Yes, I think that would need to be checked for. As an ACL is not required we'd need to check for its existence and delete.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DeleteResourcePersister.java
dbernstein: I would suggest creating a JIRA to address this issue separately.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DeleteResourcePersister.java
```diff
@@ -52,16 +57,68 @@ public void persist(final OCFLPersistentStorageSession session, final ResourceOp
...
+     * @param path Path to delete
+     * @param session Session to delete the path in.
+     * @param headers The headers for the file.
+     * @throws PersistentStorageException if can't read, write or delete a file.
+     */
+    private void deletePath(final String path, final OCFLObjectSession session, final ResourceHeadersImpl headers)
+        throws PersistentStorageException {
+        session.delete(path);
+        headers.setDeleted(true);
+        headers.setLastModifiedDate(Instant.now());
```
pwinckles, line 106: I think you might want to call `ResourceHeadersUtil.touchModificationHeaders()` here instead of directly setting the lastModifiedDate.

pwinckles:
If you do not have autoversioning enabled, I think there's a problem when you delete a resource but then can't "commit" the deletion (move it out of the mutable head and into an ocfl version).

For example:

```
curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/new-ag -H'Link: <http://www.w3.org/ns/ldp#BasicContainer>;rel="type"' -H'Link: <http://fedora.info/definitions/v4/repository#ArchivalGroup>;rel="type"' -v -H "Content-Type: text/turtle" -X PUT --data "<> <http://purl.org/dc/elements/1.1/title> 'AG Parent'"
curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/new-ag/child1 -H'Link: <http://www.w3.org/ns/ldp#BasicContainer>;rel="type"' -v -H "Content-Type: text/turtle" -X PUT --data "<> <http://purl.org/dc/elements/1.1/title> 'AG Child 1'"
curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/new-ag/child2 -H "Link: <http://www.w3.org/ns/ldp#NonRDFSource>; rel=type" -v -H "Content-Type: text/plain" -X PUT --data-binary "testing"
curl -u fedoraAdmin:fedoraAdmin -X POST -v http://localhost:8080/rest/new-ag/fcr:versions
curl -u fedoraAdmin:fedoraAdmin -X DELETE -v http://localhost:8080/rest/new-ag
curl -u fedoraAdmin:fedoraAdmin -X POST -v http://localhost:8080/rest/new-ag/fcr:versions
```

That final command returns a 410 without committing the mutable head.

whikloj:
@pwinckles So it is just responding with a 410 or is something else happening?

pwinckles:
@whikloj It's responding with a 410 and does not move the mutable HEAD to an OCFL version.

One possible solution to this problem is to automatically commit the mutable HEAD in the request the deletes a resource.

whikloj:
I think that is correct, perhaps what I should be doing (and I'm not 100% on this) is committing a version if we are not auto-versioned and there are changes in the mutable-head.

whikloj:
🤦 yes now seeing what you said. I'm not sure if people would expect a `DELETE` to create a version in a non-versioned repository. But we could document that behaviour.

whikloj:
Regarding what a deleted resource should look like in Fedora's OCFL storage root.  https://jira.lyrasis.org/browse/FCREPO-3289

PR #1672, opened 2020-05-02
Implement adding ACLs

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3134

Allows you to PUT/GET/DELETE and PATCH ACLs.

This does not ensure ACLs are enforced, that is next up.

# What's new?
I finally determined the issue with the FedoraId `resourceId` versus `fullId`. I need the resourceId to be the path as stored on disk. So for pretend things like fcr:tombstone and fcr:versions which don't exist it is the resource they hang off of, but for things that have an actual representation on disk (fcr:metadata and fcr:acl) we need their id (ie. info:fedora/object/fcr:metadata)

# How should this be tested?

Try PUTting an ACL
Try GETting the ACL
Try PATCHing an ACL
Try DELETEing the ACL

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraAcl.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraAcl.java
@@ -148,7 +148,7 @@ public class FedoraAcl extends ContentExposingResource {
             throw new BadRequestException("Content-Type (" + requestContentType + ") is invalid. Try text/turtle " +
                                           "or other RDF compatible type.");
         }
-        transaction().commit();
+        transaction().commitIfShortLived();
 
         final FedoraResource aclResource = getFedoraResource(transaction(), aclId);
         addCacheControlHeaders(servletResponse, aclResource, transaction());
@@ -201,7 +201,7 @@ public class FedoraAcl extends ContentExposingResource {
 
             LOGGER.info("PATCH for '{}'", externalPath);
             patchResourcewithSparql(aclResource, requestBody);
-            transaction().commit();
+            transaction().commitIfShortLived();
 
             addCacheControlHeaders(servletResponse, aclResource, transaction());
 
@@ -293,7 +293,7 @@ public class FedoraAcl extends ContentExposingResource {
             }
             throw new PathNotFoundRuntimeException(exc);
         } finally {
-            transaction().commit();
+            transaction().commitIfShortLived();
         }
         return noContent().build();
     }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraAcl.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraAcl.java
@@ -38,7 +38,6 @@ import static org.fcrepo.http.commons.domain.RDFMediaType.TEXT_HTML_WITH_CHARSET
 import static org.fcrepo.http.commons.domain.RDFMediaType.TEXT_PLAIN_WITH_CHARSET;
 import static org.fcrepo.http.commons.domain.RDFMediaType.TURTLE_WITH_CHARSET;
 import static org.fcrepo.http.commons.domain.RDFMediaType.TURTLE_X;
-import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_REPOSITORY_ROOT;
 import static org.fcrepo.kernel.api.FedoraTypes.FCR_ACL;
 import static org.slf4j.LoggerFactory.getLogger;
 
@@ -46,6 +45,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
+
 import javax.inject.Inject;
 import javax.servlet.http.HttpServletResponse;
 import javax.ws.rs.BadRequestException;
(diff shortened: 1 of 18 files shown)


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/identifiers/FedoraId.java
```diff
@@ -347,6 +364,31 @@ private void processIdentifier() {
...
+
+    /**
+     * Check for obvious path errors.
+     */
+    private void checkForInvalidPath() {
+        // Check for combinations of endpoints not allowed.
+        if (
+            // ID contains fcr:acl or fcr:tombstone AND fcr:metadata or fcr:versions
+            ((this.fullId.contains(FCR_ACL) || this.fullId.contains(FCR_TOMBSTONE)) &&
+                (this.fullId.contains(FCR_METADATA) || this.fullId.contains(FCR_VERSIONS))) ||
```
awoods, line 378: In the past, we supported `/fcr:metadata` having their own ACLs, no? Was it decided that this would no longer be supported?

This impacts FedoraIdTest.java updates.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
```diff
@@ -96,18 +97,20 @@ public boolean doesResourceExist(final Transaction transaction, final FedoraId f
...
 
             final PersistentStorageSession psSession = getSession(transaction);
 
             try {
-                final String id = fedoraId.isDescription() ? fedoraId.getDescriptionId() : fedoraId.getResourceId();
-                psSession.getHeaders(id, fedoraId.getMementoInstant());
-                return true;
+                // Resource ID for metadata or ACL contains their individual endopoints (ie. fcr:metadata, fcr:acl)
+                final String id = fedoraId.getResourceId();
+                final ResourceHeaders headers = psSession.getHeaders(id, fedoraId.getMementoInstant());
```
awoods, line 112: In the case that the resource is an ACL (see line:100 above), there may not be a "memento instant", no?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/identifiers/FedoraId.java
```diff
@@ -347,6 +364,31 @@ private void processIdentifier() {
...
+            // or ID contains fcr:acl AND fcr:tombstone
+            (this.fullId.contains(FCR_TOMBSTONE) && this.fullId.contains(FCR_ACL))
+        ) {
+            throw new InvalidResourceIdentifierException(String.format("Path is invalid: %s", pathOnly));
+        }
+        // Ensure we don't have 2 of any of the extensions, ie. info:fedora/object/fcr:acl/fcr:acl, etc.
+        for (final String extension : extensions) {
+            if (this.fullId.contains(extension)) {
+                final Pattern extPattern = Pattern.compile(extension);
```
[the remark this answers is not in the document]
whikloj: Fair enough, as the test is just for more than one match I can skip the contains.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/identifiers/FedoraId.java
whikloj: Did we support that, I wasn't sure. But I can move back to it.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
whikloj: Yes and the `getMementoInstant()` function would return `null` in that case.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraAcl.java
```diff
@@ -130,37 +128,32 @@ public Response createFedoraWebacAcl(@HeaderParam(CONTENT_TYPE) final MediaType
...
+                replacePropertiesService.perform(transaction().getId(), getUserPrincipal(), aclId, model);
+            } else {
+                webacAclService.create(transaction(), aclId, getUserPrincipal(), model);
+            }
         } else {
             throw new BadRequestException("Content-Type (" + requestContentType + ") is invalid. Try text/turtle " +
                                           "or other RDF compatible type.");
         }
-        transaction.commit();
+        transaction().commit();
```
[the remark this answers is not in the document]
awoods: Shutting down / committing a long-running transaction may not be the desired behavior. For example, I am creating several resources, and updating their ACLs... and boom, my transaction gets committed.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/identifiers/FedoraId.java
awoods: Let me double-check... but I believe it was supported.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraAcl.java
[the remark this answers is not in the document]
whikloj: Sorry the 👍 was that this is a correct observation and a mistake in my PR. I am replacing with
```
transaction().commitIfShortLived();
```

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/identifiers/FedoraId.java
awoods: Binaries and their descriptions can be versioned separately... but they share an ACL: the ACL on the binary.
You can disregard this comment thread :)

PR #1683, opened 2020-05-21
Adding test demonstrating PATCH with If-Unmodified-Since header

escowles:
Debugging support for `If-Unmodified-Since` header in PATCH requests.

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -749,6 +749,25 @@ public class FedoraLdpIT extends AbstractResourceIT {
     }
 
     @Test
+    public void testPatchUnmodifiedSince() throws Exception {
+        final String id = getRandomUniqueId();
+        createObjectAndClose(id);
+
+        final String location = serverAddress + id;
+        final HttpPatch updateObjectGraphMethod = patchObjMethod(id);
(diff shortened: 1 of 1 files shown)


whikloj:
Good to see you @escowles, the 4.7 branch is a little out-of-date. I think the `.travis.yml` might need a 
```
dist: trusty
```
line now to handle the Java 8 requirement. Would you mind adding that?

escowles:
Hey @whikloj — thanks for the tip on getting Travis to work with Java 8! I was mostly putting this up to demonstrate for @no-reply that the `If-Unmodified-Since` header was working, since he was seeing it not work and I couldn't find an explicit check of it (on further inspection, I think [this](https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java#L820) is checking it).

Not sure it's worth merging against the 4.x line, but I'm happy to squash and rebase if you'd like me to.

whikloj:
@escowles I have no problem adding a test to ensure a check is being processed.

There probably won't be another release of the 4.7 branch, but if there is then the test would be included. I'll see if any other @fcrepo4/committers have a thought.

I did some tracing and you were correct, the `request.evaluatePreconditions(date)` flows through jersey and does [this check](https://github.com/jersey/jersey/blob/2.25.1/core-server/src/main/java/org/glassfish/jersey/server/ContainerRequest.java#L748-L763) of the `If-Unmodified-Since` header

PR #1688, opened 2020-06-03
FCREPO-3200 - Default Digests

bbpennel:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3200

* Internal fedora binaries will always record a default digest
* Adds a repository wide default digest algorithm from a system property.
   * Can only be SHA512 or SHA256 due to OCFL restrictions
* New OCFL objects are configured to use this default digest algorithm
* Added aliases for digest algorithms to help with matching across domains
* Adds integration tests for binary persistence in OCFL

# How should this be tested?

Digests are not currently visible in http responses unfortunately, so you would need to look at the OCFL persistence to verify results:

```
curl -XPOST http://localhost:8080/rest/ -H"Slug: binary_default_digest" --data-binary @README.md -H"Content-type: text/plain" -ufedoraAdmin:fedoraAdmin

# verify that the sha512 is set (adjust for your OCFL storage root)
less /var/folders/v6/r6tq3p3n2h165cw93s39rl25cv0v_w/T/fcrepo.ocfl.storage.root.dir/84/43/86/binary_default_digest/v1/content/.fcrepo/binary_default_digest.json 

{
  "parent": "info:fedora",
  "id": "info:fedora/binary_default_digest",
  "lastModifiedDate": "2020-06-03T18:14:13.987960Z",
  "interactionModel": "http://www.w3.org/ns/ldp#NonRDFSource",
  "createdBy": "fedoraAdmin",
  "createdDate": "2020-06-03T18:14:13.987960Z",
  "lastModifiedBy": "fedoraAdmin",
  "stateToken": "4BE570CA6FBA4D7ECD6D94DB8878B3C2",
  "contentSize": 5770,
  "digests": [
    "urn:sha-512:7003d24ed80964957eb59d590f8d63961a8db755bc5760b9d55c21d97d3b011497bce353e1ceaa9374efbe9f05f4b277010468b92a55c543b0dca787300f8681"
  ],
  "filename": "",
  "deleted": false,
  "mimeType": "text/plain",
  "archivalGroup": false,
  "objectRoot": true
}
```
```
# With bad digest
curl -XPOST http://localhost:8080/rest/ -H"Slug: binary_sha1_digest" --data-binary @LICENSE -H"Content-type: text/plain" -H"digest: sha=checksumdoesntmatch" -ufedoraAdmin:fedoraAdmin
# Checksum mismatch, computed SHA digest f5c0f25223b84f167e917701f91e90c49525def0 did not match expected value checksumdoesntmatch

# With sha1 digest
curl -XPOST http://localhost:8080/rest/ -H"Slug: binary_sha1_digest" --data-binary @LICENSE -H"Content-type: text/plain" -H"digest: sha=f5c0f25223b84f167e917701f91e90c49525def0" -ufedoraAdmin:fedoraAdmin

# Verify that the sha512 and provided sha1 are present
less /var/folders/v6/r6tq3p3n2h165cw93s39rl25cv0v_w/T/fcrepo.ocfl.storage.root.dir/82/70/bb/binary_sha1_digest/v1/content/.fcrepo/binary_sha1_digest.json 

{
  "parent": "info:fedora",
  "id": "info:fedora/binary_sha1_digest",
  "lastModifiedDate": "2020-06-03T18:24:03.656085Z",
  "interactionModel": "http://www.w3.org/ns/ldp#NonRDFSource",
  "createdBy": "fedoraAdmin",
  "createdDate": "2020-06-03T18:24:03.656085Z",
  "lastModifiedBy": "fedoraAdmin",
  "stateToken": "9C8C83F606DBDA2BCB78A431E45AD3EC",
  "contentSize": 10831,
  "digests": [
    "urn:sha1:f5c0f25223b84f167e917701f91e90c49525def0",
    "urn:sha-512:05de5e29e68a7e33c86d494cf88da0c3c8c7c10ce5d3d678a548c2538991f806728d6e44495d3b59da13d279533c29952f6467b61a57415e9239d76e23201b22"
  ],
  "filename": "",
  "deleted": false,
  "mimeType": "text/plain",
  "archivalGroup": false,
  "objectRoot": true
}
```

# Notes
The SHA512 is generated both when the file is written to staging, and the OCFL client also generates and verifies it. Some efficiency should be gained later in FCREPO-3201 by passing the digest along with the OCFL client in terms of transmission fixity checking.

# Interested parties
@awoods @pwinckles

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/utils/ContentDigest.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/utils/ContentDigest.java
@@ -24,7 +24,10 @@ import static org.slf4j.LoggerFactory.getLogger;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Arrays;
+import java.util.Set;
+import java.util.stream.Collectors;
 
+import org.apache.commons.lang3.ArrayUtils;
 import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
 import org.slf4j.Logger;
 
@@ -38,20 +41,39 @@ public final class ContentDigest {
 
     private static final Logger LOGGER = getLogger(ContentDigest.class);
 
+    private final static String DEFAULT_DIGEST_ALGORITHM_PROPERTY = "fcrepo.persistence.defaultDigestAlgorithm";
+    public final static DIGEST_ALGORITHM DEFAULT_DIGEST_ALGORITHM;
+
+    static {
+        // Establish the default digest algorithm for fedora, defaulting to SHA1
+        DEFAULT_DIGEST_ALGORITHM = DIGEST_ALGORITHM.fromAlgorithm(
+                System.getProperty(DEFAULT_DIGEST_ALGORITHM_PROPERTY, DIGEST_ALGORITHM.SHA512.algorithm));
+
+        // Throw error if the configured default digest is not known to fedora
+        if (DIGEST_ALGORITHM.MISSING.equals(DEFAULT_DIGEST_ALGORITHM)) {
+            throw new IllegalArgumentException("Invalid " + DEFAULT_DIGEST_ALGORITHM_PROPERTY
+                    + " property configured: " + System.getProperty(DEFAULT_DIGEST_ALGORITHM_PROPERTY));
+        }
+    }
+
     public enum DIGEST_ALGORITHM {
-        SHA1("SHA", "urn:sha1"),
-        SHA256("SHA-256", "urn:sha-256"),
-        SHA512("SHA-512", "urn:sha-512"),
-        SHA512256("SHA-512/256", "urn:sha-512/256"),
+        SHA1("SHA", "urn:sha1", "sha-1", "sha1"),
+        SHA256("SHA-256", "urn:sha-256", "sha256"),
+        SHA512("SHA-512", "urn:sha-512", "sha512"),
+        SHA512256("SHA-512/256", "urn:sha-512/256", "sha512/256"),
         MD5("MD5", "urn:md5"),
         MISSING("NONE", "missing");
 
         final public String algorithm;
         final private String scheme;
+        final private Set<String> aliases;
 
-        DIGEST_ALGORITHM(final String alg, final String scheme) {
+        DIGEST_ALGORITHM(final String alg, final String scheme, final String... aliases) {
             this.algorithm = alg;
             this.scheme = scheme;
+            this.aliases = Arrays.stream(ArrayUtils.add(aliases, algorithm))
+                    .map(String::toLowerCase)
+                    .collect(Collectors.toSet());
         }
 
         /**
@@ -78,6 +100,20 @@ public final class ContentDigest {
         }
 
         /**
+         * Return enum value for the provided algorithm
+         *
+         * @param alg algorithm name to seek
+         * @return enum value associated with the algorithm name, or missing if not found
+         */
+        public static DIGEST_ALGORITHM fromAlgorithm(final String alg) {
+            final String seek = alg.toLowerCase();
+            return Arrays.stream(values())
+                    .filter(value -> value.aliases.contains(seek))
+                    .findFirst()
+                    .orElse(MISSING);
+        }
+
+        /**
          * Return true if the provided algorithm is included in this enum
          *
          * @param alg to test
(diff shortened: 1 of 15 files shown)


awoods:
A minor usability point: When I perform the following request (which contains correct values for each of the included algorithms):
```
curl -i -XPOST http://localhost:8080/fcrepo/rest/ -H"Slug: binary" --data-binary @README.md -H"Content-type: text/plain" -ufedoraAdmin:fedoraAdmin -H"digest: md5=f702e2340824ebc9f2f6199c6f165181, sha-1=f08cf5cf212cff8aa1937f98eb76ef69b8cd3b50, sha-256=524d6af351c67d4628ec39ae1f719f6a7899b09f4bd546255283f1a64238dc44"
```

..I get the following error:
```
400
Unsupported Digest Algorithim: md5=f702e2340824ebc9f2f6199c6f165181, sha-1=f08cf5cf212cff8aa1937f98eb76ef69b8cd3b50, sha-256=524d6af351c67d4628ec39ae1f719f6a7899b09f4bd546255283f1a64238dc44
```

It would be helpful if either in the response message of the Tomcat log there contained more detail about which algorithm is not supported.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/utils/ContentDigest.java
```diff
@@ -38,20 +41,39 @@
 
     private static final Logger LOGGER = getLogger(ContentDigest.class);
 
+    private final static String DEFAULT_DIGEST_ALGORITHM_PROPERTY = "fcrepo.persistence.defaultDigestAlgorithm";
```
awoods, line 44: Thoughts on refactoring: "fedoraconfig:defaultDigestAlgorithm"?
https://github.com/fcrepo4/fcrepo4/blob/master/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraTypes.java#L90

awoods:
Note, setting the system property, "fcrepo.persistence.defaultDigestAlgorithm" to `sha256` resulted in the default in the ".json" file being correct.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/utils/ContentDigest.java
bbpennel: Based on your slack response from the other day, I was thinking we would end up pulling the `fedoraconfig:defaultDigestAlgorithm` since it seems unnecessary with the removal of fcr:fixity and emphasis on the Want-Digest header. I can do so now, and update the documentation if you'd like.

On the other hand, it might make sense to reimagine it as a way for clients to tell fedora what digest algorithm to use for content addressing in a particular ocfl object, allowing them to override the system wide `DEFAULT_DIGEST_ALGORITHM`. But that likely doesn't fit into this PR, since it would probably come with some twists in terms of allowing clients to change existing objects. From what peter was saying I had the impression that was an allowed ocfl operation, but I would expect it would require significant changes to the object.

bbpennel:
@awoods I believe the feedback should be addressed now, and the error message updated:
```
$ curl -i -XPOST http://localhost:8080/rest/ -H"Slug: binary" --data-binary @README.md -H"Content-type: text/plain" -ufedoraAdmin:fedoraAdmin -H"digest: md5=f702e2340824ebc9f2f6199c6f165181, sha-1=f08cf5cf212cff8aa1937f98eb76ef69b8cd3b50, sha-256=524d6af351c67d4628ec39ae1f719f6a7899b09f4bd546255283f1a64238dc44"

HTTP/1.1 400 Bad Request
Unsupported Digest Algorithm: sha-1

$ curl -i -XPOST http://localhost:8080/rest/ -H"Slug-H"Content-type: text/plain" -ufedoraAdmin:fedoraAdmin -H"digest: md99=f702e2340824ebc9f2f6199c6f165181, sha-1=f08cf5cf212cff8aa1937f98eb76ef69b8cd3b50, sha-2000=524d6af351c67d4628ec39ae1f719f6a7899b09f4bd546255283f1a64238dc44"

HTTP/1.1 400 Bad Request
Unsupported Digest Algorithms: md99,sha-1,sha-2000
```

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/utils/ContentDigest.java
awoods: Thanks. Let's take a wait-and-see on `fedoraconfig:defaultDigestAlgorithm`.

PR #1690, opened 2020-06-05
FCREPO-3201 - Transmission fixity when staging to OCFL object

bbpennel:
https://jira.lyrasis.org/browse/FCREPO-3201

* During a commit, a digest is read from the header file for binaries and provided to the OCFL client for transmission fixity purposes, if there is an appropriate digest available
    * performed by first staging content files before sidecar files, so that the sidecar files are still addressable
* All fedora generated objects have first real commit at v2, due to no longer using `putObject` to create new objects.

# How should this be tested?

Things should appear to function as before. There is an integration test that attempts to simulate this situation by changing a staged file after it would have passed fcrepo transmission fixity checking, but this still isn't exactly like the real world case.

# Notes
It is very unlikely for transmission fixity failures to occur at this stage since we are using MOVE operations to stage files to OCFL, so no byte copying occurs. So it is somewhat questionable how useful this PR. It would still have some effect in the case of the fcrepo staging and ocfl storage locations being on different logical volumes. So it wouldn't hurt my feelings if we don't ultimately want to do this change.

Regarding reading digests from header files, it'd probably be possible to build a cache of digests at write time and use that instead of reading at commit time. I was a bit worried about the memory cost in a long running transaction, but it could be explored if this approach is undesirable.


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OCFLPersistentStorageUtils.java
```diff
@@ -81,7 +82,7 @@ private OCFLPersistentStorageUtils() {
      * The directory within an OCFL Object's content directory that contains
      * information managed by Fedora.
      */
-    private static final String INTERNAL_FEDORA_DIRECTORY = ".fcrepo";
+    public static final String INTERNAL_FEDORA_DIRECTORY = ".fcrepo";
```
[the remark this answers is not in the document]
bbpennel: Ah, I needed it public for an earlier version, but not the final one. Should be fixed now

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractNonRdfSourcePersister.java
```diff
@@ -88,12 +88,13 @@ protected void persistNonRDFSource(final ResourceOperation operation,
...
             outcome = null;
         } else {
+            final var transmissionDigestAlg = objectSession.getObjectDigestAlgorithm();
             final var providedDigests = nonRdfSourceOperation.getContentDigests();
             // Wrap binary stream in digest computing wrapper, requesting
             final var multiDigestWrapper = new MultiDigestInputStreamWrapper(
                     nonRdfSourceOperation.getContentStream(),
                     providedDigests,
-                    Arrays.asList(objectSession.getObjectDigestAlgorithm()));
+                    Arrays.asList(transmissionDigestAlg));
```
pwinckles, line 97: No need to change, but this can now be constructed using `List.of()`.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
```diff
@@ -449,6 +457,53 @@ private String commitUpdates(final CommitOption commitOption) {
...
+        return updater -> {
+            if (updatingObject) {
+                deletePaths.forEach(updater::removeFile);
+            }
+            if (isStagingEmpty()) {
+                return;
+            }
+
+            try {
+                Files.walk(stagingPath)
```
pwinckles, line 480: I don't think you need to walk the staged files, unless you want to. You should just be able to add the entire staging directory, and then iterate over your map of digests, calling `addFileFixity` for each.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/AbstractNonRdfSourcePersister.java
bbpennel: Sure, I'll swap it

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DefaultOCFLObjectSession.java
bbpennel: Good catch, I was still in the mindset of needing to walk files, but there's really no need at this point. I've switched to your suggestion

PR #1701, opened 2020-06-16
Simple attempt at including user types in getTypes

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3279

Separates the inclusion of user defined types from system defined types. When getting types we now have to read the user's RDF and filter for types. 

Perhaps this information should be included in the header file and eventually be cached for quick access?

As a user could add or remove a `rdf:type` at any time it seems like we might need to do this at POST/PUT/PATCH? Perhaps in the persistence layer?

# How should this be tested?

Not sure how this would expose itself through common use other than now the WebAC `acl:accessToClass` directives work because we get the user defined types back when we `getTypes()`

# Interested parties
@fcrepo4/committers


whikloj:
I thought about keeping the two statics separate and adding the needed types to a new list to return. Wasn't sure if this was more efficient then generating a new list each type to return the same types

PR #1703, opened 2020-06-18
Moves the search index into its own table.

dbernstein:
Adds support for searching by modified and created dates.
Adds support for customizing the output fields of your search.
By default all fields returned.

Resolves:
https://jira.lyrasis.org/browse/FCREPO-3349
https://jira.lyrasis.org/browse/FCREPO-3353
https://jira.lyrasis.org/browse/FCREPO-3354

* Moves the search index into its own table.
* Adds support for searching by modified and created dates.
* Adds support for customizing the output fields of your search.
* The index is kept up to date now via the internal event bus.
* Integration tests cover the following features:
   1. search by resources with modified or created date <, >, >=, or <= . Ensure that "before", "after" and "before and after" (ie between) are supported.
   2. compound queries are now supported: ie your can specify multiple conditions.

# How should this be tested?
Create a bunch of resources
```
for i in `seq 1 50`; do curl -X PUT -u fedoraAdmin:fedoraAdmin  "http://localhost:8080/rest/resource-$i"; done;
```
See that the modified and created dates are returned by default (`jq` makes the json pretty):
```
curl  -u fedoraAdmin:fedoraAdmin "http://localhost:8080/rest/fcr:search" | jq 
```
Return all available fields in results: (should be identical to the previous results )
```
curl -u fedoraAdmin:fedoraAdmin "http://localhost:8080/rest/fcr:search?fields=*" | jq
```
Return only the fedora_id and modified in results:
```
curl -u fedoraAdmin:fedoraAdmin "http://localhost:8080/rest/fcr:search?fields=fedora_id,modified" | jq
```
Return results matching fedora_id and created after using YYYY-mm-dd format. Should return one result
```
curl -u fedoraAdmin:fedoraAdmin "http://localhost:8080/rest/fcr:search?condition=fedora_id%3Dresource-1&condition=created%3E2020-06-18&fields=fedora_id,modified" | jq
```
Return results matching fedora_id and created before using YYYY-mm-dd format. Should return no  results.
```
curl -u fedoraAdmin:fedoraAdmin "http://localhost:8080/rest/fcr:search?condition=fedora_id%3Dresource-1&condition=created%3C2020-06-18&fields=fedora_id,modified" | jq
```
Return results matching fedora_id and modified after using ISO date format. Should return one result
```
curl -u fedoraAdmin:fedoraAdmin "http://localhost:8080/rest/fcr:search?condition=fedora_id%3Dresource-1&condition=modified%3E2020-06-18T19:04:10.802Z&fields=fedora_id,modified" | jq
```

# Interested parties


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraSearch.java
```diff
@@ -87,15 +89,30 @@ public Response doSearch(@QueryParam(value = "condition") final List<String> con
                 final var parsedCondition = parse(condition, identifierConverter());
                 conditionList.add(parsedCondition);
             }
-            final var params = new SearchParameters(conditionList, maxResults, offset);
+
+            List<Condition.Field> parsedFields = null;
+            if (fields == null || fields.trim().length() == 0 || fields.equals("*")) {
```
pwinckles, line 94: The first 2 clauses could be expressed as `StringUtils.isBlank(fields)`

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -0,0 +1,241 @@
...
+                status.setRollbackOnly();
+                throw new RepositoryRuntimeException("Failed add/updated the search index for : " + fullId, e);
+            }
+        });
+    }
+
+    @Override
+    public void removeFromIndex(final FedoraId fedoraId) {
+        final var txId = UUID.randomUUID().toString();
+        executeInDbTransaction(txId, status -> {
```
pwinckles, line 213: I don't think this needs a transaction, does it?

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/SearchIndexUpdater.java
```diff
@@ -0,0 +1,108 @@
...
+    @Inject
+    private PersistentStorageSessionManager persistentStorageSessionManager;
+
+    private static Logger LOGGER = LoggerFactory.getLogger(SearchIndexUpdater.class);
+    private static final Set<EventType> HANDLED_TYPES = Sets.newHashSet(RESOURCE_CREATION, RESOURCE_MODIFICATION,
+            RESOURCE_DELETION);
+
+    @Subscribe
+    public void onEvent(final Event event) {
+        LOGGER.info("event={}", event);
```
pwinckles, line 71: This seems more like a debug level log

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/SearchIndexUpdater.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Inject
+    private PersistentStorageSessionManager persistentStorageSessionManager;
+
+    private static Logger LOGGER = LoggerFactory.getLogger(SearchIndexUpdater.class);
+    private static final Set<EventType> HANDLED_TYPES = Sets.newHashSet(RESOURCE_CREATION, RESOURCE_MODIFICATION,
+            RESOURCE_DELETION);
+
+    @Subscribe
+    public void onEvent(final Event event) {
```
pwinckles, line 70: This should probably be annotated with `@AllowConcurrentEvents`

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/IndexBuilderImpl.java
```diff
@@ -63,12 +63,18 @@
...
+    @Inject
+    private SearchIndex searchIndex;
+
     @Inject
     private TransactionManager transactionManager;
 
     @Inject
     private OcflRepository ocflRepository;
 
+    public IndexBuilderImpl() {
```
pwinckles, line 75: I don't think this class needs this constructor, does it?

fcrepo-search-api/src/main/java/org/fcrepo/search/api/SearchIndex.java
```diff
@@ -17,11 +17,31 @@
...
+public interface SearchIndex {
+
+    /**
+     * @param fedoraId
+     * @param created
+     * @param modified
+     * @param size
+     * @param mimetype
+     */
+    void addUpdateIndex(final FedoraId fedoraId, final Instant created, final Instant modified,
```
pwinckles, line 38: Does this module have access to the `ResourceHeaders` interface? If so, it might be nice to have an api that just took `FedoraId` and `ResourceHeaders` rather than having to pass each value individually.

fcrepo-search-api/src/main/java/org/fcrepo/search/api/SearchParameters.java
```diff
@@ -74,6 +78,14 @@ public int getMaxResults() {
...
+    /**
+     * Returns the list of fields to display in the results.
+     * @return
+     */
+    public List<Condition.Field> getFields() {
+        return fields;
+    }
+
     @Override
     public String toString() {
```
pwinckles, line 90: Should probably update the `toString` to include the additional fields.

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -0,0 +1,241 @@
...
+        for (Condition condition : parameters.getConditions()) {
+            final var field = condition.getField();
+            final var operation = condition.getOperator();
+            var object = condition.getObject();
+            if (field.equals(Condition.Field.FEDORA_ID) &&
+                    condition.getOperator().equals(Condition.Operator.EQ)) {
+                if (!object.equals("*")) {
+                    if (object.contains("*")) {
+                        object = object.replace("*", "%");
+                        whereClauses.add(Condition.Field.FEDORA_ID + " like '" + object + "'");
```
pwinckles, line 112: Do we need to take any sql injection precautions here?

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -0,0 +1,241 @@
...
+
+        final var txId = UUID.randomUUID().toString();
+        executeInDbTransaction(txId, status -> {
+            try {
+
+                final var params = new MapSqlParameterSource();
+                params.addValue("fedora_id", fullId);
+                params.addValue("modified", modified);
+
+                if (result.size() > 0) {
```
pwinckles, line 190: Small simplification:

```java
final String sql;

if (result.size() > 0) {
    sql = "...";
} else {
    params.add...;
    sql = "...";
}

jdbcTemplate.update(sql, params);
```

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/InstantParser.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+public class InstantParser {
+
+    private static final List<DateTimeFormatter> VALID_DATE_FORMATS = new ArrayList<>();
+
+    static {
+        VALID_DATE_FORMATS.add(ISO_DATE_TIME);
+        VALID_DATE_FORMATS.add(ISO_OFFSET_DATE_TIME);
+        VALID_DATE_FORMATS.add(RFC_1123_DATE_TIME);
+        final var zoneId = ZoneId.of("UTC");
```
pwinckles, line 45: You can also get this with `ZoneOffset.UTC`

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/SearchIndexUpdater.java
```diff
@@ -0,0 +1,108 @@
...
+        LOGGER.info("event={}", event);
+        try {
+            final var types = event.getTypes();
+            for (EventType et : types) {
```
pwinckles, line 74: I'm not sure there's a guaranteed event type ordering, so if multiple events occur on the same object in the same transaction, the search index could be updated in the wrong order here.

fcrepo-search-api/src/main/java/org/fcrepo/search/api/SearchIndex.java
dbernstein: Yeah - that's valid.  I wasn't sure if we wanted to couple the interface too tightly here.

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -0,0 +1,241 @@
...
+
+    @Inject
+    private PlatformTransactionManager platformTransactionManager;
+
+    /**
+     * Setup database table and connection
+     */
+    @PostConstruct
+    public void setup() {
+        LOGGER.info("Applying ddl: {}", DDL);
```
whikloj, line 85: Since the value of `DDL` doesn't change it could either be removed or moved to a debug level.

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    @Override
+    public void addUpdateIndex(final FedoraId fedoraId, final Instant created, final Instant modified,
+                               final Long size, final String mimetype) {
+        final var fullId = fedoraId.getFullId();
+        final var selectParams = new MapSqlParameterSource();
+        selectParams.addValue("fedora_id", fullId);
+        final var result = jdbcTemplate.queryForList("SELECT fedora_id FROM " + SIMPLE_SEARCH_TABLE +
+                        " WHERE fedora_id = :fedora_id",
```
whikloj, line 179: Can use `Condition.Field.FEDORA_ID` here and the line above, ie.
```
        final var result = jdbcTemplate.queryForList("SELECT " + Condition.Field.FEDORA_ID + " FROM " + SIMPLE_SEARCH_TABLE +
                        " WHERE " + Condition.Field.FEDORA_ID + " = :fedora_id",
```

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/SearchIndexUpdater.java
whikloj: This is a good point, and because the only thing we care is that something happened to the resource maybe regardless of the event we should just see if the resource exists. If it does reindex and if not delete.

In which case it would be good to not reindex for multiple quickly occurring events. Like an aggregator which collects all the events that occur on a single resource in a specified time frame and then reduces them to a single action.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraSearch.java
dbernstein: will fix.

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
dbernstein: good catch.  I'll remove.

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/SearchIndexUpdater.java
dbernstein: Yes, I agree.

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/SearchIndexUpdater.java
dbernstein: Certainly.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/IndexBuilderImpl.java
dbernstein: Yes - somehow I unintentionally  introduced that.

fcrepo-search-api/src/main/java/org/fcrepo/search/api/SearchIndex.java
dbernstein: I'm inclined to leave it as is - perhaps @awoods  may want to weigh in.

fcrepo-search-api/src/main/java/org/fcrepo/search/api/SearchParameters.java
dbernstein: Will do.

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
dbernstein: will change to debug.

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
dbernstein: Yes - good catch - I'll use parameterSource for those.

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
dbernstein: +1

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/SearchIndexUpdater.java
dbernstein: I'm not sure that I understand the issue here.  

Are you saying that a single event can have a "delete",  a "create" and an "update" event type?

dbernstein: I'll integrate @whikloj 's suggestion.

fcrepo-search-api/src/main/java/org/fcrepo/search/api/SearchIndex.java
pwinckles: Are these the only fields that this method is going to accept, or are you planning on indexing more ResourceHeaders fields in the future as you build search out? If you're not planning on adding more, I'm okay with it. Otherwise, I think it would be better to have a different object in this method instead of an ever growing parameter list.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraSearchIT.java
```diff
@@ -124,11 +136,184 @@ public void testWildcardMatchingOnPartialFedoraId() throws Exception {
...
+                getSearchEndpoint() + "condition=" + encode(fedoraIdCondition) + "&" + greaterThanTomorrow;
+        try (final CloseableHttpResponse response = execute(new HttpGet(searchUrl))) {
+            assertEquals(OK.getStatusCode(), getStatus(response));
+            final ObjectMapper objectMapper = new ObjectMapper();
+            final SearchResult result = objectMapper.readValue(response.getEntity().getContent(),
+                    SearchResult.class);
+            assertEquals("expected no results", 0,
+                    result.getItems().size());
+            result.getItems().stream().map(x -> x.get("fedora_id"))
+                    .forEach(x -> assertTrue(x.toString().equals(fedoraId)));
```
whikloj, line 247: The size of `result.getItems().size()` is asserted as 0 above, so this stream is never run and can probably be removed.

fcrepo-search-api/src/main/java/org/fcrepo/search/api/SearchIndex.java
whikloj: It appears that this PR removed all the other fields, so it is now FEDORA_ID, MODIFIED or CREATED? Or is that not part of this?

dbernstein: So we will be adding fields  (size and mimetype) to this list.   I suppose we could have ResourceHeaders passed in.  I'm not sure why I was initially resistant to it.  

@whikloj :  the other fields will be added back in when they are actually supported by search.  Yes I did remove them to keep things tidy until we actually need them.

dbernstein: I will use ResourceHeaders.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraSearchIT.java
dbernstein: Good eye.

PR #1704, opened 2020-06-19
Don't fail with stacktrace on empty path segment in WebAC Filter

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3339

When passing a URI with an empty path segment into the WebAC Filter it ends up without roles, but we don't account for that so it does a stacktrace.
This now checks that it isn't null before proceeding, resulting in a 403 for the path with empty path segments.

# How should this be tested?

Before this PR

1. Start a Fedora 5
1. Execute `curl -utestuser:testpass http://localhost:8080/rest/bob//bob`
1. See stacktrace.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-auth-webac/src/test/java/org/fcrepo/integration/auth/webac/WebACRecipesIT.java
+++ b/fcrepo-auth-webac/src/test/java/org/fcrepo/integration/auth/webac/WebACRecipesIT.java
@@ -1951,10 +1951,15 @@ public class WebACRecipesIT extends AbstractResourceIT {
         final HttpGet getAdminRequest = getObjMethod(parent + "//" + child);
         setAuth(getAdminRequest, "fedoraAdmin");
         assertEquals(BAD_REQUEST.getStatusCode(), getStatus(getAdminRequest));
-        // User can't access the URL (because the path is not correct).
+        // Bad request, empty path in the middle.
         final HttpGet getUserRequest = getObjMethod(parent + "//" + child);
         setAuth(getUserRequest, username);
-        assertEquals(FORBIDDEN.getStatusCode(), getStatus(getUserRequest));
+        assertEquals(BAD_REQUEST.getStatusCode(), getStatus(getUserRequest));
+        // Bad request, empty path at the end.
+        final HttpGet getUserRequest2 = getObjMethod(parent + "/" + child + "//");
+        setAuth(getUserRequest2, username);
+        assertEquals(BAD_REQUEST.getStatusCode(), getStatus(getUserRequest2));
+        // On the front side extra slashes seem to get stripped off, so no testing here.
     }
 
     /**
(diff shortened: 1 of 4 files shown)


fcrepo-auth-webac/src/test/java/org/fcrepo/integration/auth/webac/WebACRecipesIT.java
```diff
@@ -1888,6 +1893,70 @@ public void testDirectRelationshipsOk() throws IOException {
...
+        assertEquals(OK.getStatusCode(), getStatus(getChild));
+
+        // Admin bypasses ACL resolution gets 409.
+        final HttpGet getAdminRequest = getObjMethod(parent + "//" + child);
+        setAuth(getAdminRequest, "fedoraAdmin");
+        assertEquals(BAD_REQUEST.getStatusCode(), getStatus(getAdminRequest));
+        // User can't access the URL (because the path is not correct).
+        final HttpGet getUserRequest = getObjMethod(parent + "//" + child);
+        setAuth(getUserRequest, username);
+        assertEquals(FORBIDDEN.getStatusCode(), getStatus(getUserRequest));
```
awoods, line 1957: This is better than a stacktrace... I am fine with 403 for the 5.1.1 patch release.

whikloj: Yeah, I didn't look to closely to figure out where the empty path exception comes from, but I feel like it comes from Modeshape. I was able to perform the matching with this
```
    private boolean hasEmptyPathSegments(final HttpServletRequest httpRequest) {
        final String requestPath = httpRequest.getContextPath() + httpRequest.getServletPath() +
                httpRequest.getRequestURI();
        final String finalTestPath;
        if (requestPath.startsWith("/") && requestPath.endsWith("/") && requestPath.length() > 1) {
            finalTestPath = requestPath.substring(1, requestPath.length() - 2);
        } else if (requestPath.startsWith("/")) {
            finalTestPath = requestPath.substring(1);
        } else if (requestPath.endsWith("/")) {
            finalTestPath = requestPath.substring(0, requestPath.length() - 1);
        } else {
            finalTestPath = requestPath;
        }
        if (finalTestPath.contains("/")) {
            final String[] paths = finalTestPath.split("/");
            if (Arrays.stream(paths).anyMatch(String::isEmpty)) {
                return true;
            }
        }
        return false;
    }
```
I'm not sure how fragile that is, but if you want I can add a commit here to return the appropriate 400 Bad Request. I'd probably add more test cases here than in the 6.0.0 location.

awoods: Thanks, @whikloj . However, I am inclined to move forward with this PR as it stands.

whikloj: Okay, so you want we to revert the last commit?

awoods: No... it is an improvement. Thanks.

PR #1709, opened 2020-07-02
Remove fedoraToOcflMapping

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3357

Remove the FedoraToOcflIndex mapping when the resource is purged (ie. tombstone is removed)

# How should this be tested?

Follow steps to cause the error in the attached ticket, with this PR you should get no error.

# Interested parties
@fcrepo4/committers


whikloj:
To account for possible dangling indexes and mappings inside transaction this will be significantly different. Closing.

PR #1710, opened 2020-07-03
Remove fedoraToOcflMapping on purge and use a new transaction table t…

whikloj:
…o avoid dangling identifier mappings.

**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3357

Removes fedoraToOcflIndex mappings when a resource is purged.

Also uses the transaction pattern of the containment index to avoid having mappings from transactions bleed into the outside world.

# How should this be tested?

There are steps in the attached ticket to cause the error this fixes.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndex.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndex.java
@@ -163,15 +163,6 @@ public class DbFedoraToOcflObjectIndex implements FedoraToOcflObjectIndex {
     );
 
     /*
-     * Add records to the mapping table that are to be added in this transaction.
-     */
-    private static final String COMMIT_ADD_RECORDS = "INSERT INTO " + MAPPING_TABLE + " ( " + FEDORA_ID_COLUMN + ", "
-            + FEDORA_ROOT_ID_COLUMN + ", " + OCFL_ID_COLUMN + " ) SELECT " + FEDORA_ID_COLUMN + ", " +
-            FEDORA_ROOT_ID_COLUMN + ", " + OCFL_ID_COLUMN + " FROM " +
-            TRANSACTION_OPERATIONS_TABLE + " WHERE " + TRANSACTION_ID_COLUMN + " = :transactionId AND " +
-            OPERATION_COLUMN + " = 'add'";
-
-    /*
      * Delete records from the mapping table that are to be deleted in this transaction.
      */
     private static final String COMMIT_DELETE_RECORDS = "DELETE FROM " + MAPPING_TABLE + " WHERE " +
@@ -204,7 +195,7 @@ public class DbFedoraToOcflObjectIndex implements FedoraToOcflObjectIndex {
 
     private PlatformTransactionManager platformTransactionManager;
 
-    private String DATABASE_PRODUCT_NAME;
+    private String databaseProductName;
 
     public DbFedoraToOcflObjectIndex(@Autowired final DataSource dataSource) {
         this.dataSource = dataSource;
@@ -223,11 +214,11 @@ public class DbFedoraToOcflObjectIndex implements FedoraToOcflObjectIndex {
 
     private String lookupDdl() {
         try (final var connection = dataSource.getConnection()) {
-            DATABASE_PRODUCT_NAME = connection.getMetaData().getDatabaseProductName();
-            LOGGER.debug("Identified database as: {}", DATABASE_PRODUCT_NAME);
-            final var ddl = DDL_MAP.get(DATABASE_PRODUCT_NAME);
+            databaseProductName = connection.getMetaData().getDatabaseProductName();
+            LOGGER.debug("Identified database as: {}", databaseProductName);
+            final var ddl = DDL_MAP.get(databaseProductName);
             if (ddl == null) {
-                throw new IllegalStateException("Unknown database platform: " + DATABASE_PRODUCT_NAME);
+                throw new IllegalStateException("Unknown database platform: " + databaseProductName);
             }
             return ddl;
         } catch (final SQLException e) {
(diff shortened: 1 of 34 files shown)


fcrepo-persistence-ocfl/src/main/resources/sql/default-ocfl-index.sql
```diff
@@ -1,9 +1,26 @@
...
+CREATE TABLE IF NOT EXISTS ocfl_id_map_session_operations (
+    fedora_id varchar(503) NOT NULL PRIMARY KEY,
+    fedora_root_id varchar(503) NULL,
+    ocfl_id varchar(503) NULL,
+    session_id varchar(255) NOT NULL,
+    operation varchar(10) NOT NULL
+);
+
+-- Create an index to speed searches for records related to adding/excluding transaction records
+CREATE INDEX IF NOT EXISTS ocfl_id_map_idx1
```
pwinckles, line 21: MySQL does not support `CREATE INDEX IF NOT EXISTS`. A MySQL specific file needs to be created, similar to the how the containment index works.

whikloj:
@pwinckles I don't understand those queries so I'll have to learn more about them. 

But in the situation where a person:
1. Creates a container in a transaction.
2. Deletes the same container in the same transaction. 

Would that not mean that at `commit()` the "delete" operation would occur (deleting nothing as the containment relationship doesn't exist yet), then the "add" occurs and adds it to the containment index?

pwinckles:
> 
> 
> @pwinckles I don't understand those queries so I'll have to learn more about them.
> 
> But in the situation where a person:
> 
>     1. Creates a container in a transaction.
> 
>     2. Deletes the same container in the same transaction.
> 
> 
> Would that not mean that at `commit()` the "delete" operation would occur (deleting nothing as the containment relationship doesn't exist yet), then the "add" occurs and adds it to the containment index?

No, because the transaction table would only contain the `delete` operation. The transaction table never contains more than one entry for any `fedora_id`, `session_id` pair. On add, an `add` operation is added to the transaction table. Then, when the resource is deleted, the same row in the transaction table is updated so that it is now a `delete` operation.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndex.java
```diff
@@ -54,67 +70,258 @@
...
+        final var ddl = lookupDdl();
+        LOGGER.info("Applying ddl: {}", ddl);
         DatabasePopulatorUtils.execute(
-                new ResourceDatabasePopulator(new DefaultResourceLoader().getResource("classpath:" + DDL)),
+                new ResourceDatabasePopulator(new DefaultResourceLoader().getResource("classpath:" + ddl)),
                 dataSource);
     }
 
+    private String lookupDdl() {
+        try (final var connection = dataSource.getConnection()) {
```
pwinckles, line 216: Ideally, the db platform names would be an enum like:

```java
public enum DbPlatform {
    POSTGRESQL("PostgreSQL"), H2("H2"), MYSQL("MySQL"), MARIADB("MariaDB");

    private final String name;

    private DbPlatform(final String name) {
        this.name = name;
    }

    private static DbPlatform fromDataSource(final DataSource dataSource) {
        try (final var connection = dataSource.getConnection()) {
            return fromString(connection.getMetaData().getDatabaseProductName());
        } catch (final SQLException e) {
            throw new RuntimeException(e);
        }
    }

    private static DbPlatform fromString(final String name) {
        for (var value : values()) {
            if (value.name.equals(name)) {
                return value;
            }
        }
        throw new IllegalArgumentException("Unknown database platform: " + name);
    }
}
```

But, it's unclear to me if there is a good common library to put that code in.

At the minimum, setting the `databaseProductName` needs to happen in the `setup()` method and not as a side effect of calling `lookupDdl`.

fcrepo-persistence-ocfl/src/main/resources/sql/default-ocfl-index.sql
```diff
@@ -1,9 +1,26 @@
...
+    ON ocfl_id_map_session_operations (fedora_id, session_id, operation);
```
pwinckles, line 22: This index should only be on `fedora_id` and `session_id`

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndex.java
pwinckles: If there isn't a good place to share the code (I believe I have duplicate code in another class). Leaving the duplicate, string based code is fine for now.

fcrepo-persistence-ocfl/src/main/resources/sql/default-ocfl-index.sql
pwinckles: Additionally, this index must be unique.

pwinckles:
I tried using PostgreSQL and MySQL, and I got stack traces from both when I tried to purge:

```
org.fcrepo.persistence.api.exceptions.PersistentStorageException: No OCFL mapping found for info:fedora/foo
	at org.fcrepo.persistence.ocfl.impl.AbstractPersister.getMapping(AbstractPersister.java:113)
	at org.fcrepo.persistence.ocfl.impl.PurgeResourcePersister.persist(PurgeResourcePersister.java:51)
	at org.fcrepo.persistence.ocfl.impl.OCFLPersistentStorageSession.persist(OCFLPersistentStorageSession.java:164)
	at org.fcrepo.kernel.impl.services.PurgeResourceServiceImpl.doAction(PurgeResourceServiceImpl.java:61)
	at org.fcrepo.kernel.impl.services.AbstractDeleteResourceService.deleteDepthFirst(AbstractDeleteResourceService.java:122)
	at org.fcrepo.kernel.impl.services.AbstractDeleteResourceService.perform(AbstractDeleteResourceService.java:74)
	at org.fcrepo.http.api.FedoraTombstones.delete(FedoraTombstones.java:89)
```

fcrepo-persistence-ocfl/src/main/resources/sql/default-ocfl-index.sql
whikloj: This index is to make it more efficient ever time we do upserts, might be able to remove `fedora_id`. But the index you are talking about is below.

pwinckles:
Closing in favor of https://github.com/fcrepo4/fcrepo4/pull/1724

PR #1712, opened 2020-07-06
Ensure files are marked as deleted

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3338

Not entirely sure what was happening, but it appears that when deleting binaries we were executing 2 delete requests. The DeleteResourceService called `resource().getDescribedResource()` to delete, but that is the binary. Changed that to `getDescription()` and removed the check to delete the binary-description from the persister.

This seemed to resolve the issue, so maybe it was a race condition?

# How should this be tested?

Follow the steps in the ticket and with this PR, the deleted children of an AG stay deleted.

# Interested parties
@fcrepo4/committers


whikloj:
Not related, rolling into #1711

PR #1726, opened 2020-07-16
Implements rdf_type parameter and output field in simple search.

dbernstein:
**Implements rdf_type parameter and output field in simple search.**
* * *

**JIRA Ticket**:  https://jira.lyrasis.org/browse/FCREPO-3378

With this PR, users can see the rdf_types of their resources in the search results.  Additionally they can perform wildcard (*) searches using the rdf_type field in conditional clauses.

# How should this be tested?
```
# create a resource
curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/test -X PUT

#  see the rdf_type in the search results
curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/fcr:search

# use rdf_type as a criterion for your search
curl -u fedoraAdmin:fedoraAdmin "http://localhost:8080/rest/fcr:search?condition=rdf_type%3D*Con*"
```

# Interested parties

### DIFF
--- a/fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACAuthorizingRealm.java
+++ b/fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACAuthorizingRealm.java
@@ -17,28 +17,6 @@
  */
 package org.fcrepo.auth.webac;
 
-import static org.fcrepo.auth.common.ServletContainerAuthFilter.FEDORA_ADMIN_ROLE;
-import static org.fcrepo.auth.common.ServletContainerAuthFilter.FEDORA_USER_ROLE;
-import static org.fcrepo.auth.webac.URIConstants.FOAF_AGENT_VALUE;
-import static org.fcrepo.auth.webac.URIConstants.WEBAC_AUTHENTICATED_AGENT_VALUE;
-import static org.fcrepo.auth.common.HttpHeaderPrincipalProvider.HttpHeaderPrincipal;
-import static org.fcrepo.auth.common.DelegateHeaderPrincipalProvider.DelegatedHeaderPrincipal;
-import static org.fcrepo.auth.webac.WebACFilter.getBaseUri;
-import static org.fcrepo.auth.webac.WebACFilter.identifierConverter;
-import static org.fcrepo.http.commons.session.TransactionConstants.ATOMIC_ID_HEADER;
-import static org.slf4j.LoggerFactory.getLogger;
-
-import java.net.URI;
-import java.security.Principal;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Map;
-import java.util.Set;
-
-import javax.inject.Inject;
-import javax.servlet.http.HttpServletRequest;
-
 import org.apache.http.auth.BasicUserPrincipal;
 import org.apache.shiro.authc.AuthenticationException;
 import org.apache.shiro.authc.AuthenticationInfo;
@@ -58,6 +36,27 @@ import org.fcrepo.kernel.api.identifiers.FedoraId;
 import org.fcrepo.kernel.api.models.FedoraResource;
 import org.fcrepo.kernel.api.models.ResourceFactory;
 import org.slf4j.Logger;
+
+import javax.inject.Inject;
+import javax.servlet.http.HttpServletRequest;
+import java.net.URI;
+import java.security.Principal;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import static org.fcrepo.auth.common.DelegateHeaderPrincipalProvider.DelegatedHeaderPrincipal;
+import static org.fcrepo.auth.common.HttpHeaderPrincipalProvider.HttpHeaderPrincipal;
+import static org.fcrepo.auth.common.ServletContainerAuthFilter.FEDORA_ADMIN_ROLE;
+import static org.fcrepo.auth.common.ServletContainerAuthFilter.FEDORA_USER_ROLE;
+import static org.fcrepo.auth.webac.URIConstants.FOAF_AGENT_VALUE;
+import static org.fcrepo.auth.webac.URIConstants.WEBAC_AUTHENTICATED_AGENT_VALUE;
+import static org.fcrepo.auth.webac.WebACFilter.getBaseUri;
+import static org.fcrepo.auth.webac.WebACFilter.identifierConverter;
+import static org.fcrepo.http.commons.session.TransactionConstants.ATOMIC_ID_HEADER;
+import static org.slf4j.LoggerFactory.getLogger;
 /**
  * Authorization-only realm that performs authorization checks using WebAC ACLs stored in a Fedora repository. It
  * locates the ACL for the currently requested resource and parses the ACL RDF into a set of {@link WebACPermission}
@@ -234,17 +233,22 @@ public class WebACAuthorizingRealm extends AuthorizingRealm {
     }
 
     private FedoraResource getResourceOrParentFromPath(final FedoraId fedoraId) {
+        final var tx = transaction();
+        final var txId = tx == null || tx.isCommitted() ? null : tx.getId();
+
         try {
             log.debug("Testing FedoraResource for {}", fedoraId.getFullIdPath());
-            return this.resourceFactory.getResource(transaction(), fedoraId);
+
+            return txId != null ? this.resourceFactory.getResource(txId, fedoraId) :
+                    this.resourceFactory.getResource(fedoraId);
         } catch (final PathNotFoundException exc) {
             log.debug("Resource {} not found getting container", fedoraId.getFullIdPath());
-            final String txID = transaction() == null ? null : transaction().getId();
-            final FedoraId containerId = containmentIndex.getContainerIdByPath(txID, fedoraId);
+            final FedoraId containerId = containmentIndex.getContainerIdByPath(txId, fedoraId);
             log.debug("Attempting to get FedoraResource for {}", fedoraId.getFullIdPath());
             try {
                 log.debug("Got FedoraResource for {}", containerId.getFullIdPath());
-                return this.resourceFactory.getResource(transaction(), containerId);
+                return txId != null ? this.resourceFactory.getResource(txId, containerId) :
+                        this.resourceFactory.getResource(containerId);
             } catch (final PathNotFoundException exc2) {
                 log.debug("Path {} does not exist, but we should never end up here.", containerId.getFullIdPath());
                 return null;
--- a/fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACAuthorizingRealm.java
+++ b/fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACAuthorizingRealm.java
@@ -30,6 +30,7 @@ import org.fcrepo.http.commons.session.TransactionProvider;
 import org.fcrepo.kernel.api.ContainmentIndex;
 import org.fcrepo.kernel.api.Transaction;
 import org.fcrepo.kernel.api.TransactionManager;
+import org.fcrepo.kernel.api.TransactionUtils;
 import org.fcrepo.kernel.api.exception.PathNotFoundException;
 import org.fcrepo.kernel.api.exception.RepositoryConfigurationException;
 import org.fcrepo.kernel.api.identifiers.FedoraId;
@@ -233,22 +234,18 @@ public class WebACAuthorizingRealm extends AuthorizingRealm {
     }
 
     private FedoraResource getResourceOrParentFromPath(final FedoraId fedoraId) {
-        final var tx = transaction();
-        final var txId = tx == null || tx.isCommitted() ? null : tx.getId();
+        final var txId = TransactionUtils.openTxId(transaction());
 
         try {
             log.debug("Testing FedoraResource for {}", fedoraId.getFullIdPath());
-
-            return txId != null ? this.resourceFactory.getResource(txId, fedoraId) :
-                    this.resourceFactory.getResource(fedoraId);
+            return this.resourceFactory.getResource(txId, fedoraId);
         } catch (final PathNotFoundException exc) {
             log.debug("Resource {} not found getting container", fedoraId.getFullIdPath());
             final FedoraId containerId = containmentIndex.getContainerIdByPath(txId, fedoraId);
             log.debug("Attempting to get FedoraResource for {}", fedoraId.getFullIdPath());
             try {
                 log.debug("Got FedoraResource for {}", containerId.getFullIdPath());
-                return txId != null ? this.resourceFactory.getResource(txId, containerId) :
-                        this.resourceFactory.getResource(containerId);
+                return this.resourceFactory.getResource(txId, containerId);
             } catch (final PathNotFoundException exc2) {
                 log.debug("Path {} does not exist, but we should never end up here.", containerId.getFullIdPath());
                 return null;
--- a/fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACAuthorizingRealm.java
+++ b/fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACAuthorizingRealm.java
@@ -234,18 +234,17 @@ public class WebACAuthorizingRealm extends AuthorizingRealm {
     }
 
     private FedoraResource getResourceOrParentFromPath(final FedoraId fedoraId) {
-        final var txId = TransactionUtils.openTxId(transaction());
-
         try {
             log.debug("Testing FedoraResource for {}", fedoraId.getFullIdPath());
-            return this.resourceFactory.getResource(txId, fedoraId);
+            return this.resourceFactory.getResource(transaction(), fedoraId);
         } catch (final PathNotFoundException exc) {
             log.debug("Resource {} not found getting container", fedoraId.getFullIdPath());
-            final FedoraId containerId = containmentIndex.getContainerIdByPath(txId, fedoraId);
+            final FedoraId containerId =
+                    containmentIndex.getContainerIdByPath(TransactionUtils.openTxId(transaction()), fedoraId);
             log.debug("Attempting to get FedoraResource for {}", fedoraId.getFullIdPath());
             try {
                 log.debug("Got FedoraResource for {}", containerId.getFullIdPath());
-                return this.resourceFactory.getResource(txId, containerId);
+                return this.resourceFactory.getResource(transaction(), containerId);
             } catch (final PathNotFoundException exc2) {
                 log.debug("Path {} does not exist, but we should never end up here.", containerId.getFullIdPath());
                 return null;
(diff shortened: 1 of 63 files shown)


fcrepo-search-impl/src/main/resources/sql/default-search-index.sql
```diff
@@ -1,8 +1,23 @@
...
     content_size bigint DEFAULT NULL,
-    mime_type varchar(255) DEFAULT NULL
-);
\ No newline at end of file
+    mime_type varchar(255) DEFAULT NULL,
+    UNIQUE KEY fedora_id (fedora_id)
+);
+
+CREATE TABLE IF NOT EXISTS search_resource_rdf_type (
+    resource_id bigint NOT NULL,
```
pwinckles, line 13: Why do these DDLs not have any foreign key relationships?

dbernstein: I ran into an issue with truncation failing when I had foreign keys on there.

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -209,20 +261,73 @@ public void addUpdateIndex(final ResourceHeaders resourceHeaders) {
         final var txId = UUID.randomUUID().toString();
         executeInDbTransaction(txId, status -> {
             try {
+                final var rdfTypes = new ArrayList<String>();
```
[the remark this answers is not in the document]
dbernstein: yes good idea.

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -209,20 +261,73 @@ public void addUpdateIndex(final ResourceHeaders resourceHeaders) {
...
+                final var fedoraResource = resourceFactory.getResource(FedoraId.create(fullId)).getDescription();
+                fedoraResource.getTriples().forEach(triple -> {
+                    if (triple.getPredicate().getURI().equals("http://www.w3.org/1999/02/22-rdf-syntax-ns#type")) {
+                        final var rdfTypeUri = triple.getObject().getURI();
+                        rdfTypes.add(rdfTypeUri);
+                    }
+                });
+
+                final var jdbcInsertRdfTypes = new SimpleJdbcInsert(this.jdbcTemplate.getJdbcTemplate());
+                jdbcInsertRdfTypes.withTableName("search_rdf_type").usingGeneratedKeyColumns(
```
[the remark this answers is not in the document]
dbernstein: will do.

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -209,20 +261,73 @@ public void addUpdateIndex(final ResourceHeaders resourceHeaders) {
         final var txId = UUID.randomUUID().toString();
         executeInDbTransaction(txId, status -> {
             try {
+                final var rdfTypes = new ArrayList<String>();
+                rdfTypes.add(resourceHeaders.getInteractionModel());
+                final var fedoraResource = resourceFactory.getResource(FedoraId.create(fullId)).getDescription();
+                fedoraResource.getTriples().forEach(triple -> {
```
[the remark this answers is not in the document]
dbernstein: I didn't realize the user types are provided as well - looks like getTypes() is a better choice.

fcrepo-search-impl/src/main/resources/sql/default-search-index.sql
pwinckles: You need to use `CASCADE` when you truncate a table with foreign keys.

dbernstein: I will try that - thanks for the tip.

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -231,6 +317,46 @@ public void addUpdateIndex(final ResourceHeaders resourceHeaders) {
...
+        final var deleteParams = new MapSqlParameterSource();
+        deleteParams.addValue(RESOURCE_ID_PARAM, resourceId);
+        jdbcTemplate.update(DELETE_RDF_TYPE_ASSOCIATIONS,
+                deleteParams);
+    }
+
+    private ArrayList<Long> findOrCreateRdfTypesInDb(final List<URI> rdfTypes) {
+        final var jdbcInsertRdfTypes = new SimpleJdbcInsert(this.jdbcTemplate.getJdbcTemplate());
+        jdbcInsertRdfTypes.withTableName(SEARCH_RDF_TYPE_TABLE).usingGeneratedKeyColumns(
+                "id");
```
pwinckles, line 343: The String `id` is used in a number of places and could be a constant for consistency

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -252,7 +378,51 @@ public void removeFromIndex(final FedoraId fedoraId) {
...
+            statement.executeBatch();
+        } catch (SQLException throwables) {
+            throwables.printStackTrace();
+        }
+    }
+
+    private List<String> toggleForeignKeyChecks(final boolean enable) {
+
+        if (isPostgres()) {
+            final var statements = new ArrayList<String>();
```
pwinckles, line 404: Minor note, only change if you're inclined, but this could be rewritten as:

```java
return List.of(
    togglePostgresTriggers(SEARCH_RESOURCE_RDF_TYPE_TABLE, enable),
    togglePostgresTriggers(SEARCH_RDF_TYPE_TABLE, enable),
    togglePostgresTriggers(SIMPLE_SEARCH_TABLE, enable),
);
```

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -252,7 +378,51 @@ public void removeFromIndex(final FedoraId fedoraId) {
...
+            statements.add(togglePostgresTriggers(SEARCH_RESOURCE_RDF_TYPE_TABLE, enable));
+            statements.add(togglePostgresTriggers(SEARCH_RDF_TYPE_TABLE, enable));
+            statements.add(togglePostgresTriggers(SIMPLE_SEARCH_TABLE, enable));
+            return statements;
+        } else {
+            return Arrays.asList("SET FOREIGN_KEY_CHECKS = " + (enable ? 1 : 0) + ";");
```
pwinckles, line 410: Similarly, using `Arrays.asList()` is worse than `List.of()` in just about every case, except if you're converting an array into a list, and even then `Arrays.asList()` can have unexpected behavior if you don't treat the resulting list as immutable.

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -252,7 +378,51 @@ public void removeFromIndex(final FedoraId fedoraId) {
...
+        }
+    }
+
+    private boolean isPostgres() {
+        return getDdlKey().equals(POSTGRESQL_KEY);
```
pwinckles, line 415: You should store this value as a field in the `setup()` method rather than looking it up every time.

fcrepo-search-impl/src/main/resources/sql/postgresql-search-index.sql
```diff
@@ -0,0 +1,24 @@
...
+    id bigserial PRIMARY KEY,
+    rdf_type_uri varchar(255) NOT NULL UNIQUE
+);
+
+CREATE TABLE IF NOT EXISTS search_resource_rdf_type (
+    resource_id bigint NOT NULL,
+    rdf_type_id bigint NOT NULL,
+    PRIMARY KEY(resource_id, rdf_type_id),
+    FOREIGN KEY (resource_id) REFERENCES simple_search(id) ON DELETE CASCADE,
+    FOREIGN KEY (rdf_type_id) REFERENCES search_rdf_type(id)
```
pwinckles, line 21: In the other ddl, this line cascades too. Is there a reason it doesn't here?

fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACAuthorizingRealm.java
```diff
@@ -234,17 +233,22 @@ public boolean supports(final AuthenticationToken token) {
...
 
     private FedoraResource getResourceOrParentFromPath(final FedoraId fedoraId) {
+        final var tx = transaction();
+        final var txId = tx == null || tx.isCommitted() ? null : tx.getId();
+
         try {
             log.debug("Testing FedoraResource for {}", fedoraId.getFullIdPath());
-            return this.resourceFactory.getResource(transaction(), fedoraId);
+
+            return txId != null ? this.resourceFactory.getResource(txId, fedoraId) :
```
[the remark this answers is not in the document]
dbernstein: Agreed.  I will fix.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
```diff
@@ -116,4 +118,8 @@ protected Transaction transaction() {
         }
         return transaction;
     }
+
+    protected String txtIdIfUncommittedOrNull(final Transaction tx) {
```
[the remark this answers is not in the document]
dbernstein: Yes - excellent idea. Will do.

fcrepo-search-impl/src/main/resources/sql/postgresql-search-index.sql
dbernstein: @pwinckles :  Just a question:  when you have ON DELETE CASCADE on a foreign key,  will the referenced record only be deleted when there are no more references in the link table?  Otherwise, I'm wondering if having  ```FOREIGN KEY (rdf_type_id) REFERENCES search_rdf_type(id) ON DELETE CASCADE ``` might cause an error.

pwinckles: @dbernstein That's an excellent question. I'm not certain what will happen.

I'm not sure if these tables should be defined with or without the cascade. I was mostly just drawing attention to the discrepancy in the definitions.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
```diff
@@ -71,28 +71,17 @@ protected HttpIdentifierConverter identifierConverter() {
...
-                return getFedoraResource(fedoraId);
-            } else {
-                return resourceFactory.getResource(transaction, fedoraId);
-            }
-        } catch (final PathNotFoundException e) {
-            throw new PathNotFoundRuntimeException(e);
-        }
+    protected FedoraResource getFedoraResource(final Transaction transaction, final FedoraId fedoraId)
+            throws PathNotFoundException {
+            return resourceFactory.getResource(TransactionUtils.openTxId(transaction), fedoraId);
```
bbpennel, line 84: It seems like every call to `ResourceFactory.getResource()` outside of `org.fcrepo.kernel.impl.models` goes through `TransactionUtils.openTxId(transaction)` or an equivalent implementation. 

Would it make sense to add back in a `ResourceFactory.getResource(Transaction tx, FedoraId resourceId)` signature which does `TransactionUtils.openTxId` internally and calls the `getResource(String txId, FedoraId resourceId)` rather than calling `TransactionUtils.openTxId` everywhere?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
```diff
@@ -238,8 +238,6 @@ private Statement createDefaultAccessToStatement(final String authSubject) {
                         createResource(currentResourcePath));
     }
 
-    protected String txId(final Transaction tx) {
-        return tx == null ? null : tx.getId();
-    }
-
-}
+    protected String txtIdIfUncommittedOrNull(final Transaction tx) {
```
bbpennel, line 241: should this be `txIdIfUncommittedOrNull` rather than txtId?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
dbernstein: Yes if not for the fact that IndexBuilderImpl does not have access to a Transaction but rather interacts directly with the  fcrepo-persistence-ocfl layer directly.  Consequently DbSearchIndexImpl doesn't want to deal in Transaction Objects, but does need to know about the transaction in which the rebuild is taking place.  @pwinckles  was in favor of removing the dependency of the Transaction from the ResourceFactory API since it encumbers the interface without providing a commensurate benefit.

dbernstein: @bbpennel ^^

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
dbernstein: This should be removed I believe.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
dbernstein: Oh - I think I misunderstood you.

dbernstein: You're saying simply add the other signature and use where appropriate yes?

bbpennel: Right, there'd be multiple `getResource` signatures, one would take a Transaction, another would take a txId. Other parts of the application would call them where appropriate and then getResource would internally do the null and committed check, since that seems to be a requirement everywhere that a transaction object is used. `doesResourceExist` could do the same thing if that's valuable, but I didn't investigate that one closely. The benefit would be that a cleaner usage and we'd be less likely to forget that when using a transaction we have to do those checks on it since it'd be centralized.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
bbpennel: I thought I saw it being referenced a few places, but if the extra `getResource(Transaction tx...)` is added then this should be unneeded

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
dbernstein: That makes sense to me

dbernstein:
@pwinckles :  I'll take a look at the h2 and postgresql issues.

fcrepo-auth-webac/src/test/java/org/fcrepo/auth/webac/WebACFilterTest.java
```diff
@@ -174,10 +174,11 @@ public void setupRequest() throws Exception {
 
         when(mockTransactionManager.get(transactionId)).thenReturn(mockTransaction);
 
-        when(mockResourceFactory.getResource(mockTransaction, testChildId))
+        when(mockResourceFactory.getResource(mockTransaction.getId(), testChildId))
```
bbpennel, line 177: Seems like this and all the other instances modified in this class could be reverted back to just passing in the `mockTransaction`

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
```diff
@@ -101,7 +90,7 @@ protected FedoraResource getFedoraResource(final Transaction transaction, final
      * @return Returns true if an object with the provided id exists
      */
     protected boolean doesResourceExist(final Transaction transaction, final FedoraId fedoraId) {
-        return resourceFactory.doesResourceExist(transaction, fedoraId);
+        return resourceFactory.doesResourceExist(TransactionUtils.openTxId(transaction), fedoraId);
```
bbpennel, line 93: Would it make sense to also add back in the `doesResourceExist(Transaction tx, ...)` signature, which internally calls `doesResourceExist(String txId, ...)`?

fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
```diff
@@ -1310,15 +1312,15 @@ public void testCreateNewBinaryWithChecksumSHAandMD5() throws MalformedRdfExcept
     @Test(expected = CannotCreateResourceException.class)
     public void testLDPRNotImplemented() throws Exception {
         final var resource = setResource(Container.class);
-        when(resourceFactory.getResource(mockTransaction, pathId.resolve("x"))).thenReturn(resource);
+        when(resourceFactory.getResource(mockTransaction.getId(), pathId.resolve("x"))).thenReturn(resource);
```
bbpennel, line 1315: Seems like this and the other `getResource(mockTransaction.getId()` calls can be reverted to just `getResource(mockTransaction` if that is more accurate to the real usage

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/models/ResourceFactoryImplTest.java
```diff
@@ -208,7 +208,7 @@ public void getResource_BasicContainer_WithParent() throws Exception {
 
         when(psSession.getHeaders(parentId, null)).thenReturn(parentHeaders);
 
-        final var resc = factory.getResource(mockTx, fedoraId);
+        final var resc = factory.getResource(mockTx.getId(), fedoraId);
```
bbpennel, line 211: Seems like all the `factory.getResource(mockTx.getId(), ...` in this test could be reverted back to `factory.getResource(mockTx, ...`

bbpennel: I don't think these ones got reverted

dbernstein: They should be good now.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/IndexBuilderImpl.java
```diff
@@ -181,6 +182,14 @@ private void indexOcflObject(final String ocflId, final String txId, final OcflO
                 fedoraToOcflObjectIndex.addMapping(txId, fedoraIdentifier, rootFedoraIdentifier, ocflId);
                 LOGGER.debug("Rebuilt fedora-to-ocfl object index entry for {}", fedoraIdentifier);
             });
+
+            headersList.forEach(headers -> {
+                if (!headers.isDeleted()) {
+                    searchIndex.addUpdateIndex(txId, headers);
```
pwinckles, line 188: Up around L117 you need to clear the search index if there's an exception during the rebuild.

pwinckles (approved):

PR #1729, opened 2020-07-22
correct class casing

pwinckles:
**JIRA Ticket**: none

Renames all of the classes in `fcrepo-persistence-ocfl` to conform to common Java naming conventions.

# How should this be tested?

No functional changes.

# Interested parties
@fcrepo4/committers


pwinckles:
I'll take a look at your list tomorrow.

Personally, I think those Jena classes are wrong, but I also don't necessarily want to start a religious war.

I tend to agree with this style guide, and I do think it is the majority, but not exclusive, practice.

https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

pwinckles:
@awoods I addressed most of the additional renames you listed. Some of them I didn't do because the code is going to be deleted in an imminent PR.

PR #1730, opened 2020-07-23
Refactor to more broadly use FedoraId instead of a string

pwinckles:
**JIRA Ticket**: none

Refactors the code to more broadly use FedoraId rather than a string to represent an identifier.

There are a handful of methods that I intentionally did not update `OcflPersistentStorageUtils` because they are removed in the next PR.

# How should this be tested?

No functionality changes. Tests should pass, and everything should continue to work the same as before.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/WebacAclServiceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/WebacAclServiceImpl.java
@@ -77,7 +77,7 @@ public class WebacAclServiceImpl extends AbstractService implements WebacAclServ
 
         final RdfSourceOperation createOp = rdfSourceOperationFactory
                 .createBuilder(fedoraId, FEDORA_WEBAC_ACL_URI)
-                .parentId(FedoraId.create(fedoraId.getBaseId()))
+                .parentId(fedoraId.asBaseId())
                 .triples(stream)
                 .relaxedProperties(model)
                 .userPrincipal(userPrincipal)
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/WebacAclServiceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/WebacAclServiceImpl.java
@@ -76,8 +76,8 @@ public class WebacAclServiceImpl extends AbstractService implements WebacAclServ
         final RdfStream stream = fromModel(model.getResource(fedoraId.getFullId()).asNode(), model);
 
         final RdfSourceOperation createOp = rdfSourceOperationFactory
-                .createBuilder(fedoraId.getResourceId(), FEDORA_WEBAC_ACL_URI)
-                .parentId(fedoraId.getBaseId())
+                .createBuilder(fedoraId, FEDORA_WEBAC_ACL_URI)
+                .parentId(FedoraId.create(fedoraId.getBaseId()))
                 .triples(stream)
                 .relaxedProperties(model)
                 .userPrincipal(userPrincipal)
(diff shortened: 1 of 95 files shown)


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OcflPersistentStorageUtils.java
```diff
@@ -180,9 +181,9 @@ public static String resolveExtensions(final String subPath, final boolean isRdf
      * @param fedoraIdentifier The fedora identifier
      * @return The resolved topic
      */
-    public static  String resolveTopic(final String fedoraIdentifier) {
-        if (fedoraIdentifier.endsWith(FEDORA_METADATA_SUFFIX)) {
-            return fedoraIdentifier.substring(0, fedoraIdentifier.indexOf(FEDORA_METADATA_SUFFIX));
+    public static FedoraId resolveTopic(final FedoraId fedoraIdentifier) {
+        if (fedoraIdentifier.isDescription()) {
+            return FedoraId.create(fedoraIdentifier.getBaseId());
```
awoods, line 186: `asBaseId()`?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OcflPersistentStorageUtils.java
```diff
@@ -180,9 +181,9 @@ public static String resolveExtensions(final String subPath, final boolean isRdf
      * @param fedoraIdentifier The fedora identifier
      * @return The resolved topic
      */
-    public static  String resolveTopic(final String fedoraIdentifier) {
-        if (fedoraIdentifier.endsWith(FEDORA_METADATA_SUFFIX)) {
-            return fedoraIdentifier.substring(0, fedoraIdentifier.indexOf(FEDORA_METADATA_SUFFIX));
+    public static FedoraId resolveTopic(final FedoraId fedoraIdentifier) {
```
awoods, line 184: It would seem that this method (which can be `private`) could be replaced by `FedoraId.asBaseId()`?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/WebacAclServiceImpl.java
```diff
@@ -76,8 +76,8 @@ public void create(final Transaction transaction, final FedoraId fedoraId, final
         final RdfStream stream = fromModel(model.getResource(fedoraId.getFullId()).asNode(), model);
 
         final RdfSourceOperation createOp = rdfSourceOperationFactory
-                .createBuilder(fedoraId.getResourceId(), FEDORA_WEBAC_ACL_URI)
-                .parentId(fedoraId.getBaseId())
+                .createBuilder(fedoraId, FEDORA_WEBAC_ACL_URI)
+                .parentId(FedoraId.create(fedoraId.getBaseId()))
```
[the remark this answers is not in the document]
pwinckles: Yes, I added the `asBaseId` method towards the end of the refactor and missed some of earlier changes. Will fix.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OcflPersistentStorageUtils.java
pwinckles: > 
> 
> Looks good. Some minor comments.
> As a note, it is becoming slightly confusing when to use which method of `FedoraId`. Can we work towards removing some of the "getters" that return Strings?

I don't disagree! Because the id is used differently in different contexts it is confusing to know which method you want or even what the current "state" of the id is. I attempted to increase the clarity in this recent PR: https://github.com/fcrepo4/fcrepo4/pull/1719, but it's still not perfect.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OcflPersistentStorageUtils.java
pwinckles: I left this one because I actually fixed it in my next PR. Let me know if you want it fixed in this PR, and I'll do some git gymnastics to get it fixed here. Otherwise, the next PR will take care of it.

PR #1734, opened 2020-08-06
allow integration tests to be run against any db

pwinckles:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3382

# What does this Pull Request do?

Enables the integration tests to be run against any db.

# How should this be tested?

First, build the entire project, and then execute the integration tests against specific dbs.

## PostgreSQL

```
docker run --name f6-postgres -e POSTGRES_USER=fcrepo-user -e POSTGRES_PASSWORD=fcrepo-pw -e POSTGRES_DB=fcrepo -p 5432:5432 -d postgres:12.3

mvn -Dfcrepo.db.url=jdbc:postgresql://localhost:5432/fcrepo -Dfcrepo.db.user=fcrepo-user -Dfcrepo.db.password=fcrepo-pw clean install -pl fcrepo-http-api

docker stop f6-postgres
docker rm f6-postgres
```

## MariaDB

```
docker run --name f6-mariadb -e MYSQL_ROOT_PASSWORD=root-pw -e MYSQL_DATABASE=fcrepo -e MYSQL_USER=fcrepo-user -e MYSQL_PASSWORD=fcrepo-pw -p 3306:3306 -d mariadb:10.5.3

mvn -Dfcrepo.db.url=jdbc:mariadb://localhost:3306/fcrepo -Dfcrepo.db.user=fcrepo-user -Dfcrepo.db.password=fcrepo-pw clean install -pl fcrepo-http-api

docker stop f6-mariadb
docker rm f6-mariadb
```

## MySQL

```
docker run --name f6-mysql -e MYSQL_ROOT_PASSWORD=root-pw -e MYSQL_DATABASE=fcrepo -e MYSQL_USER=fcrepo-user -e MYSQL_PASSWORD=fcrepo-pw -p 3306:3306 -d mysql:8.0

mvn -Dfcrepo.db.url=jdbc:mysql://localhost:3306/fcrepo -Dfcrepo.db.user=fcrepo-user -Dfcrepo.db.password=fcrepo-pw clean install -pl fcrepo-http-api

docker stop f6-mysql
docker rm f6-mysql
```

# Interested parties
@fcrepo4/committers

PR #1740, opened 2020-08-10
Run ITests against all DBs

pwinckles:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3382

Previous work: https://github.com/fcrepo4/fcrepo4/pull/1734

Adds 3 new Travis jobs to run all of the ITests against PostgreSQL, MaraiDB, and MySQL. This code is based on experimenting I did here: https://github.com/pwinckles/travis-db-test

# How should this be tested?

The Travis build should pass.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,57 @@
 
 jobs:
   include:
-    - os: linux
-      language: java
+    - name: Build - Linux
+      os: linux
       dist: trusty
       sudo: true
+      language: java
       jdk: openjdk11
-    - os: windows
+      env: DEPLOY=true
+    - name: Build - Windows
+      os: windows
       language: shell
+    - name: ITests - PostgreSQL
+      os: linux
+      dist: trusty
+      sudo: true
+      language: java
+      jdk: openjdk11
+      before_script:
+        - psql -c 'CREATE DATABASE fcrepo;' -U postgres
+      script:
+        - ./mvnw -DjdbcUrl="jdbc:postgresql://localhost:5432/fcrepo" -DdbUser="postgres" -DdbPassword="" clean install -P db-tests
+    - name: ITests - MariaDB
+      os: linux
+      dist: trusty
+      sudo: true
+      language: java
+      jdk: openjdk11
+      addons:
+        mariadb: "10.4"
+      before_script:
+        - mysql -e 'CREATE DATABASE IF NOT EXISTS fcrepo;'
+      script:
+        - ./mvnw -DjdbcUrl="jdbc:mariadb://localhost:3306/fcrepo" -DdbUser="travis" -DdbPassword="" clean install -P db-tests
+    - name: ITests - MySQL
+      os: linux
+      dist: xenial
+      sudo: true
+      language: java
+      jdk: openjdk11
+      services:
+        - mysql
+      before_script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,13 +1,69 @@
...
+      language: java
+      jdk: openjdk11
+      env:
+        - DB=mysql
+        - DB_PORT=3306
+        - ITEST=true
+      services:
+        - mysql
+      before_script:
+        - wget https://repo.mysql.com//mysql-apt-config_0.8.10-1_all.deb
```
bbpennel, line 59: doesn't seem to make a difference here, but might be a good idea to remove the double slash

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersioningIT.java
```diff
@@ -853,8 +853,8 @@ private void assertMementoLink(final Link expected, final Link actual) {
 
             final var expectedUri = expected.getUri().toString();
             final var actualUri = actual.getUri().toString();
-            assertEquals(expectedUri.substring(0, expectedUri.length() - 2),
-                    actualUri.substring(0, actualUri.length() - 2));
+            assertEquals(expectedUri.substring(0, expectedUri.length() - 3),
```
bbpennel, line 856: If i'm understanding correctly, this change basically increases wiggle room allowed between the expected timestamp and the real timestamp up to the minutes range? Sort of wondering if it can tell apart the 4 mementos created in that test since they are separated by 1 second each, but it seems like that already would have been an issue since it seems like it was trimming off the seconds from the timestamp.

I'm actually kind of surprised this test is passing since the logs for it are cranking out a lot of grizzly `Method Not Allowed` errors. But I can't really remember where versioning support stands in fedora 6 at the moment anyway.

pwinckles: Yes, it increased the wiggle room. It's not as good as it could be if the waits were longer, but I personally don't think it's worth inserting longer waits into the tests.

The comparison behavior works like this:

1. Is the timestamp exactly what we expected?
   1. If yes, we're done -- PASS
   2. If no, continue
2. Are the `Instants` within 5 seconds of each other?
   1. If yes, continue
   2. If no, we're done -- FAIL
3. Are the first parts of the memento strings the same (this is the part I just modified)?
  1. If yes, we're done -- PASS
   2. If no, we're done -- FAIL

So, in the ideal case, the strings will match exactly. But, if they don't match, it checks to see that they're at least close and that they're formatted correctly.

.travis.yml
pwinckles: fixed

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraVersioningIT.java
bbpennel: Thanks for the explanation, I hadn't followed it all the way through. That seems moderately safe, although there's obviously still some chance of false positive, but that seems outside the scope of this PR.

pwinckles:
Ack. That `testConcurrentPutsWithPairtrees` test is still having problems. :(

PR #1742, opened 2020-08-20
Adds tombstone specific exception message for 404 return exception

robyj:
- Checks if exception message contains /fcr:tombstone and returns custom excption message as per the suggestion in the ticket

Resolves: https://jira.lyrasis.org/projects/FCREPO/issues/FCREPO-3261

https://jira.lyrasis.org/projects/FCREPO/issues/FCREPO-3261

The ticket requests that if a user requests an item and the item does not exist anymore (due to deletion, etc) AND the item left a tombstone in its place, the message returned via the exception handler should be different than and item that has been removed but has no tombstone

Ideally, it should be tested by creating an item, deleting it in such a way to generate a tombstone then retrieving the deleted item as well as creating an item, deleting it in a way to not generate a tombstone and retrieving the item (to check the non-tombstone return message)

I'm sorry if the code doesn't work as intended or if the tests aren't as accurate as you'd like. I'm not sure how to create/delete/retrieve in a JUnit test.

@awoods

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/ItemNotFoundExceptionMapper.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/ItemNotFoundExceptionMapper.java
@@ -25,7 +25,6 @@ import javax.ws.rs.ext.Provider;
 import org.fcrepo.kernel.api.exception.ItemNotFoundException;
 import org.slf4j.Logger;
 
-
 /**
  * Catch ItemNotFoundException
  *
@@ -43,8 +42,19 @@ public class ItemNotFoundExceptionMapper implements
 
         LOGGER.debug("Exception intercepted by ItemNotFoundExceptionMapper: {}\n", e.getMessage());
         debugException(this, e, LOGGER);
-        return Response.status(Response.Status.NOT_FOUND).
-                entity("Error: " + e.getMessage()).build();
+        String tmp = e.getMessage();
+        LOGGER.debug("tmp is {}\n", tmp);
+        final int idx = tmp.lastIndexOf("/");
+        if (idx != -1) {
+            tmp = tmp.substring(idx);
+        }
+        if (tmp.contains("fcr:tombstone")) {
+            return Response.status(Response.Status.NOT_FOUND).
+                            entity("Discovered tombstone resource at " + tmp).build();
+        } else {
+            return Response.status(Response.Status.NOT_FOUND).
+                            entity("Resource at " + tmp + " not found").build();
+        }
     }
 }
 
--- a/fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/ItemNotFoundExceptionMapperTest.java
+++ b/fcrepo-http-commons/src/test/java/org/fcrepo/http/commons/exceptionhandlers/ItemNotFoundExceptionMapperTest.java
@@ -34,10 +34,13 @@ import org.junit.Test;
 public class ItemNotFoundExceptionMapperTest {
 
     private ItemNotFoundExceptionMapper testObj;
+    private ItemNotFoundExceptionMapper testObj2;
 
     @Before
     public void setUp() {
+
         testObj = new ItemNotFoundExceptionMapper();
+        testObj2 = new ItemNotFoundExceptionMapper();
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


pwinckles (commented):
I don't think these changes will address the problem.

1. Tombstones are already working. Try the following:

```
curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/binary -H "Link: <http://www.w3.org/ns/ldp#NonRDFSource>; rel=type" -v -H "Content-Type: text/plain" -X PUT --data-binary "testing"
curl -u fedoraAdmin:fedoraAdmin -XDELETE http://localhost:8080/rest/binary
curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/binary
```

2. I think the display problem for resources that do not exists is related to the exception construction here: https://github.com/fcrepo4/fcrepo4/blob/f7c9ede08535d14f97da6a36018b6a47ef760bd0/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java#L216

It should actually be:

```java
throw new PathNotFoundException(e.getMessage(), e);
```

whikloj:
I think this is caused because there is no `ExceptionMapper` that handles a `PersistentItemNotFoundException`. 

That exception is from the persistence layer and at the very least should be wrapped in a `ItemNotFoundException` or `PathNotFoundException`

Though I'm not sure why we need both of those exceptions either. But anyways, this is a matter of finding where the `PersistentItemNotFoundException` is getting back to the client and alter it on the way.

pwinckles:
> 
> 
> I think this is caused because there is no `ExceptionMapper` that handles a `PersistentItemNotFoundException`.
> 
> That exception is from the persistence layer and at the very least should be wrapped in a `ItemNotFoundException` or `PathNotFoundException`
> 
> Though I'm not sure why we need both of those exceptions either. But anyways, this is a matter of finding where the `PersistentItemNotFoundException` is getting back to the client and alter it on the way.

This is what I had originally thought as well, but I don't think it's actually the case, at least not for this specific example. If you notice response is prefixed with "Error: ", which means that the exception _was_ actually handled by either `ItemNotFoundExceptionMapper` or `PathNotFoundExceptionMapper` but the exceptions message was not formatted correctly, which is the case for the exception thrown here (https://github.com/fcrepo4/fcrepo4/blob/f7c9ede08535d14f97da6a36018b6a47ef760bd0/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java#L216).

robyj:
Good morning to the both of you

Thank you for giving me some ideas of where to look. I'll use those 3 curl calls and look around ResourceFactoryImpl to see if thats where the issue is

Thank you for your expertise
Jon

robyj:
The exception is line 218 in ResourceFactoryImpl and its a PathNotFoundException

So, do I add the PersistentItemNotFoundException handler, or change the PathNotFoundException error message?

pwinckles:
> 
> 
> The exception is line 218 in ResourceFactoryImpl and its a PathNotFoundException
> 
> So, do I add the PersistentItemNotFoundException handler, or change the PathNotFoundException error message?

Try changing that line to:

```java
throw new PathNotFoundException(e.getMessage(), e);
```

robyj:
After that change, the error message is: 
Error: org.fcrepo.persistence.api.exceptions.PersistentItemNotFoundException: No OCFL mapping found for info:fedora/binary2

pwinckles:
> 
> 
> After that change, the error message is:
> Error: org.fcrepo.persistence.api.exceptions.PersistentItemNotFoundException: No OCFL mapping found for info:fedora/binary2

Hmm... What does your code look like? I just tried it and I get:

```
Error: No OCFL mapping found for info:fedora/binary
```

Which is probably not the preferable text, but at least it doesn't have the exception class in it.

robyj:
Its:

throw new PathNotFoundException(w.getMessage(),e);

and PathNotFoundExceptionMapper looks like:

@Override
public Response toReponse(final PathNotFoundException e) {
  LOGGER.debug("Exception intercepted by PathNotFoundExceptionMapper: {}\n", e.getMessage());
  debugException(this, e, LOGGER);
  return Response.status(Response.Status.NOT_FOUND).
             entity("Error: " + e.getMessage()).build();
  }

pwinckles:
> 
> 
> Its:
> 
> throw new PathNotFoundException(w.getMessage(),e);
> 
> and PathNotFoundExceptionMapper looks like:
> 
> @override
> public Response toReponse(final PathNotFoundException e) {
> LOGGER.debug("Exception intercepted by PathNotFoundExceptionMapper: {}\n", e.getMessage());
> debugException(this, e, LOGGER);
> return Response.status(Response.Status.NOT_FOUND).
> entity("Error: " + e.getMessage()).build();
> }

So, did you add a constructor to `PathNotFoundException` like this?

```java
    public PathNotFoundException(final String message, final Throwable exception) {
        super(message, exception);
    }
```

In your example, what is `w`?

robyj:
doh, sorry, its e, not w. so its e.getMessage() (im doing development on a VM)

No, sorry, didnt add the new constructor, i will try that

Thank you
Jon

pwinckles:
The code should not have compiled without that constructor. Make sure you rebuild (`mvn install`) `fcrepo-kernel-impl` before launching Fedora

robyj:
Well, that took forever. But i'm getting the same error message as you now

PR #1746, opened 2020-09-06
Rewrite PATCH request repository URIs to internal IDs

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3311

We use the HttpRdfService on PUT/POST requests to ensure RDF has URIs for items in the repository converted to internal IDs. But for Sparql we didn't do this. So while the RDF appears correct the repository doesn't realize it is referencing an item inside the repository.

This adds an "UpdateVisitor" to generate a new Sparql request based on the old one but with any in domain URIs converted to internal IDs

# How should this be tested?

I'm not sure if this is testable except by checking the actual RDF stored in the OCFL objects.

1. Create a RDFSource object.
     `curl -i -ufedoraAdmin:fedoraAdmin -XPOST http://localhost:8080/rest `
1. Patch the resource with some internal IDs.
     `curl -ufedoraAdmin:fedoraAdmin -i -XPATCH -d"INSERT DATA { <http://localhost:8080/rest/resource1> <http://example.org/some/predicate> <http://localhost:8080/rest/someOtherResource> }" -H"Content-type: application/sparql-update" (the URL created above)`
1. Locate and check the `fcr-container.nt` file in the OCFL storage root. It should be in the `v2` directory.

Before the PR it will look like
```
<http://localhost:8080/rest/resource1> <http://example.org/some/predicate> <http://localhost:8080/rest/someOtherResource> .
```
and after it should look like 
```
<info:fedora/resource1> <http://example.org/some/predicate> <info:fedora/someOtherResource> .
```

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -513,7 +513,9 @@ public class FedoraLdp extends ContentExposingResource {
             evaluateRequestPreconditions(request, servletResponse, resource(), transaction);
 
             LOGGER.info("PATCH for '{}'", externalPath);
-            patchResourcewithSparql(resource(), requestBody);
+            final String newRequest = httpRdfService.patchRequestToInternalString(resource().getFedoraId().getFullId(),
+                    requestBody, identifierConverter());
+            patchResourcewithSparql(resource(), newRequest);
             transaction.commitIfShortLived();
 
             addCacheControlHeaders(servletResponse, resource(), transaction);
@@ -627,7 +629,7 @@ public class FedoraLdp extends ContentExposingResource {
         try {
             final var resource = getFedoraResource(transaction, newFedoraId);
             return createUpdateResponse(resource, true);
-        } catch (PathNotFoundException e) {
+        } catch (final PathNotFoundException e) {
             throw new PathNotFoundRuntimeException(e);
         }
     }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -515,6 +515,7 @@ public class FedoraLdp extends ContentExposingResource {
             LOGGER.info("PATCH for '{}'", externalPath);
             final String newRequest = httpRdfService.patchRequestToInternalString(resource().getFedoraId().getFullId(),
                     requestBody, identifierConverter());
+            LOGGER.debug("PATCH request translated to '{}'", newRequest);
             patchResourcewithSparql(resource(), newRequest);
             transaction.commitIfShortLived();
 
@@ -630,7 +631,7 @@ public class FedoraLdp extends ContentExposingResource {
             final var resource = getFedoraResource(transaction, newFedoraId);
             return createUpdateResponse(resource, true);
         } catch (final PathNotFoundException e) {
-            throw new PathNotFoundRuntimeException(e);
+            throw new PathNotFoundRuntimeException(e.getMessage(), e);
         }
     }
 
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
@@ -41,6 +41,9 @@ import org.apache.jena.rdf.model.StmtIterator;
 import org.apache.jena.rdf.model.RDFNode;
 import org.apache.jena.riot.Lang;
 import org.apache.jena.riot.RiotException;
+import org.apache.jena.update.Update;
+import org.apache.jena.update.UpdateFactory;
+import org.apache.jena.update.UpdateRequest;
 import org.fcrepo.http.commons.api.rdf.HttpIdentifierConverter;
 import org.fcrepo.kernel.api.exception.MalformedRdfException;
 import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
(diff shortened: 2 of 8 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/SparqlTranslateVisitor.java
```diff
@@ -0,0 +1,185 @@
...
+     * Quads insert/delete data statements don't contain variables and use QuadDataAcc to accumulate,
+     * insert {} delete {} where {} statements can't contain variables and use QuadAcc to accumulate. This function
+     * simplifies the creation of the eventual Update.
+     * @param updateClass the class of Update we are starting with.
+     * @param quadList the list of Quads to generate the above class with.
+     * @return a subclass of Update with the provided Quads.
+     */
+    private Update makeUpdate(final Class<? extends Update> updateClass, final List<Quad> quadList) {
+        try {
+            if (updateClass.equals(UpdateDeleteWhere.class)) {
```
pwinckles, line 154: In the code that you were working on last week, there was a case for `if (updateClass.equals(UpdateModify.class))` here as well. Is that not needed?

pwinckles (commented):
Just a few questions

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -627,7 +629,7 @@ public Response createObject(@HeaderParam(CONTENT_DISPOSITION) final ContentDisp
         try {
             final var resource = getFedoraResource(transaction, newFedoraId);
             return createUpdateResponse(resource, true);
-        } catch (PathNotFoundException e) {
+        } catch (final PathNotFoundException e) {
```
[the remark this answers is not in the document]
whikloj: Ok so I am adding a `PathNotFoundRuntimeExceptionMapper` and updating `PathNotFoundRuntimeException` to add the method
```
    /**
     * Wrap a PathNotFoundException in a runtime exception
     * @param message the original message.
     * @param rootCause the root cause.
     */
    public PathNotFoundRuntimeException(final String message, final Throwable rootCause) {
        super(message, rootCause);
    }
```

We should think about the use of the `ItemNotFoundException` and whether we need mappers for it and for `PathNotFoundException`. But I'm not doing that now.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/SparqlTranslateVisitor.java
```diff
@@ -0,0 +1,185 @@
...
+        final List<Quad> insertQuads = (update.hasInsertClause() ? translateQuads(update.getInsertQuads()) :
+                Collections.emptyList());
+        insertQuads.forEach(q -> newUpdate.getInsertAcc().addQuad(q));
+
+        final List<Quad> deleteQuads = (update.hasDeleteClause() ? translateQuads(update.getDeleteQuads()) :
+                Collections.emptyList());
+        deleteQuads.forEach(q -> newUpdate.getDeleteAcc().addQuad(q));
+
+        final Element where = update.getWherePattern();
+        newUpdate.setElement(where);
```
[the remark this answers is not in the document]
whikloj: Checking in to this.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/SparqlTranslateVisitor.java
whikloj: I don't think so because the above `private void translateUpdate(final UpdateModify update)` returns the correct `UpdateModify` object. This is used by the `private void translateUpdate(final Update update) {` one that returns one of 3 classes.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
```diff
@@ -161,6 +164,25 @@ public RdfStream bodyToInternalStream(final String extResourceId, final InputStr
...
 
+    /**
+     * Takes a PATCH request body and translates any subjects and objects that are in the domain of the repository
+     * to use internal IDs.
+     * @param extResourceId the internal ID of the current resource.
+     * @param requestBody the request body.
+     * @param idTranslator an ID converter for the current context.
+     * @return the converted string.
+     */
+    public String patchRequestToInternalString(final String extResourceId, final String requestBody,
```
awoods, line 175: The javadoc says "extResourceId" is the internal ID... what does the "ext" stand for?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
```diff
@@ -161,6 +164,25 @@ public RdfStream bodyToInternalStream(final String extResourceId, final InputStr
         return fromModel(model.getResource(idTranslator.toInternalId(extResourceId)).asNode(), model);
     }
 
+    /**
+     * Takes a PATCH request body and translates any subjects and objects that are in the domain of the repository
+     * to use internal IDs.
+     * @param extResourceId the internal ID of the current resource.
+     * @param requestBody the request body.
+     * @param idTranslator an ID converter for the current context.
+     * @return the converted string.
```
awoods, line 173: It may be more clear to change to: "the converted PATCH request"

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/SparqlTranslateVisitor.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    @Override
+    public void visit(final UpdateModify update) {
+        translateUpdate(update);
+    }
+
+    /**
+     * Get the new UpdateRequest based on the parsed Updates.
+     * @return the new updaterequest object.
```
awoods, line 88: minor typo: "updaterequest"

fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
```diff
@@ -1037,12 +1037,12 @@ public void testPatchBinaryDescription() throws Exception {
...
         when(resourceFactory.getResource(mockTransaction, binaryDescId))
                 .thenReturn(mockNonRdfSourceDescription);
-        testObj.updateSparql(toInputStream("xyz", UTF_8));
+        testObj.updateSparql(toInputStream("INSERT DATA { <> <http://some/predicate> \"xyz\" }", UTF_8));
         verify(updatePropertiesService).updateProperties(
                 eq(mockTransaction.getId()),
                 anyString(),
                 eq(binaryDescId),
-                eq("xyz")
+                anyString()
```
awoods, line 1045: This seems to be weakening the test. Can we check for some part of the string that should be expected?

awoods (changes requested):
Works great... some very minor comments.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
whikloj: I was stealing from the other functions here, but it appears that they might be mis-named as well.

awoods (approved):

PR #1748, opened 2020-09-09
Ensure that all Fedora modules are included in the default logback.xml

awoods:
..and finetune the logging levels across the codebase.

Resolves: https://jira.lyrasis.org/browse/FCREPO-3415

# How should this be tested?
- Start Fedora and see if you like the default logging
- Start Fedora with different logging levels (other than the default, INFO) for specific modules 
   - e.g. `-Dfcrepo.log.storage=DEBUG`

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
@@ -89,9 +89,7 @@ public class TransactionImpl implements Transaction {
             this.getEventAccumulator().emitEvents(id, baseUri, userAgent);
             this.committed = true;
         } catch (final PersistentStorageException ex) {
-            if (log.isDebugEnabled()) {
-                log.debug("Failed to commit transaction: {}\n{}", ex.getMessage(), ExceptionUtils.getStackTrace(ex));
-            }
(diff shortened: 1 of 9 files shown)


dbernstein (dismissed):
Looks good.

dbernstein:
@awoods : There's a failing check which appears to be a network issue. I restarted the build. If that passes I will merge.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
```diff
@@ -88,6 +89,10 @@ public synchronized void commit() {
             this.getEventAccumulator().emitEvents(id, baseUri, userAgent);
             this.committed = true;
         } catch (final PersistentStorageException ex) {
+            if (log.isDebugEnabled()) {
```
pwinckles, line 92: A few questions here:

1. Why are you custom formatting the log message rather than just doing `log.debug("Failed to commit transaction {}", id, ex);`?
2. Why debug? I would consider this an error. Unless it's innocuous and happens all the time, but, if that's the case, there's a different set of questions.

pwinckles (dismissed):
Thanks for improving the logging!

pwinckles (approved):

PR #1749, opened 2020-09-10
Inbound references service

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3166

Adds an API and implementation of a new service to parse incoming RDF for internal references and stores them in a new index, then allows this index to be queried.

# How should this be tested?

1. Create a resource.
1. Create another resource with any reference to the first ie. `<> <http://awoods.com/pointsTo> <http://localhost:8080/rest/firstResource>`
1. Get the first resource, see no inbound reference triples
1. Get the first resource with the header `Prefer: return=representation; include="http://fedora.info/definitions/fcrepo#PreferInboundReferences"`, see the inbound reference.

For binaries when requesting the inbound references on the `fcr:metadata` endpoint you will get both references to the actual binary URI and the `fcr:metadata` URI.

**Note**: I messed up and forgot to remove the references from the index on deletion, I am implementing this now and will make this PR live once it is added.

# Interested parties
@bbpennel  @fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -2857,6 +2857,58 @@ public class FedoraLdpIT extends AbstractResourceIT {
     }
 
     @Test
+    public void testReferenceRemovedOnDelete() throws Exception {
+        final String id = getRandomUniqueId();
+        final String resource = serverAddress + id;
+        final String resourcea = resource + "/a";
+        final String resourceb = resource + "/b";
+        final String resourcec = resource + "/c";
+
+        createObjectAndClose(id);
+        createObjectAndClose(id + "/a");
+        createObjectAndClose(id + "/b");
+        createObjectAndClose(id + "/c");
+        // Patch a to point to c.
+        final HttpPatch updateObjectGraphMethodA = patchObjMethod(id + "/a");
+        updateObjectGraphMethodA.addHeader(CONTENT_TYPE, "application/sparql-update");
+        updateObjectGraphMethodA.setEntity(new StringEntity("INSERT { <" +
+                resourcea + "> <http://purl.org/dc/terms/isPartOf> <" + resourcec + "> . } WHERE {}"));
+        assertEquals(NO_CONTENT.getStatusCode(), getStatus(updateObjectGraphMethodA));
+        // Patch b to point to c as well.
+        final HttpPatch updateObjectGraphMethodB = patchObjMethod(id + "/b");
+        updateObjectGraphMethodB.addHeader(CONTENT_TYPE, "application/sparql-update");
+        updateObjectGraphMethodB.setEntity(new StringEntity("INSERT { <" +
+                resourceb + "> <http://purl.org/dc/terms/isPartOf> <" + resourcec + "> . } WHERE {}"));
+        assertEquals(NO_CONTENT.getStatusCode(), getStatus(updateObjectGraphMethodB));
+
+        // Verify the inbound references.
+        final HttpGet getObjMethod = new HttpGet(resourcec);
+        getObjMethod.addHeader("Prefer", "return=representation; include=\"" + INBOUND_REFERENCES + "\"");
+        try (final CloseableDataset dataset = getDataset(getObjMethod)) {
+            final DatasetGraph graph = dataset.asDatasetGraph();
+            assertTrue(graph.contains(ANY,
+                    createURI(resourcea), createURI("http://purl.org/dc/terms/isPartOf"), createURI(resourcec)));
+
+            assertTrue(graph.contains(ANY,
+                    createURI(resourceb), createURI("http://purl.org/dc/terms/isPartOf"), createURI(resourcec)));
+        }
+        // Delete a.
+        final HttpDelete deleteReferrer = deleteObjMethod(id + "/a");
+        assertEquals(NO_CONTENT.getStatusCode(), getStatus(deleteReferrer));
+        // Verify only inbound reference from b
+        final HttpGet getObjMethod2 = new HttpGet(resourcec);
+        getObjMethod2.addHeader("Prefer", "return=representation; include=\"" + INBOUND_REFERENCES + "\"");
+        try (final CloseableDataset dataset = getDataset(getObjMethod2)) {
+            final DatasetGraph graph = dataset.asDatasetGraph();
+            assertFalse(graph.contains(ANY,
+                    createURI(resourcea), createURI("http://purl.org/dc/terms/isPartOf"), createURI(resourceb)));
+
+            assertTrue(graph.contains(ANY,
+                    createURI(resourceb), createURI("http://purl.org/dc/terms/isPartOf"), createURI(resourcec)));
+        }
+    }
+
+    @Test
 @Ignore
     public void testGetObjectReferencesIndirect() throws Exception {
         final String uuid = getRandomUniqueId();
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -2827,7 +2827,6 @@ public class FedoraLdpIT extends AbstractResourceIT {
     }
 
     @Test
-@Ignore
     public void testGetObjectReferences() throws IOException {
         final String id = getRandomUniqueId();
         final String resource = serverAddress + id;
@@ -2842,7 +2841,7 @@ public class FedoraLdpIT extends AbstractResourceIT {
         updateObjectGraphMethod.setEntity(new StringEntity("INSERT { <" +
                 resourcea + "> <http://purl.org/dc/terms/isPartOf> <" + resourceb + "> . \n <" +
                 resourcea + "> <info:xyz#some-other-property> <" + resourceb + "> } WHERE {}"));
-        executeAndClose(updateObjectGraphMethod);
+        assertEquals(NO_CONTENT.getStatusCode(), getStatus(updateObjectGraphMethod));
 
         final HttpGet getObjMethod = new HttpGet(resourceb);
 
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -4018,7 +4018,6 @@ public class FedoraLdpIT extends AbstractResourceIT {
     }
 
     @Test
-@Ignore
     public void testInboundLinksDoNotUpdateEtag() throws IOException {
         final String id1 = getRandomUniqueId();
         final HttpPut httpPut = putObjMethod(id1);
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -230,6 +230,9 @@ public class FedoraLdpIT extends AbstractResourceIT {
     private static final String NON_RDF_SOURCE_LINK_HEADER = "<" + NON_RDF_SOURCE.getURI() + ">; rel=\"type\"";
     private static final String VERSIONED_RESOURCE_LINK_HEADER = "<" + VERSIONED_RESOURCE.getURI() + ">; rel=\"type\"";
 
+    private static final String INBOUND_REFERENCE_PREFER_HEADER = "return=representation; include=\"" +
+            INBOUND_REFERENCES + "\"";
+
     private static final String TEST_ACTIVATION_PROPERTY = "RUN_TEST_CREATE_MANY";
 
     private static final String WANT_DIGEST = "Want-Digest";
@@ -2845,7 +2848,7 @@ public class FedoraLdpIT extends AbstractResourceIT {
 
         final HttpGet getObjMethod = new HttpGet(resourceb);
 
-        getObjMethod.addHeader("Prefer", "return=representation; include=\"" + INBOUND_REFERENCES + "\"");
+        getObjMethod.addHeader("Prefer", INBOUND_REFERENCE_PREFER_HEADER);
         try (final CloseableDataset dataset = getDataset(getObjMethod)) {
             final DatasetGraph graph = dataset.asDatasetGraph();
             assertTrue(graph.contains(ANY,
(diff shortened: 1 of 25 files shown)


escowles (dismissed):
👍 Great to see this PR come in. I can confirm that the Valkyrie test suite passes when run against this branch. Thanks @whikloj!

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/ReferenceServiceImplTest.java
```diff
@@ -0,0 +1,223 @@
...
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+/**
+ * Reference Service Tests
+ * @author whikloj
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration("/containmentIndexTest.xml")
+public class ReferenceServiceImplTest {
```
bbpennel, line 54: This looks good, there are a few extra cases that we should probably check:
* Verify that if you request references from a committed transaction (or transaction that doesn't exist), you get nothing
* Check that references deleted in a transaction are not visible within the transaction, but still are outside of it (can probably add this to `testAddAndRemove`)
* Committing without having called updateReferences

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/IndexBuilderImpl.java
```diff
@@ -148,6 +166,13 @@ private void indexOcflObject(final String ocflId, final String txId, final OcflO
                                 fedoraId.getFullId()));
                     }
                 }
+                if (!headers.getInteractionModel().equalsIgnoreCase(NON_RDF_SOURCE.toString())) {
```
bbpennel, line 169: Doesn't seem like we need a case insensitive comparison here, the interaction model uris can be case sensitive.

bbpennel (changes requested):
This looks good to me over all, just minor suggestions and a few tests

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/ReferenceServiceImplTest.java
```diff
@@ -0,0 +1,223 @@
...
+
+    private FedoraId subject2Id;
+
+    private Resource subject1;
+
+    private Resource subject2;
+
+    private Resource target;
+
+    private static Property referenceProp = ResourceFactory.createProperty("http://example.org/pointer");
```
awoods, line 72: Make `final`?

fcrepo-kernel-impl/src/test/resources/containmentIndexTest.xml
```diff
@@ -21,4 +21,9 @@
     <bean id="containmentIndex" class="org.fcrepo.kernel.impl.ContainmentIndexImpl">
         <property name="dataSource" ref="containmentIndexDataSource"/>
     </bean>
+
+    <!-- ReferenceService index to test -->
+    <bean id="referenceIndex" class="org.fcrepo.kernel.impl.services.ReferenceServiceImpl">
+        <property name="dataSource" ref="containmentIndexDataSource" />
```
awoods, line 27: You are using the `containmenIndexDataSource`??

fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/AbstractJmsIT.java
```diff
@@ -135,6 +140,12 @@
 
     protected abstract Destination createDestination() throws JMSException;
 
+    @Before
+    public void setUp() {
+        setField(createResourceService, "referenceService", referenceService);
```
awoods, line 145: It appears that setting the fields here is unnecessary. I believe the `setUp` method can be removed.

awoods (changes requested):

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/ReferenceServiceImplTest.java
whikloj: A little unsure I understand this one
> Verify that if you request references from a committed transaction ... you get nothing

You mean if there is nothing? Because the query in a transaction gets things from the general table and this specific to the current transaction.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImpl.java
```diff
@@ -180,6 +180,8 @@ public void perform(final String txId, final String userPrincipal, final FedoraI
 
         try {
             pSession.persist(createOp);
+            referenceService.updateReferences(txId, fedoraId,
+                    fromModel(model.getResource(fedoraId.getFullId()).asNode(), model));
```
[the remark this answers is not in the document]
whikloj: Not sure I understand the reasoning for that? You have a method that calls the other method. Seems to make it harder to understand what is going on. I'll do it, but if you could explain the benefits for my own sanity.

fcrepo-kernel-impl/src/test/resources/containmentIndexTest.xml
whikloj: It's just the bean name, instead of creating a second bean to connect to the database. I can change the name to be more generic.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImpl.java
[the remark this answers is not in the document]
whikloj: Oh I think I get what you're thinking here.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -2850,6 +2849,58 @@ public void testGetObjectReferences() throws IOException {
         }
     }
 
+    @Test
```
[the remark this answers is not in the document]
whikloj: Done

awoods (dismissed):

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/ReferenceServiceImplTest.java
bbpennel: So if you provide a transaction id to getInboundReferences, and there are no entries specific to that transaction, you would get back references from outside of transactions? Looks like it from the new test. That seems reasonable.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/ReferenceServiceImplTest.java
```diff
@@ -0,0 +1,256 @@
...
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+/**
+ * Reference Service Tests
+ * @author whikloj
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration("/containmentIndexTest.xml")
+public class ReferenceServiceImplTest {
```
bbpennel, line 54: One more test I'd suggest adding: that a reference to a binary is returned in a request to its description

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -2856,6 +2856,41 @@ public void testGetObjectReferences() throws IOException {
...
+        final HttpPost postBinary = postObjMethod();
+        postBinary.setHeader(CONTENT_TYPE, TEXT_PLAIN);
+        postBinary.setEntity(new StringEntity("Test text"));
+        final String binaryUri;
+        try (final CloseableHttpResponse response = execute(postBinary)) {
+            assertEquals(CREATED.getStatusCode(), getStatus(response));
+            binaryUri = getLocation(response);
+        }
+
+        final String linkRdf = "INSERT { <" + binaryUri + "> <http://awoods.com/pointsAt> <" + containerUri + "> . } " +
```
bbpennel, line 2881: Minor, but there is a helper method on the parent class of this test used for adding properties to resources which you could use here. `setProperty(binaryUri + "/" + FCR_METADATA, "http://awoods.com/pointsAt", "<" + containerUri + ">")`

bbpennel:
@whikloj I'm testing out commit d33b6152c339fe96d55fe4b65757b55ef4bb3ebd locally, but the uri for an inbound reference isn't coming back the way I was expecting. I did the following steps:
```
curl -u fedoraAdmin:fedoraAdmin -XPUT http://localhost:8080/rest/resc1 

curl -u fedoraAdmin:fedoraAdmin -XPUT http://localhost:8080/rest/bin1 -H'Link: <http://www.w3.org/ns/ldp#NonRDFSource>; rel="type"' --data-binary "@README.md"

echo 'INSERT { <> <http://awoods.com/pointsTo> <http://localhost:8080/rest/resc1> . } WHERE { }' | curl -u fedoraAdmin:fedoraAdmin -XPATCH http://localhost:8080/rest/bin1/fcr:metadata -H"Content-Type: application/sparql-update" --data-binary "@-"

curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/resc1 -H 'Prefer: return=representation; include="http://fedora.info/definitions/fcrepo#PreferInboundReferences";'
```
But the response from the last part is coming back as:
```
@prefix rdf:  <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix fedora:  <http://fedora.info/definitions/v4/repository#> .
@prefix ldp:  <http://www.w3.org/ns/ldp#> .

<http://localhost:8080/rest/resc1>
        fedora:created         "2020-09-21T21:19:34.848515Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> ;
        fedora:lastModified    "2020-09-21T21:19:34.848515Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> ;
        fedora:createdBy       "fedoraAdmin" ;
        fedora:lastModifiedBy  "fedoraAdmin" ;
        rdf:type               ldp:BasicContainer ;
        rdf:type               ldp:Resource ;
        rdf:type               fedora:Resource ;
        rdf:type               ldp:RDFSource ;
        rdf:type               fedora:Container .

<http://localhost:8080/rest/bin1/fcr:metadata>
        <http://awoods.com/pointsTo>  <http://localhost:8080/rest/resc1> .
```

From the test you added it sounds like the uri is supposed to be `http://localhost:8080/rest/bin1`. I rebuilt everything and started jetty, but am still getting that result. Is it behaving differently, or am I expecting the wrong outcome?

bbpennel:
Sorry to get a little edge casey, but what behavior would you expect if a resource references itself? Its sort of an inbound reference, but also sort of not. At the moment, it looks like you get a duplicate property:

```
curl -u fedoraAdmin:fedoraAdmin -XPUT http://localhost:8080/rest/resc_self

echo 'INSERT { <> <http://awoods.com/pointsTo> <http://localhost:8080/rest/resc_self> . } WHERE { }' | curl -u fedoraAdmin:fedoraAdmin -XPATCH http://localhost:8080/rest/resc_self -H"Content-Type: application/sparql-update" --data-binary "@-"

curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/resc_self -H 'Prefer: return=representation; include="http://fedora.info/definitions/fcrepo#PreferInboundReferences";'

<http://localhost:8080/rest/resc_self>
        <http://awoods.com/pointsTo>  <http://localhost:8080/rest/resc_self> ;
        fedora:created                "2020-09-21T21:56:29.908695Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> ;
        fedora:lastModified           "2020-09-21T21:56:33.806400Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> ;
        fedora:createdBy              "fedoraAdmin" ;
        fedora:lastModifiedBy         "fedoraAdmin" ;
        rdf:type                      ldp:BasicContainer ;
        rdf:type                      ldp:Resource ;
        rdf:type                      fedora:Resource ;
        rdf:type                      ldp:RDFSource ;
        rdf:type                      ldp:Container ;
        rdf:type                      fedora:Container ;
        <http://awoods.com/pointsTo>  <http://localhost:8080/rest/resc_self> .


curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/resc_self

<http://localhost:8080/rest/resc_self>
        <http://awoods.com/pointsTo>  <http://localhost:8080/rest/resc_self> ;
        fedora:created                "2020-09-21T21:56:29.908695Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> ;
        fedora:lastModified           "2020-09-21T21:56:33.806400Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> ;
        fedora:createdBy              "fedoraAdmin" ;
        fedora:lastModifiedBy         "fedoraAdmin" ;
        rdf:type                      ldp:BasicContainer ;
        rdf:type                      ldp:Resource ;
        rdf:type                      fedora:Resource ;
        rdf:type                      ldp:RDFSource ;
        rdf:type                      ldp:Container ;
        rdf:type                      fedora:Container .
```

bbpennel:
Did some more testing and overall things look pretty good. I tried out a few other cases and they behaved as expect:
* Changing a property from being a reference to a literal - cleaned up fine
* Inbound from a Hash URI to a resource - Adds inbound reference for the resource
* Reference from resc1 to of resc2#hash_uri - No inbound reference on the either resc2 or resc2#hash_uri (not really sure what to expect for this one)
* Web ACL - Does not produce inbound references on other resources. References to the ACL do produce inbound references on the ACL though.
* References from subjects not in the repository domain - Are not returned, which is expected I believe. I think this behavior is fine, although it might be possible someone will come up with a use case for inbound references from out of domain subjects.

whikloj:
Numbers 3 & 4 above
3. Referencing a hash uri with an inbound reference...not sure how we would want this to work. Hash Uris are not separate resolvable entities, they are just a pointer to a spot in an existing document. But the server can only provide that document. I found some ignored tests in the codebase that are trying to request a URI with a hash and fail because you can't. Perhaps we have a lack of clarity in what we can/should provide w/r/t  hash uris.
4. I have no opinion on this but it would be trivial to omit references to `fcr:acl`

awoods:
I am not going to be able to do a full review this week, but can move this PR forward if your testing shows no issues, @bbpennel ... which appears to be the case.
Regarding hash-URIs, I would rather not delay this effort on that. Let's wait to see what the community wants with regards to hash URIs and references. 
Resources not indicating that an ACL is pointing at it is probably fine. We can change later if necessary.

bbpennel (approved):

PR #1751, opened 2020-09-10
Allow any authenticated user to POST to /fcr:tx

peichman-umd:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3321

Allows any authenticated user to create a transaction by allowing POST to the /fcr:tx endpoint as long as the current user is authenticated.

# How should this be tested?

* Run the WebACRecipesIT integration tests
* Build and run using `mvn jetty:run` then test by doing: `curl -u testuser:testpass -i -X POST http://localhost:8080/rest/fcr:tx`. You should get a `201 Created` response.

# Interested parties
@fcrepo4/committers, @awoods

### DIFF
--- a/fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACFilter.java
+++ b/fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACFilter.java
@@ -300,7 +300,7 @@ public class WebACFilter extends RequestContextFilter {
     private boolean isAuthorized(final Subject currentUser, final HttpServletRequest httpRequest) throws IOException {
         final String requestURL = httpRequest.getRequestURL().toString();
         final boolean isAcl = requestURL.endsWith(FCR_ACL);
-        final boolean isTxEndpoint = requestURL.endsWith(FCR_TX);
+        final boolean isTxEndpoint = requestURL.endsWith(FCR_TX) || requestURL.endsWith(FCR_TX + "/");
         final URI requestURI = URI.create(requestURL);
         log.debug("Request URI is {}", requestURI);
         final FedoraResource resource = resource(httpRequest);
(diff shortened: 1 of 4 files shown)


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraTypes.java
```diff
@@ -99,6 +99,8 @@
 
     String FCR_TOMBSTONE = "fcr:tombstone";
 
+    String FCR_TX = "fcr:tx";
```
awoods, line 102: This is already defined in [TransactionConstants](https://github.com/fcrepo4/fcrepo4/blob/main/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/session/TransactionConstants.java#L42)... although FedoraTypes seems like a logical home. Would you be interested in refactoring these into a single constant?

fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACFilter.java
```diff
@@ -365,6 +367,10 @@ private boolean isAuthorized(final Subject currentUser, final HttpServletRequest
                 }
             }
         case "POST":
+            if (isTxEndpoint && currentUser.isAuthenticated()) {
+                log.debug("POST allowed to transaction endpoint");
```
awoods, line 371: Could you add the "currentUser"s ID to the debug message?

awoods (changes requested):
This works as desired. Nice job.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/FedoraTypes.java
peichman-umd: @awoods I've modified `TX_PREFIX` to be defined in terms of the `FCR_TX` constant.

fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACFilter.java
peichman-umd: Done.

awoods (approved):

PR #1754, opened 2020-09-15
FCREPO-3431 - Implement getChildren

bbpennel:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3431

* Implements getChildren for FedoraResource, Container and ResourceFactory
* All ContainmentIndex methods now take FedoraIds instead of FedoraResource.
* Deprecates unused `getChild` method (which seems modeshape oriented)

# How should this be tested?

New tests were added. No functional changes.

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
@@ -142,6 +142,7 @@ public interface FedoraResource {
      * @param relPath the given path
      * @return the child of this resource
      */
+    @Deprecated
     FedoraResource getChild(String relPath);
 
     /**
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
@@ -137,15 +137,6 @@ public interface FedoraResource {
     FedoraResource getAcl();
 
     /**
-     * Get the child of this resource at the given path
-     *
-     * @param relPath the given path
-     * @return the child of this resource
-     */
-    @Deprecated
-    FedoraResource getChild(String relPath);
-
-    /**
      * Does this resource have a property
      * @param relPath the given path
      * @return the boolean value whether the resource has a property
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/ContainmentTriplesServiceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/ContainmentTriplesServiceImpl.java
@@ -44,8 +44,9 @@ public class ContainmentTriplesServiceImpl implements ContainmentTriplesService 
 
     @Override
     public Stream<Triple> get(final Transaction tx, final FedoraResource resource) {
-        final Node currentNode = createURI(resource.getFedoraId().getFullId());
-        return containmentIndex.getContains(txId(tx), resource).map(c ->
+        final var fedoraId = resource.getFedoraId();
+        final Node currentNode = createURI(fedoraId.getFullId());
+        return containmentIndex.getContains(txId(tx), fedoraId).map(c ->
                 new Triple(currentNode, CONTAINS.asNode(), createURI(c)));
     }
 
--- a/fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/models/ContainerImplTest.java
+++ b/fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/models/ContainerImplTest.java
@@ -1,0 +1,82 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.fcrepo.kernel.impl.models;
+
+import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_ID_PREFIX;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import java.util.UUID;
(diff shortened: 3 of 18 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/ContainmentTriplesServiceImpl.java
```diff
@@ -44,8 +44,9 @@
 
     @Override
     public Stream<Triple> get(final Transaction tx, final FedoraResource resource) {
-        final Node currentNode = createURI(resource.getFedoraId().getFullId());
-        return containmentIndex.getContains(txId(tx), resource).map(c ->
+        final var fedoraId = resource.getFedoraId();
+        final Node currentNode = createURI(fedoraId.getFullId());
```
whikloj, line 48: I think this can be removed, I'm just reading code on my phone so I could be mistaken.

whikloj (commented):
One note for now. Will read not on phone tomorrow

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/ContainmentTriplesServiceImpl.java
bbpennel: I think I need to keep both, currentNode is used on line 50, and fedoraId is used on 49. But thank you for the phone dedication

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
```diff
@@ -142,6 +142,7 @@
      * @param relPath the given path
      * @return the child of this resource
      */
+    @Deprecated
```
whikloj, line 145: I'd vote to just remove it and not deprecate it, either that or at least return one child because the method is totally useless otherwise.

whikloj: When I say its useless, I mean because it is just returning `null` right now

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/ContainmentTriplesServiceImpl.java
whikloj: 🤦  I looked at these 5 lines for like 5 minutes and still couldn't get them correct in my mind. You are correct, please disregard my previous comment.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/models/ContainerImplTest.java
```diff
@@ -0,0 +1,82 @@
...
+    private ResourceFactory resourceFactory;
+
+    private final static String TX_ID = "transacted";
+
+    private FedoraId fedoraId;
+
+    @Before
+    public void setup() throws Exception {
+        MockitoAnnotations.initMocks(this);
+        final var fedoraIdStr = FEDORA_ID_PREFIX + "/" + UUID.randomUUID().toString();
```
whikloj, line 61: You can leave off the `FEDORA_ID_PREFIX + "/" +` and the FedoraId class will add it automatically.
So
```
- final var fedoraIdStr = FEDORA_ID_PREFIX + "/" + UUID.randomUUID().toString();
- fedoraId = FedoraId.create(fedoraIdStr);
+ fedoraId = FedoraId.create(UUID.randomUUID().toString());
```

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/models/ResourceFactoryImplTest.java
```diff
@@ -418,6 +419,67 @@ public void doesResourceExist_Exception_WithoutSession() throws Exception {
...
+        final var childrenStream = factory.getChildren(sessionId, fedoraId);
+        assertEquals(0, childrenStream.count());
+    }
+
+    @Test
+    public void getChildren_WithChildren() throws Exception {
+        populateHeaders(resourceHeaders, BASIC_CONTAINER);
+
+        final var child1IdStr = FEDORA_ID_PREFIX + "/" + UUID.randomUUID().toString();
+        final var child1Id = FedoraId.create(child1IdStr);
```
whikloj, line 442: Same as above, don't need to prefix with `FEDORA_ID_PREFIX + "/"`, can reduce to just FedoraId.create() with the UUID.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/models/ResourceFactoryImplTest.java
```diff
@@ -418,6 +419,67 @@ public void doesResourceExist_Exception_WithoutSession() throws Exception {
...
+        final var child1Headers = new ResourceHeadersImpl();
+        child1Headers.setId(child1Id);
+        populateHeaders(child1Headers, BASIC_CONTAINER);
+        when(psSession.getHeaders(child1Id, null)).thenReturn(child1Headers);
+
+        final var childNestedIdStr = FEDORA_ID_PREFIX + "/" + UUID.randomUUID().toString();
+        final var childNestedId = FedoraId.create(childNestedIdStr);
```
whikloj, line 449: Same as above, don't need to prefix with `FEDORA_ID_PREFIX + "/"`, can reduce to just FedoraId.create() with the UUID.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/models/ResourceFactoryImplTest.java
```diff
@@ -418,6 +419,67 @@ public void doesResourceExist_Exception_WithoutSession() throws Exception {
...
+        final var childNestedHeaders = new ResourceHeadersImpl();
+        childNestedHeaders.setId(childNestedId);
+        populateHeaders(childNestedHeaders, BASIC_CONTAINER);
+        when(psSession.getHeaders(childNestedId, null)).thenReturn(childNestedHeaders);
+
+        final var child2IdStr = FEDORA_ID_PREFIX + "/" + UUID.randomUUID().toString();
+        final var child2Id = FedoraId.create(child2IdStr);
```
whikloj, line 456: Same as above, don't need to prefix with `FEDORA_ID_PREFIX + "/"`, can reduce to just FedoraId.create() with the UUID.

whikloj (commented):
This is good, and builds. I've made some suggestions which are just for the tests which is why I am not requesting changes. But it would remove some extra lines. I see other existing test code doing the same thing around creating `FedoraId` strings with `FEDORA_ID_PREFIX`, so it could also be cleaned up later as well.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/models/ContainerImplTest.java
bbpennel: Ah okay, I think I copied that from another test, I'll chop these out

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/FedoraResource.java
bbpennel: I went ahead and removed it

whikloj (approved):
👍

PR #1755, opened 2020-09-16
Reject sparql-update as rdf for object creation.

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3429

Previously we allowed people to POST/PUT with application/sparql-update. This is no longer supported and as such we need to reject the mimetype. This adds a new Exception and Mapper and throws it from the HttpRdfService if we can't get a Jena RDF format from the mimeType.

Also corrects some tickets and removes an old test.

# How should this be tested?

Before the PR.
Post with application/sparql-update
```
curl -i -ufedoraAdmin:fedoraAdmin -H"Content-type: application/sparql-update" -d"INSERT { <> <http://purl.org/dc/elements/1.1/title> \"title\" } WHERE {}" -XPOST http://localhost:8080/rest
```
Get a `500 Server Error`.
Pull in this PR.
Do the same thing, get a `415 Unsupported Media Type` 

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -1730,37 +1730,23 @@ public class FedoraLdpIT extends AbstractResourceIT {
         }
     }
 
+    /**
+     * We no longer support SPARQL as a valid RDF syntax for object creation.
+     */
     @Test
-@Ignore
     public void testIngestWithNewAndSparqlQuery() throws IOException {
         final HttpPost method = postObjMethod();
         method.addHeader(CONTENT_TYPE, "application/sparql-update");
         method.setEntity(new StringEntity(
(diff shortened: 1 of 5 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -2067,21 +2054,6 @@ public void testIngestWithRDFLang() throws IOException {
         }
     }
 
-    @Test
-    // TODO It's not clear what this test is actually testing, or why it sleeps while running
```
whikloj, line 2071: For reference, this test was added back in [August 2014](https://github.com/fcrepo4/fcrepo4/commit/db1180e899c3ad9bbb65dc2f226c86b1417df7d1) related to this [PivotalTracker ticket](https://www.pivotaltracker.com/n/projects/684825/stories/77492702).

Seems it is disabled by default and so has not been run possibly ever.

pwinckles (approved):
Worked for me

PR #1757, opened 2020-09-25
Refactor referenceService to emit events.

whikloj:
Fix AbstractJmsIT to stop checking message headers.

Fix JsonLdEventMessage to have only one type for Object and Actor.

**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3434

Implements the INBOUND_REFERENCE event type when you reference an item.

Fixes the fcrepo-jms tests to stop relying on the message headers and actually decode the body of the message.

# How should this be tested?

1. Setup a JMSClient to listen to Fedora.
1. Create two resources.
1. Make one reference the other ie. `<http://localhost:8080/rest/first> <http://example.org/pointsTo> <http://localhost:8080/rest/second>`
1. Check your JMSClient for a Follow event, like
    ```
    {"id":"urn:uuid:8cea1bab-c1c1-4b9b-ac5d-a666aa6abad1","type":["Follow"],"name":"refer to resource","published":"2020-09-25T20:57:08.641850Z","actor":[{"type":"Person","id":"info:fedora/local-user#fedoraAdmin"},{"type":"Application","name":"curl/7.64.1"}],"object":{"type":["http://mementoweb.org/ns#TimeGate","http://fedora.info/definitions/v4/repository#Container","http://www.w3.org/ns/ldp#RDFSource","http://www.w3.org/ns/ldp#Resource","http://www.w3.org/ns/ldp#Container","http://mementoweb.org/ns#OriginalResource","http://fedora.info/definitions/v4/repository#Resource","http://www.w3.org/ns/ldp#BasicContainer","http://www.w3.org/ns/prov#Entity"],"id":"http://localhost:8080/rest/second","isPartOf":"http://localhost:8080/rest"},"@context":["https://www.w3.org/ns/activitystreams#",{"prov":"http://www.w3.org/ns/prov#","dcterms":"http://purl.org/dc/terms/","type":"@type","id":"@id","isPartOf":{"@id":"dcterms:isPartOf","@type":"@id"}}]}
    ```
1. Do the same but make the target of the reference a resource that does NOT exist.
1. See NO Follow event.

# Interested parties
@fcrepo4/committers

### DIFF
--- a/fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/JsonLDEventMessage.java
+++ b/fcrepo-event-serialization/src/main/java/org/fcrepo/event/serialization/JsonLDEventMessage.java
@@ -18,8 +18,10 @@
 
 package org.fcrepo.event.serialization;
 
-import static java.util.Collections.singletonList;
+import static java.time.format.DateTimeFormatter.ISO_INSTANT;
 import static java.util.stream.Collectors.toList;
+
+import static org.fcrepo.kernel.api.RdfLexicon.ACTIVITY_STREAMS_NAMESPACE;
 import static org.fcrepo.kernel.api.RdfLexicon.PROV_NAMESPACE;
 import static org.slf4j.LoggerFactory.getLogger;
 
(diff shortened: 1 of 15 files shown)


pwinckles (approved):
Works as expected.

Note, that this PR does include a "breaking change" in that it corrects the actor's type to be a string rather than an array of strings.

PR #1760, opened 2020-09-30
FCREPO-3409 - DC membership index

bbpennel:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3409

Adds and populates a membership index, which tracks the membership history of resources.

I'm not sure if a message should be emitted when membership changes, but there is not an existing test indicating so.

# How should this be tested?

The application's behaviors should be unchanged from before since the service to add membership to responses has not been completed. I could update this PR to add it, but it was a separate ticket https://jira.lyrasis.org/browse/FCREPO-3165

So it should build and the new tests should pass.

# Interested parties
@fcrepo4/committers


whikloj (commented):
I haven't read through the huge test class, so these are just my initial thoughts. Also wondering if there is any significant difference in how the system works with or without autoversioning. Mostly because I see that we only test with autoversioning enabled

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipIndexManager.java
```diff
@@ -0,0 +1,602 @@
...
+            " WHERE source_id = :sourceId" +
+                " AND end_time = :noEndTime" +
+                " AND NOT EXISTS (" +
+                    " SELECT TRUE" +
+                    " FROM membership_tx_operations mtx" +
+                    " WHERE mtx.subject_id = m.subject_id" +
+                        " AND mtx.property = m.property" +
+                        " AND mtx.object_id = m.object_id" +
+                        " AND mtx.source_id = m.source_id" +
+                        " AND mtx.operation = :deleteOp" +
```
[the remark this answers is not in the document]
bbpennel: mainly for programmer error. When I was working on it I originally had 'delete' hardcoded in the query, but ended up with mismatches between the values in different places ('delete' vs 'deleted') which failed silently since it was valid sql. By passing in a constant it means I'll get an error if I misspell `:deleteOp` since the variable would be missing.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipIndexManager.java
```diff
@@ -0,0 +1,602 @@
...
+        }
+
+        return membership.stream();
+    }
+
+    /**
+     * Perform a commit of operations stored in the specified transaction
+     * @param txId transaction id
+     */
+    public void commitTransaction(final String txId) {
```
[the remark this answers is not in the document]
bbpennel: I'll add that. Is it needed for all methods that modify the index?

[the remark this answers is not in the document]
whikloj: Just if you are performing multiple SQL statements that all either need to succeed or the whole thing fail. In the Containment index only `commit` and `reset` are transactional as they each have more than one updates that must occur.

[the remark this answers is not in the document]
bbpennel: Okay, I'll add the annotation to a few more places

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipServiceImpl.java
```diff
@@ -0,0 +1,398 @@
...
+
+    /**
+     * Creates a timeline of states for a DirectContainer, tracking changes to its
+     * properties that impact membership.
+     * @param fedoraResc resource subject of the timeline
+     * @return timeline
+     */
+    private List<DirectContainerProperties> makePropertyTimeline(final FedoraResource fedoraResc) {
+        final var entryList = fedoraResc.getTimeMap().getChildren()
+                .sorted((m1, m2) -> m1.getFedoraId().getFullId().compareTo(m1.getFedoraId().getFullId()))
```
pwinckles, line 333: `m2` is not used.

I think this could also be rewritten like: `.sorted(Comparator.comparing(m -> m.getFedoraId().getFullId()))`

pwinckles: Are you just trying to sort the mementos by date? If so, I think they should already be sorted.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipServiceImpl.java
```diff
@@ -0,0 +1,398 @@
...
+                .map(memento -> new DirectContainerProperties(memento))
+                .collect(Collectors.toList());
+
+        // Reduce timeline to entries where significant properties change
+        final var changeEntries = new ArrayList<DirectContainerProperties>();
+        var curr = entryList.get(0);
+        changeEntries.add(curr);
+        int i = 1;
+        while (i < entryList.size()) {
```
pwinckles, line 342: Why are you using a while loop rather than a for loop?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipServiceImpl.java
```diff
@@ -0,0 +1,398 @@
...
+            final var parentRdfResc = getRdfResource(parentResc);
+            final var newMembership = generateDirectMembership(txId, parentRdfResc, fedoraResc);
+            indexManager.addMembership(txId, parentResc.getFedoraId(), newMembership, fedoraResc.getCreatedDate());
+        }
+
+        // TODO check if fedoraResc is a member of any IndirectContainers
+    }
+
+    @Override
+    public void resourceModified(final String txId, final FedoraId fedoraId) {
```
pwinckles, line 95: I think you'll actually want a `@Transactional` at this level

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipServiceImpl.java
```diff
@@ -0,0 +1,398 @@
...
+        indexManager.deleteTransaction(txId);
+    }
+
+    @Override
+    public void reset() {
+        indexManager.clearIndex();
+    }
+
+    @Override
+    public void populateMembershipHistory(final String txId, final FedoraId containerId) {
```
pwinckles, line 267: `@Transactional`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipServiceImpl.java
```diff
@@ -0,0 +1,398 @@
...
+    private FedoraResource getParentResource(final FedoraResource resc) {
+        try {
+            return resc.getParent();
+        } catch (final PathNotFoundException e) {
+            throw new PathNotFoundRuntimeException(e);
+        }
+    }
+
+    @Override
+    public void resourceDeleted(final String txId, final FedoraId fedoraId) {
```
pwinckles, line 208: `@Transactional`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipIndexManager.java
```diff
@@ -0,0 +1,602 @@
...
+     *
+     * @param txId transaction id
+     * @param sourceId ID of the direct/indirect container whose membership should be ended
+     * @param membership membership triple to end
+     * @param endTime the time the resource was deleted, generally its last modified
+     */
+    public void endMembership(final String txId,  final FedoraId sourceId, final Triple membership,
+            final Instant endTime) {
+        final Map<String, Object> parameterSource = Map.of(
+                "txId", txId,
```
pwinckles, line 310: Perhaps the column names should be constants?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipIndexManager.java
```diff
@@ -0,0 +1,602 @@
...
+        // Clear any existing delete operation for this membership
+        final Map<String, Object> parametersDelete = Map.of(
+                "txId", txId,
+                "sourceId", sourceId.getFullId(),
+                "subjectId", membership.getSubject().getURI(),
+                "property", membership.getPredicate().getURI(),
+                "objectId", membership.getObject().getURI(),
+                "operation", DELETE_OPERATION);
+
+        jdbcTemplate.update(CLEAR_ENTRY_IN_TX, parametersDelete);
```
pwinckles, line 436: I think there may be some oddness here when `addMembership()` is called after calling `deleteMembershipForSource()`, but it's possible that I just don't understand the intended behavior. The problem I see is that there are two different types of deletes, with or without the `force` flag. `deleteMembershipForSource()` creates deletes with the `force` flag, and I assume that you would always want to apply them. However, this method will delete all of these operations in the transaction, so they won't be applied. Won't this result in duplicate entries (assuming there aren't any db constraints being violated)? I think the case where this matters is MembershipServiceImpl#L104.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipIndexManager.java
```diff
@@ -0,0 +1,602 @@
...
+            " WHERE EXISTS (" +
+                "SELECT TRUE" +
+                " FROM membership_tx_operations mto" +
+                " WHERE mto.tx_id = :txId" +
+                    " AND mto.operation = :deleteOp" +
+                    " AND m.source_id = mto.source_id" +
+                    " AND m.subject_id = mto.subject_id" +
+                    " AND m.property = mto.property" +
+                    " AND m.object_id = mto.object_id" +
+                " )";
```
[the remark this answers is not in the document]
bbpennel: The first `WHERE` clause is in order to get the new value for `end_time`, and the `WHERE EXISTS` is to determine which `membership` entries are being ended. But I'll add the operation check to the first subquery, just in case there are multiple operations for membership entry in the current transaction. If I'm remembering correctly, this query ended up kind of dumb looking because h2 doesn't support joins in `UPDATE` queries, so I had to have the two nearly identical subqueries.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipServiceImpl.java
```diff
@@ -0,0 +1,398 @@
...
+
+    public static final Instant NO_END_INSTANT = Instant.parse("9999-12-31T00:00:00.000Z");
+
+    @Inject
+    private MembershipIndexManager indexManager;
+
+    @Inject
+    private ResourceFactory resourceFactory;
+
+    @Value("${fcrepo.autoversioning.enabled:true}")
```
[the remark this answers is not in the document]
pwinckles: It shouldn't be a problem for any module to take a dependency on fcrepo-configs.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipServiceImpl.java
bbpennel: Yes, I'm trying to sort them by date, I'm happy to remove the sort (and its typo) if we can rely on the versions to be sorted in ascending order. If that's a guarantee, perhaps we should update the documentation for `org.fcrepo.persistence.api.PersistentStorageSession.listVersions(FedoraId)` to state that? Any other theoretic persistence implementations would need to ensure order too.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipServiceImpl.java
bbpennel: Hmm, I'd like to think I had a reason at some point, but doesn't seem like there is one, I'll switch it.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipServiceImpl.java
bbpennel: Sure. In the case of a container with lots of members this could be a very long transaction, I'm guessing that won't be problematic?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipIndexManager.java
bbpennel: The parameter names or the column names? I'm okay with constants for the parameter names

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipIndexManager.java
bbpennel: That's a good point, I believe I should exclude `force_flag` deletes from being cleaned up here.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipServiceImpl.java
pwinckles: No, I wouldn't expect that would be a problem

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipIndexManager.java
pwinckles: The keys in the map

pwinckles (approved):

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipIndexManager.java
```diff
@@ -0,0 +1,630 @@
...
+    }
+
+    /**
+     * End membership entry within a transaction
+     * @param txId transaction id
+     * @param sourceId ID of the direct/indirect container whose membership should be ended
+     * @param membership membership triple to end
+     * @param endTime the time the resource was deleted, generally its last modified
+     */
+    public void endMembershipInTx(final String txId,  final FedoraId sourceId, final Triple membership,
```
awoods, line 362: Can this method be removed? It does not appear to be called anywhere.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipServiceImpl.java
```diff
@@ -0,0 +1,438 @@
...
+            lastVersionDatetime = null;
+        } else {
+            // If at least one past version, then reindex membership involving the last version and after
+            lastVersionDatetime = mementoDatetimes.get(mementoDatetimes.size() - 1);
+        }
+        indexManager.deleteMembershipForSourceAfter(txId, dcId, lastVersionDatetime);
+        populateMembershipHistory(txId, dcResc, lastVersionDatetime);
+    }
+
+    private Triple generateDirectMembership(final String txId, final Resource dcRdfResc,
```
awoods, line 146: `txId` never used.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipServiceImpl.java
```diff
@@ -0,0 +1,438 @@
...
+                .collect(Collectors.toList());
+        } else {
+            timeline = propertyTimeline;
+        }
+
+        // get all the members of the DC and index the history for each, accounting for changes to the DC
+        fedoraResc.getChildren().forEach(member -> {
+            final var memberDeleted = member instanceof Tombstone;
+            final var memberNode = NodeFactory.createURI(member.getFedoraId().getFullId());
+            log.debug("Populating membership history for DirectContainer {}member {}",
```
awoods, line 314: Maybe add a `,` between "{}" and "member"?

awoods (changes requested):

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipServiceImpl.java
bbpennel: I know it looks goofy, but that would result in two spaces between `DirectContainer` and `member` if the member isn't deleted. I could remove the deletion state from the debug statement though

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipIndexManager.java
bbpennel: Good catch, I think this got leftover during a few refactors of `populateMembershipHistory`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipIndexManager.java
```diff
@@ -0,0 +1,630 @@
...
+                    ")";
+
+    private static final String DELETE_EXISTING_FOR_SOURCE_AFTER =
+            "INSERT INTO membership_tx_operations" +
+            " (subject_id, property, object_id, source_id, start_time, end_time, tx_id, operation, force_flag)" +
+            " SELECT subject_id, property, object_id, source_id, start_time, end_time, :txId, :deleteOp, :forceFlag" +
+            " FROM membership m" +
+            " WHERE m.source_id = :sourceId" +
+                " AND (m.start_time >= :startTime" +
+                " OR m.end_time >= :startTime)";
```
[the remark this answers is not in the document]
bbpennel: Yes, that's right. It is to deal with obnoxious history rewriting from modifying mutable head versions of DirectContainers. Sorry to do this to you, but consider the following timeline, in a repo with autoversioning turned off:

1) Create membership resc `coll1` and direct container `dc1` with membership relation `ex:hasMember1`, at time T0.
2) At time T1, resource `dc1/mem1` is created. This adds an entry to the index:
`<coll1> ex:hasMember1 <dc1/mem1>` which spans T1 to forever.
3) At time T2, `dc1` is changed to have relation `ex:hasMember2`. The index is now:
`<coll1> ex:hasMember2 <dc1/mem1>` which spans T1 to forever (the old entry is deleted since its end time was >= T2)
4) At T3, create V1 of `dc1`. No changes to the index.
5) At time T4, `dc1` is changed to have relation `ex:hasMember3`. The index is now:
`<coll1> ex:hasMember2 <dc1/mem1>` which spans T1 to T4 (T3/V1 locked ex:hasMember2 in, which changes at T4)
`<coll1> ex:hasMember3 <dc1/mem1>` which spans T4 to forever.
6) At T5, `dc1` is changed back to relation `ex:hasMember1`. History in the index is editted:
`<coll1> ex:hasMember2 <dc1/mem1>` which spans T1 to T5 (note, end time changed)
`<coll1> ex:hasMember1 <dc1/mem1>` which spans T5 to forever.

The query in question allows the last few states to be rewritten. Alternatively, I could regenerate the full history of membership produced by `dc1` whenever its modified, which I had earlier in the PR. Its simpler, but obviously costly in scenarios where clients create versions of DCs and modify them.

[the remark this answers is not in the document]
awoods: Thanks for the explanation and deep consideration of scenarios.

awoods (approved):

PR #1761, opened 2020-09-30
Enforce server managed predicates and types, add interaction model to containment index.

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3258, also continues work on https://jira.lyrasis.org/browse/FCREPO-3223

Enforces not being able to change the interaction model of a resource via
* Link header
* RDF body
* PATCH request body

# How should this be tested?

Lots of tests have been added, but now if you add RDF with a `rdf:type` in the namespace of Memento, LDP or Fedora you should get rejected. If you send a Link header different from the current interaction model you should get rejected. If you send a `rdf:type` that is not a URI you should get rejected.

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
@@ -22,7 +22,6 @@ import org.fcrepo.http.commons.api.rdf.HttpIdentifierConverter;
 import org.fcrepo.http.commons.session.TransactionProvider;
 import org.fcrepo.kernel.api.Transaction;
 import org.fcrepo.kernel.api.TransactionManager;
-import org.fcrepo.kernel.api.TransactionUtils;
 import org.fcrepo.kernel.api.exception.PathNotFoundException;
 import org.fcrepo.kernel.api.identifiers.FedoraId;
 import org.fcrepo.kernel.api.models.FedoraResource;
@@ -105,10 +104,6 @@ abstract public class FedoraBaseResource extends AbstractResource {
      */
     protected boolean isGhostNode(final Transaction transaction, final FedoraId fedoraId) {
         return resourceHelper.isGhostNode(transaction, fedoraId);
-    }
-
-    protected String getInteractionModel(final Transaction transaction, final FedoraId fedoraId) {
-        return resourceFactory.getInteractionModel(TransactionUtils.openTxId(transaction), fedoraId);
     }
 
     protected String getUserPrincipal() {
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
@@ -22,6 +22,7 @@ import org.fcrepo.http.commons.api.rdf.HttpIdentifierConverter;
 import org.fcrepo.http.commons.session.TransactionProvider;
 import org.fcrepo.kernel.api.Transaction;
 import org.fcrepo.kernel.api.TransactionManager;
+import org.fcrepo.kernel.api.TransactionUtils;
 import org.fcrepo.kernel.api.exception.PathNotFoundException;
 import org.fcrepo.kernel.api.identifiers.FedoraId;
 import org.fcrepo.kernel.api.models.FedoraResource;
@@ -35,6 +36,7 @@ import javax.ws.rs.core.Context;
 import javax.ws.rs.core.SecurityContext;
 import java.security.Principal;
 
+import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_NON_RDF_SOURCE_DESCRIPTION;
 import static org.slf4j.LoggerFactory.getLogger;
 
 /**
@@ -104,6 +106,13 @@ abstract public class FedoraBaseResource extends AbstractResource {
      */
     protected boolean isGhostNode(final Transaction transaction, final FedoraId fedoraId) {
         return resourceHelper.isGhostNode(transaction, fedoraId);
+    }
+
+    protected String getInteractionModel(final Transaction transaction, final FedoraId fedoraId) {
+        if (fedoraId.isDescription()) {
+            return FEDORA_NON_RDF_SOURCE_DESCRIPTION;
+        }
+        return resourceFactory.getInteractionModel(TransactionUtils.openTxId(transaction), fedoraId);
     }
 
     protected String getUserPrincipal() {
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
@@ -36,7 +36,6 @@ import javax.ws.rs.core.Context;
 import javax.ws.rs.core.SecurityContext;
 import java.security.Principal;
 
-import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_NON_RDF_SOURCE_DESCRIPTION;
 import static org.slf4j.LoggerFactory.getLogger;
 
 /**
@@ -109,9 +108,6 @@ abstract public class FedoraBaseResource extends AbstractResource {
     }
 
     protected String getInteractionModel(final Transaction transaction, final FedoraId fedoraId) {
-        if (fedoraId.isDescription()) {
-            return FEDORA_NON_RDF_SOURCE_DESCRIPTION;
-        }
         return resourceFactory.getInteractionModel(TransactionUtils.openTxId(transaction), fedoraId);
     }
 
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -65,6 +65,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.stream.Collectors;
 
 import javax.inject.Inject;
@@ -401,8 +402,8 @@ public class FedoraLdp extends ContentExposingResource {
             }
 
             final String resInteractionModel = getInteractionModel(transaction, fedoraId);
-            if (StringUtils.isNoneBlank(interactionModel) && StringUtils.isNoneBlank(resInteractionModel)
-                    && !resInteractionModel.equals(interactionModel)) {
+            if (StringUtils.isNoneBlank(resInteractionModel, interactionModel) &&
+                    !Objects.equals(resInteractionModel, interactionModel)) {
                 throw new InteractionModelViolationException("Changing the interaction model " + resInteractionModel
                         + " to " + interactionModel + " is not allowed!");
             }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -51,6 +51,7 @@ import static org.fcrepo.http.commons.domain.RDFMediaType.APPLICATION_OCTET_STRE
 
 import static org.fcrepo.kernel.api.RdfLexicon.ARCHIVAL_GROUP;
 import static org.fcrepo.kernel.api.RdfLexicon.INTERACTION_MODEL_RESOURCES;
+import static org.fcrepo.kernel.api.RdfLexicon.NON_RDF_SOURCE;
 import static org.fcrepo.kernel.api.RdfLexicon.VERSIONED_RESOURCE;
 import static org.fcrepo.kernel.api.services.VersionService.MEMENTO_RFC_1123_FORMATTER;
 import static org.slf4j.LoggerFactory.getLogger;
@@ -97,6 +98,7 @@ import org.fcrepo.kernel.api.FedoraTypes;
 import org.fcrepo.kernel.api.exception.AccessDeniedException;
 import org.fcrepo.kernel.api.exception.CannotCreateResourceException;
 import org.fcrepo.kernel.api.exception.GhostNodeException;
+import org.fcrepo.kernel.api.exception.InteractionModelViolationException;
 import org.fcrepo.kernel.api.exception.InvalidChecksumException;
 import org.fcrepo.kernel.api.exception.MalformedRdfException;
 import org.fcrepo.kernel.api.exception.MementoDatetimeFormatException;
@@ -398,13 +400,12 @@ public class FedoraLdp extends ContentExposingResource {
                 throw new ClientErrorException("An If-Match header is required", 428);
             }
 
-            // TODO: Check existing resources interaction model and make sure we aren't trying to change it.
-            //final String resInteractionModel = getInteractionModel(resource);
-            //if (StringUtils.isNoneBlank(interactionModel) && StringUtils.isNoneBlank(resInteractionModel)
-            //        && !resInteractionModel.equals(interactionModel)) {
-            //    throw new InteractionModelViolationException("Changing the interaction model " + resInteractionModel
-            //            + " to " + interactionModel + " is not allowed!");
-            //}
+            final String resInteractionModel = getInteractionModel(transaction, fedoraId);
+            if (StringUtils.isNoneBlank(interactionModel) && StringUtils.isNoneBlank(resInteractionModel)
+                    && !resInteractionModel.equals(interactionModel)) {
+                throw new InteractionModelViolationException("Changing the interaction model " + resInteractionModel
+                        + " to " + interactionModel + " is not allowed!");
+            }
         }
 
         if (isGhostNode(transaction(), fedoraId)) {
@@ -680,7 +681,7 @@ public class FedoraLdp extends ContentExposingResource {
                              final boolean contentPresent, final boolean contentExternal) {
         final String simpleContentType = contentPresent ? contentType : null;
         final boolean isRdfContent = isRdfContentType(simpleContentType);
-        return "ldp:NonRDFSource".equals(interactionModel) || contentExternal ||
+        return NON_RDF_SOURCE.getURI().equals(interactionModel) || contentExternal ||
                 (contentPresent && interactionModel == null && !isRdfContent);
     }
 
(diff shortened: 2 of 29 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
```diff
@@ -150,6 +150,42 @@ public FedoraResource getContainer(final String transactionId, final FedoraId re
...
 
+    @Override
+    public String getInteractionModel(final String transactionId, final FedoraId fedoraId) {
+        if (fedoraId.isRepositoryRoot()) {
+            return BASIC_CONTAINER.getURI();
+        }
+        if (!(fedoraId.isMemento() || fedoraId.isAcl())) {
+            // containment index doesn't handle versions and only tells us if the resource (not acl) is there,
+            // so don't bother checking for them.
+            return containmentIndex.getInteractionModel(transactionId, fedoraId);
```
pwinckles, line 161: I'm not sure if I follow why the interaction model is being added to the containment index. Is it just a place to stick it so that the resource headers don't need to be read?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
```diff
@@ -92,6 +94,13 @@ protected boolean doesResourceExist(final Transaction transaction, final FedoraI
         return resourceFactory.doesResourceExist(transaction, fedoraId);
     }
 
+    protected String getInteractionModel(final Transaction transaction, final FedoraId fedoraId) {
+        if (fedoraId.isDescription()) {
```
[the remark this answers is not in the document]
whikloj: My understanding, but I could be wrong that binaries exist in the containment index but not binary descriptions.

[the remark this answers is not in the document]
pwinckles: But, aren't you asking the ResourceFactory and not the ContainmentIndex?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
pwinckles: Hm... I think we should think about this more. It feels like data is being tacked on here just because it's a place to put it.

In the cases where `getInteractionModel()` is used, how often are the ResourceHeaders still loaded as part of processing the request? If the headers are still being loaded most of the time, then it would seem to me that a better optimization would be to cache the resource headers rather than sticking the interaction model in the database.

If we do want this information in the database, then I think we should consider instead putting all of the resource headers into the database as a new table.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraBaseResource.java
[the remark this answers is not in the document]
whikloj: Fair point, I was having a problem which I may have mis-diagnosed. I'll dig in after the call today.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
whikloj: Ok containment index is left alone.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -395,13 +397,12 @@ public Response createOrReplaceObjectRdf(
...
 
-            // TODO: Check existing resources interaction model and make sure we aren't trying to change it.
-            //final String resInteractionModel = getInteractionModel(resource);
-            //if (StringUtils.isNoneBlank(interactionModel) && StringUtils.isNoneBlank(resInteractionModel)
-            //        && !resInteractionModel.equals(interactionModel)) {
-            //    throw new InteractionModelViolationException("Changing the interaction model " + resInteractionModel
-            //            + " to " + interactionModel + " is not allowed!");
-            //}
+            final String resInteractionModel = getInteractionModel(transaction, fedoraId);
+            if (StringUtils.isNoneBlank(interactionModel) && StringUtils.isNoneBlank(resInteractionModel)
```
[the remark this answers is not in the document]
whikloj: I had to keep the `StringUtils.isNoneBlank` but I did add the `Objects.equals` too. Because I can only compare if they both are not blank.

pwinckles (dismissed):
Works for me.

One minor nitpick, if you attempt to delete a rdf type you get a message like "Changing this resource's interaction model to http://www.w3.org/ns/ldp#Resource is not allowed", which is not what the operation was attempting. That said, I'm fine with it.

whikloj:
I can make the triple check code return the exception and let the calling code capture those exceptions and decide how to word the final exception.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -3203,13 +3195,16 @@ public void testCreateResourceWithoutContentType() {
         assertEquals(CREATED.getStatusCode(), getStatus(new HttpPut(serverAddress + getRandomUniqueId())));
     }
 
+    /*
+     * TODO: This was originally expecting a 415 Unsupported Media Type, but because we assume text/turtle now it
```
awoods, line 3199: Is there actually anything "todo"?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -3262,7 +3253,7 @@ public void testBinarySetBadMimeType() throws IOException {
                 "INSERT { <" + subjectURI + "> ebucore:hasMimeType \"-- invalid syntax! --\" } WHERE {}")
         );
 
-        assertEquals(BAD_REQUEST.getStatusCode(), getStatus(patch));
```
awoods, line 3265: Are you saying that a malformed SPARQL-Update request should silently fail? It seems like `BAD_REQUEST` is a more correct response. Am I missing something?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
```diff
@@ -112,4 +112,12 @@ public FedoraResource getResource(final String transactionId, final FedoraId fed
...
      */
     public Stream<FedoraResource> getChildren(final String transactionId, final FedoraId resourceId);
+
+    /**
+     * Get the interaction model of the resource, assumes you've checked it exists already.
+     * @param transactionId The transaction id
+     * @param fedoraId Identifier of the resource.
+     * @return The interaction model.
+     */
+    public String getInteractionModel(final String transactionId, final FedoraId fedoraId);
```
awoods, line 122: It certainly can work here; however, it seems like "getInteractionModel" would more appropriately be a "FedoraResource" method, rather than a "ResourceFactory" method.

The following is available on "FedoraResourceImpl"
```
psSession.getHeaders(fedoraId, fedoraId.getMementoInstant());
```

Would it make sense for "FedoraLdp" to create the known-to-exist resource, then request its interaction-model?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
```diff
@@ -86,6 +88,12 @@
     public static final Predicate<String> isManagedNamespace = p -> p.equals(REPOSITORY_NAMESPACE) ||
             p.equals(LDP_NAMESPACE) || p.equals(MEMENTO_NAMESPACE);
 
+    /**
+     * Tests if we are trying to set the rdf:type to an object with a restricted namespace.
```
awoods, line 92: Maybe update this comment to describe what qualifies as a "restricted namespace", vs talking about "setting" rdf:types... which this predicate is not doing.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
```diff
@@ -259,21 +267,25 @@
     /**
      * Fedora defined JCR node type with supertype of nt:file with two nt:folder named fedora:timemap and
      * fedora:binaryTimemap inside.
+     * TODO: Remove modeshape-ism?
```
awoods, line 270: Nice finds. Can you create a ticket to purge these relics?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
```diff
@@ -150,6 +150,35 @@ public FedoraResource getContainer(final String transactionId, final FedoraId re
         }
     }
 
+    @Override
+    public String getInteractionModel(final String transactionId, final FedoraId fedoraId) {
+        if (fedoraId.isRepositoryRoot()) {
```
awoods, line 155: The headers for the RepositoryRoot also indicate its interaction model. Is the potential performance optimization here worth the redundant logic?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
```diff
@@ -116,8 +116,8 @@
      * Insert a parent child relationship to the transaction operation table.
      */
     private static final String INSERT_CHILD_IN_TRANSACTION = "INSERT INTO " + TRANSACTION_OPERATIONS_TABLE +
-            " ( " + PARENT_COLUMN + ", " + FEDORA_ID_COLUMN + ", " + TRANSACTION_ID_COLUMN + ", " + OPERATION_COLUMN +
-            " ) VALUES (:parent, :child, :transactionId, 'add')";
+            " ( " + PARENT_COLUMN + ", " + FEDORA_ID_COLUMN + ", " + TRANSACTION_ID_COLUMN + ", " +
```
awoods, line 119: It does not look like this is a function update. Can you revert the reformatting?

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/ContainmentIndexImplTest.java
```diff
@@ -78,6 +79,8 @@
     private final Map<String, FedoraResource> id_to_resource = new HashMap<>();
     private final Map<String, Transaction> id_to_transaction = new HashMap<>();
 
+    private final String ixnModel = BASIC_CONTAINER.getURI();
```
awoods, line 82: I believe all updates to this file can be reverted.

awoods (changes requested):

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
```diff
@@ -117,6 +122,7 @@ public Model bodyToInternalModel(final String extResourceId, final InputStream s
 
         while (stmtIterator.hasNext()) {
             final Statement stmt = stmtIterator.nextStatement();
+            checkForDisallowedRdf(stmt);
```
bbpennel, line 125: It seems like some of the functionality here is duplicating some checks in `AbstractService`:
https://github.com/fcrepo/fcrepo/blob/main/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java#L154
Were you thinking we would remove the check in AbstractService?

One advantage to where it is in AbstractService is that it is validating the model after a sparql update query has been applied to the model, in the case of PATCH. I don't think this method is called for PATCH currently?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
whikloj: Sorry I just wanted to keep track of these little changes to the interactions. If going from 415 to 400 is fine, then I can remove the `TODO`.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
whikloj: Its actually not invalid syntax unless we want to parse all ontologies for their targets and domains.

This was a problem because we relied on this particular triple to send the binary response. Now we store the mime-type in the headers file and this is user rdf. If they want the triple `<> ebucore:hasMimeType "I want my baby back baby back baby back"` that is up to them. 

I could modify this test to check that the binary is still retrievable and the user rdf says whatever or I can remove the entire test.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
whikloj: I'm fine with that.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
bbpennel: I don't know if this is a concern or not, but it might be unpredictable which error response you get from a request if the input contains both an interaction model change and a server managed triple. By checking for all the errors in the same loop (or not deferring the error thrown), whichever error is encountered first will be the result. Maybe that's fine, I was just concerned about that when looking at some tests, since the order of RDF is not generally guaranteed.

For what its worth, I stupidly hadn't noticed this PR yesterday and had started implementing checks for interaction model changes in my work on membership triples since so many of the Direct/Indirect container tests are about interaction model changes. I had added this method to `ReplacePropertiesServiceImpl`:
```
private void ensureInteractionModelUnchanged(final FedoraId fedoraId, final Model model,
            final String existingInteractionModel) {
        final var rdfResc = model.getResource(fedoraId.getFullId());
        final StmtIterator it = rdfResc.listProperties(RDF.type);
        while (it.hasNext()) {
            final Statement st = it.next();
            if (!st.getObject().isURIResource()) {
                continue;
            }
            final var objUri = st.getResource().getURI();

            if (INTERACTION_MODELS_FULL.contains(objUri) && !objUri.equals(existingInteractionModel)) {
                throw new InteractionModelViolationException("Changing the interaction model "
                        + existingInteractionModel + " to " + objUri + " is not allowed!");
            }
        }
    }
```
I was running this before `checkForSmtsLdpTypes`. On the other hand adding more iterations through the model will add up.

 but wanted to make sure that if the request was changing the interaction model, that we would get the error for that over an error for

bbpennel (changes requested):

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
```diff
@@ -230,4 +236,34 @@ protected static Model parseBodyAsModel(final InputStream requestBodyStream,
...
+     * Several tests for invalid or disallowed RDF statements.
+     * @param triple the triple to check.
+     */
+    public static void checkTripleForDisallowed(final Triple triple) {
+        if (triple.getPredicate().equals(type().asNode()) && !triple.getObject().isURI()) {
+            // The object of a rdf:type triple is not a URI.
+            final var newType = triple.getObject().toString();
+            throw new MalformedRdfException(
+                    String.format("Invalid rdf:type: %s", newType));
+        } else if (restrictedType.test(triple)) {
```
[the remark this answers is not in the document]
whikloj: 🤦 I forgot to add the lenient handling. Mea culpa.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
```diff
@@ -86,6 +88,12 @@
     public static final Predicate<String> isManagedNamespace = p -> p.equals(REPOSITORY_NAMESPACE) ||
             p.equals(LDP_NAMESPACE) || p.equals(MEMENTO_NAMESPACE);
 
+    /**
+     * Tests if we are trying to set the rdf:type to an object with a restricted namespace.
+     */
+    public static final Predicate<Triple> restrictedType = s -> s.getPredicate().equals(type().asNode()) &&
```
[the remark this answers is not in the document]
whikloj: `public static final Property type = Init.type();`

So I'm not sure it matters, but I can change it.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
awoods: The change in response code here is probably ok.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
whikloj: You are right about some duplication at the AbstractService layer. I had forgotten about that. As we are already looping through the entire incoming RDF/Sparql-update at the Http layer perhaps removing those checks from the lower level would be better?

This does check PATCH requests as we also loop through them to translate any external URIs to internal IDs.

As for storing up exceptions...I'm a believer in here is your first problem, come back later. But also I am wondering is if you try to change a server managed triple and change the interaction model in a single request. Do you expect a 409 or 400?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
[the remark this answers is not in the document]
bbpennel: I doubt it has much impact, but `RDF.Init.type()` is a call to `ResourceFactory.createProperty` while it seems like RDF.type is a constant of that result.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
bbpennel: Seems like we either need to define precedence for exception types (maybe prioritize specificity?), or leave it as a free for all. Flapping tests seem possible without any sort of precedence, but that doesn't seem to have been a problem so far. Not sure if that's a function of jena not changing significantly, the order in which checks happen in Fedora 4/5, or how the tests are written.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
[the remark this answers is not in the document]
whikloj: This is now allowed if you provide `handling=lenient` because we just remove those triples from the provided RDF.

awoods:
Looks like a merge conflict

awoods (approved):

bbpennel (approved):

PR #1762, opened 2020-10-05
cleanup exceptions

pwinckles:
**JIRA Ticket**: none

1. Stacktraces are never returned by default
2. Server exceptions are always logged by default (client exceptions remain on debug only)
3. Removed custom exception constructors that do not accept an exception message. The problem with these constructors is that Java appends the exception class name to the wrapped exception message if a message is not provided. This creates a problem for Fedora's HTTP exception mapping, which tends to rely on calling `getMessage()`.

# How should this be tested?

Do some bad stuff and make sure you like how the exceptions are handled.

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/ExternalContentAccessExceptionMapper.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/ExternalContentAccessExceptionMapper.java
@@ -45,7 +45,7 @@ public class ExternalContentAccessExceptionMapper
 
     @Override
     public Response toResponse(final ExternalContentAccessException exception) {
-        debugException(this, exception, LOGGER);
+        LOGGER.warn("Failed to read external content", exception);
         return status(BAD_GATEWAY).entity(exception.getMessage()).type(TEXT_PLAIN_WITH_CHARSET)
             .build();
     }
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/InvalidPrefixExceptionMapper.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/InvalidPrefixExceptionMapper.java
@@ -17,18 +17,17 @@
  */
 package org.fcrepo.http.commons.exceptionhandlers;
 
-import static javax.ws.rs.core.Response.status;
-import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
-import static org.fcrepo.http.commons.domain.RDFMediaType.TEXT_PLAIN_WITH_CHARSET;
-import static org.slf4j.LoggerFactory.getLogger;
+import org.fcrepo.kernel.api.exception.InvalidPrefixException;
+import org.slf4j.Logger;
 
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
 import javax.ws.rs.ext.Provider;
 
-import org.fcrepo.kernel.api.exception.InvalidPrefixException;
-
-import org.slf4j.Logger;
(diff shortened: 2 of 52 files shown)


fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/ExternalContentAccessExceptionMapper.java
```diff
@@ -45,7 +45,7 @@
 
     @Override
     public Response toResponse(final ExternalContentAccessException exception) {
-        debugException(this, exception, LOGGER);
```
awoods, line 48: This class no longer really "implements" `ExceptionDebugLogging`. Should we systematically refactor the use of `ExceptionDebugLogging`?

It seems like this PR does a partial job of that.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/InvalidPrefixExceptionMapper.java
```diff
@@ -44,7 +43,6 @@
 
     @Override
     public Response toResponse(final InvalidPrefixException e) {
-        LOGGER.error("FedoraInvalidPrefixExceptionMapper caught an exception: {}", e.getMessage());
```
awoods, line 47: You seem to have taken a different approach with this update than the main thrust of this PR. Was that intentional?

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/ItemNotFoundExceptionMapper.java
```diff
@@ -40,8 +41,6 @@
 
     @Override
     public Response toResponse(final ItemNotFoundException e) {
-
-        LOGGER.debug("Exception intercepted by ItemNotFoundExceptionMapper: {}\n", e.getMessage());
```
awoods, line 44: Is there a rationale for which exception-mappers now have `warn` exceptions instead of calling `debugException`, and which mappers simply remove the current `debug` logging?

awoods (changes requested):

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/ExternalContentAccessExceptionMapper.java
pwinckles: That did occur to me. I'm happy to remove the ExceptionDebugLogging from these classes. I decided to leave it so that all of the mappers were the same, and in case someone wanted to use that function in the future, but I have no problem removing it, if that's your vote.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/InvalidPrefixExceptionMapper.java
pwinckles: Yeah, I removed it because it seemed like a client exception rather than a server exception. The approach I had was that client exceptions should be logged at debug and server exceptions should be warn (or higher).

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/exceptionhandlers/ItemNotFoundExceptionMapper.java
pwinckles: Yep, see my previous comment for the rational. Additionally, I removed some duplicate debug logs, like the one commented on here. There's no reason to both have a debug call in this class and then call debugException as well.

awoods (approved):

PR #1763, opened 2020-10-05
get fedora working against s3

pwinckles:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3436

Allows you to configure Fedora to write directly to S3 rather than the filesystem.

Note: This integration seems VERY SLOW. I have not done any investigation into the root cause, but I suspect it's a combination of inefficiencies in Fedora (eg repeatedly reading the resource headers) and inefficiencies in the ocfl-java S3 implementation, which works but is not as optimized as it could be.

Another note, ocfl-java only supports H2 and Postgres, and really the H2 support only exists for testing. The database is necessary to work around eventual consistency issues. After the troubles I've seen here when working with MySQL, I'm less inclined to add support for additional databases...

# How should this be tested?

1. Create a AWS user and bucket with the [necessary permissions](https://github.com/UW-Madison-Library/ocfl-java#amazon-s3).
2. [Setup your credentials](https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/setup-credentials.html)
3. Start Fedora, something along these lines: `mvn -Dfcrepo.storage=ocfl-s3 -Dfcrepo.aws.region=us-east-2 -Dfcrepo.ocfl.s3.bucket=pwinckles-ocfl -Dfcrepo.ocfl.s3.prefix=fcrepo-1 clean jetty:run`
4. Use Fedora normally

[rocfl](https://github.com/pwinckles/rocfl/) supports S3. For example, `rocfl -b pwinckles-ocfl -R us-east-2 -r fcrepo-1 ls -l`

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
@@ -27,6 +27,7 @@ import javax.annotation.PostConstruct;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Objects;
 
 /**
  * Fedora's OCFL related configuration properties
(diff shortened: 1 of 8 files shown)


awoods (approved):
Looks good. There should likely be a follow-on ticket for adding wiki documentation for using the feature.

pwinckles:
> 
> 
> Looks good. There should likely be a follow-on ticket for adding wiki documentation for using the feature.

https://jira.lyrasis.org/browse/FCREPO-3442

PR #1764, opened 2020-10-06
fixed exceptions

pwinckles:
**JIRA Ticket**: none

Fixes a couple of exceptions with changed signatures.

# How should this be tested?

Should build

# Interested parties
@fcrepo/committers


whikloj:
I wonder if we should remove the exception only constructor while in here and leave a note as to why?

pwinckles:
> 
> 
> I wonder if we should remove the exception only constructor while in here and leave a note as to why?

I removed that constructor yesterday from all custom exceptions

bbpennel (approved):

PR #1765, opened 2020-10-06
Make ghost nodes immutable - FCREPO-3256

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3256

Makes it so you can't create a real resource at the address of a ghost node.

# How should this be tested?

Testing steps in the ticket or
1. Create a resource `a/b/c` - `curl -i -ufedoraAdmin:fedoraAdmin -XPUT http://localhost:8080/rest/a/b/c`
1. Try to create resource `a/b` - `curl -i -ufedoraAdmin:fedoraAdmin -XPUT http://localhost:8080/rest/a/b`
    Without this PR get a `201 Created`, with this PR get a `409 Conflict`.
1. Delete resource `a/b/c` - `curl -i -ufedoraAdmin:fedoraAdmin -XDELETE http://localhost:8080/rest/a/b/c`
1. With the PR, you still can't create at `a/b` due to the tombstone. Running step 2 still returns a 409 Conflict.
1. Delete the `a/b/c/fcr:tombstone` - `curl -i -ufedoraAdmin:fedoraAdmin -XDELETE http://localhost:8080/rest/a/b/c/fcr:tombstone`
1. Now you can create a resource at `a/b` - `curl -i -ufedoraAdmin:fedoraAdmin -XPUT http://localhost:8080/rest/a/b`

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -408,7 +408,7 @@ public class FedoraLdp extends ContentExposingResource {
         }
 
         if (isGhostNode(transaction(), fedoraId)) {
-            throw new GhostNodeException("Resource path " + externalPath() + " is an immutable node.");
+            throw new GhostNodeException("Resource path " + externalPath() + " is an immutable resource.");
         }
 
         // TODO: Refactor to check preconditions
@@ -808,7 +808,7 @@ public class FedoraLdp extends ContentExposingResource {
         final FedoraId fullTestPath = fedoraId.resolve(pid);
 
         if (doesResourceExist(transaction(), fullTestPath) || isGhostNode(transaction(), fullTestPath)) {
-            LOGGER.debug("Resource with path {} already exists or is ghost node; minting new path instead",
+            LOGGER.debug("Resource with path {} already exists or is an immutable resource; minting new path instead",
                     fullTestPath);
             return mintNewPid(fedoraId, null);
         }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -96,6 +96,7 @@ import org.fcrepo.http.commons.domain.PATCH;
 import org.fcrepo.kernel.api.FedoraTypes;
 import org.fcrepo.kernel.api.exception.AccessDeniedException;
 import org.fcrepo.kernel.api.exception.CannotCreateResourceException;
+import org.fcrepo.kernel.api.exception.GhostNodeException;
 import org.fcrepo.kernel.api.exception.InvalidChecksumException;
 import org.fcrepo.kernel.api.exception.MalformedRdfException;
 import org.fcrepo.kernel.api.exception.MementoDatetimeFormatException;
@@ -404,6 +405,10 @@ public class FedoraLdp extends ContentExposingResource {
             //    throw new InteractionModelViolationException("Changing the interaction model " + resInteractionModel
             //            + " to " + interactionModel + " is not allowed!");
             //}
+        }
+
+        if (isGhostNode(transaction(), fedoraId)) {
+            throw new GhostNodeException("Resource path " + externalPath() + " is an immutable node.");
         }
 
         // TODO: Refactor to check preconditions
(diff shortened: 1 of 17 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -404,6 +405,10 @@ public Response createOrReplaceObjectRdf(
             //}
         }
 
+        if (isGhostNode(transaction(), fedoraId)) {
+            throw new GhostNodeException("Resource path " + externalPath() + " is an immutable node.");
```
awoods, line 409: Maybe change "immutable node" to "immutable resource"?

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
```diff
@@ -112,4 +112,17 @@ public FedoraResource getResource(final String transactionId, final FedoraId fed
...
+     * of the URI of another resource? For example:
+     *
+     * http://localhost/rest/a/b - does exist
+     * http://localhost/rest/a - does not exist and is therefore a ghost node.
+     *
+     * @param transaction The transaction
+     * @param resourceId Identifier of the resource
+     * @return Whether the resource does not exist, but has
+     */
+    public boolean isGhostNode(final Transaction transaction, final FedoraId resourceId);
```
awoods, line 127: This seems more like a method that would be on `FedoraResource` than `ResourceFactory`.

..such as `isAcl()` and `isMemento()`.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/ContainmentIndex.java
```diff
@@ -124,4 +124,11 @@
      */
     void reset();
 
+    /**
+     * Find whether there are any resources that starts with the ID provided.
+     * @param txID The transaction id, or null if no transaction.
+     * @param fedoraId The ID to use to look for other IDs.
+     * @return Are there any matching IDs.
+     */
+    boolean hasResourcesStartingWith(final String txID, final FedoraId fedoraId);
```
awoods, line 133: Maybe as a separate clean-up task, this interface should normalize on using `ID` or `Id`. Currently it is a complete mix.

awoods (changes requested):

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
whikloj: The use of `isGhostNode` in the mintNewPid, seems to make adding this method to `FedoraResource` make less sense. Alternatively I can move the actual check down to the `AbstractService` which also has access to the `ContainmentIndex`.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/ContainmentIndex.java
whikloj: I'll clean it up now.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/models/ResourceFactory.java
whikloj: Actually the use of `isGhostNode` in the `mintNewPid` means this needs to exist here, because in that case we don't fail we just mint a new pid. So I think the current location is best and the use of `ResourceFactory` is just because we don't Inject the `ContainmentIndex` to the Http layer.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/ContainmentIndex.java
```diff
@@ -105,23 +105,30 @@
     /**
      * Check if the resourceID exists in the containment index. Which should mean it exists.
      *
-     * @param txID The transaction id, or null if no transaction
+     * @param txId The transaction id, or null if no transaction
      * @param fedoraID The resource's FedoraId.
      * @return True if it is in the index.
      */
-    boolean resourceExists(final String txID, final FedoraId fedoraID);
+    boolean resourceExists(final String txId, final FedoraId fedoraID);
```
awoods, line 112: Very minor: one last `fedoraID`

whikloj: Doh 🤦

awoods (approved):

PR #1772, opened 2020-10-11
configured to use resource headers cache

pwinckles:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3453

This PR depends on this PR https://github.com/fcrepo-exts/fcrepo-storage-ocfl/pull/11, and it will not build until the other PR has been committed.

Configures Fedora to cache resource headers by default. Throughout the lifetime of a request a resource's headers are read repeatedly. This cache is intended to reduce some of this overhead at the cost of memory. This is particularly important when the storage layer is in S3.

# How should this be tested?

There should be no behavior changes in Fedora. If you run Fedora with either of the following flags you will be able to see the cache in action: `-Dfcrepo.log.storage=trace -Dfcrepo.metrics.enable=true`.

The first should produce caching related log messages, and the second enables metrics collection. The cache metrics can be inspected if you configure [metrics collection](https://wiki.lyrasis.org/display/FEDORA6x/Metrics).

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-auth-common/src/test/resources/application.properties
+++ b/fcrepo-auth-common/src/test/resources/application.properties
@@ -1,2 +1,3 @@
 # Set the Fedora home location for tests
 fcrepo.home=target/fcrepo-home
+fcrepo.resource-header-cache.enable=false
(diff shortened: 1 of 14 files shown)


whikloj (approved):
I tested this and the `fcrepo-storage-ocfl` changes. I tried running a simple JMeter container creation for 6 minutes with cache enabled and disabled. 

There isn't much of a difference overall but the `readHeaders` average had a fastest time of 111 pico-seconds without the cache and a slowest time of 103 pico-seconds with the cache.

Screen Shot 2020-10-11 at 2 50 51 PM


So I'm good with the changes here.

PR #1773, opened 2020-10-11
Reduce number of queries and make updates batched in SearchIndex

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3455

Reduces the number of `for` loops with individual queries or inserts in favour of single batchs of multiple records.

# How should this be tested?

The ol' @awoods batch ingest test would be best.

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
+++ b/fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
@@ -125,8 +125,21 @@ public class DbSearchIndexImpl implements SearchIndex {
     /*
      * Insert a new RDF type into the RDF type table.
      */
-    private static final String INSERT_RDF_TYPE = "INSERT INTO " + SEARCH_RDF_TYPE_TABLE + " (" + RDF_TYPE_URI_PARAM +
-            ") VALUES (:rdf_type_uri)";
+    private static final String INSERT_RDF_TYPE_MYSQLMARIA = "INSERT IGNORE INTO " + SEARCH_RDF_TYPE_TABLE + " (" +
+            RDF_TYPE_URI_PARAM + ") VALUES (:rdf_type_uri)";
+
+    private static final String INSERT_RDF_TYPE_POSTGRES = "INSERT INTO " + SEARCH_RDF_TYPE_TABLE + " (" +
+            RDF_TYPE_URI_PARAM + ") VALUES (:rdf_type_uri) ON CONFLICT DO NOTHING";
+
+    private static final String INSERT_RDF_TYPE_H2 = "MERGE INTO " + SEARCH_RDF_TYPE_TABLE + " (" +
+            RDF_TYPE_URI_PARAM + ") KEY (" + RDF_TYPE_URI_PARAM + ") VALUES (:rdf_type_uri)";
+
+    private static final Map<DbPlatform, String> INSERT_RDF_TYPE = Map.of(
+            MYSQL, INSERT_RDF_TYPE_MYSQLMARIA,
+            MARIADB, INSERT_RDF_TYPE_MYSQLMARIA,
+            POSTGRESQL, INSERT_RDF_TYPE_POSTGRES,
+            H2, INSERT_RDF_TYPE_H2
+    );
 
     @Inject
     private DataSource dataSource;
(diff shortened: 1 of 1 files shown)


whikloj:
I should note I stole some lines from @awoods's [PR](https://github.com/fcrepo/fcrepo/pull/1771) ([1](https://github.com/fcrepo/fcrepo/pull/1773/files#diff-7d435ff5d6f51b1cc53f74079ae3badcR136) and [2](https://github.com/fcrepo/fcrepo/pull/1773/files#diff-7d435ff5d6f51b1cc53f74079ae3badcR161-R163))

whikloj:
Actually hold on to this. I'm still seeing some early RepositoryRuntimeExceptions, wondering if this is a thread-safe issue.

whikloj:
Ok so apparently `main` throws an Exception at the start of a JMeter run and this PR still throws the exact same Exception.

Not sure why but my guess is 2 requests being handled, neither sees the other and so they both try to add a resource. Seemingly one fails to insert but then would query the ID that the other inserted.

Not sure how we handle that or if we need to, so long as one has inserted the rdf type then the other should query the ID out of the database.

This does seem to speed up the `addUpdateIndex` method though. This PR is on the left and current main is on the right for 6 minutes of 2 thread container creation.

Screen Shot 2020-10-11 at 11 14 18 PM

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -346,33 +363,69 @@ private void deleteRdfTypeAssociations(final Long resourceId) {
...
+            final List<MapSqlParameterSource> parameterSourcesList = new ArrayList<>();
+            missingUris.forEach(u -> {
+                final var assocParams = new MapSqlParameterSource();
+                assocParams.addValue(RDF_TYPE_URI_PARAM, u);
+                LOGGER.debug("Adding rdf type uri: " + u);
+                parameterSourcesList.add(assocParams);
+            });
+            final MapSqlParameterSource[] psArray = parameterSourcesList.toArray(new MapSqlParameterSource[0]);
+            // Batch insert all the records.
+            jdbcTemplate.batchUpdate(INSERT_RDF_TYPE, psArray);
```
pwinckles, line 389: As you noted, this can fail if multiple threads attempt to concurrently add the same types. Does it not throw an exception in that case? If so, I think you could safely swallow it and proceed because you'll get a failure on line 394 if the types weren't actually created.

pwinckles: What happens if there are two concurrent requests, one is inserting types A, B, and C and the other is inserting B, C, and D? Since it's a batch update, will the end result be that the database either contains A, B, and C or B, C, and D? Or will it contain all of the types?

whikloj: It looks like we need to go back and use the standard jdbc batch which should return a BatchUpdateException and return success and failure counts. I'll give this a try next

whikloj:
Ok went to using the database specific way of Ignoring conflicts, except H2 which has no way to do that. In which case I basically just update the existing record with the same URI.

No more multi-threaded exception appearing so this also handles https://jira.lyrasis.org/browse/FCREPO-3449

Screen Shot 2020-10-12 at 4 32 35 PM

Showing `addUpdateIndex` p99 and average for 6 minutes each with MySQL, Mariadb, Postgres and then H2

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -346,33 +376,69 @@ private void deleteRdfTypeAssociations(final Long resourceId) {
...
+
+        final List<RdfType> results = jdbcTemplate.query(SELECT_RDF_TYPE_ID,
+                Map.of(RDF_TYPE_URI_PARAM, rdfTypes_str), RDF_TYPE_ROW_MAPPER);
+        // List of existing type ids.
+        final var rdfTypeIds = results.stream().map(RdfType::getTypeId).collect(Collectors.toList());
+        // List of existing type uris.
+        final var rdfTypeUris = results.stream().map(RdfType::getTypeUri).collect(Collectors.toList());
+        // Type uris that don't already have a record. Needs to be a set to avoid inserting the same URI and
+        final var missingUris = rdfTypes_str.stream().filter(t -> !rdfTypeUris.contains(t))
+                .collect(Collectors.toSet());
```
pwinckles, line 390: Doesn't matter that much because I suspect we're only ever dealing with a handful of types, but this can be optimized a little:

```java
final var rdfTypeIds = new ArrayList<Long>(rdfTypes.size());
final var rdfTypeUris = new HashSet<String>();

for (final var type : results) {
    rdfTypeIds.add(type.getTypeId());
    rdfTypeUris.add(type.getTypeUri());
}

final var missingUris = rdfTypes.stream().map(URI::toString).filter(t -> !rdfTypeUris.contains(t)).collect(Collectors.toSet());
```

pwinckles (commented):
Looks great! Thanks for fighting the dbs!

pwinckles (approved):

PR #1778, opened 2020-10-14
Reflect autoversioning state in HTML UI

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3446

If autoversioning is enabled (as it is by default) then the "Create Version" button is replaced by the text "Autoversioning is enabled. Versions are created with each update to a resource." 

Versioning of binaries is still disabled from the HTML UI, though as binary and binary description versioning is now tied together we might be able to enable this.

Also includes a little CSS change as long binary checksum were breaking the HTML.
Screen Shot 2020-10-14 at 11 13 24 AM

# How should this be tested?

Build this branch and run it. Make a container and see the new message.
Stop Fedora
Run again with `-Dfcrepo.autoversioning.enabled=false` and see the same object show the "Create Version" button.

The button **only** creates a version if something has changed. We will need to make sure this is clear to users of the HTML UI.

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-http-api/pom.xml
+++ b/fcrepo-http-api/pom.xml
@@ -14,6 +14,7 @@
 
   <properties>
     <osgi.import.packages>
+        org.fcrepo.config.*,
         org.fcrepo.http.commons.*,
         org.fcrepo.kernel.api.*,
         org.fcrepo.search.api.*,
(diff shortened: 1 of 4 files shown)


pwinckles (dismissed):
Works. A couple of notes:

1. The UI caches that autoversioning is enabled, so it takes some hard refreshes to show otherwise after changing the settings. This is annoying, but the caching is a bigger problem when it does not show newly added children without a hard refresh.
2. The "delete version" button still exists, which is currently not a feature that will be supported in F6.

whikloj:
When you say "change settings" you mean stopping the server and restarting it correct?

pwinckles:
> 
> 
> When you say "change settings" you mean stopping the server and restarting it correct?

Yeah. I don't think you need to do anything about it here.

whikloj:
Yeah, nothing that can be done. Some browsers (<ahem> Chrome) are much more vigilant at using the browser cache. A shift-refresh should force them to check the server again. I've removed the Delete button though

pwinckles (approved):

PR #1780, opened 2020-10-15
Add an additional index if the lc_collate is not C

whikloj:
**JIRA Ticket**: n/a

Tries to add an additional index for Postgres if your LC_TYPE is not C so that LIKE queries can use this index.

# How should this be tested?

Build and run and watch that the `hasResourcesStartingWith` does not slow incrementally with database size.

# Interested parties
@fcrepo/committers (@awoods)


awoods (approved):
It would be nice to have similar optimizations for MariaDB and MySQL...

whikloj:
It is unclear if Maria and MySQL need those optimizations. We probably need to run the same test against all 3 and see how they compare.

PR #1782, opened 2020-10-19
Reload resource after a PATCH request, enforce request preconditions.

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3223

Another iteration on enabling FedoraLdpIT tests.

# How should this be tested?

This is for internal tests.

Preconditions on a PUT have been enabled, and literal whitespaces are translated to %20 in FedoraId to ensure our IRIs are valid.

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -797,6 +797,11 @@ public class FedoraLdp extends ContentExposingResource {
     private FedoraId mintNewPid(final FedoraId fedoraId, final String slug) {
         final String pid;
 
+        if (!isBlank(slug) && slug.contains(" ")) {
+            // Can't use a literal space
+            throw new CannotCreateResourceException("Slug cannot include literal spaces");
+        }
+
         if (!isBlank(slug)) {
             pid = slug;
         } else if (pidMinter != null) {
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -407,14 +407,12 @@ public class FedoraLdp extends ContentExposingResource {
                 throw new InteractionModelViolationException("Changing the interaction model " + resInteractionModel
                         + " to " + interactionModel + " is not allowed!");
             }
+            evaluateRequestPreconditions(request, servletResponse, resource(), transaction);
         }
 
         if (isGhostNode(transaction(), fedoraId)) {
             throw new GhostNodeException("Resource path " + externalPath() + " is an immutable resource.");
         }
-
-        // TODO: Refactor to check preconditions
-        //evaluateRequestPreconditions(request, servletResponse, resource, transaction);
 
         final var providedContentType = getSimpleContentType(requestContentType);
 
@@ -529,7 +527,7 @@ public class FedoraLdp extends ContentExposingResource {
             patchResourcewithSparql(resource(), newRequest);
             transaction.commitIfShortLived();
 
-            addCacheControlHeaders(servletResponse, resource(), transaction);
+            addCacheControlHeaders(servletResponse, reloadResource(), transaction);
 
             return noContent().build();
         } catch (final IllegalArgumentException iae) {
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -797,11 +797,6 @@ public class FedoraLdp extends ContentExposingResource {
     private FedoraId mintNewPid(final FedoraId fedoraId, final String slug) {
         final String pid;
 
-        if (!isBlank(slug) && slug.contains(" ")) {
-            // Can't use a literal space
-            throw new CannotCreateResourceException("Slug cannot include literal spaces");
-        }
-
         if (!isBlank(slug)) {
             pid = slug;
         } else if (pidMinter != null) {
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/identifiers/FedoraId.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/identifiers/FedoraId.java
@@ -20,6 +20,7 @@ package org.fcrepo.kernel.api.identifiers;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 import org.apache.commons.lang3.StringUtils;
+import org.fcrepo.kernel.api.exception.IdentifierConversionException;
 import org.fcrepo.kernel.api.exception.InvalidMementoPathException;
 import org.fcrepo.kernel.api.exception.InvalidResourceIdentifierException;
 
@@ -87,7 +88,7 @@ public class FedoraId {
      * @throws IllegalArgumentException If ID does not start with expected prefix.
      */
     private FedoraId(final String fullId) {
-        this.fullId = ensurePrefix(fullId).replaceAll("\\s", "%20").replaceAll("/+$", "");
+        this.fullId = ensurePrefix(fullId).replaceAll("/+$", "");
         // Carry the path of the request for any exceptions.
         this.fullPath = this.fullId.substring(FEDORA_ID_PREFIX.length());
         checkForInvalidPath();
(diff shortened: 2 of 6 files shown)


pwinckles:
I'm not sure what I'm doing wrong, but I can't get it to work. When I do:

`curl -u fedoraAdmin:fedoraAdmin 'http://localhost:8080/rest/with some spaces' -H "Link: <http://www.w3.org/ns/ldp#NonRDFSource>; rel=type" -v -H "Content-Type: text/plain" -X PUT --data-binary "testing"`

I get back:

`<h1>Bad Message 400</h1><pre>reason: Illegal character SPACE=' '</pre>`

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/identifiers/FedoraId.java
```diff
@@ -87,7 +87,7 @@
      * @throws IllegalArgumentException If ID does not start with expected prefix.
      */
     private FedoraId(final String fullId) {
-        this.fullId = ensurePrefix(fullId).replaceAll("/+$", "");
+        this.fullId = ensurePrefix(fullId).replaceAll("\\s", "%20").replaceAll("/+$", "");
```
pwinckles, line 90: Perhaps there can be a FedoarId test for this behavior?

whikloj:
The test is doing a POST, this seems to be caught by Jersey if you try to use a space in the URI, whereas the Slug it doesn't care.
```
> curl -i -ufedoraAdmin:fedoraAdmin -XPOST -H"Slug: spaces 2" "http://localhost:8080/rest"
HTTP/1.1 201 Created
Date: Wed, 21 Oct 2020 14:50:47 GMT
Set-Cookie: JSESSIONID=node0sf37nxmy000j1cj11m81c6gdw0.node0; Path=/
Set-Cookie: rememberMe=deleteMe; Path=/; Max-Age=0; Expires=Tue, 20-Oct-2020 14:50:47 GMT
ETag: W/"31F065CBCB64E94AA2681035DF7A4C03"
X-State-Token: 31F065CBCB64E94AA2681035DF7A4C03
Last-Modified: Wed, 21 Oct 2020 14:50:47 GMT
Link: <http://localhost:8080/rest/spaces%202>; rel="timegate"
Link: <http://localhost:8080/rest/spaces%202>; rel="original"
Link: <http://localhost:8080/rest/spaces%202/fcr:versions>; rel="timemap"
Link: <http://www.w3.org/ns/ldp#BasicContainer>; rel="type"
Link: <http://www.w3.org/ns/ldp#Resource>; rel="type"
Link: <http://fedora.info/definitions/v4/repository#Resource>; rel="type"
Link: <http://mementoweb.org/ns#OriginalResource>; rel="type"
Link: <http://mementoweb.org/ns#TimeGate>; rel="type"
Link: <http://www.w3.org/ns/ldp#Container>; rel="type"
Link: <http://fedora.info/definitions/v4/repository#Container>; rel="type"
Link: <http://localhost:8080/rest/spaces%202/fcr:acl>; rel="acl"
Location: http://localhost:8080/rest/spaces%202
Content-Type: text/plain
Content-Length: 37
Server: Jetty(9.4.24.v20191120)

http://localhost:8080/rest/spaces%202

curl -i -ufedoraAdmin:fedoraAdmin -XPATCH -H"Content-type: application/sparql-update" -d "INSERT { <> <http://purl.org/dc/elements/1.1/title> \"test\" } WHERE { }" http://localhost:8080/rest/spaces%202
HTTP/1.1 400 Bad Request
Date: Wed, 21 Oct 2020 14:52:03 GMT
Set-Cookie: JSESSIONID=node0aize2n8mb19m1r9ahtax4r63c1.node0; Path=/
Set-Cookie: rememberMe=deleteMe; Path=/; Max-Age=0; Expires=Tue, 20-Oct-2020 14:52:03 GMT
Content-Length: 622
Server: Jetty(9.4.24.v20191120)

Encountered " "<" "< "" at line 2, column 3.
Was expecting one of:
    <IRIref> ...
    <PNAME_NS> ...
```

whikloj:
Ok this is altered to now reject literal spaces in a `Slug` with a 400 Bad Request error. I did this at the HTTP layer but also had the `FedoraId` throw an exception if it encountered an id with a space.

pwinckles:
I don't think this is working quite right yet.

```
curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest -H "Link: <http://www.w3.org/ns/ldp#NonRDFSource>; rel=type" -v -H "Content-Type: text/plain" -X POST -H"Slug: has%20spaces%201" --data-binary "testing"
curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/has%20spaces%201
```

The GET fails, but it shouldn't

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -799,6 +797,11 @@ private void handleRequestDisallowedOnMemento() {
     private FedoraId mintNewPid(final FedoraId fedoraId, final String slug) {
         final String pid;
 
+        if (!isBlank(slug) && slug.contains(" ")) {
+            // Can't use a literal space
+            throw new CannotCreateResourceException("Slug cannot include literal spaces");
```
pwinckles, line 802: Why are spaces handled differently from any other character that might otherwise be url encoded? For example, is it okay if a slug contains a literal tab character? If so, what is it about spaces that requires special handling? (I'm not saying it doesn't require special handling, just trying to understand.)

Generally speaking, slugs are treated as string literals, yes? So, if I create an object with a slug of `%D0%BE%D0%B1%D1%8A%D0%B5%D0%BA%D1%82` is that the object's literal id or is it `объект`?

pwinckles (commented):
I'm still getting a case where it's not behaving as expected:

```
curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest -H "Link: <http://www.w3.org/ns/ldp#NonRDFSource>; rel=type" -v -H "Content-Type: text/plain" -X POST -H"Slug: has+spaces+2" --data-binary "testing"

curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/has%20spaces%202
curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/has+spaces+2
```

As far as I can tell, that object is essentially inaccessible.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
whikloj: Sorry I misunderstood and misspoke. My understanding is we allow `-H"Slug: has%20spaces"` but we don't allow `-H:"Slug: has spaces"` because literal spaces are not allowed as part of a valid URI as defined in [RFC-3986](https://tools.ietf.org/html/rfc3986)

whikloj: I think what you are asking is "are we using the URL encoded or decoded form of the path". We are using the decoded form with the exception of spaces. We can move to using the encoded form instead, but currently Jersey decodes on the way in for us. Then we need to work out how to generate a URI without double encoding the path.

pwinckles: I don't see where https://tools.ietf.org/html/rfc3986 says you can't have spaces in URIs.

It would seem to me that a slug needs to be treated either as a string literal or as a url encoded value. It's odd to me that only spaces are required to be encoded. If Jersey is treating it as a url encoded value and it is decoding it before injecting the value in FedoraLdp, wouldn't there always be a " " in a slug that contains `+` or `%20`?

whikloj: My reading is the section 2.2 specifies what characters have a reserved purpose in a URI.
> reserved    = gen-delims / sub-delims
>
> gen-delims  = ":" / "/" / "?" / "#" / "[" / "]" / "@"
>
> sub-delims  = "!" / "$" / "&" / "'" / "(" / ")" / "*" / "+" / "," / ";" / "="

and section 2.3 specifies which characters have an unreserved purpose in a URI.
> Characters that are allowed in a URI but do not have a reserved
   purpose are called unreserved.  These include uppercase and lowercase
   letters, decimal digits, hyphen, period, underscore, and tilde.
> 
> unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"

Spaces are not defined in either of those sections and so have no purpose in a URI.

Jersey automatically (I think) handles it when you try to add spaces to the URI you are attempting to access, as you [encountered above](https://github.com/fcrepo/fcrepo/pull/1782#issuecomment-713555728). As `Slug`s are passed in as headers which do allow spaces but they are to meant be part of the URI so we enforce this requirement to have the end-user percent encode their spaces.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/identifiers/FedoraId.java
whikloj: I have reverted this change for a later attempt.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
whikloj: I have reverted this change for a later attempt.

pwinckles (approved):

awoods (approved):

PR #1783, opened 2020-10-23
Parallelize index rebuilding

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3503

Splits the reindexing of OCFL objects into individual tasks which can be shared amongst threads to speed up reindexing.

This is all still done in a single "transaction" for each service which is then committed at the end.

The unit tests for the manager and service might be too weak so I am open to adding more.

# How should this be tested?

Create a repo with a bunch of objects, remove the ocflIndex database. Restart Fedora and see how long it takes to finish reindexing with and before this PR.

# Interested parties
@fcrepo/committers


pwinckles:
The problem with this approach is that it will flood the executor service with tasks. So, if you take @awoods's repo as a for instance, there would be 15 million (or however big that repository is) tasks queued in memory waiting to be processed.

In this case, I think we can be more memory efficient if we just spin up N worker threads and have these threads ask the manager for work to do rather than queuing up tasks in an executor service.

So, it would look something like this:

```java

class Worker extends Thread {

  Manager manager;
  Indexer indexer;

  void run() {
    while (true) {
      var ids = manager.getIds();
      if (ids.isEmpty()) {
        break;
      }
      ids.forEach(id -> {
         indexer.index(id);
      });
      // perhaps collect successes and failures and report them back the manager
    }
  }
}

class Manager {

  // you'll probably want to create these a constructor
  List<Worker> workers;
  Iterator<String> ocflIdIterator;

  void index() {
    workers.forEach(Thread:start);
    // you could also spin up a thread to start counting objects here, but you'd need to use a different iterator
    workers.forEach(Thread:join);
  }

  synchonized List<String> getIds() {
    // it would be best to batch the ids that are given to the workers so they don't need to keep on asking
    // if you wanted to be really fancy, you could prefetch N (100, 200, 1000, it depends on how many ids you give out with each call to this method) ids and buffer them in the manager, so that the worker does
   //  not need to wait, or you could just pull of some here. if you do decide to prefetch, one way to do this is to have
   // the manager do this rather than going straight to blocking on the join. it could work populating a buffer till it was full
   // and then block till getIds() notifies it that some ids have been taken....
  }
}
```

Memory issues aside, this PR also has a change in behavior in that it does not fail on error, unless I'm mistaken. Is this the behavior we want?

pwinckles:
That said, you can still use an executor service and tasks. If you want to do this, perhaps the manager can create tasks of N object ids to process, submit some number of them to the executor service, and then wait until they start to complete before submitting more. It's okay to submit more tasks than you have worker threads, but ideal the queue would be kept to a reasonable size. This should be more compatible with your existing code.

whikloj:
Ok, thanks for the feedback. I'll give it another try.

pwinckles:
@whikloj I did some back of the envelop calculations on how much memory might be used to store all of those tasks in memory. I expect that it'd be between 0.5 and 1GB for @awoods's repository. So, it's not catastrophic. Because the IDs are streamed from ocfl-java it is possible to implement this using a lot less memory. However, I'm comfortable going along with the general approach that you have here and later we can put some back pressure on the filling of the task queue as needed.

Let me know if this is the direction you want to go, and, if so, I'll leave a couple of comments on the code.

whikloj:
No I think you're right that moving to a subscriber thread type setup is better. I can see how it works in pseudo code. Just have to see if I can write it. If you'd rather you can always go ahead and I'll just play with this for my own education.

pwinckles:
> No I think you're right that moving to a subscriber thread type setup is better. I can see how it works in pseudo code. Just have to see if I can write it. If you'd rather you can always go ahead and I'll just play with this for my own education.

Sounds good. You don't need to do the fancy pre-fetching of the ids. That should simplify things.

whikloj:
Two questions I have is 1) if I wanted to do the pre-fetching (not sure I will) does that method (ie. `prefetchIds()`) also have to be `synchronized` and 2) the `listObjects()` method returns a `Stream<String>` so can you effectively iterate over that, or are you thinking we then store that to a list which we iterate over?

pwinckles:
> Two questions I have is 1) if I wanted to do the pre-fetching (not sure I will) does that method (ie. `prefetchIds()`) also have to be `synchronized` and 2) the `listObjects()` method returns a `Stream<String>` so can you effectively iterate over that, or are you thinking we then store that to a list which we iterate over?

Answering the second question first, get and store an iterator off the stream. Then, when you fetch ids, you call `next()` on the iterator to get however many ids that you want.

For the first question, it depends on who's calling the method for fetching the ids. If it's only ever called by one thread it doesn't need to be synchonized. For example, the manager, or a separate thread, could feed ids from the id stream into a concurrent collection until it's "full". Then, it could go to sleep for a little bit (not using an actual "sleep" but by using wait and notify). When a worker retrieves a new batch of ids, this method would then notify the thread that's prefetching the ids to wake up and fetch some more ids. I'm sure there are other ways to implement this. Just one idea that comes to mind.

whikloj:
Ok, I hope I did this correctly.

whikloj:
Actually, I left the `ReindexTask` in there and I have another thought about not storing all the result states until the end and just have the work report back with each batch. I'll push up another commit so you can let this sit.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexManager.java
```diff
@@ -0,0 +1,144 @@
...
+
+    private static final Logger LOGGER = getLogger(ReindexManager.class);
+
+    private final String transactionId;
+
+    private final List<ReindexWorker> workers;
+
+    private final Iterator<String> ocflIter;
+
+    private final Map<String, String> resultStates = new HashMap<>();
```
pwinckles, line 46: This seems to be less of a state map and more of a map of errors, the successes have an empty value. It also has the potential to grow very large. How are we planning on using these results? For me, what makes sense is to have a count of the objects that were successfully processed, and either a collection or a map of exceptions that occurred during processing. Or, we could just keep a count of both and immediately log errors so that we do not need to collect them. I think keeping the counts is worth while because it will allow us to do periodic reporting during the processing.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/IndexBuilderImpl.java
```diff
@@ -120,148 +117,40 @@ private void rebuild() {
...
-                return null;
-            });
-
-            execQuietly("Failed to rollback membership index transaction " + txId, () -> {
-                membershipService.rollbackTransaction(txId);
-                return null;
-            });
-            throw e;
+            reindexManager.start();
+            reindexManager.commit();
```
pwinckles, line 129: What happens on failure? Isn't this going to leave the db transaction in place?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexService.java
```diff
@@ -0,0 +1,256 @@
...
+     * Commit the records added from transaction.
+     * @param transactionId the id of the transaction.
+     */
+    public void commit(final String transactionId) {
+        try {
+            LOGGER.debug("Performing commit");
+            containmentIndex.commitTransaction(transactionId);
+            ocflIndex.commit(transactionId);
+            referenceService.commitTransaction(transactionId);
+            indexMembership(transactionId);
```
pwinckles, line 171: Indexing the membership does seem like a distinct operation from committing the db transactions. Perhaps there could be a method for committing the db transactions, one for rolling back the db transactions, and another that kicks of the membership indexing? I haven't looked at it in detail, but does the membership indexing use much of else that's in this class? It the surface it seems like it could easily go into its own class.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexWorker.java
```diff
@@ -0,0 +1,98 @@
...
+    @Override
+    public void run() {
+        while (running) {
+            final List<String> ids = manager.getIds();
+            if (ids.isEmpty()) {
+                stopThread();
+                break;
+            }
+            final Map<String, String> states = new HashMap<>();
+            for (final var id : ids) {
```
pwinckles, line 74: You should have a check in this loop to see if the thread should stop.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexWorker.java
```diff
@@ -0,0 +1,98 @@
...
+                try {
+                    service.indexOcflObject(manager.getTransactionId(), id);
+                    states.put(id, "");
+                } catch (final Exception e) {
+                    states.put(id, e.getMessage());
+                    if (failOnError) {
```
pwinckles, line 80: The problem with this behavior is that it will only ever stop _this_ thread on error. The rest of the threads will proceed. One solution to this problem is to change the status reporting, as previously discussed, and have the `updateComplete` method trigger stopping all of the threads when an error is reported and `failOnError` is true.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexManager.java
```diff
@@ -0,0 +1,144 @@
...
+     */
+    public ReindexManager(final Stream<String> ids, final ReindexService reindexService, final boolean failOnError,
+                          final int batchSize) {
+        this.ocflIter = ids.iterator();
+        this.reindexService = reindexService;
+        this.batchSize = batchSize;
+        transactionId = UUID.randomUUID().toString();
+        workers = new ArrayList<>();
+        final int availableProcessors = Runtime.getRuntime().availableProcessors();
+        final int threads = availableProcessors > 1 ? availableProcessors - 1 : 1;
```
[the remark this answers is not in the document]
whikloj: I can make this configurable but then what should be the default? Single threaded?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/IndexBuilderImpl.java
whikloj: No, [`ReindexManager.commit()`](https://github.com/fcrepo/fcrepo/pull/1783/files#diff-685980790009aa46aa9d033609dbc6da47d661c9f8c251e9b80f6bda4745bdc3R140-R142) calls [`ReindexService.commit()`](https://github.com/fcrepo/fcrepo/pull/1783/files#diff-b834cc79180db3731825fef4c20b6e309d24a06db2379de2799a8c063fdc8204R165) which does the commit and rollback on failure work.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexWorker.java
whikloj: Because each worker has a reference to the ReindexManager, I was just going to add a call to `ReindexManager.stop()` which stops all the threads. Do you see any problems there?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexManager.java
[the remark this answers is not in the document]
whikloj: Also are you thinking this is specifically configurable w/r/t reindexing?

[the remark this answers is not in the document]
pwinckles: You can have the default be exactly what you have here. There are a number of ways that you could write it. Probably the easiest is to write a static function that does the thread computation and then write something like this is in the config:

```java
@Value("${fcrepo.reindex.threads:" + computeDefaultThreads() + "}")
private int reindexThreads;
```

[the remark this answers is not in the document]
pwinckles: > 
> 
> Also are you thinking this is specifically configurable w/r/t reindexing?

Not sure

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/IndexBuilderImpl.java
pwinckles: Right, but what happens if the call to `reindexManager.start()` throws an exception?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexWorker.java
pwinckles: Yes, that should be fine

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/IndexBuilderImpl.java
whikloj: Ok I see what you're saying the DB transaction would just sit there.

pwinckles (commented):
Looks good! Just a couple more minor comments

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexWorker.java
```diff
@@ -0,0 +1,105 @@
...
+            for (final var id : ids) {
+                if (!running) {
+                    break;
+                }
+                try {
+                    service.indexOcflObject(transactionId, id);
+                    completed += 1;
+                } catch (final Exception e) {
+                    errors += 1;
+                    if (failOnError) {
```
pwinckles, line 85: If `failOnError` is not set, I think it might make sense to WARN/ERROR the exception here

whikloj:
1 thread
```
INFO 09:10:21.156 (IndexBuilderImpl) Index rebuild completed 13645 objects successfully and 0 objects had errors in 3 mins, 3 seconds
```
11 threads
```
INFO 09:05:39.707 (IndexBuilderImpl) Index rebuild completed 13645 objects successfully and 0 objects had errors in 1 mins, 13 seconds 
```

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexWorker.java
```diff
@@ -89,6 +94,7 @@ public void run() {
                         service.cleanupSession(transactionId);
                         throw e;
                     }
+                    LOGGER.error("Reindexing of OCFL id {} failed: {}", id, e.getMessage());
```
pwinckles, line 97: Can you change this to:

```java
LOGGER.error("Reindexing of OCFL id {} failed", id, e);
```

It'll make it a lot easier to figure out what went wrong

pwinckles (dismissed):
👍

fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
```diff
@@ -92,12 +92,31 @@
     @Value("${fcrepo.resource-header-cache.expire-after-seconds:600}")
     private long resourceHeadersCacheExpireAfterSeconds;
 
+    @Value("${fcrepo.ocfl.reindex.threads:-1}")
```
awoods, line 95: We will need to either document these configuration elements in the [wiki](https://wiki.lyrasis.org/display/FEDORA6x/Configuration+Options+Inventory) now, or as a follow-on ticket.

awoods (changes requested):

fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
whikloj: Updated
https://wiki.lyrasis.org/display/FEDORA6x/Configuration+Options+Inventory#ConfigurationOptionsInventory-RebuildfromOCFLpersistentstorage

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/IndexBuilderImpl.java
```diff
@@ -114,154 +84,45 @@ public void rebuildIfNecessary() {
...
-
-                // Results are paged, so step through pages until we reach the last one
-                offset += membershipPageSize;
-            } while (numResults == membershipPageSize);
-
-        } catch (final InvalidQueryException e) {
-            throw new RepositoryRuntimeException("Failed to repopulate membership history", e);
-        }
-        membershipService.commitTransaction(txId);
+    public void setMembershipQueryPageSize(final int pageSize) {
```
awoods, line 116: Is this method ever used?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/IndexBuilderImpl.java
```diff
@@ -114,154 +84,45 @@ public void rebuildIfNecessary() {
...
     }
 
     /**
-     * @param pageSize number of results to use when querying for membership producing resources
+     * Change the number of OCFL ids given out to workers at a time.
+     * @param batchSize number of OCFL ids per batch.
      */
-    public void setMembershipQueryPageSize(final int pageSize) {
-        this.membershipPageSize = pageSize;
+    public void setReindexBatchSize(final int batchSize) {
```
awoods, line 124: Is this method ever used?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexWorker.java
```diff
@@ -0,0 +1,111 @@
...
+    private static final Logger LOGGER = getLogger(ReindexWorker.class);
+    private Thread t;
+    private ReindexManager manager;
+    private ReindexService service;
+    private String transactionId;
+    private boolean running = true;
+    private boolean failOnError;
+
+    /**
+     * Basic Constuctor
```
awoods, line 41: typo: 'Constuctor'

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexService.java
```diff
@@ -0,0 +1,290 @@
...
+        membershipService.reset();
+    }
+
+    /**
+     * Commit the records added from transaction.
+     * @param transactionId the id of the transaction.
+     */
+    public void commit(final String transactionId) {
+        try {
+            LOGGER.debug("Performing commit");
```
awoods, line 185: Maybe include the txnId in the log message?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexService.java
```diff
@@ -0,0 +1,290 @@
...
+            containmentIndex.commitTransaction(transactionId);
+            ocflIndex.commit(transactionId);
+            referenceService.commitTransaction(transactionId);
+            LOGGER.debug("Finished commit");
```
awoods, line 189: Maybe include the txnId in the log message?

awoods (changes requested):

awoods (approved):

PR #1784, opened 2020-10-27
Fix travis postgres issues

pwinckles:

Fix the travis postgres build. The build broke because travis decided to include the newly released Postgres 13 in the image by default.

# How should this be tested?

The ci build should pass.

# Interested parties
@fcrepo/committers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,10 +25,10 @@ jobs:
             - postgresql-client-12
       env:
         - DB=postgresql
-        - DB_PORT=5432
+        - DB_PORT=5433
         - ITEST=true
       before_script:
-        - psql -c 'CREATE DATABASE fcrepo;' -U postgres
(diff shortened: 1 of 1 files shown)


whikloj:
Not sure how we want to handle new versions of databases. Are we going to follow the latest version or stick to the one we were building against? Anyways, as an option we can lock the db version down if we do desire. https://docs.travis-ci.com/user/database-setup/#using-a-different-postgresql-version

pwinckles:
> 
> 
> Anyways, as an option we can lock the db version down if we do desire. https://docs.travis-ci.com/user/database-setup/#using-a-different-postgresql-version

We are already using those settings and travis still installed 13.

Since 13 was just released last month, I think it makes sense to continue to test against 12. It's unlikely that anyone is using 13 yet.

awoods (approved):

PR #1786, opened 2020-11-04
Adds support for using a single configuration file to specify fedora …

dbernstein:
…properties as an alternative to command-line properties.

**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3504

With this update you may now  use  the -Dfcrepo.config.file=/path/to/config.properties file  in order to configure fedora as an alternative to using command-line( -Dparam.name=value).   

If  you prefer not to pass the file via the command-line, fedora will look for a file this location:
```<cwd>/fcrepo-home/config/fcrepo.properties```

You can override default values by placing them in the config file.  If you wish  to override values in the config file use the command line.

So to be clear, the precedence is as follows
1. set default values  (defined in code)
2. apply overrides defined in <cwd>/fcrepo-home/config/fcrepo.properties if exists
3. apply overrides defined by -Dfcrepo.config.file  (if file exists)
4. finally apply any -D parameters 

# How should this be tested?
Experiment by adding a properties file via the commandline - ie via the  `-Dfcrepo.config.file=/path/to/config.properties`
for example:  
```
fcrepo.home=file_defined_home
```
Then run fedora: 

```
java -Dfcrepo.config.file=/path/to/config.properties -jar fcrepo-webapp/target/fcrepo-webapp-6.0.0-SNAPSHOT-jetty-console.jar --headless
```
Verify that the directory `file_defined_home` is present in your working directory.

Perform the same test except copy the property file to <cwd>/fcrepo-home/config/fcrepo.properties

Then try overriding that value like this: 
```
java -Dfcrepo.home=overriden_fedora_home  -Dfcrepo.config-file=/path/to/config.properties -jar fcrepo-webapp/target/fcrepo-webapp-6.0.0-SNAPSHOT-jetty-console.jar --headless
```

Verify that the directory `overriden_fedora_home` is present in your working directory.

# Interested parties

### DIFF
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/BasePropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/BasePropsConfig.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to DuraSpace under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information
+ * regarding copyright ownership.
+ *
+ * DuraSpace licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.fcrepo.config;
+
+import org.springframework.context.annotation.PropertySource;
+import org.springframework.context.annotation.PropertySources;
+
+/**
+ * A base class for property configs
+ *
+ * @author dbernstein
+ * @since 6.0.0
+ */
+@PropertySources({
+        @PropertySource(value = FedoraPropsConfig.DEFAULT_FCREPO_CONFIG_FILE_DEFAULT_PROP_SOURCE,
+                ignoreResourceNotFound = true),
+        @PropertySource(value = FedoraPropsConfig.FCREPO_CONFIG_FILE_PROP_SOURCE, ignoreResourceNotFound = true)
+})
+abstract class BasePropsConfig {
+
+    public static final String DEFAULT_FCREPO_CONFIG_FILE_DEFAULT_PROP_SOURCE =
+            "file:fcrepo-home/config/fcrepo.properties";
+    public static final String FCREPO_CONFIG_FILE_PROP_SOURCE = "file:${" + FedoraPropsConfig.FCREPO_CONFIG_FILE + "}";
+
+}
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/BasePropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/BasePropsConfig.java
@@ -28,14 +28,14 @@ import org.springframework.context.annotation.PropertySources;
  * @since 6.0.0
  */
 @PropertySources({
-        @PropertySource(value = FedoraPropsConfig.DEFAULT_FCREPO_CONFIG_FILE_DEFAULT_PROP_SOURCE,
+        @PropertySource(value = BasePropsConfig.DEFAULT_FCREPO_CONFIG_FILE_PROP_SOURCE,
                 ignoreResourceNotFound = true),
-        @PropertySource(value = FedoraPropsConfig.FCREPO_CONFIG_FILE_PROP_SOURCE, ignoreResourceNotFound = true)
+        @PropertySource(value = BasePropsConfig.FCREPO_CONFIG_FILE_PROP_SOURCE, ignoreResourceNotFound = true)
 })
 abstract class BasePropsConfig {
 
-    public static final String DEFAULT_FCREPO_CONFIG_FILE_DEFAULT_PROP_SOURCE =
+    public static final String DEFAULT_FCREPO_CONFIG_FILE_PROP_SOURCE =
             "file:fcrepo-home/config/fcrepo.properties";
-    public static final String FCREPO_CONFIG_FILE_PROP_SOURCE = "file:${" + FedoraPropsConfig.FCREPO_CONFIG_FILE + "}";
+    public static final String FCREPO_CONFIG_FILE_PROP_SOURCE = "file:${fcrepo.config.file}";
 
 }
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/BasePropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/BasePropsConfig.java
@@ -34,8 +34,10 @@ import org.springframework.context.annotation.PropertySources;
 })
 abstract class BasePropsConfig {
 
+    public static final String FCREPO_HOME_PROPERTY = "fcrepo.home";
+    public static final String DEFAULT_FCREPO_HOME_VALUE = "fcrepo-home";
     public static final String DEFAULT_FCREPO_CONFIG_FILE_PROP_SOURCE =
-            "file:fcrepo-home/config/fcrepo.properties";
+            "file:${" + FCREPO_HOME_PROPERTY + ":" + DEFAULT_FCREPO_HOME_VALUE + "}/config/fcrepo.properties";
     public static final String FCREPO_CONFIG_FILE_PROP_SOURCE = "file:${fcrepo.config.file}";
 
 }
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
@@ -22,6 +22,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.PropertySource;
 
 import javax.annotation.PostConstruct;
 import java.io.IOException;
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Comment by pwinckles** [comment]
Haven't finished reviewing this yet, but I think it would be nice if it looked for the config file in a default location that can be overriden. By default, the following directory structure is setup:

```
fcrepo-home/
 \- data/
     |- staging/
     |- ocfl-root/
     \- ocfl-temp/
```

One option that comes to mind is to augment this structure as follows:

```
fcrepo-home/
 |- config/
 |   \- fcrepo.properties
 \- data/
     |- staging/
     |- ocfl-root/
     \- ocfl-temp/
```

Ideally, you should be able to start Fedora without having to use any `-D` arguments.

File: fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java [hunk 1]
```diff
@@ -35,20 +36,54 @@
...
-
+    public static final String FCREPO_JMS_HOST = "fcrepo.jms.host";
+    public static final String FCREPO_DYNAMIC_JMS_PORT = "fcrepo.dynamic.jms.port";
+    public static final String FCREPO_DYNAMIC_STOMP_PORT = "fcrepo.dynamic.stomp.port";
+    public static final String FCREPO_ACTIVEMQ_CONFIGURATION = "fcrepo.activemq.configuration";
+    public static final String FCREPO_NAMESPACE_REGISTRY = "fcrepo.namespace.registry";
+    public static final String FCREPO_EXTERNAL_CONTENT_ALLOWED = "fcrepo.external.content.allowed";
     private static final String DATA_DIR = "data";
+    private static final String ACTIVE_MQ_DIR = "ActiveMQ/kahadb";
+    private static final String FCREPO_ACTIVEMQ_DIRECTORY = "fcrepo.activemq.directory";
```
pwinckles, line 54: These variables are getting confused. Some of them are property names and some of them are default property values. They should be visually separate in the file.

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
```diff
@@ -35,20 +36,54 @@
  * @since 6.0.0
  */
 @Configuration
+@PropertySource (value = "file:${" + FedoraPropsConfig.FCREPO_CONFIG_FILE + "}", ignoreResourceNotFound = true)
 public class FedoraPropsConfig {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(FedoraPropsConfig.class);
 
+    public static final String FCREPO_CONFIG_FILE = "fcrepo.config-file";
```
awoods, line 44: For consistency, can we rename this property to not include a `-`. Maybe `fcrepo.config.file` or `fcrepo.config` or `fcrepo.config.location`... or something else?

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
dbernstein: I will change that up.

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
dbernstein: sure - no problem.

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
```diff
@@ -94,4 +134,66 @@ public void setFedoraData(final Path fedoraData) {
         this.fedoraData = fedoraData;
     }
 
+    /**
+     * @return The JMS host
+     */
+    public String getJmsHost() {
```
awoods, line 140: Of all of the following "getters", only `getVelocityLog()` appears to be used. Should the others be removed?

awoods (changes requested):

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
dbernstein: I believe all of them are referenced in SpEL ( i.e. "#{fcrepoPropsConfig.member}" references in other parts of the codebase.  So not recommended.

fcrepo-configs/src/main/java/org/fcrepo/config/BasePropsConfig.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+@PropertySources({
+        @PropertySource(value = FedoraPropsConfig.DEFAULT_FCREPO_CONFIG_FILE_DEFAULT_PROP_SOURCE,
+                ignoreResourceNotFound = true),
+        @PropertySource(value = FedoraPropsConfig.FCREPO_CONFIG_FILE_PROP_SOURCE, ignoreResourceNotFound = true)
+})
+abstract class BasePropsConfig {
+
+    public static final String DEFAULT_FCREPO_CONFIG_FILE_DEFAULT_PROP_SOURCE =
+            "file:fcrepo-home/config/fcrepo.properties";
```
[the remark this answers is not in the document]
dbernstein: The idea here is that if the user fires up fedora without any configuration file,  fcrepo will create <cwd>/fcrepo-home.  Then the decide they want to override some props so they create  <cwd>/fcrepo-home/config/fcrepo.properties.
No need to change any command-line args.

If they decide they want fcrepo.home somewhere else they can use fcrepo.config.file and point to their new config file which contains a definition  fcrepo.home.   Even if they leave <cwd>/fcrepo-home/config/fcrepo.properties, any properties defined in the new prop file will override those in <cwd>/fcrepo-home/config/fcrepo.properties.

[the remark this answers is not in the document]
dbernstein: I believe there is a problem of a circular dependency. We need to know where to look for the default file.  But if the location of the default file depends on the value of fcrepo.home which may be defined in the property file.

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
```diff
@@ -35,20 +35,60 @@
...
+            ACTIVE_MQ_DIR_DEFAULT_VALUE + "').toAbsolutePath().toString()}}")
+    private String activeMqDirectory;
+
+    @Value("${" + FCREPO_NAMESPACE_REGISTRY + ":classpath:/namespaces.yml}")
+    private String namespaceRegistry;
+
+    @Value("${" + FCREPO_EXTERNAL_CONTENT_ALLOWED + ":#{null}}")
+    private String externalContentAllowed;
+
+    @Value("${fcrepo.session.timeout:180000}")
```
awoods, line 85: Is there a reason why the `fcrepo.session.timeout` has not be extracted into a static constant like the ones above?
Same question for `fcrepo.velocity.runtime.log` below.

awoods:
As a note, the functionality of this PR works as advertised.

fcrepo-configs/src/main/java/org/fcrepo/config/BasePropsConfig.java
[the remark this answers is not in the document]
awoods: When you say:
> The idea here is that if the user fires up fedora without any configuration file, fcrepo will create /fcrepo-home

..you are referring to "firing up" Fedora with either the one-click or jetty:run, no? In both cases, those Fedora's are strictly for short-term testing. Is there a need to have a "default configuration file location" in the context of short-term testing? Is it not just as easy to start Fedora in either of these ways with a system property that points to the desired config file?
```
MAVEN_OPTS="-Dfcrepo.config.file=/tmp/awoods/fcrepo.props"; mvn jetty:run -pl fcrepo-webapp/
```
or
```
java -Dfcrepo.config.file=/tmp/awoods/fcrepo.props -jar fcrepo-webapp/target/fcrepo-webapp-6.0.0-SNAPSHOT-jetty-console.jar --headless
```

I may be missing a use case, but it seems like it may be clearer to remove this default.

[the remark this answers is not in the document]
pwinckles: I believe that @dbernstein added this on my request. I stand by it as a good idea, though I do agree that the Fedora home directory should not be hard coded. In my experience setting up applications, it is normal to override a default application home directory and have it resolve your data, config, etc locations relative to that. In my mind, this is a normal install experience and if you're going require setting one `-D` parameter then it should be `fcrepo.home` and not whatever parameter sets the location of a configuration file. In most cases, the configuration file should be in a known location relative to the application home.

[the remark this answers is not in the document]
awoods: @pwinckles : I like the idea, it just seemed like the default config location should be relative to wherever the user set `fcrepo.home`... I believe is what you are also saying.
If the default config only supports jetty:run and the one-click, I am not sure I see the value.

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
dbernstein: It's just that I don't think they're referenced anywhere else.   I'll make them consistent.

fcrepo-configs/src/main/java/org/fcrepo/config/BasePropsConfig.java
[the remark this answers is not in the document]
dbernstein: @pwinckles : Do you know of a way to get around the circular reference  issue? If so let's keep it in.  But if not, I'm thinking it may be better to remove the default properties file location (<ie <cwd>/fcrepo-home/config) and require the user to specify the location of the config file  using the fcrepo.config.file java option.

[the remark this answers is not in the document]
dbernstein: @pwinckles :  actually - scratch that.  I think I have a good work around.

awoods (approved):

PR #1788, opened 2020-11-05
Memento containment

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3450

Contains the past containment relationships for those things that did exist and are now deleted (but not purged).

**Note**: this does not include the repository rebuild processing. I figured I'd work that once this seemed correct.

**Note note**: I renamed the `resources` table to `containment` and the `transaction_operations` table to `containment_transactions`. The old names were not super clear and we have a lot more tables now.

# How should this be tested?

Create a resource (parent)
Create a child resource (child)
Create a version of parent (I patch it)
Delete child.

Get the memento you created, before there were no ldp:contains, now you should see child.

# Interested parties

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
@@ -35,6 +35,9 @@ import javax.annotation.Nonnull;
 import javax.annotation.PostConstruct;
 import javax.inject.Inject;
 import javax.sql.DataSource;
+
+import java.sql.Timestamp;
+import java.time.Instant;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -45,6 +48,8 @@ import static org.slf4j.LoggerFactory.getLogger;
 
 /**
  * @author peichman
+ * @author whikloj
+ * @since 6.0.0
  */
 @Component("containmentIndexImpl")
 public class ContainmentIndexImpl implements ContainmentIndex {
@@ -56,9 +61,11 @@ public class ContainmentIndexImpl implements ContainmentIndex {
 
     private NamedParameterJdbcTemplate jdbcTemplate;
 
-    public static final String RESOURCES_TABLE = "resources";
+    private DbPlatform dbPlatform;
 
-    private static final String TRANSACTION_OPERATIONS_TABLE = "transaction_operations";
+    public static final String RESOURCES_TABLE = "containment";
+
+    private static final String TRANSACTION_OPERATIONS_TABLE = "containment_transactions";
 
     public static final String FEDORA_ID_COLUMN = "fedora_id";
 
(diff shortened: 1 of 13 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
```diff
@@ -325,21 +377,27 @@ private NamedParameterJdbcTemplate getNamedParameterJdbcTemplate() {
 
     @Override
     public Stream<String> getContains(final String txId, final FedoraId fedoraId) {
-        final String resourceId = fedoraId.getFullId();
+        final String resourceId = fedoraId.isMemento() ? fedoraId.getBaseId() : fedoraId.getFullId();
```
pwinckles, line 380: Would `getResouceId()` not do what you want here?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
```diff
@@ -591,4 +660,18 @@ public void setDataSource(final DataSource dataSource) {
...
+     * issues with memento datetimes.
+     * @param instant
+     * @return the timestamp
+     */
+    private Timestamp formatInstant(final Instant instant) {
+        if (instant == null) {
+            return null;
+        }
+        final var timestamp = Timestamp.from(instant);
+        timestamp.setNanos(0);
```
pwinckles, line 674: Since mementos have second level granularity, perhaps it would make sense to truncate to the second?

pwinckles (commented):
I haven't played around with it yet, but the code looks good. Will try it out shortly

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
whikloj: `getResourceId()` handles ACLs and NonRdfSourceDescriptions, but not Mementos.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
whikloj: Is that not what this is doing? Setting the nanoseconds to 0? Perhaps @bbpennel could help as I stole this from [his code](https://github.com/fcrepo/fcrepo/blob/main/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/MembershipIndexManager.java#L548-L558)

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
pwinckles: But, the output of `getResourceId` can never contain a memento because it constructs an id based on the `baseId`.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
pwinckles: I see. It reads as if it's truncating to milliseconds, but I see now that it's just removing all fractions of a second.

pwinckles (approved):
Tested both atomic and archival group resources and it seems to work

👍

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
whikloj: I understand, I can change this.

pwinckles (approved):

PR #1792, opened 2020-11-05
Don't load a tombstone for ACLs

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3516

Stops trying to load the ACL roles for a tombstone.

# How should this be tested?

1. Create resource
    `curl -i -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/box -XPUT`
1. Delete resource
    `curl -i -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/box -XDELETE`

# Interested parties


whikloj:
Actually hold on to this, I think I have a solution to this and https://jira.lyrasis.org/browse/FCREPO-3516

whikloj:
Ok so this solves both FCREPO-3515 and FCREPO-3516 though it appears that FCREPO-3516 was an expected exception. This removes the stacktrace and instead you get the tombstone message. The tombstone appears regardless of inherited ACLs above the deleted resource because (I think) once the resource is gone it can't report on its container anymore. If this is a concern we can look at it again.

awoods:
When logging is set to `DEBUG` I am still seeing a stracktrace when performing the test detailed above.

awoods (approved):

PR #1798, opened 2020-11-11
big dep upgrade

pwinckles:
**Jira**: https://jira.lyrasis.org/browse/FCREPO-3526

Upgrades fcrepo dependencies to latest versions when possible.

# How should this be tested?

Everything should still work.

# Interested parties

@fcrepo/committers

### DIFF
--- a/fcrepo-auth-webac/src/test/java/org/fcrepo/integration/auth/webac/WebACRecipesIT.java
+++ b/fcrepo-auth-webac/src/test/java/org/fcrepo/integration/auth/webac/WebACRecipesIT.java
@@ -21,9 +21,9 @@ import static java.util.Arrays.stream;
 
 import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
 import static javax.ws.rs.core.Response.Status.CREATED;
+
 import static javax.ws.rs.core.Response.Status.FORBIDDEN;
 import static javax.ws.rs.core.Response.Status.OK;
-
 import static org.apache.http.HttpStatus.SC_CREATED;
 import static org.apache.http.HttpStatus.SC_FORBIDDEN;
 import static org.apache.http.HttpStatus.SC_NOT_FOUND;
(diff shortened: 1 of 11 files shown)


fcrepo-auth-webac/src/test/java/org/fcrepo/integration/auth/webac/WebACRecipesIT.java
```diff
@@ -2084,70 +2081,6 @@ public void testBinaryAndDescriptionAllowed() throws Exception {
 
     }
 
-    @Test
-    public void testRequestWithEmptyPath() throws Exception {
```
pwinckles, line 2088: This test is removed because it appears that as a result of one of the upgrades, likely Jersey, empty path segments, `//`, are normalized, so `parent/child` and `parent//child` are indistinguishable to the application.

awoods: It appears that HttpClient changed in the 4.5.7 release to remove empty paths by default. Can we leave this test in-place, replacing the last portion with:
```
        // Ensure HttpClient does not remove empty paths
        final RequestConfig config = RequestConfig.custom().setNormalizeUri(false).build();

        // Admin bypasses ACL resolution gets 409.
        final HttpGet getAdminRequest = getObjMethod(parent + "//" + child);
        setAuth(getAdminRequest, "fedoraAdmin");
        getAdminRequest.setConfig(config);
        assertEquals(BAD_REQUEST.getStatusCode(), getStatus(getAdminRequest));
        // User
        final HttpGet getUserRequest = getObjMethod(parent + "//" + child);
        getUserRequest.setConfig(config);
        setAuth(getUserRequest, username);
        assertEquals(BAD_REQUEST.getStatusCode(), getStatus(getUserRequest));
```

As a note, I tried this config update at the `AbstractResourceIT.getObjMethod(id)` level, but it breaks many of the other tests in `WebACRecipesIT` that include double slashes in their test fixtures.

awoods (changes requested):
Thanks for this. One minor comment.

fcrepo-auth-webac/src/test/java/org/fcrepo/integration/auth/webac/WebACRecipesIT.java
pwinckles: Nice! Thanks for figuring that out.

awoods (approved):

PR #1799, opened 2020-11-12
Return RepositoryRoot as system type

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3422

Currently when the repository is reinitialized the root resource is created with the RepositoryRoot type as user rdf. So it returns as a user type. This removes that and returns it as a system type.

# How should this be tested?

Before GETting the root resource should return the triple in the RDF but not the headers. Now it should appear as both.

# Interested parties


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/FedoraResourceImpl.java
```diff
@@ -250,6 +251,9 @@ public boolean hasType(final String type) {
             // ldp:Resource is on all resources
             types.add(RESOURCE_URI);
             types.add(FEDORA_RESOURCE_URI);
+            if (getFedoraId().isRepositoryRoot()) {
+                types.add(create(REPOSITORY_ROOT.getURI()));
```
pwinckles, line 255: Rather than creating the URI every time, just create it once as a static constant

pwinckles (approved):

PR #1801, opened 2020-11-13
Rebuild containment index to support deleted objects

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3450

Alters the ReindexService to add the start time to the containment index for all resources and add an end time for deleted resources.

# How should this be tested?

Make some resources in Fedora.
Delete some of them.
Shut down Fedora
Blow away the indexes.
Restart Fedora
Check the containment index to see that the resources are all there and have a start time. Deleted resources also have an end time.

# Interested parties


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
```diff
@@ -395,7 +395,7 @@ private NamedParameterJdbcTemplate getNamedParameterJdbcTemplate() {
 
     @Override
     public Stream<String> getContains(final String txId, final FedoraId fedoraId) {
-        final String resourceId = fedoraId.getResourceId();
+        final String resourceId = fedoraId.isMemento() ? fedoraId.getBaseId() : fedoraId.getFullId();
```
whikloj, line 398: I had to switch this back as `getResourceId` only deals with things that have an actual file backing them.

A get to a timemap returned both the mementos but also the containment from the Original Resource. So I switched back to this method

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexService.java
```diff
@@ -118,17 +118,23 @@ public void indexOcflObject(final String txId, final String ocflId) {
...
-                        if (content.isPresent()) {
-                            final RdfStream rdf = parseRdf(fedoraId, content.get());
-                            this.referenceService.updateReferences(txId, fedoraId, null, rdf);
+                    final var created = headers.getCreatedDate();
+                    if (!headers.isDeleted()) {
+                        if (!headers.getInteractionModel().equals(NON_RDF_SOURCE.toString())) {
+                            final Optional<InputStream> content = session.readContent(fedoraId.getFullId())
+                                    .getContentStream();
+                            if (content.isPresent()) {
+                                final RdfStream rdf = parseRdf(fedoraId, content.get());
```
pwinckles, line 127: Recognizing that this is not new code, I believe this input stream is not being closed.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/ReferenceServiceImpl.java
```diff
@@ -302,6 +302,8 @@ public void updateReferences(@Nonnull final String txId, final FedoraId resource
             LOGGER.warn("Unable to update reference index for resource {} in transaction {}: {}",
                     resourceId.getFullId(), txId, e.getMessage());
             throw new RepositoryRuntimeException("Unable to update reference index", e);
+        } finally {
+            rdfStream.close();
```
pwinckles, line 306: This will work, but it's slightly unexpected behavior. Generally speaking, streams are closed on the same level that they're opened on. For example,

```java
try (final stream = content.get()) {
    final RdfStream rdf = parseRdf(fedoraId, stream);
    this.referenceService.updateReferences(txId, fedoraId, null, rdf);
}
```

There are a variety of advantages to this approach, not the least of which is that it's clear when a stream is closed and you don't need to inspect every method that the stream is passed to to see if it closes it.

whikloj: So its going to end up looking like this
```java
try (final var stream = content.get()) {
    final RdfStream rdf = parseRdf(fedoraId, stream);
    this.referenceService.updateReferences(txId, fedoraId, null, rdf);
} catch (final IOException e) {
    LOGGER.warn("Content stream for {} closed prematurely, inbound references skipped.",
        fedoraId.getFullId());
    throw new RepositoryRuntimeException(e.getMessage(), e);
}
```

(I think) because its already in a `try {}` enclosure. Do I also need to close the RdfStream?

pwinckles: Yes, that looks correct.

I poked around in the code, and it looks like what's going on is the InputStream is read into an RDF model object and then that model object is turned into an RdfStream. The RdfStream has no relation to the InputStream. It is a stream in the Java 8 Streaming since, and very likely does not need to be closed. Streams can be closed, but, unless the Stream is IO related, it is usually not necessary.

pwinckles (approved):
Worked for me

PR #1803, opened 2020-11-14
Actions

pwinckles:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3532

Replaces Travis with Github Actions

# How should this be tested?

This is difficult to test. Here is an example of the build workflow: https://github.com/pwinckles/fcrepo4/actions/runs/363679840

I have not tested the deploy workflow. I think it should work, but it may need tweaking. All I can say for certain is that it is syntactically correct.

The deploy workflow only runs on pushes to `main` and the following secrets must be defined in the Github settings:

* OSSRH_USERNAME
* OSSRH_TOKEN
* GPG_PASSPHRASE
* GPG_PRIVATE_KEY
* DOCKER_USERNAME
* DOCKER_PASSWORD

**I have not defined these secrets.**

# Interested parties
@fcrepo/committers


pwinckles:
One idea for testing the deployment workflow is to add the necessary secrets and then merge these changes into a branch rather than `main`. I can update the deployment workflow to apply to the branch, which would allow us to see if it works before merging into `main`.

awoods:
@pwinckles : That seems like a good test step.

pwinckles:
@awoods Okay, I added the trigger. The deploy workflow will now run on `main` and `deploy-test`.

awoods:
Thanks, @pwinckles . I will wait for @dbernstein to add the "secrets".

dbernstein:
@awoods : I'm looking into the secrets now.

dbernstein:
Secrets in place: https://github.com/fcrepo/fcrepo/settings/secrets/actions

pwinckles:
This PR should be good to go. Here's a successful run, including a deploy, on a test branch: https://github.com/fcrepo/fcrepo/actions/runs/371243618.

Note, I got rid of the deploy workflow, moving the deploy job into the build workflow so that I could make it depend on all of the other jobs. Deploy will only run on `main`.

awoods (approved):

PR #1804, opened 2020-11-16
Stream containment and membership via paged DB queries

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3507

Instead of querying for a list, which could be too large to return in memory this creates a stream backed by a paged query of the database.

# How should this be tested?

Should have absolutely no discernable effect except that objects with an extremely large number of contained children or membership properties should still be retrieveable in a timely fashion.

# Interested parties
@fcrepo/committers (@bbpennel as it was your ticket)

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
@@ -45,8 +45,7 @@ import java.util.NoSuchElementException;
 import java.util.Queue;
 import java.util.Spliterator;
 import java.util.Spliterators;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.function.Consumer;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
@@ -406,7 +405,7 @@ public class ContainmentIndexImpl implements ContainmentIndex {
         return new NamedParameterJdbcTemplate(getDataSource());
     }
 
-    public void setContainsLimit(final int limit) {
+    void setContainsLimit(final int limit) {
         containsLimit = limit;
     }
 
@@ -416,7 +415,6 @@ public class ContainmentIndexImpl implements ContainmentIndex {
         final Instant asOfTime = fedoraId.isMemento() ? fedoraId.getMementoInstant() : null;
         final MapSqlParameterSource parameterSource = new MapSqlParameterSource();
         parameterSource.addValue("parent", resourceId);
-        parameterSource.addValue("containsLimit", containsLimit);
 
         LOGGER.debug("getContains for {} in transaction {} and instant {}", resourceId, txId, asOfTime);
 
@@ -443,7 +441,6 @@ public class ContainmentIndexImpl implements ContainmentIndex {
         final String resourceId = fedoraId.getFullId();
         final MapSqlParameterSource parameterSource = new MapSqlParameterSource();
         parameterSource.addValue("parent", resourceId);
-        parameterSource.addValue("containsLimit", containsLimit);
 
         final String query;
         if (txId != null) {
@@ -719,8 +716,8 @@ public class ContainmentIndexImpl implements ContainmentIndex {
      * If this needs to be run in parallel we will have to override trySplit() and determine a good method to split on.
      */
     private class ContainmentIterator extends Spliterators.AbstractSpliterator<String> {
-        final Queue<String> children = new LinkedBlockingQueue<>();
-        final AtomicInteger numOffsets = new AtomicInteger(0);
+        final Queue<String> children = new ConcurrentLinkedQueue<>();
+        int numOffsets = 0;
         final String queryToUse;
         final MapSqlParameterSource parameterSource;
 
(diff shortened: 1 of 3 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
```diff
@@ -696,4 +712,38 @@ private Timestamp formatInstant(final Instant instant) {
...
         return timestamp;
     }
+
+    /**
+     * Private class to back a stream with a paged DB query.
+     *
+     * If this needs to be run in parallel we will have to override trySplit() and determine a good method to split on.
+     */
+    private class ContainmentIterator extends Spliterators.AbstractSpliterator<String> {
+        final Queue<String> children = new LinkedBlockingQueue<>();
```
pwinckles, line 722: I think you might want a `ConcurrentLinkedQueue` here instead: https://www.baeldung.com/java-queue-linkedblocking-concurrentlinked

Incidentally, I think your original implementation of the parallelized reindex code would have worked if you'd backed the executor service with a `LinkedBlockingQueue`. I'm sorry I didn't think of it at the time. :(

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
```diff
@@ -696,4 +712,38 @@ private Timestamp formatInstant(final Instant instant) {
...
+            queryToUse = query;
+            parameterSource = parameters;
+        }
+
+        @Override
+        public boolean tryAdvance(final Consumer<? super String> action) {
+            try {
+                action.accept(children.remove());
+            } catch (final NoSuchElementException e) {
+                parameterSource.addValue("offSet", numOffsets.getAndIncrement() * containsLimit);
```
pwinckles, line 738: Is there a reason to not set the `containsLimit` parameter here too?

I see, because there's no reason to set it every time. Perhaps set it in the constructor then?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
```diff
@@ -696,4 +712,38 @@ private Timestamp formatInstant(final Instant instant) {
...
+                children.addAll(jdbcTemplate.queryForList(queryToUse, parameterSource, String.class));
```
pwinckles, line 739: Out of curiosity, does the ordering ever matter? I ask because this implementation is trying to be thread-safe, but I'm not sure if it is entirely. Is it possible that `tryAdvance` could be called concurrently? If so, then this code will have a race to load the next page. Assuming there are sufficient items remaining, the competing threads will correctly load different pages, but the order they are added to the queue will be non-deterministic.

All that to say, I'm not sure if this implementation needs to be thread-safe. I do not believe that we have any code that uses parallel streaming features.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
```diff
@@ -393,50 +406,56 @@ private NamedParameterJdbcTemplate getNamedParameterJdbcTemplate() {
         return new NamedParameterJdbcTemplate(getDataSource());
     }
 
+    public void setContainsLimit(final int limit) {
```
pwinckles, line 409: Is this just for testing purposes? If so, it could probably be package-private and optionally annotated with `@VisibleForTesting`

pwinckles (commented):
Haven't tested yet, but it looks good with a few minor comments

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
whikloj: My understanding, which is new, is that to allow your stream to be operated on in parallel you override the `trySplit` method which handles splitting the stream up into chunks for individual threads. 

In this case, no we don't operate on it in parallel ever. At least not right now so order is not really important except in that we need the query to generate the same order so our OFFSET and LIMIT work properly.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
whikloj: I was setting it in the calling method where I set all of the other parameters (ie. parent, startTime, child). I can move it into this method if that makes it clearer. But yes, LIMIT is a constant and we just advance the OFFSET parameter.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
pwinckles: In that case, do we need to use `AtomicInteger`?

whikloj: Ohhhhhhhhhh. No, I needed that when I had this class just embedded in the `StreamSupport.generate`. I can change that to an `int`

pwinckles:
The updates look good to me. The same changes could be applied to the membership spliterator too. Sorry, I didn't note this in my original review. I was trying to avoid leaving duplicate comments.

pwinckles (approved):
Great!

pwinckles (approved):

PR #1806, opened 2020-11-17
Added exception handling for identifier line that is too long, plus c…

robyj:
…ustom message

Resolves https://jira.lyrasis.org/browse/FCREPO-3374

* * *

**JIRA Ticket**: (link)

A brief description of what the intended result of the PR will be and/or what problem it solves.

# How should this be tested?

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -202,6 +202,8 @@ import com.google.common.collect.Iterators;
 import nu.validator.htmlparser.sax.HtmlParser;
 import nu.validator.saxtree.TreeBuilder;
 
+import org.fcrepo.kernel.api.exception.InvalidResourceIdentifierException;
+
 /**
  * @author cabeer
  * @author ajs6f
@@ -5009,6 +5011,21 @@ public class FedoraLdpIT extends AbstractResourceIT {
             assertTrue(graph.contains(ANY, rootNode, type.asNode(), CONTAINER.asNode()));
             assertTrue(graph.contains(ANY, rootNode, type.asNode(), FEDORA_CONTAINER.asNode()));
             assertTrue(graph.contains(ANY, rootNode, type.asNode(), BASIC_CONTAINER.asNode()));
+        }
+    }
+
+    @Test
+    public void testLongIdentifier() throws InvalidResourceIdentifierException, IOException {
+        String url = serverAddress;
+        for (int cnt = 0; cnt < 200; cnt++) {
+            url = url + "/" + Integer.toString(cnt);
+            final HttpPost postMethod = new HttpPost(url);
+            postMethod.setHeader("Slug", Integer.toString(cnt));
+            try (final CloseableHttpResponse response = execute(postMethod)) {
+                if ((cnt < 148) && (getStatus(response) == BAD_REQUEST.getStatusCode())) {
+                    fail("failed at try");
+                }
+            }
         }
     }
 }
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -146,6 +146,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.NoSuchElementException;
 import java.util.Optional;
+import java.util.Random;
 import javax.ws.rs.core.Link;
 import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.core.Variant;
@@ -202,7 +203,7 @@ import com.google.common.collect.Iterators;
 import nu.validator.htmlparser.sax.HtmlParser;
 import nu.validator.saxtree.TreeBuilder;
 
-import org.fcrepo.kernel.api.exception.InvalidResourceIdentifierException;
+//import org.fcrepo.kernel.api.exception.InvalidResourceIdentifierException;
 
 /**
  * @author cabeer
@@ -5014,18 +5015,37 @@ public class FedoraLdpIT extends AbstractResourceIT {
         }
     }
 
+    public String genLongUrl(final String prefix, final int numChars) {
+        String ret = prefix;
+        String lastret = ret;
+        final Random rand = new Random();
+        int cnt = prefix.length();
+
+        while (cnt <= numChars) {
+            lastret = ret;
+            ret = ret + "/" + (char)(rand.nextInt(26) + 'a');
+            cnt = ret.length();
+        }
+        return lastret;
+    }
+
     @Test
-    public void testLongIdentifier() throws InvalidResourceIdentifierException, IOException {
-        String url = serverAddress;
-        for (int cnt = 0; cnt < 200; cnt++) {
-            url = url + "/" + Integer.toString(cnt);
-            final HttpPost postMethod = new HttpPost(url);
-            postMethod.setHeader("Slug", Integer.toString(cnt));
-            try (final CloseableHttpResponse response = execute(postMethod)) {
-                if ((cnt < 148) && (getStatus(response) == BAD_REQUEST.getStatusCode())) {
-                    fail("failed at try");
-                }
-            }
+    public void testLongIdentifier1() throws IOException {
+        final String url = genLongUrl(serverAddress,478);
+        final HttpPost postMethod = new HttpPost(url);
+        postMethod.setHeader("Slug", getRandomUniqueId());
+        try (final CloseableHttpResponse response = execute(postMethod)) {
+            assertEquals(CREATED.getStatusCode(), getStatus(response));
+        }
+    }
+
+    @Test
+    public void testLongIdentifier2() throws IOException {
+        final String url = genLongUrl(serverAddress,479);
+        final HttpPost postMethod = new HttpPost(url);
+        postMethod.setHeader("Slug", getRandomUniqueId());
+        try (final CloseableHttpResponse response = execute(postMethod)) {
+            assertEquals(BAD_REQUEST.getStatusCode(), getStatus(response));
         }
     }
 }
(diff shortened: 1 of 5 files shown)


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndex.java
```diff
@@ -265,7 +266,11 @@ private void upsert(final String transactionId, final FedoraId fedoraId, final S
         parameterSource.addValue("ocflId", ocflId);
         parameterSource.addValue("transactionId", transactionId);
         parameterSource.addValue("operation", operation);
-        jdbcTemplate.update(UPSERT_MAPPING_TX_MAP.get(dbPlatform), parameterSource);
+        try {
+            jdbcTemplate.update(UPSERT_MAPPING_TX_MAP.get(dbPlatform), parameterSource);
+        } catch (Exception e) {
```
pwinckles, line 271: This logic is not specific enough. It's possible for the `update` call to fail for reason unrelated to the length of the id.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndex.java
```diff
@@ -265,7 +266,11 @@ private void upsert(final String transactionId, final FedoraId fedoraId, final S
...
+            throw new InvalidResourceIdentifierException("ID path too long");
```
pwinckles, line 272: It is almost always a good idea to include the cause when throwing a new exception. You should do so here, unless there's a good reason not to.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndex.java
robyj: so, should I check the length of the id, rather than just wrap it in an exception handler?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndex.java
robyj: you mean change the message to be more descriptive?

pwinckles: I mean include the cause in the new exception like:

```java
throw new InvalidResourceIdentifierException("message", e);
```

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndex.java
pwinckles: Checking the length may be reasonable, though, if we go that route, I think it would make more sense if the id length enforcement happened up in the HTTP layer.

Catching the exception here is not a problem. You just need to be more targeted about it. What is the type of exception that's thrown when the id is too long? What is the message of the exception? Are these consistent across database implementations? Using that information, you should be able to inspect the exception and only throw the id length exception when appropriate.

awoods (commented):
This looks good. To @pwinckles ' [point](https://github.com/fcrepo/fcrepo/pull/1806#discussion_r526181085), an integration test with an ID that is too long would ensure that this solution works across all three databases.

robyj:
@awoods - Ok, I've modified the error message (maybe better? maybe worse?) and made some of the code changes @pwinckles suggested. And I'm currently working on the integration test

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndexTest.java
```diff
@@ -175,6 +176,23 @@ public void testRollback() throws Exception {
         }
     }
 
+    @Test
+    public void identifierTooLong() throws InvalidResourceIdentifierException {
```
awoods, line 180: This is a good test, and we should keep it. However, it only tests against H2:
https://github.com/fcrepo/fcrepo/blob/30a516136e8b1009d9266ae795e17a581ba2fd0a/fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndexTest.java#L53-L54

If we add a similar integration test to FedoraLdpIT, the GitHub Actions builds will test it against all three supported databases. You should be able to add a test in here:
https://github.com/fcrepo/fcrepo/blob/main/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java

..and you will be good to go.
Sorry for the lack of specificity in my previous comment. The integration tests (those that end in **IT.java) run against a deployed Fedora application... and GitHub Actions are configured to run the tests three times.

awoods (changes requested):

robyj:
the mariadb travis test generated a 500 http error and i have no idea if thats related to the test or travis weirdness

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndex.java
```diff
@@ -265,7 +267,11 @@ private void upsert(final String transactionId, final FedoraId fedoraId, final S
         parameterSource.addValue("ocflId", ocflId);
         parameterSource.addValue("transactionId", transactionId);
         parameterSource.addValue("operation", operation);
-        jdbcTemplate.update(UPSERT_MAPPING_TX_MAP.get(dbPlatform), parameterSource);
+        try {
+            jdbcTemplate.update(UPSERT_MAPPING_TX_MAP.get(dbPlatform), parameterSource);
+        } catch (final DataIntegrityViolationException e) {
```
whikloj, line 272: MariaDB is dying because it is throwing a different Exception and so its not getting caught here. 

```
org.springframework.jdbc.BadSqlGrammarException: PreparedStatementCallback; bad SQL grammar [INSERT INTO ocfl_id_map_session_operations (fedora_id, fedora_root_id, ocfl_id, session_id, operation) VALUES (?, ?, ?, ?, ?) ON DUPLICATE KEY UPDATE fedora_root_id = VALUES(fedora_id), ocfl_id = VALUES(ocfl_id), operation = VALUES(operation)]; nested exception is java.sql.SQLSyntaxErrorException: (conn=33) Data too long for column 'fedora_id' at row 1
```

So you'll need to catch both Exceptions here, ala 
```java
} catch (final DataIntegrityViolationException | BadSqlGrammarException e) {
```

whikloj (changes requested):
Good stuff, couple small suggestions and one part I'm a little unsure I understand. Might need another set of 👀

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -5011,4 +5014,38 @@ public void testRepositoryRootTypes() throws Exception {
...
     }
+
+    public String genLongUrl(final String prefix, final int numChars) {
+        String ret = prefix;
+        String lastret = ret;
+        final Random rand = new Random();
+        int cnt = prefix.length();
+
+        while (cnt <= numChars) {
+            lastret = ret;
```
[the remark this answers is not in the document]
robyj: Hmmm, I am using lastret to store the last string before it goes over the length required, so when it completes, lastret would contain the string that is upto the length required. Of course, I see now that the "<=" probably would return a string that is almost always less than the required length.
I used the / between the elements in the path as the original bug report script iterated a path until it was too big and the first version of the test did the same, a for loop iterated through the path. I wasn't sure if the problem was a length issue or a path "deepness" issue

whikloj (dismissed):
👍 I'm good here

awoods (commented):
Nice work, @robyj .
This looks good to go as long as we have confidence that the only exception coming from the JDBC upsert is a result of an identifier that is too long:
https://github.com/fcrepo/fcrepo/pull/1806/files#diff-1d1714acc853285d46ffd91d0f1d1443cd07acda2f09b8b04b0da014b676c939R273-R274

I suspect this is not the case, but do not have a good suggestion for determining the exact nature of the JDBC exception.

pwinckles:
Here are the exceptions from each of the dbs:

**Postgres**

```
org.springframework.dao.DataIntegrityViolationException: PreparedStatementCallback; SQL [INSERT INTO ocfl_id_map_session_operations ( fedora_id, fedora_root_id, ocfl_id, session_id, operation) VALUES (?, ?, ?, ?, ?) ON CONFLICT (fedora_id, session_id) DO UPDATE SET fedora_root_id = EXCLUDED.fedora_root_id, ocfl_id = EXCLUDED.ocfl_id, operation = EXCLUDED.operation]; ERROR: value too long for type character varying(503); nested exception is org.postgresql.util.PSQLException: ERROR: value too long for type character varying(503)
```

**MariaDB**

```
org.springframework.jdbc.BadSqlGrammarException: PreparedStatementCallback; bad SQL grammar [INSERT INTO ocfl_id_map_session_operations (fedora_id, fedora_root_id, ocfl_id, session_id, operation) VALUES (?, ?, ?, ?, ?) ON DUPLICATE KEY UPDATE fedora_root_id = VALUES(fedora_id), ocfl_id = VALUES(ocfl_id), operation = VALUES(operation)]; nested exception is java.sql.SQLSyntaxErrorException: (conn=48) Data too long for column 'fedora_id' at row 1
```

**MySQL**

```
org.springframework.dao.DataIntegrityViolationException: PreparedStatementCallback; SQL [INSERT INTO ocfl_id_map_session_operations (fedora_id, fedora_root_id, ocfl_id, session_id, operation) VALUES (?, ?, ?, ?, ?) ON DUPLICATE KEY UPDATE fedora_root_id = VALUES(fedora_id), ocfl_id = VALUES(ocfl_id), operation = VALUES(operation)]; Data truncation: Data too long for column 'fedora_id' at row 1; nested exception is com.mysql.cj.jdbc.exceptions.MysqlDataTruncation: Data truncation: Data too long for column 'fedora_id' at row 1
```

At the minimum, I think you should check that the exception message contains the string "too long"

awoods (approved):

whikloj (approved):
👍

PR #1810, opened 2020-11-18
Alter ETag based on contained resources

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3541

We don't change the ETag when a container has a new or deleted contained resource.  This attempts to fix that by generating a hash of all the children of the requested resource.

I think this might be slowing down ingest as I was creating 5,000 children of a single parent and it seemed to average about 0.14 per resource (which is slower than normal on my machine).

Also there is probably a limit to how many contained resources this might work with. Probably need the big test machine to see if and when this crashes.

# How should this be tested?

Create a container, note it's ETag and Last Modified time.
Create a child of the container, note it's ETag changes, but it's Last Modified is unchanged.

# Interested parties

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
@@ -17,6 +17,7 @@
  */
 package org.fcrepo.kernel.impl.models;
 
+import org.apache.commons.codec.digest.DigestUtils;
 import org.fcrepo.kernel.api.ContainmentIndex;
 import org.fcrepo.kernel.api.Transaction;
 import org.fcrepo.kernel.api.TransactionUtils;
@@ -171,7 +172,7 @@ public class ResourceFactoryImpl implements ResourceFactory {
 
             final var rescImpl = constructor.newInstance(instantiationId, transactionId,
                     persistentStorageSessionManager, this);
-            populateResourceHeaders(rescImpl, headers, versionDateTime);
+            populateResourceHeaders(transactionId, rescImpl, headers, versionDateTime);
 
             if (headers.isDeleted()) {
                 final var rootId = FedoraId.create(identifier.getBaseId());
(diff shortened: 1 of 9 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
```diff
@@ -193,14 +194,18 @@ private FedoraResource instantiateResource(final String transactionId,
...
-                                         final Instant version) {
+    private void populateResourceHeaders(final String transactionId, final FedoraResourceImpl resc,
+                                         final ResourceHeaders headers, final Instant version) {
         resc.setCreatedBy(headers.getCreatedBy());
         resc.setCreatedDate(headers.getCreatedDate());
         resc.setLastModifiedBy(headers.getLastModifiedBy());
         resc.setLastModifiedDate(headers.getLastModifiedDate());
         resc.setParentId(headers.getParent());
-        resc.setEtag(headers.getStateToken());
+        final Instant containmentUpdated = containmentIndex.containmentLastUpdated(transactionId, resc.getFedoraId());
```
pwinckles, line 204: This code is only relevant when loading a container, correct? If so, I think we should only do this lookup when necessary.

This code would also be a good candidate for caching in the future.

pwinckles: I'm guessing that there will be similar code to this for membership? It might be nice to abstract the etag computation out into its own class, but you don't necessarily need to do so here.

pwinckles (commented):
I tried it locally and it mostly works as expected. However, when I tried reindexing the eTags changed. Is it possible to keep the eTags consistent? And, if not, is this a problem?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/models/ResourceFactoryImpl.java
whikloj: One of the reasons I liked the previous implementation is that if the containment didn't change then neither did the eTag. But this was more consistent across DBs. Perhaps we just need to stream all containment and generate a hash in code rather than trying to do it in the database.

whikloj:
So I'm playing with this, but I could probably use some help. Here is my current idea.

On commit, check which resources are having their containment updated (like in the current PR). But add those to a list, then asynchronously (so a separate thread) have it pull the `getContains` result, chop them to 100 characters max, concatenate them together and generate a SHA-256 hash, store that back into the DB. Inside a long running transaction we'd have to calculate the hash on the fly, but 🤷

pwinckles (approved):

PR #1811, opened 2020-11-18
Display root resource/ACL correctly in HTML UI

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3424

HTML UI performs extra processing to allow interactions so we load the resource we are looking at, for "fake" resources like the default ACL this fails to load. Handle this more cleanly.

# How should this be tested?

Build PR, browse to `http://localhost:8080/rest/fcr:acl` with a web browser. See the results. Use cURL for alternate serializations and still see the results. Put a new default ACL and see it displays too.

# Interested parties


awoods (changes requested):
Works like a charm.
Would you mind adding an integration test? https://github.com/fcrepo/fcrepo/blob/main/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraAclIT.java

awoods (dismissed):

awoods:
..waiting on GitHub Actions

awoods (approved):

PR #1812, opened 2020-11-18
Fixity Service

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3535

Implements a fixity checking service.

# How should this be tested?

Put a binary in, then go to it's `fcr:fixity` endpoint. Try both cUrl and a web browser.
Assuming the result is `SUCCESS` try altering your binary and run the request again.
See the `BAD_CHECKSUM` result. 

In Fedora 5 you would get a `BAD_SIZE` outcome as well, I feel like if the size changes the checksum changes and we aren't all that concerned about ever getting a `BAD_SIZE` without a `BAD_CHECKSUM`. But if desired we'd have to add this functionality to the fixity service.

# Interested parties
@fcrepo/committers (specifically @bbpennel) and @dwilcox

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/FixityServiceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/FixityServiceImpl.java
@@ -32,6 +32,7 @@ import org.apache.jena.rdf.model.Model;
 import org.apache.jena.rdf.model.Resource;
 import org.fcrepo.kernel.api.RdfStream;
 import org.fcrepo.kernel.api.exception.InvalidChecksumException;
+import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
 import org.fcrepo.kernel.api.exception.UnsupportedAlgorithmException;
 import org.fcrepo.kernel.api.models.Binary;
 import org.fcrepo.kernel.api.rdf.DefaultRdfStream;
@@ -40,6 +41,7 @@ import org.fcrepo.kernel.api.utils.ContentDigest.DIGEST_ALGORITHM;
 import org.fcrepo.persistence.common.MultiDigestInputStreamWrapper;
 import org.springframework.stereotype.Component;
 
+import java.io.IOException;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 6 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/FixityServiceImpl.java
```diff
@@ -49,8 +73,37 @@
...
+        existingDigestList.add(binary.getContentDigest());
+
+        final var digestAlgs = existingDigestList.stream()
+                .map(URI::toString)
+                .map(a -> a.replace("urn:", "").split(":")[0])
+                .map(DIGEST_ALGORITHM::fromAlgorithm)
+                .collect(Collectors.toList());
+
+        final MultiDigestInputStreamWrapper digestWrapper = new MultiDigestInputStreamWrapper(
+                binary.getContent(), existingDigestList, digestAlgs);
```
pwinckles, line 98: Haven't looked at this PR in depth. Was just browsing and noticed this. `binary.getContent()` should be called in a try-with block so that it is closed.

pwinckles (commented):
Haven't looked at this in depth. Was just browsing an

pwinckles:
Are there any ITests that could be renabled for this?

whikloj:
Good question, I'll have a look

whikloj:
No, no there aren't. Looks like there is a Memento fixity test that uses this end point. All other tests in FedoraFixityIT use the `Want-Digest` to get fixity, not check the existing fixity. I can try to write one, not sure about messing with data on disk to verify a failure though.

pwinckles:
Looks good to me, and it works for me locally. I do think it would be good to have a negative integration test as well. This could should help with corrupting a binary in an itest:

```java
private OcflPropsConfig ocflConfig;
    private OcflRepository ocflRepo;

    @Before
    public void setup() {
        ocflConfig = getBean(OcflPropsConfig.class);
        ocflRepo = getBean(OcflRepository.class);
    }

    @Test
    public void badChecksumTest() throws IOException {
        final var id = UUID.randomUUID().toString();

        // setup the binary

        // change the file on disk
        final var ocflDesc = ocflRepo.describeVersion(
                ObjectVersionId.head(FedoraId.create(id).getResourceId()));
        final var storagePath = ocflDesc.getFile(id).getStorageRelativePath();

        Files.writeString(ocflConfig.getOcflRepoRoot().resolve(storagePath),
                "corrupted!", StandardCharsets.UTF_8, StandardOpenOption.APPEND);

        // run fixity check
    }
```
I have not run it, so let me know if you have any problems with it.

pwinckles (approved):

PR #1815, opened 2020-11-20
Removes need to disable triggers on PostgresQL

dbernstein:
Removes need to disable triggers on PostgresQL which was causing a problem database administrator does not have super admin control as is the case on AWS RDS.

**JIRA Ticket**:  https://jira.lyrasis.org/browse/FCREPO-3540

No functional change.  
* Removes the need to disable triggers on truncation in postgresql
# How should this be tested?
Integration tests should already cover it.
I build a new docker image and deployed it to fcrepo/fcrepo:6.0.0-SNAPSHOT-3540-test and verified that postgresql now works when I deploy using this PR of fcrepo-aws-deployer: https://github.com/fcrepo-exts/fcrepo-aws-deployer/pull/3

You can test these changes in AWS yourself using the instructions in the aforementioned PR. 

# Interested parties
@awoods

### DIFF
--- a/fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
+++ b/fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
@@ -59,6 +59,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 
 import static java.time.format.DateTimeFormatter.ISO_INSTANT;
+import static java.util.Collections.EMPTY_LIST;
 import static org.fcrepo.common.db.DbPlatform.H2;
 import static org.fcrepo.common.db.DbPlatform.MARIADB;
 import static org.fcrepo.common.db.DbPlatform.MYSQL;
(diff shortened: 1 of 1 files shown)


fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -471,11 +472,7 @@ public void reset() {
...
 
         if (isPostgres()) {
-            return List.of(
-                    togglePostgresTriggers(SEARCH_RESOURCE_RDF_TYPE_TABLE, enable),
-                    togglePostgresTriggers(SEARCH_RDF_TYPE_TABLE, enable),
-                    togglePostgresTriggers(SIMPLE_SEARCH_TABLE, enable)
-            );
+            return EMPTY_LIST;
         } else {
             return List.of("SET FOREIGN_KEY_CHECKS = " + (enable ? 1 : 0) + ";");
```
whikloj, line 477: Your description says this is now only needed for H2, so could we stop doing this for MySQL and MariaDB too. Maybe repurpose the `isPostgres()` into `isH2()`?

dbernstein: @whikloj :  I was wrong - we need the foreign key enable/disable on mysql, h2, and mariadb after all.

whikloj (approved):
👍

PR #1816, opened 2020-11-24
all db commits within the same db transaction

pwinckles:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3529

Fixes the transaction commit code so that all db commits are within the same transaction so that either they are all applied or none of them are.

This PR only fixes the db transactions. It will no cleanup already persisted OCFL changes. This should be handled by https://jira.lyrasis.org/browse/FCREPO-3130

# How should this be tested?

This is difficult to test. You'd need to create a Fedora transaction and then corrupt a db table before committing the transaction so that the Fedora transaction does not commit cleanly.

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/TransactionsIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/TransactionsIT.java
@@ -17,6 +17,30 @@
  */
 package org.fcrepo.integration.http.api;
 
+import org.apache.commons.lang3.StringUtils;
+import org.apache.http.Header;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpDelete;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.methods.HttpHead;
+import org.apache.http.client.methods.HttpPatch;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.client.methods.HttpPut;
+import org.apache.http.entity.StringEntity;
+import org.fcrepo.http.commons.test.util.CloseableDataset;
+import org.fcrepo.kernel.api.ContainmentIndex;
+import org.fcrepo.kernel.api.identifiers.FedoraId;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.springframework.test.context.TestExecutionListeners;
+
+import javax.ws.rs.core.Response.Status;
+import java.io.IOException;
+import java.time.format.DateTimeParseException;
+import java.util.UUID;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 import static java.lang.Math.min;
 import static java.lang.Thread.sleep;
 import static java.time.Duration.ofMinutes;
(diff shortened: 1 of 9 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/TransactionsIT.java
```diff
@@ -194,6 +196,47 @@ public void testCreateDoStuffAndRollbackTransaction() throws IOException {
...
+        containmentIndex.addContainedBy("bogus", FedoraId.getRepositoryRootId(), FedoraId.create(resourceId));
+        containmentIndex.commitTransaction("bogus");
+
+        // Commit transaction -- should fail
+        assertEquals(CONFLICT.getStatusCode(), getStatus(new HttpPut(txLocation)));
+
+        assertEquals("Rolled back transaction should be gone",
+                GONE.getStatusCode(), getStatus(new HttpGet(txLocation)));
+
+        assertEquals("Expected to not find our object after rollback",
```
whikloj, line 234: I think this `NOT_FOUND` response is coming from the containment index, which failed and should have rolled back. Could we have one of the other services fail instead, like the ocfl Index, reference service or membership service? Just to ensure all the transactions are being rolled back.

pwinckles: The Fedora to OCFL mapping is persisted first, and the containment index. This test does correctly confirm that the OCFL mapping is rolled back when the containment index update fails -- except in MySQL which is apparently not rolling back for some reason.

pwinckles: On GET resource requests, the the OCFL mapping table is accessed when the resource headers are loaded before the containment table is checked.

pwinckles:
The MySQL build problem was because the containment table did not have a unique constraint of `fedora_id`. At some point, I think it would be a good idea to review all of our table definitions? Perhaps we could even get some of Aaron's time to go over them?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/TransactionsIT.java
whikloj: My mistake, you are correct.

fcrepo-kernel-impl/src/main/resources/sql/mysql-containment.sql
```diff
@@ -3,10 +3,11 @@
 
 -- Holds the ID and its parent.
 CREATE TABLE IF NOT EXISTS containment (
-    fedora_id  varchar(503) NOT NULL,
+    fedora_id varchar(503) NOT NULL,
     parent varchar(503) NOT NULL,
     start_time timestamp NOT NULL,
-    end_time timestamp NULL
+    end_time timestamp NULL,
+    UNIQUE KEY (fedora_id)
```
whikloj, line 10: You can just add the `PRIMARY KEY` to the `fedora_id` line, I'm not sure why it disappeared but clearly I made a mistake.

pwinckles: Yeah, I can change that. I originally wrote it that way because I was browsing some discussion about varchar primary keys, which a lot of folks advise against for performance reasons. In this case, the key difference between using `PRIMARY KEY` vs `UNIQUE KEY` is that the former creates a clustered index and the later creates a non-clustered index. Because we will not be adding keys sequentially, this is less ideal for clustered indexes.

That said, I don't mind changing it. All of the other databases are currently defined with `PRIMARY KEY`, lookup should still be faster with the clustered index, and it's unclear how much of a difference it even makes.

whikloj: I agree that having someone with more indepth DB experience have a look at our schemas (<wink> <wink> @birkland) would be helpful. If you want to leave this as is, I'm fine. Just don't want people to think there is a specific reason that MySQL is different.

whikloj (approved):
👍

PR #1819, opened 2020-11-27
Remove fedora:TimeMap type

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3007

Removes the `fedora:TimeMap` type from the link headers of timemaps. `memento:TimeMap` exists in its place.

# How should this be tested?

1. Make an object (ie. `curl -i -ufedoraAdmin:fedoraAdmin -XPUT http://localhost:8080/rest/test`)
1. Get the object's timemap (ie. `curl -i -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/test/fcr:versions`)
1. See two Timemap headers
     ```
     Link: <http://fedora.info/definitions/v4/repository#TimeMap>; rel="type"
     Link: <http://mementoweb.org/ns#TimeMap>; rel="type"
     ```
1. Build this PR, do the GET again and see only one header
    ```
     Link: <http://mementoweb.org/ns#TimeMap>; rel="type"
    ```

# Interested parties

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/RdfLexicon.java
@@ -127,8 +127,6 @@ public final class RdfLexicon {
             createResource(REPOSITORY_NAMESPACE + "Pairtree");
     public static final Resource ARCHIVAL_GROUP =
             createResource(REPOSITORY_NAMESPACE + "ArchivalGroup");
-    public static final Resource TIME_MAP =
-            createResource(REPOSITORY_NAMESPACE + "TimeMap");
 
     // Linked Data Platform
     public static final Property PAGE =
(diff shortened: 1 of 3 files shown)


pwinckles:
@whikloj I noticed the following:

```
Link: <http://www.w3.org/ns/ldp#Resource>; rel="type"
Link: <http://www.w3.org/ns/ldp#Container>; rel="type"
Link: <http://mementoweb.org/ns#TimeMap>; rel="type"
```

```
        rdf:type               ldp:Resource ;
        rdf:type               ldp:RDFSource .
```

Is there a reason TimeMap isn't an RDF type too? Obviously not something that as affected by this PR, but I'm curious.

whikloj:
Some are only exposed as Link headers, can't remember the history. But I think we tried to not add a bunch of these to the rdf body. It's why we have that `forRdf` on `getSystemTypes`. I can't quite remember the full back story. Perhaps @dbernstein or @awoods  might.

pwinckles (approved):

PR #1822, opened 2020-12-02
modified the version of the H2 database library reference in pom.xml,…

robyj:
… from 1.4.200 down to 1.4.195

as per https://github.com/h2database/h2database/issues/2356

Resolves: https://jira.lyrasis.org/browse/FCREPO-3368

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     <narayana-jta.version>5.9.0.Final</narayana-jta.version>
     <jetty.version>9.4.34.v20201102</jetty.version>
     <guava.version>30.0-jre</guava.version>
-    <h2database.version>1.4.200</h2database.version>
+    <h2database.version>1.4.195</h2database.version>
     <hk2.version>2.6.1</hk2.version>
     <htmlunit.version>2.33</htmlunit.version>
     <httpclient.version>4.5.13</httpclient.version>


pwinckles:
@robyj Can you confirm if this issue is just with H2 or if it affects the other platforms as well? For example, if you start a one-click or tomcat instance against Postgres are similar exceptions logged?

If the answer is no, then this fix is sufficient. If the answer is yes, then it would seem that there might be a more general root cause fix to pursue.

awoods:
I can confirm that this PR does indeed fix the issue of extraneous messages when using H2.

pwinckles:
@robyj Okay, I'm good with this. I just tried Tomcat + Postgres. There is an error in the logs when you shut it down:

```
03-Dec-2020 08:01:48.960 WARNING [main] org.apache.catalina.loader.WebappClassLoaderBase.clearReferencesJdbc The web application [fcrepo] registered the JDBC driver [org.postgresql.Driver] but failed to unregister it when the web application was stopped. To prevent a memory leak, the JDBC Driver has been forcibly unregistered.
03-Dec-2020 08:01:48.963 WARNING [main] org.apache.catalina.loader.WebappClassLoaderBase.clearReferencesThreads The web application [fcrepo] appears to have started a thread named [Resource Destroyer in BasicResourcePool.close()] but has failed to stop it. This is very likely to create a memory leak. Stack trace of thread:
03-Dec-2020 08:01:48.976 INFO [main] org.apache.coyote.AbstractProtocol.stop Stopping ProtocolHandler ["http-nio-8080"]
03-Dec-2020 08:01:48.980 INFO [main] org.apache.coyote.AbstractProtocol.destroy Destroying ProtocolHandler ["http-nio-8080"]
```

But, it's not nearly as dramatic as the H2 one.

👍

robyj:
Hi

Ok, I'm glad as I spent yesterday afternoon trying to get postgresql to work with the one-click solution and I seemed to be doing something wrong as it wasn't noticing postgresql. But glad the new error message isn't as severe

awoods (approved):

PR #1823, opened 2020-12-02
add resource locking

pwinckles:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3123

**This PR requires https://github.com/fcrepo/fcrepo-storage-ocfl/pull/25 to work**

Adds resource locking so that a resource cannot be concurrently modified in different transaction.

In this implementation,

* A transaction holds a lock on a resource for the duration of the transaction
* The same transaction may modify a resource that it holds a lock on multiple times
* AGs are locked when any of their children are modified
* Binaries are locked when their description is modified
* Binary descriptions are locked when their binary is modified

# How should this be tested?

1. Create a resource
2. Start a transaction
3. Modify the resource in a transaction, but do not commit
4. Attempt to modify the same resource outside of the transaction. This should fail with a `409`.

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
@@ -22,6 +22,8 @@ import org.fcrepo.kernel.api.ContainmentIndex;
 import org.fcrepo.kernel.api.Transaction;
 import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
 import org.fcrepo.kernel.api.exception.TransactionClosedException;
+import org.fcrepo.kernel.api.identifiers.FedoraId;
+import org.fcrepo.kernel.api.lock.ResourceLockManager;
 import org.fcrepo.kernel.api.observer.EventAccumulator;
 import org.fcrepo.kernel.api.services.MembershipService;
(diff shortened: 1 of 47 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
```diff
@@ -238,6 +255,12 @@ private void failIfRolledback() {
         }
     }
 
+    private void releaseLocks() {
+        execQuietly("Failed to release resource locks cleanly. You may need to restart Fedora.", () -> {
```
whikloj, line 259: Is this just going to be a log message? Seems like it could be a bigger problem or am I over estimating its severity?

pwinckles: It is highly unlikely to happen. I'm not even sure how to engineer it to happen. But, yes, if locks were not released, then the resources that were not unlocked would remain locked until Fedora was restarted.

whikloj (dismissed):
This works as described along with the fcrepo-storage-ocfl PR. 👍

pwinckles:
Okay, the flapping _should_ be resolved now

whikloj (approved):
👍

PR #1826, opened 2020-12-02
Adds support for side-loading of new resources.

dbernstein:
**Adds support for side-loading of new resources**
* * *

**JIRA Ticket**: Resolves: https://jira.lyrasis.org/browse/FCREPO-3332

Two follow on tickets for reindexing existing resources and ensuring that newly created, updated, or deleted resources emit events when they are reindexed.
* https://jira.lyrasis.org/browse/FCREPO-3553
* https://jira.lyrasis.org/browse/FCREPO-3554

This PR adds the /fcr:reindex endpoint on all resources. So for  example http://localhost:8080/my-resource/fcr:reindex is now exposed.
Eventually this endpoint will service on demand reindexing. With this PR,  this endpoint is used to accomplish "side-loading" which is the addition of new resources to your Fedora repository by adding OCFL objects directly to the OCFL layout.

# How should this be tested?
There is an integration test covering this feature.  However if you want try it manually, you can start up a fresh fedora and then copy the entire contents of `fcrepo-http-api/src/test/resources/reindex-test` into your ocfl-root directory (fcrepo-home/data/ocfl-root).

The id of the new ocfl resource is "info:fedora/container1" (as you'll notice in the inventory.json file).

To side load the resource, simply do the following:
```
curl -v -u fedoraAdmin:fedoraAdmin -X POST http://localhost:8080/rest/container1/fcr:reindex
```
It should return a 204. 

Then do
```
curl -v -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/container1
```
to verify that it was successfully side loaded.

# Interested parties


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/AbstractResourceIT.java
```diff
@@ -634,6 +635,12 @@ protected static void assertDeleted(final String id) {
         assertThat("Expected object to be deleted", getStatus(new HttpGet(location)), is(GONE.getStatusCode()));
     }
 
+    protected static void assertNotFound(final String id) {
+        final String location = serverAddress + id;
+        assertThat("Expected object to return 404", getStatus(new HttpHead(location)), is(NOT_FOUND.getStatusCode()));
+        assertThat("Expected object to return 4040", getStatus(new HttpGet(location)), is(NOT_FOUND.getStatusCode()));
```
pwinckles, line 641: Typo `4040`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/ReindexServiceImpl.java
```diff
@@ -0,0 +1,48 @@
...
+    private PersistentStorageSessionManager persistentStorageSessionManager;
+
+    @Inject
+    private ReindexResourceOperationFactory resourceOperationFactory;
+
+    @Override
+    public void reindexByFedoraId(final String txId, final String principal, final FedoraId fedoraId) {
+        final var operation = resourceOperationFactory.create(fedoraId).userPrincipal(principal).build();
+        final var psession = persistentStorageSessionManager.getSession(txId);
+        psession.persist(operation);
```
pwinckles, line 46: This operation will need to acquire a lock on the resource. Locking coming in: #1823

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraReindex.java
```diff
@@ -0,0 +1,95 @@
...
+        try {
+            final var transaction = transaction();
+            final var id = FedoraId.create(externalPath);
+            if (doesResourceExist(transaction, id, false)) {
+                //TODO : remove this block once reindexing of existing resources is supported.
+                return status(HttpStatus.SC_CONFLICT, "Reindexing of existing resources is not currently supported. " +
+                        "Only resources that have not yet been indexed are allowed.").build();
+            } else {
+                this.reindexService.reindexByFedoraId(transaction().getId(), getUserPrincipal(), id.asBaseId());
+                transaction.commitIfShortLived();
```
pwinckles, line 84: It seems like there might not be a great solution for this currently, but I believe reindexing is applying live changes to the search index that will not be rollback if the commit ultimately fails.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraReindexIT.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+@TestExecutionListeners(
+        listeners = {LinuxTestIsolationExecutionListener.class},
+        mergeMode = TestExecutionListeners.MergeMode.MERGE_WITH_DEFAULTS)
+public class FedoraReindexIT extends AbstractResourceIT {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(FedoraReindexIT.class);
+
+    @Test
+    public void testSideLoading() throws Exception {
```
pwinckles, line 43: This is a happy path test. Other test ideas:

1. What happens if you attempt to reindex an object that does not exist?
2. What happens if you attempt to reindex an archival group part rather than the ag? For example, `info:fedora/ag/child` rather than `info:fedora/ag`
3. What happens if the object exists but is invalid?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraReindex.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    @POST
+    @Produces({APPLICATION_JSON + ";qs=1.0",
+            TEXT_PLAIN_WITH_CHARSET})
+    public Response reindexObject() {
+        LOGGER.info("receiving reindex request for fedora_id = {}", externalPath);
+        try {
+            final var transaction = transaction();
+            final var id = FedoraId.create(externalPath);
+            if (doesResourceExist(transaction, id, false)) {
```
pwinckles, line 78: I think this should be `doesResourceExist(transaction, id, true)`. Otherwise I think it will return `false` for resources that are deleted by not purged.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexResourcePersister.java
```diff
@@ -0,0 +1,63 @@
...
+    public boolean handle(final ResourceOperation operation) {
+        return operation != null && REINDEX.equals(operation.getType());
+    }
+
+    @Override
+    public void persist(final OcflPersistentStorageSession session, final ResourceOperation operation)
+            throws PersistentStorageException {
+        final var ocflId = operation.getResourceId().getBaseId();
+        try {
+            this.reindexService.indexOcflObject(session.getId(), ocflId);
```
pwinckles, line 58: This indexing does not do any sort of validation. I am okay giving that a pass for this PR, but, if we don't do validation here, then we need a follow up ticket to validate side-loaded objects. I don't know if it's a beta blocker (seems very important to me because it's likely people will create invalid objects by hand) or not, but definitely a release blocker.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraReindex.java
dbernstein: We're passing in the transactionId to the reindex service which ultimately gets passed down to the DbSearchIndex.   Isn't it a matter of adding transaction support to the search index was we're doing with other indices like the FedoraToOcflObjectIndex (which could be added in a downstream ticket)?

pwinckles: Yes, if the search table was updated in the same way as the other tables that would solve the problem. I guess I had thought that you didn't want the search index to have a sibling transaction table? Won't that impact all of the queries?

pwinckles: Or are you thinking that changes in the transaction table will never be visible until they're merged into the main table?

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexResourcePersister.java
dbernstein: https://jira.lyrasis.org/browse/FCREPO-3560

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraReindex.java
dbernstein: I'm thinking the search table would work like the others.  I'll create a jira for that.

dbernstein: https://jira.lyrasis.org/browse/FCREPO-3561

pwinckles (commented):
I tested locally and it is working.

Outstanding issues that should be fixed in this PR;

* Need to acquire the resource lock
* It looks like the Windows build failed on one of the new tests -- not sure if it's transient

Outstanding issues to be addressed in future PRs:

* Object validation
* Search transactions

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraReindex.java
```diff
@@ -0,0 +1,125 @@
...
+            final var transaction = transaction();
+            final var id = FedoraId.create(externalPath);
+            if (doesResourceExist(transaction, id, true)) {
+                //TODO : remove this block once reindexing of existing resources is supported.
+                return status(HttpStatus.SC_CONFLICT, "Reindexing of existing resources is not currently supported. " +
+                        "Only resources that have not yet been indexed are allowed.").build();
+            } else {
+                try {
+                    final var baseId = id.asBaseId();
+                    transaction.lockResource(baseId);
```
pwinckles, line 88: This is inconsistent with how locking is done in the other services. It will work, but, to me, having the kernel acquire a lock rather than the http layer seems more appropriate.

dbernstein: @pwinckles :  I'm assuming the lock should also be released in the kernel layer as well (ie in the same routine as it is being done here in the http layer)?

pwinckles: The way all of the other services work is that the kernel layer acquires the lock and the http layer initiates the transaction commit. The locks cannot be released before the transaction is committed, so so long as the transaction is committed in the http layer, the locks must also be released there.

If it is possible to push the commit down to the kernel, that does seem better to me, but I think we need to be consistent so that it's easy to reason about.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexService.java
```diff
@@ -146,7 +145,11 @@ public void indexOcflObject(final String txId, final String ocflId) {
             });
 
             if (rootId.get() == null) {
-                throw new IllegalStateException(String.format("Failed to find root resource in object %s", ocflId));
+                throw new IllegalStateException(String.format("Failed find to the root resource in object identified" +
```
awoods, line 148: Jumbled grammar: "Failed find to the root resource"

awoods (changes requested):
Minor comments.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/BaseTestExecutionListener.java
```diff
@@ -0,0 +1,57 @@
...
+import org.fcrepo.kernel.api.ContainmentIndex;
+import org.fcrepo.persistence.ocfl.api.FedoraToOcflObjectIndex;
+import org.fcrepo.search.api.SearchIndex;
+import org.springframework.test.context.TestContext;
+import org.springframework.test.context.support.AbstractTestExecutionListener;
+
+/**
+ * A base class test execution listener class.
+ *
+ * @author pwinkles
```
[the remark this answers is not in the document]
awoods: If so, you misspelled your name.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/BaseTestExecutionListener.java
```diff
@@ -0,0 +1,57 @@
...
+public class BaseTestExecutionListener extends AbstractTestExecutionListener {
+
+    protected void cleanDb(final TestContext testContext) {
+        final ContainmentIndex containmentIndex = getBean(testContext, "containmentIndex");
+        final FedoraToOcflObjectIndex ocflIndex = getBean(testContext, "ocflIndex");
+        final SearchIndex searchIndex = getBean(testContext, "searchIndex");
+
+        containmentIndex.reset();
+        ocflIndex.reset();
+        searchIndex.reset();
```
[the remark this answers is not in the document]
dbernstein: Thanks - I'll take a look.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/ReindexExecutionListener.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Isolate ReindexITs from the rest of the IT contexts as well as tests within ReindexIT.
+ *
+ * @author dbernstein
+ */
+public class ReindexExecutionListener extends BaseTestExecutionListener {
+
+    @Override
+    public void afterTestMethod(final TestContext testContext) throws Exception {
+        FileUtils.deleteDirectory(Path.of("target/fcrepo-home/data/ocfl-root").toFile());
```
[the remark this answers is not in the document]
dbernstein: Okay - that all sounds good. I'll try it.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/BaseTestExecutionListener.java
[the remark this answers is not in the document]
dbernstein: Since it was a refactoring and it included only your previous work,  I wanted to give credit where it was due.  I'll fix the spelling!

awoods (approved):

pwinckles (approved):

PR #1829, opened 2020-12-04
Encode internal URIs in stored RDF documents.

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3539

* Adds a new `getEncodedFullId()` method to the `FedoraId` class to ensure when clients use spaces in their IDs they don't mess up the RDF.  
* It stops encoding the path as it is translated from internal to external URIs.
* Removes the `bodyToInternalStream()` method as it was only referenced by its own test.

# How should this be tested?

Create objects with complex and strange URIs both as POST and PUT and ensure that the resulting graph looks correct.

# Interested parties


pwinckles (approved):
👍

PR #1835, opened 2020-12-08
fixed MySQL transaction problems and improved commit robustness

pwinckles:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3581 and https://jira.lyrasis.org/browse/FCREPO-3543

* MySQL can fail if a competing transaction updates the same record. The only solution is to retry. This PR adds retries to the transaction commit.
* Reorders the commit operations so that all of the db changes are committed before the ocfl changes

# How should this be tested?

Everything should still work and MySQL should not fail due to a race condition. I ran the test that exposed the problem, `FedoraLdpIT.testConcurrentPutsWithPairtrees()`, 20+ times with this PR and it did not fail.

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/TransactionsIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/TransactionsIT.java
@@ -291,7 +291,7 @@ public class TransactionsIT extends AbstractResourceIT {
     }
 
     @Test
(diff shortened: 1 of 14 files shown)


pwinckles:
I'm going to try adding a few more rollback tests that are triggered by ocfl write failures.

pwinckles:
Okay, this is ready for review again

awoods (approved):

PR #1851, opened 2021-01-16
Alter preferTags to use enum (Minimal container triples)

whikloj:
Setup new ITs.

**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3334

Makes an new enum to allow for identifying include, exclude or silent prefer headers.

Made new `PrefersLdpIT` to hold prefer header tests. Moved several tests from `FedoraLdpIT`  to `PrefersLdpIT`

# How should this be tested?

As this is essentially about minimal container triples.
Make a resource, add some Rdf, and a contained resource.
GET the resource.
See user triples, server managed triples and containment triples.
GET the resource with `Prefer: return=representation; include="https://www.w3.org/ns/ldp#PreferMinimalContainer"`
See user triples and server managed triples.
GET the resource with `Prefer: return=representation; omit="https://www.w3.org/ns/ldp#PreferMinimalContainer"`
See containment triples.

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/EtagService.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/EtagService.java
@@ -74,23 +74,23 @@ public class EtagService {
 
         // Factor in preferences which change which triples are included in the response
         etag.append('|');
-        if (!prefers.prefersContainment().equals(LdpTriplePreferences.preferChoice.EXCLUDE)) {
+        if (prefers.displayContainment()) {
             final var lastUpdated = containmentIndex.containmentLastUpdated(txId, resource.getFedoraId());
             if (lastUpdated != null) {
                 etag.append(lastUpdated);
             }
         }
         etag.append('|');
-        if (!prefers.prefersMembership().equals(LdpTriplePreferences.preferChoice.EXCLUDE)) {
+        if (prefers.displayMembership()) {
             final var lastUpdated = membershipService.getLastUpdatedTimestamp(txId, resource.getFedoraId());
             if (lastUpdated != null) {
                 etag.append(lastUpdated);
             }
         }
-        addComponent(etag, prefers.prefersEmbed());
-        addComponent(etag, prefers.preferMinimal());
-        addComponent(etag, prefers.prefersReferences());
-        addComponent(etag, prefers.prefersServerManaged());
+        addComponent(etag, prefers.displayEmbed());
+        addComponent(etag, prefers.displayUserRdf());
+        addComponent(etag, prefers.displayReferences());
+        addComponent(etag, prefers.displayServerManaged());
 
         // Compute a digest of all these components to use as the etag
         final String etagMd5 = DigestUtils.md5Hex(etag.toString()).toUpperCase();
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/EtagService.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/EtagService.java
@@ -74,21 +74,21 @@ public class EtagService {
 
         // Factor in preferences which change which triples are included in the response
         etag.append('|');
-        if (prefers.prefersContainment()) {
+        if (!prefers.prefersContainment().equals(LdpTriplePreferences.preferChoice.EXCLUDE)) {
             final var lastUpdated = containmentIndex.containmentLastUpdated(txId, resource.getFedoraId());
             if (lastUpdated != null) {
                 etag.append(lastUpdated);
             }
         }
         etag.append('|');
-        if (prefers.prefersMembership()) {
+        if (!prefers.prefersMembership().equals(LdpTriplePreferences.preferChoice.EXCLUDE)) {
             final var lastUpdated = membershipService.getLastUpdatedTimestamp(txId, resource.getFedoraId());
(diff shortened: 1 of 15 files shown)


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/LdpTriplePreferences.java
```diff
@@ -25,37 +25,48 @@
 public interface LdpTriplePreferences {
 
     /**
-     * @return Whether to return a minimal container.
+     * What the prefer tag choice is.
      */
-    boolean getMinimal();
+    enum preferChoice {
```
awoods, line 30: This `enum` declaration can be simplified to:
```
    enum preferChoice {
        INCLUDE,
        EXCLUDE,
        SILENT
    }
```

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/EtagService.java
```diff
@@ -74,21 +74,21 @@ public String getRdfResourceEtag(final String txId, final FedoraResource resourc
 
         // Factor in preferences which change which triples are included in the response
         etag.append('|');
-        if (prefers.prefersContainment()) {
+        if (!prefers.prefersContainment().equals(LdpTriplePreferences.preferChoice.EXCLUDE)) {
```
awoods, line 77: The double-negative probably warrants a clarifying, inline comment:
```
// Should containment be included, i.e. not excluded?
```

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/EtagService.java
```diff
@@ -74,21 +74,21 @@ public String getRdfResourceEtag(final String txId, final FedoraResource resourc
...
             final var lastUpdated = containmentIndex.containmentLastUpdated(txId, resource.getFedoraId());
             if (lastUpdated != null) {
                 etag.append(lastUpdated);
             }
         }
         etag.append('|');
-        if (prefers.prefersMembership()) {
+        if (!prefers.prefersMembership().equals(LdpTriplePreferences.preferChoice.EXCLUDE)) {
```
awoods, line 84: // Should membership be included, i.e. not excluded?

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/PrefersLdpIT.java
```diff
@@ -0,0 +1,704 @@
...
+        getReq.setHeader("Prefer", preferLink(
+                null,
+                PREFER_SERVER_MANAGED.getURI()));
+        try (final var dataset = getDataset(getReq)) {
+            final var graph = dataset.asDatasetGraph();
+            assertTrue(graph.contains(userRdfTriple));
+            assertFalse(graph.contains(systemManagedPropertyTriple));
+            // Containment is consider system managed.
+            assertFalse(graph.contains(containmentTriple));
+            assertTrue(graph.contains(membershipTriple));
```
awoods, line 291: It seems that "membership" is just as "server-managed" as "containment", no?
..which argues for omitting "membership" triples here as well.

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/rdf/LdpTriplePreferences.java
```diff
@@ -25,37 +25,48 @@
...
+            this.value = value;
+        }
+    }
 
     /**
-     * @return Whether this prefer tag demands membership triples.
+     * @return Whether this preference tag asks to include or omit a minimal container or is silent.
      */
-    boolean prefersMembership();
+    preferChoice preferMinimal();
```
awoods, line 45: The logic of the pre-PR interface is much simpler and readable than the `!preferences.preferX().equals(EXCLUDE)`, etc. Would it be possible to keep the existing interface and bury the new enum-based logic in the `LdpPreferTag` implementation?

awoods (changes requested):

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/PrefersLdpIT.java
whikloj: I was going off of past tests. As there is a specific tag for including or excluding containment I think removing them them from the "system managed" category would be clearer. But 🤷

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/ldp/LdpPreferTag.java
```diff
@@ -99,32 +101,58 @@ private preferChoice getChoice(final Property tag) {
...
+        // If we did ask for references. (Not shown by default).
+        return references.equals(INCLUDE);
     }
 
     @Override
-    public preferChoice prefersEmbed() {
-        return embed;
+    public boolean displayEmbed() {
+        // If we did ask for embedded resources. (Not shown by default).
+        return embed.equals(preferChoice.INCLUDE);
```
awoods, line 136: May as well join the club, and drop the `preferChoice` qualifier.

awoods (changes requested):

awoods (approved):

PR #1852, opened 2021-01-20
add rebuild validation

pwinckles:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3560

**This PR will not build without https://github.com/fcrepo/fcrepo-storage-ocfl/pull/28**

Adds a validation step to rebuild/reindex. OCFL objects that are not valid F6 objects will not be accepted.

# How should this be tested?

Create an invalid F6 and either attempt to start F6 on top of it or attempt to side load it. The object should not be indexed and there should be errors in the logs indicating why the object was not accepted.

# Interested parties
@fcrepo/committers


pwinckles:
The objects currently produced by migration-utils may contain bad digests. `-Dfcrepo.rebuild.validation.fixity=false` can be used to skip the fixity check.

dbernstein (dismissed):
@pwinckles : Looks good.  Just to confirm here:  if a validation fails, a runtime exception will be immediately (ie on the first object to fail), all problems will be written to the log, and Fedora  will start up anyway.  So unless the user scans the log,  they won't know of validation failures.  Which shouldn't be a problem since we're planning to add validation on read and write operations which will alert the user because operations will fail.    On the other hand, we were talking about just validating the resource headers file rather than the entire object on read/write.  In the case that the resource header is fine,  would that mean that the only way the user could get a list of valid objects would be to grep the logs after starting up with a  fresh rebuild?

pwinckles:
> 
> 
> @pwinckles : Looks good. Just to confirm here: if a validation fails, a runtime exception will be immediately (ie on the first object to fail), all problems will be written to the log, and Fedora will start up anyway. So unless the user scans the log, they won't know of validation failures. Which shouldn't be a problem since we're planning to add validation on read and write operations which will alert the user because operations will fail. On the other hand, we were talking about just validating the resource headers file rather than the entire object on read/write. In the case that the resource header is fine, would that mean that the only way the user could get a list of valid objects would be to grep the logs after starting up with a fresh rebuild?

Because the rebuild is a process that runs on server start, there really isn't another way to report these errors other than in the logs.

On the other hand, the error messages could be exposed in the response to HTTP requests, if that it something we wanted to do.

Currently, the only HTTP request that the validation logic runs on is reindex, and I do not believe it is returning the specific validation problems. So, yes, in that case, the user would need to check the logs after getting a reindex error.

dbernstein (approved):
Excellent.   Thank you!

PR #1856, opened 2021-01-29
fix flapping indirect container test

pwinckles:
I think the problem was that the resource was created before it was added to the indirect container, so whether or not the original version was a member of the indirect container varied based on whether or not the resource and the proxy were created in the same second.


pwinckles:
nope

bbpennel (approved):
Looks good, should help to guarantee consistent behavior. I wonder if my 1.5 second sleeps are overkill, but its probably fine.

PR #1858, opened 2021-02-03
Fixes broken ExternalContentHandlerIT tests.

dbernstein:
It appears that <uid>.example.com now responds with a 201.  This update changes those invalid external content links with a non-existent domain.

**WAIT TO MERGE THIS: Apparently others are not seeing this problem - let's wait to see if others begin experiencing it**

# How should this be tested?
mvn clean install

# Interested parties


pwinckles:
Seems like this is a non-issue

PR #1859, opened 2021-02-05
Account for servletpath of index.html

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3625

Inside Tomcat if you browse to the root of the context, Tomcat assumes a servletPath of `/index.html`. This caused the existing logic (which assumed a path and not filename) would break.

# How should this be tested?

Deploy fedora to Tomcat at context `fcrepo` (for example), browse to `http://localhost:8080/fcrepo` and get a 500 Server Error.
Rebuild with this PR and re-deploy. Successfully get the splash page.

# Interested parties


pwinckles:
The splash page in `main` works for @dbernstein and I in Tomcat. I tried the PR code, and it still works. Is more investigation needed, or do you just want me to merge?

whikloj:
No if this is only occurring for me then adding this seems like a bad idea. I'll close this until/if we get another person with the issue.

PR #1863, opened 2021-02-08
Make the default persistence digest algorithm configurable from the p…

whikloj:
…roperties file.

**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3520

Enables the use of `fcrepo.persistence.defaultDigestAlgorithm` in the configuration file to change the default OCFL digest algorithm

# How should this be tested?

1. Try creating a new Fedora repository and add a binary object. View the binary description and see the `premis:hasMessageDigest` starts wtith `urn:sha-512:`
1. Create a text file with the line
     ```
     fcrepo.persistence.defaultDigestAlgorithm=sha256
     ```
1. Restart Fedora and add another binary object. View the description for this second object and see that now the `premis:hasMessageDigest` starts wtith `urn:sha-256:`

# Interested parties

### DIFF
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
@@ -18,16 +18,19 @@
 
 package org.fcrepo.config;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import javax.annotation.PostConstruct;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Configuration;
-
-import javax.annotation.PostConstruct;
-import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.Objects;
 
 /**
  * Fedora's OCFL related configuration properties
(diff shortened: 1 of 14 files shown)


fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
```diff
@@ -133,6 +151,16 @@ private void postConstruct() throws IOException {
             LOGGER.info("Fedora OCFL S3 bucket: {}", ocflS3Bucket);
             LOGGER.info("Fedora OCFL S3 prefix: {}", ocflS3Prefix);
         }
+        FCREPO_DIGEST_ALGORITHM = DigestAlgorithm.fromAlgorithm(FCREPO_DIGEST_ALGORITHM_VALUE);
```
pwinckles, line 154: I feel a little uneasy about moving `DigestAlgorithm` into `fcrepo-config`. The module is not really supposed to have "code" in it. What about moving it to `fcrepo-common` instead, and doing the digest resolution/validation logic inside `OcflPersistenceConfig`?

whikloj: `DigestAlgorithm` is an `enum`, much like [`Storage`](https://github.com/fcrepo/fcrepo/blob/main/fcrepo-configs/src/main/java/org/fcrepo/config/Storage.java) which already exists in `fcrepo-config` there which is why I thought that was allowed. 
I need to validate the configuration which is why I moved this into fcrepo-config, the DigestAlgorithm processing allows people more flexibility in the text they use (ie. sha512, SHA-512, sha-512). 

So I can do string comparisons or I can have a dependency on fcrepo-common (which I thought was a bad idea) if you'd like the `enum` moved

pwinckles: That's fair. Let's leave it.

pwinckles (approved):
👍

PR #1866, opened 2021-02-10
Only update mementoCreatedDate on new version

whikloj:
**JIRA Ticket**: Follow on from https://jira.lyrasis.org/browse/FCREPO-3608

Updates the CreateVersionPersister to only update the mementoCreatedDate on a new version.
Also adds a new test to ensure the memento and original resource remain the same.

# How should this be tested?

Create a resource, get the body of the resource.
Create a memento of the resource, get the body of the Memento.
They should match.
Get a new copy of the resource body.
It should also match the two previous. 

# Interested parties


dbernstein (approved):
Nicely done @whikloj :  you are an unstoppable force of nature.  I verified that the related testsuite tests are passing.

PR #1869, opened 2021-02-12
Enable updating SMTs when in relaxed mode

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3225

Alters the HTTP layer to not throw exceptions for SMTs if we are in relaxed mode and instead continue passing the model down.
Then it changes the kernel layer to filter any SMTs so they are not persisted to the user RDF and are only used by the relaxedProperties method.
Lastly it uses some existing work from @mikedurbin because the tests we had put `Date`s in the RDF but when we actually get the RDF back it is a `XSDDateTime` and these methods properly parse that.

# How should this be tested?

Switch the server to relaxed mode by adding a `fcrepo.properties.management=relaxed` to your configuration file. Then try PUT/POST or PATCH with server managed triples.
ie.
```
curl -ufedoraAdmin:fedoraAdmin -H"Slug: test" -XPOST -H"Content-type: text/turtle" -d" <> <http://purl.org/dc/elements/1.1/title> \"Some title\" ; <http://fedora.info/definitions/v4/repository#created> \"1900-01-01T01:00:00Z\"^^<http://www.w3.org/2001/XMLSchema#dateTime> ." http://localhost:8080/rest
```
Then GET `http://localhost:8080/rest/test` and see a very old object 😉 

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
@@ -145,28 +145,27 @@ public class HttpRdfService {
                 } catch (final ServerManagedPropertyException | ServerManagedTypeException exc) {
                     if (!fedoraPropsConfig.getServerManagedPropsMode().equals(RELAXED)) {
                         exceptions.add(exc);
+                        continue;
                     }
-                } finally {
-                    // Still need to convert the model, incase we are not throwing the exceptions above.
-                    if (stmt.getSubject().isURIResource()) {
-                        final String originalSubj = stmt.getSubject().getURI();
-                        final String subj = idTranslator.translateUri(originalSubj);
+                }
+                if (stmt.getSubject().isURIResource()) {
+                    final String originalSubj = stmt.getSubject().getURI();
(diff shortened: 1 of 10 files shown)


bbpennel (changes requested):
It looks like the behavior of relaxed mode is a little more relaxed now than in fcrepo5. The outcomes of the following commands are different in fedora 5 vs 6 with relaxed mode enabled:
```
echo 'insert { <> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://fedora.info/definitions/v4/repository#Binary> } where {}' |
   curl -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/test -XPATCH -H"Content-Type: application/sparql-update" --data-binary @-
```
In 6, it responds with a 204 but ignores the new property. In 5, it gives a 409 and says that it rejects the property. This holds true for PUT requests too, where I'm now able to PUT the GET response directly back to the server.

Personally I prefer the new behavior, particularly for PUT, but wanted to note that this is a change in behavior in case others care.

Otherwise, this seems to work as expected, good work.

bbpennel (approved):

PR #1870, opened 2021-02-15
add web.xml with auth disabled

pwinckles:
**Jira**: https://jira.lyrasis.org/browse/FCREPO-3634

* Adds a version of the web.xml with auth disabled that is bundled in the war

# How should this be tested?

Everything should work the same as before. The auth disabled web.xml must be manually specified: https://wiki.lyrasis.org/display/FEDORA6x/Disable+Authentication+and+Authorization

# Interested parties
@fcrepo/committers


whikloj (approved):
👍

PR #1874, opened 2021-02-17
add rebuild progress message

pwinckles:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3467

Adds a log message that reports on the status of the rebuild process

# How should this be tested?

Start Fedora on a large migrated repo and see it report its progress in the logs.

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexManager.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexManager.java
@@ -180,7 +180,8 @@ public class ReindexManager {
                     final var now = Instant.now();
                     final var duration = Duration.between(startTime, now);
                     LOGGER.info("Index rebuild progress: Complete: {}; Errored: {}; Time: {}; Rate: {}/s",
-                            complete, errored, getDurationMessage(duration), (complete + errored) / duration.getSeconds());
+                            complete, errored, getDurationMessage(duration),
+                            (complete + errored) / duration.getSeconds());
                 }
             } catch (InterruptedException e) {
                 // processing has completed exit normally
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexManager.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexManager.java
@@ -25,11 +25,14 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Stream;
 
 import org.fcrepo.config.OcflPropsConfig;
 import org.slf4j.Logger;
+
+import com.google.common.base.Stopwatch;
 
 /**
  * Class to coordinate the index rebuilding tasks.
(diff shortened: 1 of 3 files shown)


dbernstein (approved):
@pwinckles : this looks good.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexWorker.java
```diff
@@ -70,18 +77,29 @@ public void start() {
...
         while (running) {
             final List<String> ids = manager.getIds();
             if (ids.isEmpty()) {
                 stopThread();
                 break;
             }
+
             int completed = 0;
             int errors = 0;
+            long reportingInterval = REPORTING_INTERVAL_SECS + jitter();
```
dbernstein, line 90: @pwinckles  Just curious what the expected outcome is here - is the idea that adding random seconds should smooth out the reporting interval - ie make it more regular?

dbernstein (approved):

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexWorker.java
pwinckles: Jitter is generally useful to try to reduce resource contention. So, in this case, if multiple works report their status at the same time, one of them will have to wait, briefly. Whereas, if they report at different times, neither have to wait. Here, I don't think it ultimately matters that much with or without it. I can remove it, if you'd prefer.

dbernstein: No I think it's fine.  I was just curious.  I appreciate you introducing me to that technique.

dbernstein: @pwinckles :  would you mind just adding a comment indicating the purpose for the jitter() in case someone else is unfamiliar that technique?

dbernstein (changes requested):

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexWorker.java
pwinckles: I decided to change to use a reporting thread. Now the status will be reported every 5 minutes and the jitter is irrelevant

dbernstein (approved):

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexWorker.java
dbernstein: 👍 : I'll try it out now.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexManager.java
```diff
@@ -160,4 +169,38 @@ public void rollback() {
     public void shutdown() {
         ocflStream.close();
     }
+
+    private Thread startReporter() {
+        final var reporter = new Thread(() -> {
+            final var startTime = Instant.now();
```
dbernstein, line 175: An log statement here (or wherever you think is appropriate) communicating to the users that they can expect progress updates every 5 minutes (or whatever the reporting interval is) so they're not left hanging for 5 minutes not knowing what's going on.

dbernstein (changes requested):

dbernstein (approved):

PR #1877, opened 2021-02-18
make the event bus async

pwinckles:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3587

Makes the guava event bus that's used to emit messages asynchronous, which is, I believe, how people were already assuming it was operating.

# How should this be tested?

Fire up Fedora, listen for events, create a resource, and see that events still work.

# Interested parties
@fcrepo/committers


whikloj (approved):
👍

PR #1878, opened 2021-02-18
Throw exceptions for managed types when in RELAXED mode

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3645

To more closely mimic Fedora 5s behaviour this alters Fedora 6 to throw an exception when a user adds a managed type in their RDF when the server is in relaxed mode. Previously it just ignored the triple.

# How should this be tested?

1. Start Fedora in relaxed mode. ie. (`fcrepo.properties.management=relaxed`)
1. Create a resource a new resource.
1. GET the resource without server managed properties, or remove them from the result to make it easier to see what is happening.
1. Add a server managed type to the body, like `<> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/ns/ldp#NonRDFSource>`
1. Try to PUT this back to the server.
1. In `main` you will succeed but nothing changes.
1. In this PR you will get a 409 Conflict.

# Interested parties
@fcrepo/committers, specifically @bbpennel

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
@@ -21,7 +21,7 @@ package org.fcrepo.http.api.services;
 import static org.apache.jena.rdf.model.ModelFactory.createDefaultModel;
 import static org.apache.jena.rdf.model.ResourceFactory.createProperty;
 import static org.apache.jena.riot.RDFLanguages.contentTypeToLang;
-import static org.fcrepo.config.ServerManagedPropsMode.RELAXED;
+import static org.fcrepo.config.ServerManagedPropsMode.STRICT;
 import static org.fcrepo.kernel.api.RdfLexicon.isManagedPredicate;
 import static org.fcrepo.kernel.api.RdfLexicon.restrictedType;
 import static org.fcrepo.kernel.api.utils.RelaxedPropertiesHelper.checkTripleForDisallowed;
@@ -136,17 +136,26 @@ public class HttpRdfService {
 
         while (stmtIterator.hasNext()) {
             final Statement stmt = stmtIterator.nextStatement();
-            if (lenientHandling && stmtIsServerManaged(stmt)) {
+            if (lenientHandling && stmtIsServerManaged(stmt) &&
+                    fedoraPropsConfig.getServerManagedPropsMode().equals(STRICT)) {
                 // Remove any statement that touches a server managed property or namespace.
                 stmtIterator.remove();
             } else {
                 try {
                     checkForDisallowedRdf(stmt);
-                } catch (final ServerManagedPropertyException | ServerManagedTypeException exc) {
-                    if (!fedoraPropsConfig.getServerManagedPropsMode().equals(RELAXED)) {
+                } catch (final ServerManagedPropertyException exc) {
+                    if (fedoraPropsConfig.getServerManagedPropsMode().equals(STRICT)) {
                         exceptions.add(exc);
                         continue;
                     }
+                } catch (final ServerManagedTypeException exc) {
+                    if (lenientHandling) {
+                        // Remove the invalid statement because client specified lenient handling.
+                        stmtIterator.remove();
+                    } else {
+                        exceptions.add(exc);
+                    }
+                    continue;
                 }
                 if (stmt.getSubject().isURIResource()) {
                     final String originalSubj = stmt.getSubject().getURI();
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
@@ -41,6 +41,7 @@ import org.fcrepo.kernel.api.RdfStream;
 import org.fcrepo.kernel.api.exception.ConstraintViolationException;
 import org.fcrepo.kernel.api.exception.MalformedRdfException;
 import org.fcrepo.kernel.api.exception.MultipleConstraintViolationException;
+import org.fcrepo.kernel.api.exception.RelaxableServerManagedPropertyException;
 import org.fcrepo.kernel.api.exception.RepositoryRuntimeException;
 import org.fcrepo.kernel.api.exception.ServerManagedPropertyException;
 import org.fcrepo.kernel.api.exception.ServerManagedTypeException;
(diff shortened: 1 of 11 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraRelaxedLdpIT.java
```diff
@@ -351,6 +363,108 @@ public void testRoundtripping() throws IOException {
...
+                "<" + CREATED_DATE + "> ?date . } INSERT { <> <" + RDF.type + "> <" + DIRECT_CONTAINER + "> ; " +
+                "<" + CREATED_DATE + "> \"" + newDate + "\"^^<" +
+                XSDDateTimeType.XSDdateTime.getURI() + "> . } WHERE { <> <" + RDF.type + "> ?type ; " +
+                "<" + CREATED_DATE + "> ?date . } "));
+        assertEquals(CONFLICT.getStatusCode(), getStatus(patchBinaryDescription));
+        // Check that the date didn't change either.
+        try (final var dataset = getDataset(new HttpGet(binaryUri + "/" + FCR_METADATA))) {
+            final var graph = dataset.asDatasetGraph();
+            // Assert it has a NonRDFSource type
+            assertTrue(graph.contains(ANY, binaryNode, RDF.type.asNode(), NON_RDF_SOURCE.asNode()));
```
bbpennel, line 460: Should probably also check that it doesn't have the `DIRECT_CONTAINER` type

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraRelaxedLdpIT.java
```diff
@@ -351,6 +363,108 @@ public void testRoundtripping() throws IOException {
...
+            assertTrue(graph.contains(ANY, binaryNode, CREATED_DATE.asNode(),
+                    createLiteral(oldCreatedDate, XSDDateTimeType.XSDdateTime)));
+            assertFalse(graph.contains(ANY, binaryNode, CREATED_DATE.asNode(),
```
bbpennel, line 463: might be safest to verify that there is only one CREATED_DATE, rather than that this specific value isn't present. Just suggesting this since I thought there was another PR recently that was verify timestamps matched within a certain precision range. If that's not a concern, then fine to ignore this

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/utils/RelaxedPropertiesHelper.java
```diff
@@ -147,8 +147,9 @@ private static Calendar parseExpectedXsdDateTimeValue(final RDFNode node) {
      * @param triple the triple to check.
      */
     public static void checkTripleForDisallowed(final Triple triple) {
-        if (triple.getPredicate().equals(type().asNode()) && !triple.getObject().isURI()) {
-            // The object of a rdf:type triple is not a URI.
+        if (triple.getPredicate().equals(type().asNode()) && !triple.getObject().isVariable() &&
```
bbpennel, line 150: Hadn't considered variables being supplied for RDF types... Is this for PATCH requests that set types based on a variable? Are there any tests that demonstrate this restriction?

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
```diff
@@ -136,17 +136,26 @@ public Model bodyToInternalModel(final FedoraId extResourceId, final InputStream
...
                 try {
                     checkForDisallowedRdf(stmt);
-                } catch (final ServerManagedPropertyException | ServerManagedTypeException exc) {
-                    if (!fedoraPropsConfig.getServerManagedPropsMode().equals(RELAXED)) {
+                } catch (final ServerManagedPropertyException exc) {
+                    if (fedoraPropsConfig.getServerManagedPropsMode().equals(STRICT)) {
                         exceptions.add(exc);
                         continue;
                     }
+                } catch (final ServerManagedTypeException exc) {
```
bbpennel, line 151: This approach still allows me to provide non-relaxed SMTs and have them get silently ignored rather than 409'ed without specifying lenient behavior. For instance:
```
echo "<> <http://www.loc.gov/premis/rdf/v1#hasMessageDigest> 'helloworld'" |
curl -ufedoraAdmin:fedoraAdmin -XPUT http://localhost:8080/rest/binary_with_digests/fcr:metadata --data-binary @- -H"Content-Type: text/turtle" -H"Prefer: handling=lenient; received='minimal'"
```
Gives a 204, but ignores the digest property. 

What would you think about changing `ServerManagedTypeException` to be a subclass of `ServerManagedPropertyException`, and then introducing another subclass named `RelaxedServerManagedPropertyException` which is only thrown when relaxed properties are encountered? Then for blocks like this you could just handle `RelaxedServerManagedPropertyException` vs `ServerManagedPropertyException`, since my understanding is that all other SMTs should be treated the same in terms of when to throw exceptions or not.

bbpennel (changes requested):

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/utils/RelaxedPropertiesHelper.java
whikloj: Only PATCH requests seem to have variables but this was needed based on the test I wrote which had something like
```
DELETE { <> rdf:type ?type } INSERT { <> rdf:type <some rdf type> } WHERE { <> rdf:type ?type }
```
In that case this method gets a Variable as the object of the triple. So I didn't want to throw an error for it.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
whikloj: Ok I see what you are saying, I'll add a new exception and split out the tests.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraRelaxedLdpIT.java
whikloj: You are right there was a problem but that was from creating a date from an Instant and then converting an RDF value into an Instant. It lost some precision. I was going to do CREATED_BY, but without WebAC that triple never exists. 
So if I check my concern is that you shouldn't have updated the CREATED_DATE, but testing for only a single value won't confirm it didn't change. I'll change the new date to waaaaay in the past and then I can do more general comparison.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/services/HttpRdfService.java
whikloj: Also, in your example you **did** specify lenient behaviour.

bbpennel: Sorry, I ran it both ways to verify the behavior, the correct command is
```
echo "<> <http://www.loc.gov/premis/rdf/v1#hasMessageDigest> 'helloworld'" |
curl -ufedoraAdmin:fedoraAdmin -XPUT http://localhost:8080/rest/binary_with_digests/fcr:metadata --data-binary @- -H"Content-Type: text/turtle"
```
which gives the same 204 response

whikloj:
I didn't add any tests when you try to update a non-relaxable SMT but also specified `handling=lenient`, it should succeed but possibly not change anything. You want I should add that?

bbpennel:
I think its alright as is. And the behavior for non-relaxed properties is working as I would expect now

bbpennel (approved):

PR #1882, opened 2021-02-19
reject paths that contain the 'fcr:' prefix

pwinckles:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3650

* Removes the `fedora:` prefix path restriction
* Adds a restrictions that paths cannot contain parts prefixed with `fcr:` that are not `fcr:acl` or `fcr:metadata`

# How should this be tested?

Try to create/update/delete resources with parts with `fcr:` prefixes.

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
@@ -4061,31 +4061,40 @@ public class FedoraLdpIT extends AbstractResourceIT {
     }
 
     @Test
-    public void testPostFedoraSlug() throws IOException {
+    public void testPostFcrSlug() throws IOException {
         final HttpPost httpPost = postObjMethod("/");
-        httpPost.addHeader("Slug", "fedora:path");
+        httpPost.addHeader("Slug", "fcr:path");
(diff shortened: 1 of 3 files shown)


fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
```diff
@@ -4554,7 +4563,7 @@ private static String genLongUrl(final String prefix, final int numChars) {
 
     @Test
     public void testLongIdentifier1() throws IOException {
-        final String url = genLongUrl(serverAddress,476);
+        final String url = genLongUrl(serverAddress,475);
```
dbernstein, line 4566: What's this change about?  Also I'm surprised this works since presumably we're posting to a non-existent resource no?  Just curious.

dbernstein (approved):
Looks good.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraLdpIT.java
pwinckles: I don't know understand what was going on. For some reason that test started failing. I thought maybe the random id was a character longer for some reason....

dbernstein (approved):

PR #1883, opened 2021-02-19
Reject disallowed fcr:fixity methods

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3649

Disallows many methods on the `fcr:fixity` endpoints.

# How should this be tested?

See various examples in the ticket, but essentially you should no longer get any response except a 405 Method Not Allowed for any of `DELETE`, `HEAD`, `PUT`, `POST` or `PATCH`

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraFixity.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraFixity.java
@@ -17,6 +17,7 @@
  */
 package org.fcrepo.http.api;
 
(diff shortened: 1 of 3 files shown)


pwinckles (commented):
Works for me!

Would you be willing to change line 110? It currently produces an error message like this:

> org.fcrepo.kernel.impl.models.ContainerImpl@4fce5b37 is not a binary

pwinckles (dismissed):
👍

pwinckles (approved):
:+1:

PR #1889, opened 2021-03-12
corrected slugs to be interpreted as url encoded strings

pwinckles:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3668

* Changes the `Slug` header to be interpreted as a url-encoded string, allowing it to support unicode characters.

# How should this be tested?

The following does not work before this PR and works after.

```
echo "hello world" | curl -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/ -H'Link: <http://www.w3.org/ns/ldp#NonRDFSource>; rel="type"' -H"Slug: %E2%98%9E%E2%98%95%E2%98%9C%E2%98%91" --data-binary @- -XPOST -v

curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/%E2%98%9E%E2%98%95%E2%98%9C%E2%98%91
```

# Interested parties

@fcrepo/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -60,6 +60,7 @@ import static org.slf4j.LoggerFactory.getLogger;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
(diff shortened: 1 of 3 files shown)


dbernstein (approved):
Thank you @pwinckles

PR #1892, opened 2021-04-01
Create automatic version when deleting a resource in a manual versioned repository

whikloj:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3289

Forces a new version to be generated when you delete an atomic resource or an Archival Group (but not a part of an Archival Group)

# How should this be tested?

Start a manually versioned repository.
Create a resource.
Delete the resource.
Investigate the OCFL on disk to see that the only version in the OCFL object has the object as not deleted.
Pull in this PR and repeat.
See that now the object will have a second version persisting the changes to the header's sidecar file.

# Interested parties
@fcrepo/committers


pwinckles (approved):
👍

PR #1894, opened 2021-04-08
expose properties for customizing the s3 endpoint

pwinckles:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3690

Exposes two new properties for customizing the S3 endpoint:

1. `fcrepo.s3.endpoint`: URL of a non-aws s3 endpoint
2. `fcrepo.s3.path.style.access`: Set to `true` if your endpoint needs path style access

# How should this be tested?

Running Fedora against a non-aws endpoint should work if you start Fedora [configured to use s3](https://wiki.lyrasis.org/display/FEDORA6x/Amazon+S3), set `fcrepo.s3.endpoint` to the custom endpoint, and it should work.

# Interested parties

@fcrepo/committers

### DIFF
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
@@ -115,6 +115,9 @@ public class OcflPropsConfig extends BasePropsConfig {
     @Value("${" + FCREPO_PERSISTENCE_ALGORITHM + ":sha512}")
     private String FCREPO_DIGEST_ALGORITHM_VALUE;
 
+    @Value("${fcrepo.ocfl.s3.db.enabled:true}")
+    private boolean ocflS3DbEnabled;
+
     private DigestAlgorithm FCREPO_DIGEST_ALGORITHM;
 
     /**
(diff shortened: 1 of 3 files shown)


pwinckles:
I updated the PR to expose another ocfl s3 related property: `fcrepo.ocfl.s3.db.enabled`. This is `true` by default and can be set to `false` to disable the inventory db cache. 

Disabling the cache will likely have negligible performance impacts, S3 is already really slow. The cache originally existed to solve S3's eventual consistency problem, but S3 is strongly consistent now, though non-AWS implementations may not be.

dbernstein (approved):
The code looks good.  I haven't tested it with a non-aws endpoint but I can see no reason why it shouldn't work.  I'm inclined to merged it in order to get it into the hands of interested users so they can try it out.   @pwinckles  what do you think?

pwinckles:
> 
> 
> The code looks good. I haven't tested it with a non-aws endpoint but I can see no reason why it shouldn't work. I'm inclined to merged it in order to get it into the hands of interested users so they can try it out. @pwinckles what do you think?

Fine with me. I'll update the docs after the merge

noahwsmith:
> I'm inclined to merged it in order to get it into the hands of interested users so they can try it out.

That would be awesome. The way ISLE makes this available to the Islandora 8 community (currently) is by grabbing a named release and building it into their `fcrepo6` image. If you were to issue a `beta-2` with this as part of it, we could test and comment very easily. Thanks again for the quick work on this, however it is to be released!

PR #1896, opened 2021-04-29
Do not use the tx tables for short-lived transactions

pwinckles:
**JIRA Ticket**: https://jira.lyrasis.org/browse/FCREPO-3695

Short-lived transactions do not use the transaction db tables when reading or writing.

# How should this be tested?

Everything should work the same as before. Reads and writes outside of a long-running tx should be faster than before.

# Interested parties

@fcrepo/committers


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexManager.java
```diff
@@ -164,37 +164,18 @@ public int getErrorCount() {
...
-            attempt += 1;
+    private void indexMembership() {
+        try {
+            reindexService.indexMembership(transaction());
+        } catch (final RuntimeException e) {
             try {
-                TimeUnit.SECONDS.sleep(backoff * attempt);
-            } catch (final InterruptedException e) {
-                // Continue on if we are interrupted.
+                reindexService.rollbackMembership(transaction());
```
whikloj, line 173: Using a shortlived transaction (line 225 below) means that the only action that can be rolled back by the `ReindexService` is resetting the search index, all others are now skipped. Not saying this is right or wrong, but we could probably remove this whole `try catch`.

whikloj (changes requested):
Some comments and questions. I restricted my review to @pwinckles 7 commits and I would request @bbpennel have a look at the Membership stuff.

fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/services/CreateResourceServiceImplTest.java
```diff
@@ -217,7 +217,7 @@ public void testNoParentRdf() throws Exception {
     public void testParentAg() {
         final FedoraId fedoraId = FedoraId.create(UUID.randomUUID().toString());
         final FedoraId childId = fedoraId.resolve("child");
-        containmentIndex.addContainedBy(transaction, fedoraId, childId);
+        containmentIndex.addContainedBy(transaction, FedoraId.getRepositoryRootId(), fedoraId);
```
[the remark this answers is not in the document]
pwinckles: The original test was wrong because it was adding the containment relationship for the child before the child was created when it should have been adding the relationship for its parent. The child containment relationship is created when the resource is created in the test.

The test used to pass because the records were merged into the db using an upsert. However, they are now simply being inserted, which exposed the bug in the test.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexManager.java
pwinckles: You're saying remove membership rollback?

whikloj: Doh, I read `rollback` but you removed that method. You can ignore this.

bbpennel (approved):

pwinckles:
I can't physically approve this, but I will verbally. Thanks for working on this @whikloj and @bbpennel!

whikloj (approved):

PR #1897, opened 2021-04-30
enable scheduling

whikloj:
**JIRA Ticket**: n/a

Enables scheduling

# How should this be tested?

1. Enabled TRACE logging and start your fedora, after 3 minutes see no messages.
1. Pull in this PR.
1. Enabled TRACE logging and start your fedora, after 3 minutes see at least
     * `TRACE 19:54:06.699 (TransactionManagerImpl) Cleaning up expired transactions`

# Interested parties

### DIFF
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
@@ -30,6 +30,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.core.io.Resource;
+import org.springframework.scheduling.annotation.EnableScheduling;
 
 /**
  * General Fedora properties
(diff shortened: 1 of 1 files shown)


fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
```diff
@@ -38,6 +39,7 @@
  * @since 6.0.0
  */
 @Configuration
+@EnableScheduling
```
pwinckles, line 42: What do you think about adding this to `WebappConfig` instead?

whikloj: I have no particular feelings on this, other than wondering why is this required now?

pwinckles: My best guess is that I broke it in this commit: https://github.com/fcrepo/fcrepo/commit/81ac61427376db639b2bc12c83520e70c1ad2ff2

This was the old definition: 

```xml
    <task:scheduler id="taskScheduler" />
    <task:executor id="taskExecutor" pool-size="1" />
    <task:annotation-driven executor="taskExecutor" scheduler="taskScheduler" />
```

It's possible that one of those lines was also enabling scheduling.

Personally, I think it makes more sense for the scheduling to be enabled as part of the webapp config.

pwinckles (approved):

PR #1898, opened 2021-04-30
add option to enable unsafe ocfl writing

pwinckles:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3711

**This PR will not build until https://github.com/fcrepo/fcrepo-storage-ocfl/pull/38 is merged**

Adds an option for enabling unsafe OCFL writes. When this option is enabled, the OCFL client will not calculate a digest for files that are added to objects. This should be faster, but is less safe as the object will be corrupted if the digest is incorrect.

Set `fcrepo.ocfl.unsafe.write.enabled=true` to enable unsafe writes.

# How should this be tested?

Fedora should behave the same with or without this feature. It should be impossible to make Fedora create a bad object, including by intentionally giving it an incorrect digest like:

```
curl -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/binary -H "Link: <http://www.w3.org/ns/ldp#NonRDFSource>; rel=type" -v -H "Content-Type: text/plain" -H"digest: sha-512=cb1a576f22e8e3e110611b616e3e2f5ce9bdb941" -X PUT --data-binary "testing"
```

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
@@ -118,6 +118,9 @@ public class OcflPropsConfig extends BasePropsConfig {
     @Value("${fcrepo.ocfl.s3.db.enabled:true}")
     private boolean ocflS3DbEnabled;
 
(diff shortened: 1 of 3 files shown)


whikloj (approved):
👍

PR #1899, opened 2021-05-03
Add cache to FedoraToOcflIndex getMapping

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3712

Relies on: https://github.com/fcrepo/fcrepo-storage-ocfl/pull/39

Add a caffeine cache to the `DbFedoraToOcflIndex` for the `getMapping` calls.

# How should this be tested?

There should  be no functional difference, though I hope it is faster.

# Interested parties

### DIFF
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
@@ -121,10 +121,10 @@ public class OcflPropsConfig extends BasePropsConfig {
     @Value("${fcrepo.ocfl.unsafe.write.enabled:false}")
     private boolean unsafeWriteEnabled;
 
-    @Value("${fcrepo.ocfl.id_map_cache.size:1024}")
+    @Value("${fcrepo.cache.db.ocfl.id_map.size.entries:1024}")
     private long fedoraToOcflCacheSize;
 
-    @Value("${fcrepo.ocfl.id_map_cache.timeout:30}")
+    @Value("${fcrepo.cache.db.ocfl.id_map.timeout.minutes:30}")
     private long fedoraToOcflCacheTimeout;
 
     private DigestAlgorithm FCREPO_DIGEST_ALGORITHM;
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
@@ -121,6 +121,12 @@ public class OcflPropsConfig extends BasePropsConfig {
     @Value("${fcrepo.ocfl.unsafe.write.enabled:false}")
     private boolean unsafeWriteEnabled;
 
+    @Value("${fcrepo.ocfl.id_map_cache.size:1024}")
+    private long fedoraToOcflCacheSize;
+
+    @Value("${fcrepo.ocfl.id_map_cache.timeout:30}")
+    private long fedoraToOcflCacheTimeout;
+
     private DigestAlgorithm FCREPO_DIGEST_ALGORITHM;
 
     /**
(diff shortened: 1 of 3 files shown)


whikloj:
This will fail until the associated PR on fcrepo-storage-ocfl is merged.

pwinckles (commented):
Whoops, I thought I submitted this yesterday, but apparently did not.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndex.java
```diff
@@ -188,6 +202,11 @@
     public DbFedoraToOcflObjectIndex(@Autowired final DataSource dataSource) {
         this.dataSource = dataSource;
         this.jdbcTemplate = new NamedParameterJdbcTemplate(dataSource);
+        final var cache = Caffeine.newBuilder()
+                .maximumSize(512)
+                .expireAfterAccess(10, TimeUnit.MINUTES)
```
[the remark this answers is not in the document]
bbpennel: Give that it depends on usage patterns, it may make sense to make cache options be configurable by administrators

whikloj:
Made the cache configurable, but I didn't do any value checking. So let me know if we should be limiting what the cache sizes can be set to. Ran a small test (11,000 containers) and with this change the ocf_id_map select happens and average of 3 times per transaction (versus 12 before).

fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
```diff
@@ -121,6 +121,12 @@
     @Value("${fcrepo.ocfl.unsafe.write.enabled:false}")
     private boolean unsafeWriteEnabled;
 
+    @Value("${fcrepo.ocfl.id_map_cache.size:1024}")
```
pwinckles, line 124: I find it's often helpful to include the unit in things like this. eg `fcrepo.ocfl.id_map_cache.size.bytes` and `fcrepo.ocfl.id_map_cache.timeout.minutes`. Or something along those lines.

We're planning on adding more db caches, yeah? Like for containment? I wonder if they should all be grouped along the lines of `fcrepo.db.cache.ocfl.size...` and `fcrepo.db.cache.containment.size...`? Or perhaps `fcrepo.cache.db.ocfl.size...`

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndex.java
```diff
@@ -289,6 +315,7 @@ public void reset() {
         try {
             jdbcTemplate.update(TRUNCATE_MAPPINGS, Collections.emptyMap());
             jdbcTemplate.update(TRUNCATE_TRANSACTIONS, Collections.emptyMap());
+            this.mappingCache.invalidateAll();
```
pwinckles, line 318: The `removeMapping()` method should immediately invalidate the key when it's a short-lived tx

fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
whikloj: I did think about that and went back and forth. We also have a cache in the OcflPersistentSessionStorage that could be configurable so maybe `fcrepo.cache....` to keep them together.

As for the `size`, it is the [maximum number of elements](https://github.com/ben-manes/caffeine/blob/e1e0c84d860dc0646c5599c15fa853cefeb53c98/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Caffeine.java#L344-L348). Soooooo `fcrepo.cache.db.ocfl.size.elements` or maybe `fcrepo.cache.db.ocfl.size.entries`?

pwinckles: Ah, right! It's the number of entries. :)

I like `fcrepo.cache...` and either of the elements or entries options.

pwinckles (approved):
👍

PR #1900, opened 2021-05-04
fix rebuild to commit transactions

pwinckles:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3713

Fixes the rebuild process to commit the transactions that they create. This fixes the bug where resources are removed when the tx manager rollsback expired transactions.

# How should this be tested?

1. Start Fedora
2. Verify you can get the root
3. Wait a few minutes
4. Verify you can still get the root

# Interested parties
@fcrepo/committers


dbernstein (approved):
🥇

PR #1901, opened 2021-05-04
refactor tx to have state

pwinckles:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3697

1. Adds a state to transactions
2. Ensures that changes cannot be made to a tx when it is no longer open
3. Waits for in progress changes to finish before committing a tx
4. If a request that's part of a long running tx fails, then the entire tx is marked as failed and cannot be committed

# How should this be tested?

1. Create a tx, create resources in the tx, add a resource that fails (such as an ag child of an ag), and see that the tx cannot be committed
2. Create a tx, concurrently add a large resource and a small resource in the tx, commit the tx before the large resource is finished, and see that the commit waits for it to finish before committing

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-http-api/pom.xml
+++ b/fcrepo-http-api/pom.xml
@@ -73,6 +73,12 @@
     </dependency>
 
     <dependency>
+      <groupId>org.fcrepo</groupId>
+      <artifactId>fcrepo-common</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
     </dependency>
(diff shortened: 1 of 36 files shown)


pwinckles:
Force push to resolve merge conflicts

dbernstein (commented):
Code looks good. Testing it now.

dbernstein:
@pwinckles : I ran the first test above.  A couple of questions:
1.  The response code returned when trying to  commit the transaction (after the attempt to nest the AG failed) is `410 Gone` :  is that what we want?   Or should it also be a  409 Conflict  since we can still delete the transaction? 
2.  What happens if the user never issues a DELETE on the transaction?

dbernstein:
Test 2 works as advertised.

pwinckles:
> 
> 
> @pwinckles : I ran the first test above. A couple of questions:
> 
>     1. The response code returned when trying to  commit the transaction (after the attempt to nest the AG failed) is `410 Gone` :  is that what we want?   Or should it also be a  409 Conflict  since we can still delete the transaction?
> 
>     2. What happens if the user never issues a DELETE on the transaction?

@dbernstein 

1. Yeah, that's code is odd. I'll tweak the exceptions/return codes tomorrow.
2. If a user does not delete a transaction, it will be rolledback and deleted by the transaction manager when it expires.

dbernstein (approved):

PR #1909, opened 2021-05-18
fix db retries on startup; fix mysql search deadlock issues

pwinckles:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3720

1. Fixes the the repository initialization so that it runs after the retry and scheduling infrastructure has been initialized. This is necessary so that the application does not hang waiting for an exception to be retried.
2. Fixes the search index so that it does not deadlock as much in MySQL. To do this, rdf types are added _outside_ of the current transaction.

# How should this be tested?

Fedora should start successfully when using MySQL and initialized on top of migrated content. Test data set should contain at least 100 resources and the reindexer should use multiple threads.

# Interested parties
@fcrepo/committers


fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -577,7 +568,16 @@ private void insertRdfTypes(final List<URI> rdfTypes) {
             parameterSourcesList.add(assocParams);
         }
         final MapSqlParameterSource[] psArray = parameterSourcesList.toArray(new MapSqlParameterSource[0]);
-        jdbcTemplate.batchUpdate(INSERT_RDF_TYPE, psArray);
+
+        // RDF types are upserted outside of the tx to avoid concurrent update contention between txs, especially
+        // in MySQL. This means that they will not be rolled back if the tx fails.
+        noTransactionTemplate.executeWithoutResult(status -> {
+            try {
+                jdbcTemplate.batchUpdate(INSERT_RDF_TYPE, psArray);
```
dbernstein, line 576: @pwinckles :  if the first item in the array fails due to a DuplicateKeyException will subsequent arrays be inserted?

pwinckles: Ah, good catch. I forgot that was a batch update

dbernstein (approved):

PR #1910, opened 2021-05-19
refactor db transaction handling

pwinckles:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3721

Refactors the db transaction management to no longer use Spring's annotations, though it still use's Spring's transaction manager. This change addresses some hypothetical edge case bugs. Specifically, the annotations used to wrap the entire request, which meant that if there were any exceptions during the request it would rollback the changes to the db. This is generally desirable, except in the case of an exception that occurs after the operation was completed and while the response is being generated.

With this change, the transaction itself is more narrowly scoped, and exceptions that occur after resource changes have been committed to disk will never be rolled back.

# How should this be tested?

The bugs this PR addresses are rare, and I can't think of a good way to trigger them. The service should otherwise work the same as before.

# Interested parties

@fcrepo/committers


whikloj (dismissed):
Makes sense and solves existing concurrency issues in MySQL/MariaDB

pwinckles:
This PR also addresses: https://fedora-repository.atlassian.net/browse/FCREPO-3724

I added a concurrency itest

pwinckles:
Added another commit to the PR that addresses https://fedora-repository.atlassian.net/browse/FCREPO-3725.

With this commit all deadlocking issues (table and connection) should be resolved, and you should be able to execute an arbitrary number of concurrent requests against Fedora.

bbpennel (approved):

whikloj (approved):
This still makes sense and solves my deadlock issues with both container and binaries in all DBs I tested.

PR #1911, opened 2021-05-21
Cache WebAC effectiveAcl locations and authorizations

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3322

It adds a cache to the WebAC layer which stores the resource that contains the ACL for the given target resource and list of all the authorizations in that ACL.

These authorizations are still processed against the target resource to filter for applicable authorizations (if any).

# How should this be tested?

This PR should have no functional difference and the performance implications are un-clear.

# Interested parties

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractDeleteResourceService.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractDeleteResourceService.java
@@ -131,6 +131,9 @@ abstract public class AbstractDeleteResourceService extends AbstractService {
                 // Flush ACL cache on any ACL creation/update/deletion.
                 authHandleCache.invalidateAll();
             }
+        } else {
+            // Flush ACL cache on any ACL creation/update/deletion.
+            authHandleCache.invalidateAll();
         }
 
         //delete/purge the resource itself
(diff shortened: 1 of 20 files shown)


dbernstein (approved):
@whikloj :  this looks good to me and builds locally. I'll wait to merge in case any one else wants to take a look.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractDeleteResourceService.java
```diff
@@ -126,6 +134,8 @@ private void deleteDepthFirst(final Transaction tx, final PersistentStorageSessi
             final FedoraResource acl = fedoraResource.getAcl();
             if (acl != null) {
                 doAction(tx, pSession, acl.getFedoraId(), userPrincipal);
+                // Flush ACL cache on any ACL creation/update/deletion.
+                authHandleCache.invalidateAll();
```
bbpennel, line 138: Is it possible to delete the ACL directly? It seems like the acl cache would get flushed if you delete a resource with an acl, but not if you delete the acl by itself.

whikloj: Ohh good catch, yes let me add for that.

bbpennel (dismissed):
I haven't tested this, but it looks good from a review perspective

bbpennel (approved):

dbernstein (approved):

PR #1914, opened 2021-05-25
Add resourceExists cache

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3708

Adds a second cache to the containment index to track `resourceExists` requests. 

I left this to use the same settings as the existing containment cache, we could just update the documentation to make clear this is happening or I can separate the settings into two sets.

I also left this under the original containment cache ticket (which was closed), but I couldn't find another ticket and wasn't sure if a new one was needed.

# How should this be tested?

Should have no functional difference, just some performance improvements from less DB queries.

# Interested parties

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
@@ -803,6 +803,11 @@ public class ContainmentIndexImpl implements ContainmentIndex {
             return !jdbcTemplate.queryForList(queryToUse,
                     Map.of("child", resourceId, "transactionId", tx.getId()), String.class).isEmpty();
         } else if (includeDeleted) {
+            final Boolean exists = resourceExistsCache.getIfPresent(resourceId);
+            if (exists != null && exists) {
+                // Only return true, false values might change once deleted resources are included.
+                return true;
+            }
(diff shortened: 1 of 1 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/ContainmentIndexImpl.java
```diff
@@ -775,18 +797,19 @@ public boolean resourceExists(@Nonnull final Transaction tx, final FedoraId fedo
...
-        final String queryToUse;
         if (tx.isOpenLongRunning()) {
-            queryToUse = includeDeleted ? RESOURCE_OR_TOMBSTONE_EXISTS_IN_TRANSACTION :
+            final var queryToUse = includeDeleted ? RESOURCE_OR_TOMBSTONE_EXISTS_IN_TRANSACTION :
                     RESOURCE_EXISTS_IN_TRANSACTION;
-            parameterSource.addValue("transactionId", tx.getId());
+            return !jdbcTemplate.queryForList(queryToUse,
+                    Map.of("child", resourceId, "transactionId", tx.getId()), String.class).isEmpty();
+        } else if (includeDeleted) {
+            return !jdbcTemplate.queryForList(RESOURCE_OR_TOMBSTONE_EXISTS,
```
pwinckles, line 806: I understand that you don't want to load the cache here, but what do you think about checking the cache for a hit before doing this query? If there's a hit, you can return, and if it's a miss then you do the query without putting the result in the cache.

pwinckles (approved):

PR #1917, opened 2021-05-28
make binaries non-retryable

pwinckles:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3728

When a MySQL deadlock occurs, Fedora retries the request. However, if the initial attempt consumed the request's input stream, the retry will fail. This PR fixes this problem by:

1. Moving the logic that parses RDF outside of the retryable block so that the stream is only read once
2. Makes non-RDF requests non-retryable

# How should this be tested?

The concurrency itest that was flapping should flap no more

# Interested parties

@fcrepo/committers


dbernstein (approved):
🥇

PR #1922, opened 2021-06-10
Improves the /rest/fcr:search endpoint retrieval speed on mysql for l…

dbernstein:
…arger (1M+)  datasets.

**Improves the /rest/fcr:search endpoint retrieval speed on mysql.**
* * *

**JIRA Ticket**: N/A

Improves the search retrieval on larger repos on mysql.
# How should this be tested?
Create a repository with 1  million + resources.  Verify that the /fcr:search loads within 1 second.
```
http://fcrepo-staging.eba-mnmpg8gf.us-east-1.elasticbeanstalk.com/fcrepo/fcr:search
```
# Interested parties

### DIFF
--- a/fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
+++ b/fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
@@ -120,9 +120,9 @@ public class DbSearchIndexImpl implements SearchIndex {
             "search_rdf_type rt ," + SIMPLE_SEARCH_TABLE + " s " +
             "WHERE rrt.rdf_type_id = rt.id group by rrt.resource_id) r ";
 
-    private static final String POSTGRES_GROUP_CONCAT_FUNCTION = "STRING_AGG(rt.rdf_type_uri, ',')";
-    private static final String DEFAULT_GROUP_CONCAT_FUNCTION = "GROUP_CONCAT(distinct rt.rdf_type_uri " +
-            "ORDER BY rt.rdf_type_uri ASC SEPARATOR ',')";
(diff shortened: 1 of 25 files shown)


pwinckles:
Shouldn't this PR be targeting `main`?

fcrepo-search-impl/src/main/java/org/fcrepo/search/impl/DbSearchIndexImpl.java
```diff
@@ -411,20 +411,84 @@ private StringBuilder createSearchQuery(final SearchParameters parameters,
...
-        final var containsRDFTypeField = queryFields.contains(RDF_TYPE.toString());
 
-        final var groupByFields = fields.stream().filter(f->!f.equals(RDF_TYPE.toString())).collect(toList());
-        final var returnFields = new ArrayList<>(groupByFields);
+        if (isPostgres()) {
+            return createPostgresOptimizedQuery(parameterSource, whereClauses, conditions, fields,
+                    rdfTypeConditionValue);
+        } else {
+            return createDefaultQuery(parameterSource, whereClauses, conditions, fields, rdfTypeConditionValue);
+        }
```
whikloj, line 420: Instead of `return`ing right from here, could you log the SQL returned from either of the two functions above. This could be at either `DEBUG` or `TRACE` level. It's really hard to tell what query is being generated here. This would help us see the result now and in future.

dbernstein: Good idea.

dbernstein: Added.

whikloj (approved):
👍

PR #1926, opened 2021-06-18
move db cleanup to run after each test

pwinckles:
**JIRA Ticket**: na

Moves the db cleanup to run after each test rather than before in an attempt to fix some issues with tests affecting each other across test class boundaries.

# How should this be tested?

All tests should pass

# Interested parties

@fcrepo/committers


whikloj (approved):
:+1:

PR #1927, opened 2021-07-09
Update modeshape version to 5.5.0.fcr

bbpennel:
Updates modeshape version to 5.5.0.fcr

# How should this be tested?
There shouldn't be any visible functionality changes. Running the compatibility test suite should produce the same result as before.


dbernstein (approved):
Builds and passes all the tests.

PR #1928, opened 2021-07-16
added field for specifying mime type in web console

pwinckles:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3741

1. Adds a field for specifying mime type when adding/updating binaries
2. Fixes bug where content type was being set to an empty string

Ideally, the type would be pre-populated with the existing mime type on update, but I don't know how to do that.

# How should this be tested?

Do the following for adding and updating binaries:

1. Open the console
2. Add/update a binary _without_ a file extension and _without_ a mime type, and see it default to `application/octet-stream`
3. Add/update a binary _with_ a file extension and _without_ a mime type and see it guess the mime type
4. Add/update a binary _without_ a file extension and _with_ a mime type and see it use the specified type

# Interested parties

@fcrepo/committers

### DIFF
--- a/fcrepo-http-api/src/main/resources/views/common-node-actions.vsl
+++ b/fcrepo-http-api/src/main/resources/views/common-node-actions.vsl
@@ -18,6 +18,12 @@
           <label for="update_file" class="control-label">File</label>
           <input type="file" id="update_file"/>
         </div>
+        <div class="form-group">
+          <label for="update_mime_type" class="control-label">
+            MIME Type
+          </label>
+          <input type="text" id="update_mime_type" placeholder="application/octet-stream" name="update_mime_type" class="form-control"/>
+        </div>
         <input type="submit" id="binary_update_content" class="btn btn-primary" value="Update">
       </form>
     #end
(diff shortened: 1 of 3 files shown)


bbpennel (commented):
This looks good and works well, the only part that feels a little weird to me is the placeholder text that says `application/octet-stream` even when the browser is going to provide a different mimetype.

It looks like there are a number of approaches and libraries available for trying to guess the File type, although to be honest most of it seems out of scope given that direct use of the API does not attempt to guess the mimetype. It may be that using the javascript File/blob api will yield roughly the same value as the browser provides during upload:
https://stackoverflow.com/questions/18299806/how-to-check-file-mime-type-with-javascript-before-upload
Or we could change the placeholder text to somehow indicate that the browser is going to guess, but I'm having a hard time thinking of what it would say.

demiankatz:
@bbpennel, should the placeholder text say something like "leave blank to autodetect"? If the file extension is the only thing being used for autodetection, we could even say "leave blank to autodetect from file extension" to be more clear about what's going on... or, if we want to leave things more open-ended, we could say "specifying a MIME type is highly recommended." But maybe I'm getting a little ambitious about how much text we can fit in the box... in that case, perhaps simply "autodetect" or "autodetect (unreliable)" might do the job.

pwinckles:
@bbpennel I updated it to automatically set the mime type in the input box when the type is known

bbpennel:
The implemented approach seems to work well for me, I didn't encounter any situations where the javascript couldn't come up with a mimetype while the browser could. I was testing in firefox and chrome on macOS big sur. The webconsole definitely has some weird caching behaviors, I had to do forced reloads in chrome for the updates to show up, but that's unrelated to this PR.

So I'm inclined to go ahead with the `application/octet-stream` placeholder text since it appears to be accurate in so far as what gets submitted by the browser.

bbpennel (approved):

PR #1929, opened 2021-07-21
Add Link header for the ArchivalGroup of a resource

mikejritter:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3295

Searches the ancestry of a resource to see if any parent is an Archival Group and adds a Link header if one exists

# How should this be tested?

* Create an Archival Group
* Create some resources underneath the Archival Group
* Perform some HEAD requests on the child resources and check for the Archival Group Link header

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -62,6 +62,7 @@ import static org.fcrepo.http.commons.session.TransactionConstants.TX_PREFIX;
 import static org.fcrepo.kernel.api.FedoraTypes.FCR_ACL;
 import static org.fcrepo.kernel.api.FedoraTypes.FCR_METADATA;
 import static org.fcrepo.kernel.api.FedoraTypes.FEDORA_ID_PREFIX;
+import static org.fcrepo.kernel.api.RdfLexicon.ARCHIVAL_GROUP;
 import static org.fcrepo.kernel.api.models.ExternalContent.COPY;
 import static org.fcrepo.kernel.api.models.ExternalContent.PROXY;
 import static org.fcrepo.kernel.api.models.ExternalContent.REDIRECT;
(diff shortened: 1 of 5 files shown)


pwinckles:
I don't doubt that this works, however, I fear that it will be quite slow. Archival part resource actually contain a reference to their containing archival group resource in their [ResourceHeaders](https://github.com/fcrepo/fcrepo/blob/55634df48fc2c1ed58e70f636a7f6396ca48f256/fcrepo-persistence-common/src/main/java/org/fcrepo/persistence/common/ResourceHeadersImpl.java#L101). It should be much more efficient to use this field to solve this issue. I would expect that exposing it on `FedoraResource` would be fine.

mikejritter:
> I don't doubt that this works, however, I fear that it will be quite slow. Archival part resource actually contain a reference to their containing archival group resource in their [ResourceHeaders](https://github.com/fcrepo/fcrepo/blob/55634df48fc2c1ed58e70f636a7f6396ca48f256/fcrepo-persistence-common/src/main/java/org/fcrepo/persistence/common/ResourceHeadersImpl.java#L101). It should be much more efficient to use this field to solve this issue. I would expect that exposing it on `FedoraResource` would be fine.

It looks like `isArchivalGroup` already exists as a field in `FedoraResourceImpl` as well, so this makes sense. I'll have an update ready soon with changes reflecting this.

I'm curious though, was the main concern around `resource.getTypes()`? I had assumed these would have been populated when the `FedoraResource` was created but looking at the impl this doesn't appear to be the case.

pwinckles:
> I'm curious though, was the main concern around `resource.getTypes()`? I had assumed these would have been populated when the `FedoraResource` was created but looking at the impl this doesn't appear to be the case.

No, I was referencing `getArchivalGroupId()`. So, you ask the resource what archival group it is part of rather than traversing up the resource's ancestors.

mikejritter:
> No, I was referencing `getArchivalGroupId()`. So, you ask the resource what archival group it is part of rather than traversing up the resource's ancestors.

I completely missed that from the `ResourceHeaders`, got a bit too excited when I saw the `isArchiveGroup` boolean already in the `FedoraResourceImpl`. Updated again to expose the `archivalGroupId` if it exists.

pwinckles (approved):
👍

PR #1933, opened 2021-09-01
FCREPO-3752: Altering the external content URL does not update the binary metadata's last-modified date

mikejritter:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3752

This updates the metadata handling for server managed properties to be applied to the binary headers rather than the binary description. This allows for the ManagedPropertiesService to resolve all managed properties from the binary.

# Changes

* Create a new operation for updating only the binary headers
* Create a new persister for updating the binary headers
* When updating a description, create two operations for managed and unmanaged properties
* Resolve managed properties from the Binary when possible

# How should this be tested?

There are two things to test here - the binary metadata being pulled from the correct source and that updates to server managed properties are done to the correct file.

## Binary Metadata

* Start fedora
* Create a binary
  ```
  curl -X PUT -H "Content-Type: text/plain" --data "somedata" -H "Link: <http://www.w3.org/ns/ldp#NonRDFSource>; rel=\"type\"" --user fedoraAdmin:fedoraAdmin http://localhost:8080/rest/test1 
  ```
* Get the metadata
  ```
  curl --user fedoraAdmin:fedoraAdmin http://localhost:8080/rest/binary/fcr:metadata
  ```
* Do a second put over top of the binary
  ```
  curl -X PUT -H "Content-Type: text/plain" --data "somemoredata" -H "Link: <http://www.w3.org/ns/ldp#NonRDFSource>; rel=\"type\"" --user fedoraAdmin:fedoraAdmin http://localhost:8080/rest/test1 
  ```
* Check that the metadata is updated properly

## Server Managed Properties

* Start fedora with relaxed properties mode
* Create a binary
  ```
  curl -X PUT -H "Content-Type: text/plain" --data "somedata" -H "Link: <http://www.w3.org/ns/ldp#NonRDFSource>; rel=\"type\"" --user fedoraAdmin:fedoraAdmin http://localhost:8080/rest/binary 
  ```
* Get the metadata for the binary and see that it matches what is the in the binary headers
  ```
  curl --user fedoraAdmin:fedoraAdmin http://localhost:8080/rest/binary/fcr:metadata
  ```
* Update the server managed properties on the binary
  ```
  curl -X PATCH --data "PREFIX fedora: <http://fedora.info/definitions/v4/repository#>
  INSERT {
      <> fedora:createdDate "2021-08-26T15:04:43.517Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> ;
      fedora:lastModified "2021-08-26T15:04:43.517Z"^^<http://www.w3.org/2001/XMLSchema#dateTime> .
  }
  WHERE  { }" -H "Content-Type: application/sparql-update" --user fedoraAdmin:fedoraAdmin http://localhost:8080/rest/descupdate/fcr:metadata 
  ```
* Get the metadata for the binary and see that it matches the update + what is the in the v2 binary headers
  ```
  curl --user fedoraAdmin:fedoraAdmin http://localhost:8080/rest/binary/fcr:metadata
  ```

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/ReplacePropertiesServiceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/ReplacePropertiesServiceImpl.java
@@ -113,7 +113,6 @@ public class ReplacePropertiesServiceImpl extends AbstractService implements Rep
     private void updateBinaryHeaders(final Transaction tx,
                                      final PersistentStorageSession pSession,
                                      final ResourceOperation operation) {
-        // tx.lockResource(operation.getResourceId());
         pSession.persist(operation);
         recordEvent(tx, operation.getResourceId(), operation);
     }
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/ReplacePropertiesServiceImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/ReplacePropertiesServiceImpl.java
@@ -37,6 +37,8 @@ import javax.inject.Inject;
 
 import static org.fcrepo.kernel.api.rdf.DefaultRdfStream.fromModel;
 
+import java.util.Optional;
+
 /**
  * This class mediates update operations between the kernel and persistent storage layers
  * @author bseeger
@@ -56,7 +58,7 @@ public class ReplacePropertiesServiceImpl extends AbstractService implements Rep
                         final FedoraId fedoraId,
                         final Model inputModel) throws MalformedRdfException {
         try {
-            final PersistentStorageSession pSession = this.psManager.getSession(tx);
+            final PersistentStorageSession pSession = psManager.getSession(tx);
 
             final var headers = pSession.getHeaders(fedoraId, null);
             final var interactionModel = headers.getInteractionModel();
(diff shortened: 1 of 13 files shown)


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/UpdateRdfHeadersPersister.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public UpdateRdfHeadersPersister(final FedoraToOcflObjectIndex fedoraOcflIndex) {
+        super(RdfSourceOperation.class, UPDATE_HEADERS, fedoraOcflIndex);
+    }
+
+    @Override
+    public void persist(final OcflPersistentStorageSession session, final ResourceOperation operation)
+        throws PersistentStorageException {
+        final var resourceId = operation.getResourceId();
+        log.info("persisting {} headers to {}", resourceId, session);
```
pwinckles, line 54: I think this should probably be debug. The other persisters are using debug, yeah?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/ReplacePropertiesServiceImpl.java
```diff
@@ -56,35 +58,63 @@ public void perform(final Transaction tx,
...
         } catch (final PersistentStorageException ex) {
             throw new RepositoryRuntimeException(String.format("failed to replace resource %s",
                     fedoraId), ex);
         }
     }
+
+    private void updateBinaryHeaders(final Transaction tx,
+                                     final PersistentStorageSession pSession,
+                                     final ResourceOperation operation) {
+        tx.lockResource(operation.getResourceId());
```
pwinckles, line 116: I don't think you need this lock. I believe the resource is already locked on line 98

pwinckles (commented):
Looks good!

A couple of minor comments. Additionally, can you add an itest to the `fcrepo-http-api` module?

I'll test it locally shortly, but my expectation is that it will work as expected.

mikejritter:
@pwinckles Thanks for the review, I just got to it this morning. I added an integration test for checking the `LastModified` of the binary description after an update to a binary. The `FedoraRelaxedLdpIT` has one for server managed predicates so I opted not to add another for that.

pwinckles (approved):
👍

whikloj (changes requested):
This seems 👍 just the one line that can be removed.

whikloj (approved):

PR #1935, opened 2021-09-09
add user rdf types cache

pwinckles:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3732

Adds a cache for user rdf types so that they do not need to be repeatedly loaded from disk.

# How should this be tested?

Everything should work the same as before except triples should never be loaded from disk when creating or updating resources. I added a new debug log that will log every time a resource's triples are loaded. If you enable debug logging you should be able to confirm that the triples are no longer read when creating resources.

Key things to look at:

1. Triples are not read when creating/updating any resource
2. Updating resources in transactions works and does not bleed outside the tx

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
@@ -131,6 +131,12 @@ public class FedoraPropsConfig extends BasePropsConfig {
     @Value("${fcrepo.cache.db.containment.timeout.minutes:10}")
     private long containmentCacheTimeout;
 
+    @Value("${fcrepo.cache.types.size.entries:1024}")
+    private long userTypesCacheSize;
+
+    @Value("${fcrepo.cache.types.timeout.minutes:10}")
+    private long userTypesCacheTimeout;
+
     @Value("${fcrepo.cache.webac.acl.size.entries:1024}")
     private long webacCacheSize;
 
@@ -341,6 +347,20 @@ public class FedoraPropsConfig extends BasePropsConfig {
      */
     public long getContainmentCacheTimeout() {
         return containmentCacheTimeout;
+    }
+
+    /**
+     * @return The number of entries in the user types cache.
+     */
+    public long getUserTypesCacheSize() {
+        return userTypesCacheSize;
+    }
+
+    /**
+     * @return The number of minutes before items in the user types cache expire.
+     */
+    public long getUserTypesCacheTimeout() {
+        return userTypesCacheTimeout;
     }
 
     /**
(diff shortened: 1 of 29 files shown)


fcrepo-kernel-impl/src/test/java/org/fcrepo/kernel/impl/cache/UserTypesCacheImplTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+    @Test
+    public void manuallyCachedTypesAreOnlyAvailableInSession() {
+        cache.cacheUserTypes(fedoraId, List.of(ANIMAL_TYPE), sessionId);
+
+        var types = cache.getUserTypes(fedoraId, sessionId, noCallSupplier());
+        assertThat(types, containsInAnyOrder(ANIMAL_TYPE));
+
+        types = cache.getUserTypes(fedoraId, sessionId2, rdfSupplier(ANIMAL_TYPE, CAT_TYPE));
+        assertThat(types, containsInAnyOrder(ANIMAL_TYPE, CAT_TYPE));
```
whikloj, line 94: I think I get what you mean here, but just wondering. If you called
```
types = cache.getUserTypes(fedoraId, sessionId2, noCallSupplier());
```
would `types == emptyList()`?

pwinckles: No, because the `noCallSupplier()` throws an exception when it's called. It's just used in the tests to ensure that the values are returned from the cache and were not newly loaded.

whikloj (approved):
👍

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
```diff
@@ -131,6 +131,12 @@
     @Value("${fcrepo.cache.db.containment.timeout.minutes:10}")
     private long containmentCacheTimeout;
 
+    @Value("${fcrepo.cache.types.size.entries:1024}")
```
dbernstein, line 134: Did these settings make it into the documentation?

dbernstein (approved):

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
pwinckles: Not yet, I was waiting for the merge, but thanks for the reminder.

dbernstein: It looks like not.  I'll add them.

dbernstein: No worries.  I'll add them before I merge.

dbernstein (approved):

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
pwinckles: Thank you!

PR #1941, opened 2021-09-29
Display blank node triples in HTML UI

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3773

When displaying triples in the HTML UI we expect the subject of all triples to be a URI. Blank nodes (while not recommended) are supported and don't have a URI in Jena.

This checks for the type of resource and finds something to use.

# How should this be tested?

Make a RDFResource with a blank node, get it with `Accept: text/turtle` and see the output. Get it with the HTML UI and get a 500 server error.
Pull in this PR and be able to see the blank node resource in the HTML UI.

There is a test as the first commit if you want to verify it fails before the actual fix.

# Interested parties
@fcrepo/committers


pwinckles (approved):
👍

PR #1954, opened 2021-11-19
Suppress events on rebuild.

dbernstein:
**Suppress events on rebuild.**
* * *

**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3792

This PR causes Fedora to suppress all events during the rebuild process.  Prior to this change,  object creation events were not being generated on rebuild since they are recorded during creation of the ocfl resource.  As a result only events that involve index updates (such as inbound reference updates) were being generated.  Since event generation is connected to the request (from which the base url is resolved) this was causing the NPE symptom since rebuild make use of the fcrepo.jms.baseUrl property.  

Since we neither want to require the use of fcrepo.jms.baseUrl nor generate a subset of repo events, the most straightforward solution I believe is to simply suppress all events for now during rebuild.

# How should this be tested?
1. fire up fedora
2. Do the following

```
curl -X PUT -u fedoraAdmin:fedoraAdmin "http://localhost:8080/rest/test1"
curl -X PUT -u fedoraAdmin:fedoraAdmin "http://localhost:8080/rest/test2"

curl -u fedoraAdmin:fedoraAdmin "http://localhost:8080/rest/test1" -X PATCH -H "Content-Type: application/sparql-update" --data "INSERT DATA { <> <http://relation/pointat> <http://localhost:8080/rest/test2>. }"

curl -u fedoraAdmin:fedoraAdmin "http://localhost:8080/rest/test2" -X PATCH -H "Content-Type: application/sparql-update" --data "INSERT DATA { <> <http://relation/pointat> <http://localhost:8080/rest/test1>. }"
```
3. delete everything but ocfl-root in your data directory
4. restart fedora and validate that it comes up without the NPE.

# Interested parties

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
@@ -100,7 +100,10 @@ public class TransactionImpl implements Transaction {
             updateState(TransactionState.COMMITTED);
             if (!this.suppressEvents) {
                 this.getEventAccumulator().emitEvents(this, baseUri, userAgent);
+            } else {
+                this.getEventAccumulator().clearEvents(this);
             }
+
             releaseLocks();
             log.debug("Committed transaction {}", id);
         } catch (final Exception ex) {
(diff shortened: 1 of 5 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
```diff
@@ -97,7 +98,9 @@ public synchronized void commit() {
             }
 
             updateState(TransactionState.COMMITTED);
-            this.getEventAccumulator().emitEvents(this, baseUri, userAgent);
+            if (!this.suppressEvents) {
+                this.getEventAccumulator().emitEvents(this, baseUri, userAgent);
```
pwinckles, line 102: This may cause a memory leak due to the events never being cleared from the accumulator.

pwinckles (approved):
👍

PR #1956, opened 2021-11-24
Make a query param to change content-disposition from attachment to inline

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3791 & https://fedora-repository.atlassian.net/browse/FCREPO-1843

Adds a query parameter `inline`, that if set to `true` will result in the Content-Disposition being set to `inline`. Otherwise it remains `attachment`

This is also changed for HEAD requests, mementos and external content.

# How should this be tested?

Add a binary to a repository using main. 
GET the binary
* Through the HTML UI, see that pressing the "Download Content" button causes a save file dialog
* Through curl see the `Content-Disposition: attachment; size=79; creation-date="Wed, 24 Nov 2021 14:52:17 GMT"; modification-date="Wed, 24 Nov 2021 14:52:17 GMT"` header

Pull in this PR
GET the binary
* Through the HTML UI you cannot easily see this behaviour but if you go to a binary (which actually takes you to the description), remove `/fcr:metadata` and add `?inline=true` you will immediately gets the content.
* Through curl also add `?inline=true` and see the `Content-Disposition: inline; size=79; creation-date="Wed, 24 Nov 2021 14:52:17 GMT"; modification-date="Wed, 24 Nov 2021 14:52:17 GMT"` header (might have to do a HEAD if you use an image as your binary)

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -428,7 +428,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
                 final var uri = new URI(identifierConverter().toExternalId(agFedoraId.getFullId()));
                 final var link = buildLink(uri, "archival-group");
                 servletResponse.addHeader(LINK, link);
-            } catch (URISyntaxException e) {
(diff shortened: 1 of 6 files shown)


dbernstein (approved):
Code looks good, builds well, and manual tests work as advertised.

PR #1957, opened 2021-11-25
Lock ghost nodes when creating a resource

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3584

Locks ghost nodes above a resource that is being created so another transaction cannot replace them while the transaction is not committed.

# How should this be tested?

Before PR:
1. Create a transaction.
     ```
      > curl -i -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/fcr:tx -X POST

     HTTP/1.1 201 Created
     ...
     Location: http://localhost:8080/rest/fcr:tx/a5912bd9-a479-4e23-9c97-bc4e85f59539
     Link: <http://localhost:8080/rest/fcr:tx/a5912bd9-a479-4e23-9c97-bc4e85f59539>; rel="http://fedora.info/definitions/v4/transaction#commitEndpoint"
     ```
1. Create a binary with ghost nodes inside the transaction.
     ```
     > curl -i -u fedoraAdmin:fedoraAdmin -H "Atomic-ID: http://localhost:8080/rest/fcr:tx/a5912bd9-a479-4e23-9c97-bc4e85f59539" -X PUT -H "Content-type: text/plain" --data-binary "SOME CONTENT" http://localhost:8080/rest/a/b/c

     HTTP/1.1 201 Created
     ...

     http://localhost:8080/rest/a/b/c
     ```
1. Outside the transaction create a binary at one of the ghost nodes.
      ```
      > curl -i -u fedoraAdmin:fedoraAdmin -X PUT -H "Content-type: text/plain" --data-binary "SOME OTHER CONTENT" http://localhost:8080/rest/a

     HTTP/1.1 201 Created
     ...

     http://localhost:8080/rest/a
     ```
1. Commit the transaction 
     ```
     > curl -i -u fedoraAdmin:fedoraAdmin http://localhost:8080/rest/fcr:tx/a5912bd9-a479-4e23-9c97-bc4e85f59539 -XPUT
      
     HTTP/1.1 204 No Content
     ```
1. See that a binary has a binary child
     ```
     > curl -u fedoraAdmin:fedoraAdmin -i http://localhost:8080/rest/a

     HTTP/1.1 200 OK
     ...
     Content-Type: text/plain
     Link: <http://www.w3.org/ns/ldp#NonRDFSource>; rel="type"

     SOME OTHER CONTENT
     ```
     ```
     > curl -u fedoraAdmin:fedoraAdmin -i http://localhost:8080/rest/a/b/c

     HTTP/1.1 200 OK
     ...
     Content-Type: text/plain
     Link: <http://www.w3.org/ns/ldp#NonRDFSource>; rel="type"

     SOME CONTENT
     ```

Pull in this PR, re-do the above steps. Note the divergence at step 3

3. Outside the transaction create a binary at one of the ghost nodes.
      ```
      > curl -i -u fedoraAdmin:fedoraAdmin -X PUT -H "Content-type: text/plain" --data-binary "SOME OTHER CONTENT" http://localhost:8080/rest/a

     HTTP/1.1 409 Conflict
     Date: Thu, 25 Nov 2021 14:57:04 GMT
     Set-Cookie: JSESSIONID=node0ms1ix09jq0ag1m2863u0x1pn02.node0; Path=/
     Set-Cookie: rememberMe=deleteMe; Path=/; Max-Age=0; Expires=Wed, 24-Nov-2021 14:57:04 GMT; SameSite=lax
     Content-Length: 79
     Server: Jetty(9.4.44.v20210927)

     Cannot update info:fedora/a because it is being updated by another transaction.
     ```
# Interested parties

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
@@ -277,7 +277,7 @@ public class TransactionImpl implements Transaction {
 
     @Override
     public void lockResource(final FedoraId resourceId) {
-        getResourceLockManger().acquire(getId(), resourceId);
+        getResourceLockManger().acquire(this, resourceId);
     }
 
     @Override
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
@@ -5,8 +5,12 @@
  */
 package org.fcrepo.kernel.impl;
 
+import static java.util.stream.Collectors.toList;
+
 import java.time.Duration;
 import java.time.Instant;
+import java.util.Arrays;
+import java.util.List;
 import java.util.concurrent.Phaser;
 
 import org.fcrepo.common.db.DbTransactionExecutor;
@@ -25,6 +29,7 @@ import org.fcrepo.kernel.api.services.MembershipService;
 import org.fcrepo.kernel.api.services.ReferenceService;
 import org.fcrepo.persistence.api.PersistentStorageSession;
 import org.fcrepo.search.api.SearchIndex;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -277,7 +282,36 @@ public class TransactionImpl implements Transaction {
 
     @Override
     public void lockResource(final FedoraId resourceId) {
-        getResourceLockManger().acquire(this, resourceId);
+        getResourceLockManger().acquire(getId(), resourceId);
+        lockGhostNodesOfResource(resourceId);
+    }
+
+    /**
+     * If you create an object with ghost nodes above it, we need to lock those paths as well to ensure
+     * no other operation alters them while the current transaction is in process.
+     *
+     * @param resourceId the resource we are creating
+     */
+    private void lockGhostNodesOfResource(final FedoraId resourceId) {
+        // NonRdfSourceDescriptions will resolve to their binary path, so skip them.
+        if (!resourceId.isDescription()) {
+            final var resourceIdStr = resourceId.getResourceId();
+            final String estimateParentPath = resourceIdStr.indexOf('/') > -1 ?
+                    resourceIdStr.substring(0,resourceIdStr.lastIndexOf('/')) : resourceIdStr;
+            final var actualParent = getContainmentIndex().getContainerIdByPath(this, resourceId, false);
+            if (!estimateParentPath.equals(actualParent.getResourceId())) {
+                // If the expected parent does not match the actual parent, then we have ghost nodes.
+                // Lock them too.
+                final List<String> ghostPaths = Arrays.stream(estimateParentPath
+                        .replace(actualParent.getResourceId(), "")
+                        .split("/")).filter(a -> !a.isBlank()).collect(toList());
+                FedoraId tempParent = actualParent;
+                for (final String part : ghostPaths) {
+                    tempParent = tempParent.resolve(part);
+                    getResourceLockManger().acquire(getId(), tempParent);
+                }
+            }
+        }
     }
 
     @Override
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
@@ -283,7 +283,6 @@ public class TransactionImpl implements Transaction {
     @Override
     public void lockResource(final FedoraId resourceId) {
         getResourceLockManger().acquire(getId(), resourceId);
-        lockGhostNodesOfResource(resourceId);
     }
 
     /**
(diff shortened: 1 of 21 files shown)


fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/lock/InMemoryResourceLockManager.java
```diff
@@ -60,11 +72,29 @@ public void acquire(final String txId, final FedoraId resourceId) {
                         String.format("Cannot update %s because it is being updated by another transaction.",
                                 resourceIdStr));
             }
+            final String estimateParentPath = resourceIdStr.substring(0, resourceIdStr.lastIndexOf('/'));
```
pwinckles, line 75: In the case of the root resource, `resourceIdStr.lastIndexOf('/')` will return `-1`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/lock/InMemoryResourceLockManager.java
```diff
@@ -60,11 +72,29 @@ public void acquire(final String txId, final FedoraId resourceId) {
...
+            final var actualParent = containmentIndex.getContainerIdByPath(tx, resourceId, false);
+
+            if (!estimateParentPath.equalsIgnoreCase(actualParent.getResourceId())) {
```
pwinckles, line 78: I thought resources are case-sensitive. Are they not?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/lock/InMemoryResourceLockManager.java
```diff
@@ -60,11 +72,29 @@ public void acquire(final String txId, final FedoraId resourceId) {
...
+                // If the expected parent does not match the actual parent, then we have ghost nodes.
+                // Add them as well.
+                LOG.debug("Getting lock for ghost parents as well.");
+                final List<String> ghostPaths = Arrays.stream(estimateParentPath
+                        .replace(actualParent.getResourceId(), "")
+                        .split("/")).filter(a -> !a.isBlank()).collect(Collectors.toList());
+                FedoraId tempParent = actualParent;
+                for (final String part : ghostPaths) {
+                    tempParent = tempParent.resolve(part);
+                    lockedResources.add(tempParent.getResourceId());
```
pwinckles, line 88: I don't think locking ghost nodes here is safe. The problem is that this code is being executed within a synchronized block for the _child_ resource, which means that there's nothing preventing another transaction from concurrently locking the same resources.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/TransactionImpl.java
```diff
@@ -278,6 +283,35 @@ public void refresh() {
     @Override
     public void lockResource(final FedoraId resourceId) {
         getResourceLockManger().acquire(getId(), resourceId);
+        lockGhostNodesOfResource(resourceId);
```
whikloj, line 286: I was unsure if these two should be wrapped in a `try { } catch` for a `ConcurrentUpdateException` to release all the locks if one fails 🤷

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
```diff
@@ -286,7 +286,7 @@ protected void lockArchivalGroupResource(final Transaction tx,
                                              final FedoraId fedoraId) {
         final var headers = pSession.getHeaders(fedoraId, null);
         if (headers.getArchivalGroupId() != null) {
-            tx.lockResource(headers.getArchivalGroupId());
+            tx.lockResourceAndGhostNodes(headers.getArchivalGroupId());
```
pwinckles, line 289: Do you think we need to look for ghosts here? Let's say that we have a ghost node `a` that contains the AG `a/b` and we're adding `a/b/c`. I think that locking `a/b` and `a/b/c` is sufficient because the ghost node is not actually being created here.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
```diff
@@ -296,13 +296,51 @@ protected void lockArchivalGroupResourceFromParent(final Transaction tx,
         if (parentId != null && !parentId.isRepositoryRoot()) {
             final var parentHeaders = pSession.getHeaders(parentId, null);
             if (parentHeaders.isArchivalGroup()) {
-                tx.lockResource(parentId);
+                tx.lockResourceAndGhostNodes(parentId);
             } else if (parentHeaders.getArchivalGroupId() != null) {
-                tx.lockResource(parentHeaders.getArchivalGroupId());
+                tx.lockResourceAndGhostNodes(parentHeaders.getArchivalGroupId());
```
pwinckles, line 301: I'm not sure if we need to look for ghosts here either. If a parent exists and we lock it, I don't think we care if the parent has any ghosts in its ancestry, do we? Don't we only care about ghosts between the child and its parent?

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
```diff
@@ -296,13 +296,51 @@ protected void lockArchivalGroupResourceFromParent(final Transaction tx,
...
+        final var headers = pSession.getHeaders(checkId, null);
+        if (checkId.isRepositoryRoot() || (!headers.isArchivalGroup() && headers.getArchivalGroupId() == null &&
+                !checkId.isAcl() && !checkId.isDescription())) {
+            // The resource's check ID is the root or is not an AG or part of an AG and is not an ACL or binary
+            // description. So we lock it with ghost node checking
+            tx.lockResourceAndGhostNodes(resourceId);
+        } else {
+            // The checkID is an AG or part of an AG that should have been locked by a call to
+            // lockArchivalGroupResource or lockArchivalGroupResourceFromParent.
+            tx.lockResource(resourceId);
```
pwinckles, line 328: I think we _do_ want to look for ghosts if it is an AG part. For example, if you have AG `a` and add part `a/b/c` then you'll want to lock the ghost node `a/b`

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
whikloj: So because this resource is part of an AG, then that AG must already be created or locked?

pwinckles: Yes, I think that we do not need to look for ghosts in the ancestry of the parent of the target resource. I think we only care about ghosts that are between the parent and the target. But, maybe I'm wrong.

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
whikloj: Above we lock ghost nodes if `headers.getArchivalGroupId() != null` (ie. is part of an AG). You did not think that was necessary.

Here we lock ghost nodes if `headers.getArchivalGroupId() == null` (ie. is not part of an AG) and you do think its necessary.

The inversion of `==` to `!=` is confusing me a little so my logic could be incorrect. Also the inclusion of the `lockArchivalGroupResource` and `lockArchivalGroupResourceFromParent` methods cause me some hesitation.

whikloj: Or wait... I think I worked out what you mean

fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/services/AbstractService.java
```diff
@@ -290,15 +290,17 @@ protected void lockArchivalGroupResource(final Transaction tx,
...
             final var parentHeaders = pSession.getHeaders(parentId, null);
-            if (parentHeaders.isArchivalGroup()) {
-                tx.lockResource(parentId);
-            } else if (parentHeaders.getArchivalGroupId() != null) {
+            if (parentHeaders.getArchivalGroupId() != null) {
+                // parent is part of an AG, so lock the AG
                 tx.lockResource(parentHeaders.getArchivalGroupId());
+            } else {
+                // otherwise lock the parent.
+                tx.lockResource(parentId);
```
pwinckles, line 303: Is this right? Won't this lock all atomic container parents?

whikloj: Right, so this is only used in the `CreateResourceServiceImpl` for creating a new resource and in the `WebacAclServiceImpl` also for creating a new ACL. Because of that I assumed we wouldn't want anyone... mostly deleting the parent while a new child is being created.

pwinckles: Hmm... It's not immediately clear to me what would happen in the existing code if you concurrently added a child while deleting its parent. I would hope it would fail, but I'm not sure where. I am a little hesitant to lock the parent in these cases because it means that you can't concurrently add children to a container, which used to only be a limitation of AGs. However, we might not have a choice if the behavior change here is actually fixing a bug.

If we do need to lock, perhaps we need a followup ticket for different kinds of locks? That is to say, locking the parent so that it can't be deleted, but allowing multiple sessions to concurrently add children.

whikloj: Test added

pwinckles (approved):
👍

PR #1963, opened 2022-01-10
Fix itest reset

pwinckles:

Fixes the itests so that they are reset correctly between tests. Without this, GH Actions runs consistently fail. The problem is related to [the change](https://github.com/fcrepo/fcrepo/commit/5503287dd7b62eaaf3d6f4031cae82994c8dfc9d) that made Fedora start before the repo was initialized. In some instances the tests were being run while the initialization was still going. The code was changed to wait for the initialization to complete.

# How should this be tested?

Tests should pass

# Interested parties
@fcrepo/committers


whikloj (commented):
👍

whikloj (approved):
Oops 👍

PR #1966, opened 2022-01-12
Move ACL cache clear to ReplacePropertiesService to catch both PATCH …

whikloj:
…and PUT requests

**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3795

We have an ACL cache to hold the resolved set of rules to avoid locating them multiple times. This cache is cleared when we update/delete an ACL.

Except we were only handling updates via PATCH and not PUT.

# How should this be tested?

Check out the two new integration tests, but basically.

This works:

1. Create a container.
2. Add an ACL giving a non-admin user Write or Append (or both) permissions.
3. See you can create a child to the above container as the non-admin user.
4. Update (PATCH) the ACL to give the non-admin user ONLY Read access to the container from step 1.
5. See you can not create another child to the above container.

This doesn't:

1. Create a container.
2. Add an ACL giving a non-admin user Write or Append (or both) permissions.
3. See you can create a child to the above container as the non-admin user.
4. Update (PUT) the ACL to give the non-admin user ONLY Read access to the container from step 1.
5. See you _can_ create another child to the above container.

This PR fixes the second case.

# Interested parties

### DIFF
--- a/fcrepo-auth-webac/src/test/java/org/fcrepo/integration/auth/webac/WebACRecipesIT.java
+++ b/fcrepo-auth-webac/src/test/java/org/fcrepo/integration/auth/webac/WebACRecipesIT.java
@@ -5,6 +5,7 @@
  */
 package org.fcrepo.integration.auth.webac;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Arrays.stream;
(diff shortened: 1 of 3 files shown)


dbernstein (approved):
The code looks good. I haven't yet tested manually.

PR #1968, opened 2022-01-18
Allow on demand reindexing of resources

mikejritter:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3553

Updates the ContainmentIndex and FedoraToOcflObjectIndex implementations to have `on conflict`/`on duplicate key` clauses in order for the reindexing to run on resources which already exist.

# How should this be tested?

* Create a resource in Fedora
* Update the resource outside of Fedora 
  * I'm not sure of the best way to do this -- I did it by hand which wasn't bad for a small resource but not really ideal with how many changes are needed for ocfl.
  * Ideally the updated resource should also have changes so that the `ReferenceService` is hit.
* Issue a POST to the reindex endpoint: `http://localhost:8080/rest/<fedora-id>/fcr:reindex`
* The reindex should succeed and any changes to the ContainmentIndex, FedoraToOCflObjectIndex, and ReferenceService should be visible.

# Notes

As I'm making this I remembered some discussion about limiting this based on the interaction model, which seems like something we would still want. I imagine there are ways to break this as it currently stands, so at least for an initial implementation we could add some restrictions before merging.

The changes to the `FedoraReindexIT` were fairly minimal, and the only check that's really done is that the http call returns a 204. It might be good to hit the endpoints to see changes in the resource being reflected.

# Interested parties

@fcrepo/committers


dbernstein:
@mikejritter : the code looks good.  Here's what I've run into so far on the testing.

When I do the following,  I run into a validation error on reindex.  If I restart fedora after swapping out the old ocfl object for the new one and before I call reindex, then everything works.   Perhaps there is some caching in the ocfl layer that needs to be cleared before reindexing?

1.  create a  basic container "test-container"
2. copy the fcrepo-home to fcrepo-home-1
3. modify the container by adding a title
4. copy the fcrepo-home to fcrepo-home-2
5.  in the fcrepo-home/data/ocfl-root delete the tests-container resource object root and copy in the resource from fcrepo-home-1
6. do reindex of existing resource
```
Caused by: java.nio.file.NoSuchFileException: /Users/danielbernstein/code/fcrepo/fcrepo-webapp/target/fcrepo-home/data/ocfl-root/d55/2d0/d00/d552d0d00e9e775fcc59f6f047ee3d4afd374fe983fab3826af57ce350e60aa5/v2/content/.fcrepo/fcr-root.json
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:116)
	at java.base/sun.nio.fs.UnixFileSystemProvider.newByteChannel(UnixFileSystemProvider.java:215)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:370)
	at java.base/java.nio.file.Files.newByteChannel(Files.java:421)
	at java.base/java.nio.file.spi.FileSystemProvider.newInputStream(FileSystemProvider.java:420)
	at java.base/java.nio.file.Files.newInputStream(Files.java:155)
	at edu.wisc.library.ocfl.core.storage.filesystem.FileSystemOcflFileRetriever.retrieveFile(FileSystemOcflFileRetriever.java:59)
	... 125 common frames omitted
ERROR 11:46:57.177 [qtp1285487049-74] (BadRequestExceptionMapper) BadRequestExceptionMapper caught an exception: OCFL object info:fedora/test-container is not a valid Fedora 6 object. The following problems were identified:
  1. Unexpected failure validating version v2: NoSuchFileException: /Users/danielbernstein/code/fcrepo/fcrepo-webapp/target/fcrepo-home/data/ocfl-root/d55/2d0/d00/d552d0d00e9e775fcc59f6f047ee3d4afd374fe983fab3826af57ce350e60aa5/v2/content/.fcrepo/fcr-root.json
```

7. restart fedora
8. perform reindex:  all is well.

pwinckles:
ocfl-java caches deserialized inventories. This is likely what the problem is. The cache can be cleared with: https://github.com/UW-Madison-Library/ocfl-java/blob/main/ocfl-java-api/src/main/java/edu/wisc/library/ocfl/api/OcflRepository.java#L294

dbernstein:
@pwinckles : thanks - that's very helpful.

dbernstein (changes requested):
Code is good overall.  See comments about invalidating caches.

dbernstein (approved):
This looks good and works as advertised after building https://github.com/fcrepo/fcrepo-storage-ocfl/pull/50 .  It looks like there is a comment on that other PR, yes?  Once that is resolved and merged then this one can be merged.

dbernstein (approved):

PR #1975, opened 2022-02-11
Version Listing for Binaries

mikejritter:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3517

* Add an additional describedby link header for the memento description
* Update the HTML UI to show Binary versions and link to them using the memento description

# How should this be tested?

* Start Fedora
* Create a binary
* Update the binary
* Use the `View Versions` to navigate to the Binary version listing
* Browse to either version for the Binary and validate that:
  * the metadata is correct for the version
  * the page shows it's a historical version and is read only

# Interested parties
@fcrepo/committers


dbernstein (approved):
Looks and works great.  Thanks @mikejritter

dbernstein (changes requested):
There is one thing missing:  you need to provide a download button on the memento page that allows the user to download the historic binary (since the page is just displaying the metadata).

dbernstein (approved):
Code looks good and works as expected. Thanks @mikejritter

PR #1992, opened 2022-05-19
Use Set to deduplicate any extra versions

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3655

Fixes the `application/link-format` output when there are more than one version in a second. The RDF formats seem to automatically deduplicate, but we manually generate this format.

# How should this be tested?

This is most obvious with an automated script because you need to generate more than one version in a single second.

1. Once you have generated multiple versions in a single second, then `GET` the Timemap (`/fcr:versions`) as `text/turtle` and see only one version per second.
1. Get the Timemap with `Accept: application/link-format` and see duplicate versions.
1. Pull in this PR and do the second GET again and see only one per second

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersioning.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersioning.java
@@ -38,10 +38,10 @@ import java.io.IOException;
 import java.net.URI;
 import java.time.Instant;
 import java.time.ZoneId;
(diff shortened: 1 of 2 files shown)


bbpennel (approved):
Looks good in local testing, and a tricky bug to trigger, good work.

dbernstein:
I accidentally merged this into 6.2.0-RC .  I will push this to main right now.

PR #2001, opened 2022-07-19
remove transaction UI elements from HTML user interface

robyj:
- removed the JS listener calls in common.js
- removed the menu items in common-header.vsl

Resolves: https://fedora-repository.atlassian.net/browse/FCREPO-3829


bbpennel (approved):
Looks good, menu is gone for me and I don't see any unexpected lingering transaction code in the ui

PR #2002, opened 2022-07-28
Remove unused JS functions

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3835

Removes listeners referencing HTML IDs that have already been removed and functions they called.

# How should this be tested?

As these functions are already removed this PR should have no effect on the HTML UI

# Interested parties
@fcrepo/committers


dbernstein (approved):
Looks good.

PR #2009, opened 2022-10-13
Generate new HttpIdentifierConverter for new baseURIs

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3818

When accessing Fedora through a reverse proxy and using the HTML UI you can get a state where the first request comes from the external URI, but a second comes from the internal URI. Because we generate a HttpIdentifierConverter from the external URI it won't match the internal URI.

This generates a Map of HttpIdentifierConverters (though at most you should only create 2).

# How should this be tested?

The only real way is to setup a reverse proxy and ensure you have `ProxyPreserveHost On` in it.

Then when accessing the HTML UI with a browser (or an `Accept: text/html` header) you should get a stack trace in Fedora like
```
WARN 17:24:48.854 [qtp1749712938-32] (WildcardExceptionMapper) Unmapped exception
java.lang.IllegalArgumentException: Cannot translate NULL path extracted from URI http://web_scripts.localhost:8111/rest/
	at org.fcrepo.http.commons.api.rdf.HttpIdentifierConverter.toInternalId(HttpIdentifierConverter.java:83)
	at org.fcrepo.http.commons.api.rdf.HttpIdentifierConverter.toInternalId(HttpIdentifierConverter.java:58)
	at org.fcrepo.http.api.responses.StreamingBaseHtmlProvider.writeTo(StreamingBaseHtmlProvider.java:205)
	at org.fcrepo.http.api.responses.StreamingBaseHtmlProvider.writeTo(StreamingBaseHtmlProvider.java:83)
	at org.glassfish.jersey.message.internal.WriterInterceptorExecutor$TerminalWriterInterceptor.invokeWriteTo(WriterInterceptorExecutor.java:242)
	at org.glassfish.jersey.message.internal.WriterInterceptorExecutor$TerminalWriterInterceptor.aroundWriteTo(WriterInterceptorExecutor.java:227)
	at org.glassfish.jersey.message.internal.WriterInterceptorExecutor.proceed(WriterInterceptorExecutor.java:139)
	at org.glassfish.jersey.server.internal.JsonWithPaddingInterceptor.aroundWriteTo(JsonWithPaddingInterceptor.java:85)
	at org.glassfish.jersey.message.internal.WriterInterceptorExecutor.proceed(WriterInterceptorExecutor.java:139)
	at org.glassfish.jersey.server.internal.MappableExceptionWrapperInterceptor.aroundWriteTo(MappableExceptionWrapperInterceptor.java:61)
	at org.glassfish.jersey.message.internal.WriterInterceptorExecutor.proceed(WriterInterceptorExecutor.java:139)
	at org.glassfish.jersey.message.internal.MessageBodyFactory.writeTo(MessageBodyFactory.java:1116)
	at org.glassfish.jersey.server.ServerRuntime$Responder.writeResponse(ServerRuntime.java:635)
	at org.glassfish.jersey.server.ServerRuntime$Responder.processResponse(ServerRuntime.java:373)
	at org.glassfish.jersey.server.ServerRuntime$Responder.process(ServerRuntime.java:363)
	at org.glassfish.jersey.server.ServerRuntime$1.run(ServerRuntime.java:258)
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:248)
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:244)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:292)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:274)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:244)
```

Pull in this PR and it should work.

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
@@ -5,18 +5,18 @@
  */
 package org.fcrepo.http.api.responses;
 
+import static com.google.common.collect.ImmutableMap.builder;
 import static java.util.stream.Stream.of;
 import static javax.ws.rs.core.MediaType.TEXT_HTML_TYPE;
-import static com.google.common.collect.ImmutableMap.builder;
 import static org.apache.jena.graph.Node.ANY;
 import static org.apache.jena.sparql.util.graph.GraphUtils.multiValueURI;
 import static org.apache.jena.vocabulary.RDF.type;
 import static org.fcrepo.http.commons.domain.RDFMediaType.TEXT_HTML_WITH_CHARSET;
 import static org.fcrepo.http.commons.session.TransactionConstants.ATOMIC_ID_HEADER;
+import static org.fcrepo.kernel.api.RdfCollectors.toModel;
 import static org.fcrepo.kernel.api.RdfLexicon.ARCHIVAL_GROUP;
 import static org.fcrepo.kernel.api.RdfLexicon.NON_RDF_SOURCE;
 import static org.fcrepo.kernel.api.RdfLexicon.RDF_SOURCE;
-import static org.fcrepo.kernel.api.RdfCollectors.toModel;
 import static org.fcrepo.kernel.api.RdfLexicon.REPOSITORY_ROOT;
 import static org.slf4j.LoggerFactory.getLogger;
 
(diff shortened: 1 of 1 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/responses/StreamingBaseHtmlProvider.java
```diff
@@ -117,12 +120,8 @@ private Transaction transaction() {
...
 
     private HttpIdentifierConverter identifierConverter() {
-        if (identifierConverter == null) {
-            final UriBuilder uriBuilder =
-                    uriInfo.getBaseUriBuilder().clone().path(FedoraLdp.class);
-            identifierConverter = new HttpIdentifierConverter(uriBuilder);
-        }
-        return identifierConverter;
+        return identifierConverters.computeIfAbsent(uriInfo.getBaseUri(),
+                i -> new HttpIdentifierConverter(uriInfo.getBaseUriBuilder().clone().path(FedoraLdp.class)));
```
claussni, line 124: Why cloning the BaseUriBuilder?

claussni:
> When accessing Fedora through a reverse proxy and using the HTML UI you can get a state where the first request comes from the external URI, but a second comes from the internal URI.

1. We never had a reverse proxy setup when we noticed this. Yet we do access resources via curl and browser UI from the same machine. Would this have a similiar effect?
2. I don’t understand the scope of the StreamingBaseHtmlProvider . So I’m guessing from the code that it holds a map of HttpIdentifierConverterinstaces for each request that comes in with a new base URI as long as StreamingBaseHtmlProvider is in used? There are no annotaions on this provider. So what is its scope?

whikloj:
1. StreamingBaseHtmlProvider handles any `text/html` output for Fedora, so if you are using the HTML UI then yes you are using this class. I know this mimic'ed the issue you were experiencing on a intermittent basis, but the reverse proxy setup made the issue reproducible.

     So I am not 100% sure this will solve the issue you initially opened, but I do think it is conceivable the it was a similar case of the stored HttpIdentifierConverter using an incorrect pattern for subsequent requests.
2. To my understanding JAX-RS Providers are Application Scoped. They are initialized when the application starts and then the overridden `writeTo()` method is called to map the Java object to a specific representation.

     See section 4.2.2 for more on MessageBodyWriters - https://download.oracle.com/otn-pub/jcp/jaxrs-1.0-fr-eval-oth-JSpec/jaxrs-1.0-final-spec.pdf?AuthParam=1665756161_0606e4a689f55531241a47f979aa26bf

dbernstein (approved):

PR #2014, opened 2022-11-04
Upgrade OCFL version 1.1 and allow choice for upgrading old objects

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3861 and https://fedora-repository.atlassian.net/browse/FCREPO-3862

Upgrades the version of OCFL Fedora writes from 1.0 to 1.1. 
It also adds a configuration property `fcrepo.ocfl.upgrade.enabled` to configure whether older (1.0) OCFL objects will get new versions written in 1.1 or stay with 1.0. It defaults to `false` (i.e. write in the old version)

# How should this be tested?

1. How I did this was wrote out a new OCFL repo with Fedora built from `main` (i.e. OCFL 1.0) and added a 2 or 3 objects.
1. Then I copied that directory to a second directory.
1. Then I stood this version up on top of one of the directories with no property configured. 
       * I wrote a new object then verified the on-disk representation was in OCFL 1.1
       * I edited an existing object and verified that the object and new version were still OCFL 1.0
1. Then I stood this version up on top of one of the directories with `fcrepo.ocfl.upgrade.enabled=true` configured. 
       * I wrote a new object then verified the on-disk representation was in OCFL 1.1
       * I edited an existing object and verified that the object and new version were now OCFL 1.1

# Interested parties

### DIFF
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
@@ -114,6 +114,9 @@ public class OcflPropsConfig extends BasePropsConfig {
     @Value("${fcrepo.cache.db.ocfl.id_map.timeout.minutes:30}")
     private long fedoraToOcflCacheTimeout;
 
+    @Value("${fcrepo.ocfl.upgrade.enabled:false}")
+    private boolean ocflUpgradeOnWrite;
+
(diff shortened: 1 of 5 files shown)


whikloj:
This will also need updates to the [Properties wiki page](https://wiki.lyrasis.org/display/FEDORA6x/Properties) before it is considered done.

dbernstein (approved):
Nice work on  this one.

PR #2015, opened 2022-11-14
Add ocfl verify inventory prop

mikejritter:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3823

* Add prop for setting verify inventory flag for ocfl

# Interested parties

### DIFF
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
@@ -38,6 +38,7 @@ public class FedoraPropsConfig extends BasePropsConfig {
     private static final String FCREPO_ACTIVEMQ_DIRECTORY = "fcrepo.activemq.directory";
     private static final String FCREPO_SESSION_TIMEOUT = "fcrepo.session.timeout";
     private static final String FCREPO_VELOCITY_RUNTIME_LOG = "fcrepo.velocity.runtime.log";
+    private static final String FCREPO_REBUILD_VALIDATION = "fcrepo.rebuild.validation";
     private static final String FCREPO_REBUILD_VALIDATION_FIXITY = "fcrepo.rebuild.validation.fixity";
     private static final String FCREPO_REBUILD_ON_START = "fcrepo.rebuild.on.start";
     private static final String FCREPO_JMS_BASEURL = "fcrepo.jms.baseUrl";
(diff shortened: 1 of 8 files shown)


dbernstein (approved):

mikejritter:
A question for something I just looked at: do we want to skip object validation entirely as well if this flag is set? Currently the ObjectValidator in the ReindexSerivce would still run as it uses a separate flag that only impacts if file digests are run (`fcrepo.rebuild.validation.fixity`).

dbernstein:
@mikejritter :  I think so, no?  So if I'm understanding it, the user would have 2 options for rebuild:
1. all validation on (slowest, but high certainty)
2. fixity off but other object validations on (faster, less certainty)
3. all validations off (fastest, but least amount of certainty).
Is that your understanding?  If so I think we should have the option to completely disable all validations in the rebuild.

Perhaps we should be adding a new flag that applies just to the rebuild: 
fcrepo.rebuild.validation.enabled=(true|false) default = true

So that way you can disable validation after rebuild or just disable for the rebuild and leave everything validating.   I'm not sure how this ticket came to be.  I don't recall how it came up.  Apparently I created the ticket.

mikejritter:
> @mikejritter : I think so, no? So if I'm understanding it, the user would have 3 options for rebuild:
> 
>     1. all validation on (slowest, but high certainty)
> 
>     2. fixity off but other object validations on (faster, less certainty)
> 
>     3. all validations off (fastest, but least amount of certainty).
>        Is that your understanding?  If so I think we should have the option to completely disable all validations in the rebuild.

That's what it seems like to me, at least from looking at the ticket.
 
> Perhaps we should be adding a new flag that applies just to the rebuild: fcrepo.rebuild.validation=(true|false) default = true
> 
> So that way you can disable validation after rebuild or just disable for the rebuild and leave everything validating. I'm not sure how this ticket came to be. I don't recall how it came up. Apparently I created the ticket.
> 
> If fcrepo.rebuild.validation=false, but fcrepo.rebuild.validation.fixity=true we should report an exception and shutdown, I would think. Or if fcrepo.rebuild.validation=false then fcrepo.rebuild.validation.fixity would be ignored? What do you think is clearer?

I think an exception would make sense there, just in case one was set incorrectly by mistake.

dbernstein (approved):

dbernstein:
Just waiting on fix to broken tests with the latest update.

PR #2022, opened 2023-01-13
FCREPO-3689: Switch URL Decoder

mikejritter:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3869

Switch from the Java [URLDecoder](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URLDecoder.html) to Spring [StringUtils](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/StringUtils.html#uriDecode(java.lang.String,java.nio.charset.Charset)) when decoding uris for internal ids.

# How should this be tested?

* Build + Run Fedora
* Using a snippet of the script in the issue, the test cases can be run
```
#!/bin/sh

info_fedora="http://localhost:8080/rest"
put="curl -XPUT -u fedoraAdmin:fedoraAdmin -H'Content-Type:application/ld+json'"

for n in 'c&h' 'c+h' 'c_h'; do
 encoded=$(jq -sRr --arg x $n '$x|@uri');
 v=$(echo -n "info:fedora/$n" | shasum -a 256); echo $n $encoded $v;
 curl --no-progress-meter -v -XPUT -u fedoraAdmin:fedoraAdmin -H'Content-Type:application/ld+json' --data "{\"@id\":\"\",\"http://schema.org/name\":\"$n\"}" "${info_fedora}/$encoded"
done

for n in 'c&h' 'c+h' 'c_h'; do
 encoded=$(jq -sRr --arg x $n '$x|@uri');
 echo "=$n or $encoded ="
 curl --no-progress-meter -XGET -u fedoraAdmin:fedoraAdmin -H'Accept:application/ld+json' --head ${info_fedora}/$encoded | grep HTTP
 curl --no-progress-meter -XGET -u fedoraAdmin:fedoraAdmin -H'Accept:application/ld+json' ${info_fedora}/$encoded | jq -r '[.[]["@id"],.[]["http://schema.org/name"][]["@value"]]'
 echo '==html=='
 curl --no-progress-meter -XGET -u fedoraAdmin:fedoraAdmin -H'Accept:text/html' --head ${info_fedora}/$encoded | grep HTTP
done
```
* As a note: jq was waiting for input when doing the encoding, so I had to `ctrl+d` through each step.
* If `jq` isn't available, the encoded versions are `%2B` for `+` and `%26` for `&`, though only `+` is affected by this change. The third curl command would also need to be adjusted.

It might be good to test an encoded UTF-8 string as well. I did include a test in the IT for this though. 

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/StreamingBaseHtmlProviderIT.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/StreamingBaseHtmlProviderIT.java
@@ -1,0 +1,84 @@
+/*
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree.
+ */
+package org.fcrepo.integration.http.api;
+
+import static javax.ws.rs.core.Response.Status.OK;
+import static org.junit.Assert.assertEquals;
+
+import java.io.IOException;
+
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.methods.HttpPut;
+import org.junit.Test;
+import org.springframework.test.context.TestExecutionListeners;
+
+/**
+ * @author mikejritter
+ */
+@TestExecutionListeners(
+    listeners = { TestIsolationExecutionListener.class },
+    mergeMode = TestExecutionListeners.MergeMode.MERGE_WITH_DEFAULTS)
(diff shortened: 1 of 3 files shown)


awoods (approved):
Visual inspection looks great... and the ratio of unit and integration tests to functional updates is admirable.

It would be helpful to get @qjhart 's review/testing.

whikloj (approved):
👍 makes sense and my test of this works

PR #2026, opened 2023-01-26
Fcrepo 3867 - Add JDK 17 to build matrix for GitHub Actions

Surfrdan:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3867

Adds a matrix of JDK Versions and adds 17 to the build job, while being marked as experimental and allowed to fail. 

# How should this be tested?
Monitor other pushes and watch for errors in the remote build within GitHub Actions

Apologies for the many commits. Please squash. These were required fo remote testing 

# Interested parties
@fcrepo/committers

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
-        java: ['17']
+        java: ['11', '17']
     steps:
     - name: Git support longpaths
       run: git config --global core.longpaths true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 # This workflow will build a Java project with Maven
 # For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-java-with-maven
-#
+
 name: Build
 
 on:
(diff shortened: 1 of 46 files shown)


.github/workflows/build.yml
```diff
@@ -13,19 +13,28 @@ on:
 jobs:
   build:
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
+      max-parallel: 1
```
whikloj, line 18: It feels like this should not be required, do you know why it is?

LlGC-dof: It appeared to lessen the flapping tests by working in serial, however I concede this could have been co-incidental. It does have an impact on the processing time, but is this a time dependant task?

Surfrdan: It appeared to lessen the flapping tests by working in serial, however I concede this could have been co-incidental. It does have an impact on the processing time, but is this a time dependant task?

whikloj: No they are not really time sensitive (until Microsoft starts charging for cycles) but changes like these will stay forever until a problem pops up because we will forget why its there and assume it is required. I'm fine leaving it in, just noting that I will promptly forget this entire discussion and treat it like gospel 🤣

Surfrdan: In that case, maybe it would be better to remove the serialisation and work on fixing the flapping tests instead.

whikloj (approved):
max-parallel is actually not having a slow down effect in fact Windows seems faster so lets leave it in

PR #2028, opened 2023-02-15
Fcrepo 3839 

Surfrdan:
**Add Banner with registration pitch and link to registration page to All HTML UI pages**
* * *

**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3839

* Part of https://fedora-repository.atlassian.net/browse/FCREPO-3840

Adds a HTML UI banner with text and a link to the repository registration site.

# How should this be tested?
View in a variety of common browsers and confirm the new configuration property ```fcrepo.banner.display=on/off``` is working as expected. THe UI feature should be visible at the top of all pages within the UI other than the entry 'splash screen' (as found at /fcrepo)

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
@@ -135,8 +135,8 @@ public class FedoraPropsConfig extends BasePropsConfig {
     @Value("${fcrepo.cache.webac.acl.timeout.minutes:10}")
     private long webacCacheTimeout;
 
-    @Value("${fcrepo.register.banner:on}")
-    private String fcrepoRegisterBanner;
+    @Value("${fcrepo.banner.enabled:true}")
+    private boolean bannerEnabled;
 
 
     @PostConstruct
@@ -409,8 +409,8 @@ public class FedoraPropsConfig extends BasePropsConfig {
     /**
      * @return whether the repository registration banner should be displayed
      */
-    public String getFcrepoRegisterBanner() {
-        return fcrepoRegisterBanner;
+    public boolean getBannerEnabled() {
+        return bannerEnabled;
     }
 
 }
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
@@ -135,6 +135,10 @@ public class FedoraPropsConfig extends BasePropsConfig {
     @Value("${fcrepo.cache.webac.acl.timeout.minutes:10}")
     private long webacCacheTimeout;
 
+    @Value("${fcrepo.register.banner:on}")
+    private String fcrepoRegisterBanner;
+
+
     @PostConstruct
     private void postConstruct() throws IOException {
         LOGGER.info("Fedora home: {}", fedoraHome);
@@ -401,4 +405,12 @@ public class FedoraPropsConfig extends BasePropsConfig {
     public long getWebacCacheTimeout() {
         return webacCacheTimeout;
     }
+
+    /**
+     * @return whether the repository registration banner should be displayed
+     */
+    public String getFcrepoRegisterBanner() {
+        return fcrepoRegisterBanner;
+    }
+
 }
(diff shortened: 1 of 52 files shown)


whikloj:
Building this now, I do think using `true` or `false` here as you mentioned in [this comment](https://github.com/fcrepo/fcrepo/pull/2029#issuecomment-1434911543) makes more sense as it aligns with the other properties.

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
```diff
@@ -135,6 +135,10 @@ public class FedoraPropsConfig extends BasePropsConfig {
     @Value("${fcrepo.cache.webac.acl.timeout.minutes:10}")
     private long webacCacheTimeout;
 
+    @Value("${fcrepo.register.banner:on}")
+    private String fcrepoRegisterBanner;
```
whikloj, line 139: Testing instructions suggested the property was `fcrepo.banner.display` but it is actually `fcrepo.register.banner`. I think going closer to the first and using the `enabled` that we do for other things (drawing in the idea of using a boolean). Something like `fcrepo.banner.enabled=true|false`

fcrepo-http-api/src/main/resources/views/common-register-header.vsl
```diff
@@ -0,0 +1,5 @@
+#if($fedoraProps.getFcrepoRegisterBanner() == "on")
+<div class="alert alert-info" role="alert">
+Fedora is an open source project supported entirely by it's users. To continue the on-going maintenance of the software and ensure Fedora is meeting the needs of the community, we are working to gather information on current installations. We strongly encourage all users to register their instance in the DuraSpace Community Supported Program Registry hosted by LYRASIS. <a class="btn btn-primary" href="https://registry.lyrasis.org/" role="button">Register your installation today</a>
+</div>
```
whikloj, line 4: This is just a suggestion, but maybe wrapping the paragraph of text and the button in two separate `<p></p>` tags so they get some space from each other. 
```html
<div class="alert alert-info" role="alert">
<p>Fedora is an open source project supported entirely by it's users. To continue the on-going maintenance of the software and ensure Fedora is meeting the needs of the community, we are working to gather information on current installations. We strongly encourage all users to register their instance in the DuraSpace Community Supported Program Registry hosted by LYRASIS.</p>
<p><a class="btn btn-primary" href="https://registry.lyrasis.org/" role="button">Register your installation today</a></p>
</div>
```

fcrepo-http-api/src/main/resources/views/binary.vsl
```diff
@@ -17,6 +17,7 @@
 
 <body class="fcrepo_binary">
 <div  id="main" class="container" resource="$topic.getURI()">
+    #parse("views/common-register-header.vsl")
     #parse("views/common-node-header.vsl")
```
whikloj, line 21: The banner is not showing up on fixity (http://localhost:8080/rest/some-binary/fcr:fixity) or version (http://localhost:8080/rest/some-example/fcr:versions) pages. 

It might make more sense to insert the `#parse("views/common-register-header.vsl")` into the `common-node-header.vsl` and `common-header.vsl` templates instead of into each individual page. Those two seem to be included in all the other pages.

whikloj (changes requested):
2 changes and 1 suggestion above....or below 🤦

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
Surfrdan: Good points. I'll also remove the superflous 'fcrepo' from the java attribute names and getter so that it becomes ```bannerEnabled```  to follow the other property to attribute mappings within the class.

whikloj: No I would not remove the `fcrepo` as if someone assigns this on the java command line it could be unclear what it is for.
```bash
> ... -Dbanner.enabled=false
```
versus
```bash
> ... -Dfcrepo.banner.enabled=false
```

Surfrdan: I am refering to the java attributes, not the java properties. So:
```fcrepoRegisterBanner```
vs
```registerEnable``` 
within the java source ```fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java``` 
Agreed that the fcrepo prefix should stay for the property

Hope that makes sense

whikloj: Ahh okay

whikloj (approved):
Looks good and works as described.

PR #2029, opened 2023-02-17
Fcrepo 3841

Surfrdan:
**Add text and registration link to registration page to all API responses**
* * *

**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3841

* part of https://fedora-repository.atlassian.net/browse/FCREPO-3840

Adds 2 configurable headres to all API responses to link to the repo registration page

# How should this be tested?

* Ensure that the headers appear and are valid whilst also being removable by the ```fcrepo.register.api.banner``` property being set to ```false```

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
@@ -135,6 +135,10 @@ public class FedoraPropsConfig extends BasePropsConfig {
     @Value("${fcrepo.cache.webac.acl.timeout.minutes:10}")
     private long webacCacheTimeout;
 
+    @Value("${fcrepo.register.api.banner:true}")
+    private boolean fcrepoRegsiterApiBanner;
+
+
     @PostConstruct
     private void postConstruct() throws IOException {
         LOGGER.info("Fedora home: {}", fedoraHome);
@@ -400,5 +404,12 @@ public class FedoraPropsConfig extends BasePropsConfig {
      */
     public long getWebacCacheTimeout() {
         return webacCacheTimeout;
+    }
+
+    /**
+     * @return whether to display the registration link banner in API responses or not.
+     */
+    public boolean getFcrepoRegsiterApiBanner() {
+        return fcrepoRegsiterApiBanner;
     }
 }
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -161,6 +161,11 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
 
     static final String HTTP_HEADER_ACCEPT_PATCH = "Accept-Patch";
 
+    static final String REGISTER_REPO_LINK = "https://registry.lyrasis.org/";
+    static final String REGISTER_REPO_TEXT = "We strongly encourage all users to register their instance " +
+                                             "in the DuraSpace Community Supported Program Registry hosted " +
+                                             "by LYRASIS. Register your installation today:";
+
     private static final String FCR_PREFIX = "fcr:";
     private static final Set<String> ALLOWED_FCR_PARTS = Set.of(FCR_METADATA, FCR_ACL);
 
@@ -589,6 +594,14 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
 
     protected void addResourceHttpHeaders(final FedoraResource resource) {
         addResourceHttpHeaders(resource, false);
+    }
+
+
+    protected void addRegisterHeaders() {
+        if (fedoraPropsConfig.getFcrepoRegsiterApiBanner() == true) {
+            servletResponse.addHeader("USER-REGISTRATION", REGISTER_REPO_TEXT + " " + REGISTER_REPO_LINK);
+            servletResponse.addHeader("Link", "<" + REGISTER_REPO_LINK + ">; rel=\"user-regisration\"");
+        }
     }
 
     /**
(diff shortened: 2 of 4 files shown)


Surfrdan:
Note that I used a boolean in this property but a string 'on' in the previous UI ticket. I'd initially done this becasue I'd read that the solution to use a boolean in the @Value annotation was a bit hacky but it seems to work fine here. I'd consider moving the other property to a boolean too to maintain consistancy.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -591,6 +596,14 @@ protected void addResourceHttpHeaders(final FedoraResource resource) {
         addResourceHttpHeaders(resource, false);
     }
 
+
+    protected void addRegisterHeaders() {
+        if (fedoraPropsConfig.getFcrepoRegsiterApiBanner() == true) {
+            servletResponse.addHeader("USER-REGISTRATION", REGISTER_REPO_TEXT + " " + REGISTER_REPO_LINK);
+            servletResponse.addHeader("Link", "<" + REGISTER_REPO_LINK + ">; rel=\"user-regisration\"");
```
whikloj, line 603: To be valid against the rules of RFC 8288 for [extension relations](https://www.rfc-editor.org/rfc/rfc8288#section-2.1.2) the `rel` needs to be in the form of a URI. Sort of like how we do for [external content](https://wiki.lyrasis.org/display/FEDORA6x/External+Content)

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
```diff
@@ -401,4 +405,11 @@ public long getWebacCacheSize() {
     public long getWebacCacheTimeout() {
         return webacCacheTimeout;
     }
+
+    /**
+     * @return whether to display the registration link banner in API responses or not.
+     */
+    public boolean getFcrepoRegsiterApiBanner() {
```
whikloj, line 412: Typo `getFcrepoRegisterApiBanner()`

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
```diff
@@ -135,6 +135,10 @@ public class FedoraPropsConfig extends BasePropsConfig {
     @Value("${fcrepo.cache.webac.acl.timeout.minutes:10}")
     private long webacCacheTimeout;
 
+    @Value("${fcrepo.register.api.banner:true}")
+    private boolean fcrepoRegsiterApiBanner;
```
whikloj, line 139: Typo `fcrepoRegisterApiBanner`

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
```diff
@@ -135,6 +135,10 @@ public class FedoraPropsConfig extends BasePropsConfig {
     @Value("${fcrepo.cache.webac.acl.timeout.minutes:10}")
     private long webacCacheTimeout;
 
+    @Value("${fcrepo.register.api.banner:true}")
```
whikloj, line 138: I feel like we probably should just have this use the same property as in https://github.com/fcrepo/fcrepo/pull/2028, I can't think of a person that would want to only disable this on the HTML UI or the HTTP API and not both.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -591,6 +596,14 @@ protected void addResourceHttpHeaders(final FedoraResource resource) {
         addResourceHttpHeaders(resource, false);
     }
 
+
+    protected void addRegisterHeaders() {
+        if (fedoraPropsConfig.getFcrepoRegsiterApiBanner() == true) {
+            servletResponse.addHeader("USER-REGISTRATION", REGISTER_REPO_TEXT + " " + REGISTER_REPO_LINK);
```
whikloj, line 602: I personally think this is not a great idea as we are creating a new HTTP Header field, that carries a bunch of text on every response which is only human actionable (unlike the Link header which is machine actionable). Might be good to get some other's input at a Thursday call. 🤷

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -789,6 +789,7 @@ protected void addResourceHttpHeaders(final FedoraResource resource, final boole
             servletResponse.addHeader(LINK, "<" + canonical + ">;rel=\"canonical\"");
 
         }
+        addRegisterHeaders();
```
whikloj, line 792: This (`addResourceHttpHeaders`) only adds the headers to requests that return a resource body, so a `GET` request (in most cases) and `HEAD`. So `DELETE`, `PATCH`, `OPTIONS`, `POST` and `PUT` requests don't get them.

Also no requests for `fcr:acl`, `fcr:versions`, `fcr:fixity` and `fcr:tx`. These are handled by other classes in fcrepo-http-api (i.e. `FedoraAcl`, `FedoraTombstones`, `FedoraFixity`, etc).

whikloj (changes requested):

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
mikejritter: Going to piggy back on this to ask if we really want additional text like this in the headers at all? There doesn't seem to be a good fit ([from the mdn reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers)), and they're more for the client/server rather than end user anyway. 

It seems to me that having the link header pointing to the registry should suffice so long as the `rel` describes it well enough.

Surfrdan: I tend to agree, but I presume the request came from the team somewhere when the jira ticket was created. Was there a reason at that time to do it this way?

whikloj: You would have to check with @dbernstein as this came out of discussions with the leadership group.

Surfrdan:
Governance agreed with the tech team that his would not be useful in the API response. We can close this

PR #2030, opened 2023-02-21
FCREPO-3830: Update ACL for Transaction Endpoint

mikejritter:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3830

Updates the ACL logic to allow for checks against the fcr:tx endpoint

# Changes

* Iterate default acls by subject to create multiple WebACAuthorizations
* Allow getting of roles for non-resource endpoints
* Check for match on fcr:tx when getting roles
* Update WebACFilter logic for transactions

# How should this be tested?

* Verify build works and all tests pass
  * mvn clean verify
* Run Fedora in your preferred way
* Test that with the default ACL, only fedoraAdmins should be able to interact with the fcr:tx endpoint
* Stop Fedora and update the default root ACL to include fcr:tx permissions, e.g.
```
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .
@prefix fedora: <http://fedora.info/definitions/v4/repository#> .
@prefix webac: <http://fedora.info/definitions/v4/webac#> .

<info:fedora/fcr:acl> a webac:Acl .

<info:fedora/fcr:acl#authz> a acl:Authorization ;
   rdfs:label "Root Authorization" ;
   rdfs:comment "By default, all non-Admin agents (foaf:Agent) only have read access (acl:Read) to the repository" ;
   acl:agentClass foaf:Agent ;
   acl:mode acl:Read ;
   acl:accessTo <info:fedora> ;
   acl:default <info:fedora> .

<info:fedora/fcr:tx#authz_read_write> a acl:Authorization ;
   rdfs:label "Test Tx Authorization" ;
   rdfs:comment "Provide read write access to the transaction endpoint" ;
   acl:agentClass acl:AuthenticatedAgent ;
   acl:mode acl:Read, acl:Write ;
   acl:accessTo <info:fedora/fcr:tx> ;
   acl:default <info:fedora/fcr:tx> .
``` 
* Run Fedora again and verify that authenticated users can now interact with the fcr:tx endpoint

# Notes

I left the default root acl the same. I've been thinking about whether or not we want to update it but haven't come to a decision on my own. Either way we'll want to update the documentation to showcase this scenario.

# Interested parties


fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACFilter.java
```diff
@@ -453,6 +453,14 @@ private boolean isAuthorized(final Subject currentUser, final HttpServletRequest
                     log.debug("DELETE prohibited without {} permission", toControl);
                     return false;
                 }
+            } else if (isTxEndpoint) {
```
mikejritter, line 456: I was considering not doing this so that there isn't any transaction specific logic, however the `isAuthorizedForMembershipResource` throws a NPE as both the resource and container are null. Not really a big deal but something I wanted to mention.

whikloj:
So I ran this and it works as described in the instructions, but I also tried to `PUT` an Acl to the root via the REST API. 

That seems to work for creating resources but didn't seem to work for creating transactions. 

I have to formalize some tests and see exactly what is allowed in `main` right now, but you can `PUT` something to `http://localhost:8080/rest/fcr:acl`, `GET` it out and it did seem to work for changing some rights.

mikejritter:
@whikloj I was able to get back to this today and did some testing and see the issue where if you `PUT` to the root it doesn't get used. I can see what's going wrong in my changes so I'll need to make adjustments for that.

mikejritter:
@whikloj I just updated to check against the root resource first before reading the default acls from disk. There's only one change that I'm not entirely sure about that I'm going to add a comment for.

I also rebased against `main` so had to force push.

fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACAuthorizingRealm.java
```diff
@@ -185,10 +186,17 @@ private Map<String, Collection<String>> getRolesForPath(final String path) {
...
     private Map<String, Collection<String>> getRolesForId(final FedoraId id) {
         Map<String, Collection<String>> roles = null;
-        final FedoraResource fedoraResource = getResourceOrParentFromPath(id);
-        if (fedoraResource != null) {
-            // check ACL for the request URI and get a mapping of agent => modes
-            roles = rolesProvider.getRoles(fedoraResource, transaction());
+
+        final var txId = FEDORA_ID_PREFIX + "/" + FCR_TX;
+        if (id.getResourceId().startsWith(txId)) {
+            final FedoraResource rootResource = getResourceOrParentFromPath(FedoraId.getRepositoryRootId());
```
mikejritter, line 192: I'm considering changing this to `getResourceOrParentFromPath(id)`. In the event the id matches a tx, it will fallback to `info:fedora` since that will be first parent in its path. Probably not necessary just something I'm on the fence about.

fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACRolesProvider.java
```diff
@@ -425,23 +484,29 @@ Optional<ACLHandle> getEffectiveAcl(final FedoraResource resource, final boolean
...
-                nodeToStringStream(triple.getObject()).forEach(values::add);
-                if (predicate.equals(WEBAC_AGENT_VALUE)) {
-                    additionalAgentValues(triple.getObject()).forEach(values::add);
+        final var defaultAcls = getDefaultAcl(null, authPropsConfig.getRootAuthAclPath());
+        final var aclSubjects = defaultAcls.listSubjects();
+
+        aclSubjects.forEach(aclResource -> {
+            final Map<String, List<String>> aclTriples = new HashMap<>();
+            aclResource.listProperties().mapWith(Statement::asTriple).forEach(aclTriple -> {
+                if (hasAclPredicate.test(aclTriple)) {
```
whikloj, line 495: Just wondering if there is a reason you didn't add this as a `filter(hasAclPredicate)` between the `.mapWith(Statement::asTriple)` and `forEach(aclTriple ->..` ? 
Not sure it has any impact on performance and there might been a reason that I'm missing.

whikloj (approved):
This looks great and works against my testing. Just a question, otherwise I'm good to merge.

fcrepo-auth-webac/src/main/java/org/fcrepo/auth/webac/WebACRolesProvider.java
mikejritter: Looking back, this is similar to how the logic was before (with the `hashAclPredicate.test` separate). I'm not really sure why, my only thought is that this is using iterators from jena rather than a stream so it's possible filter wasn't available when this was originally done. That said jena provides `filterDrop` and `filterKeep` so I'll see about using one of those.

whikloj: Don't worry about it, I really just wanted to know if there was a reason I wasn't seeing. If we find a delay we can review later.

PR #2035, opened 2023-04-05
FCREPO-3876: NPEs when using Archive Groups

mikejritter:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3876

This updates fixes NPEs in two API calls - one for fcr:versions and the other using `Accept-Datetime` header. The NPE was being caused by trying to get the memento datetime on a tombstone, which obviously does not exist. Adding an additional check to make sure the resources are mementos resolves this issue.

Additionally, after fixing the NPE for the `Accept-Datetime` call an HTTP 500 was being returned. This was being caused by the `Content-Length` header being set and having no body in the response. Normally we would expect to see a 302 so I wasn't entirely sure what was going on, but jetty was not happy with this when finishing the response and was creating the exception. To resolve this I opted to pass a flag `isRedirect` so that we can skip setting any headers when not needed. I'm not sure if this is the best option but it simply is one. 

Also I don't have an Integration test created for this yet, but can do so if required. 

# How should this be tested?

Use the script from https://fedora-repository.atlassian.net/browse/FCREPO-3866 for the two bugs

or 

* Create an archive group
* Create a resource in the AG
* Deleted the resource + tombstone
* Recreate the resource in the AG
* Test the API calls for
  * `http://localhost:8080/fcrepo/rest/archivegroup/resource/fcr:versions`
  * `Accept-Datetime: $(date) http://localhost:8080/fcrepo/rest/archivegroup/resource`

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -591,12 +591,18 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
         addResourceHttpHeaders(resource, false);
     }
 
+    protected void addResourceHttpHeaders(final FedoraResource resource, final boolean dispositionInline) {
+        addResourceHttpHeaders(resource, dispositionInline, false);
+    }
+
     /**
      * Add any resource-specific headers to the response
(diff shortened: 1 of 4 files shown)


whikloj (approved):
This works as described

PR #2037, opened 2023-05-08
FCREPO-3890: Add fedora version info endpoint

mikejritter:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3890

* Creates new endpoint /systeminfo for displaying version info
* Adds version and git commit to web ui
* Update the header on the root ui page to be consistent with other pages

# How should this be tested?
* Build Fedora, e.g.
  * `mvn clean verify -Pone-click`
* Run Fedora, e.g.
  * `java -jar fcrepo-webapp/target/fcrepo-webapp-6.5.0-SNAPSHOT-jetty-console.jar`
* Check the `/fcr:systeminfo` info returns the correct response
  * `curl -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/fcr:systeminfo`
* Check the web ui and make sure the footer exists on all pages

# Other Notes

I haven't tested acls on the new endpoint but might be worth trying.

The splash page also displays the version and commit, however does it in a different way (seems like it gets set during the build, and uses a different plugin for getting the git commit).

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraSystemInfo.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraSystemInfo.java
@@ -1,0 +1,49 @@
+/*
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree.
+ */
+package org.fcrepo.http.api;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+import java.util.Map;
+import javax.inject.Inject;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.Response;
+
+import io.micrometer.core.annotation.Timed;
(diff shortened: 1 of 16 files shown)


Surfrdan:
Tested in tomcat and both UI and CLI responses look good to me. I tried every UI page I was aware of.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraSystemInfo.java
```diff
@@ -0,0 +1,49 @@
...
+    public FedoraSystemInfo() {
+        super();
+    }
+
+    @GET
+    @Produces(APPLICATION_JSON)
+    public Response getSystemInfo() {
+        final var systemInfo = Map.of("version", config.getImplementationVersion(),
+                                      "commit", config.getGitCommit());
+        return Response.ok().entity(systemInfo).build();
```
Surfrdan, line 46: Would it be worth returning a non-200 response if unable to assertain the version or commit ? The endpoint could form part of a rudimentary health endpoint at a later stage too

Surfrdan (approved):

PR #2049, opened 2023-10-14
Return mementos and timemaps on Tombstones

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3882

Adds a Link header to the Timemap of the original resource to the Tombstone Exception returned.
Allows access to get the timemap resource of the deleted objects.
Allows access to get the mementos of the deleted objects.

# How should this be tested?

1. Create an object (either RDF or binary)
2. Make a change to that object
3. Delete the object 

| Step | Before PR | After PR |
| --- | --- | --- |
| 4. Get the object | Receive Tombstone with Tombstone Link header | Receive Tombstone with Tombstone and Timemap Link header |
| 5. Get the timemap | Tombstone returned | Timemap is returned |
| 6. Get any memento | Tombstone returned | Memento is returned |

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersioning.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersioning.java
@@ -169,7 +169,13 @@ public class FedoraVersioning extends ContentExposingResource {
         TEXT_HTML_WITH_CHARSET, APPLICATION_LINK_FORMAT })
     public Response getVersionList(@HeaderParam("Accept") final String acceptValue) throws IOException {
 
-        final FedoraResource theTimeMap = resource().getTimeMap();
+        FedoraResource theTimeMap;
+        try {
+            theTimeMap = resource().getTimeMap();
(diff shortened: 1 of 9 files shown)


fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraVersioning.java
```diff
@@ -243,4 +244,16 @@ public Response delete() {
     protected String externalPath() {
         return externalPath;
     }
+
+    @Override
+    protected FedoraResource resource() {
```
mikejritter, line 249: If we POST on `/fcr:versions` when the head is a tombstone, is that ok? It seems like before it would have been returning a 410 but this would now be returning a 201.

I haven't actually tried this yet though (or know what behavior we should have), just something I was thinking about when looking at this file.

mikejritter (commented):
@whikloj Looks good and works as expected. Just one question about a potential side effect on the fcr:versions endpoint.

mikejritter (approved):

PR #2061, opened 2023-11-07
bumping ActiveMQ to non-exploited version

Surfrdan:
**Bumping ActiveMQ to non-exploited version**
* * *

**JIRA Ticket**: https://fedora-repository.atlassian.net/jira/software/c/projects/FCREPO/issues/FCREPO-3911

only bumps minor version of activemq

# Interested parties

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
     <project_organization>fcrepo</project_organization>
     <fcrepo-storage-ocfl.version>6.4.0-SNAPSHOT</fcrepo-storage-ocfl.version>
     <!-- Dependency version properties -->
-    <activemq.version>5.16.3</activemq.version>
+    <activemq.version>5.16.7</activemq.version>
     <aws.version>2.17.50</aws.version>
     <caffeine.version>3.0.4</caffeine.version>
     <commons-codec.version>1.15</commons-codec.version>


tomcbe (approved):

pom.xml
```diff
@@ -24,7 +24,7 @@
     <project_organization>fcrepo</project_organization>
     <fcrepo-storage-ocfl.version>6.4.0-SNAPSHOT</fcrepo-storage-ocfl.version>
     <!-- Dependency version properties -->
-    <activemq.version>5.16.3</activemq.version>
+    <activemq.version>5.16.7</activemq.version>
```
whikloj, line 27: Wondering if we should try for a more recent version like `5.17.6` or `5.18.3`?

tomcbe: If we update, I think it would make sense to go to 5.18.3 directly.

Surfrdan: My only worry about that was the possible breaking changes by jumping 2 major versions rather than just minor versions.

whikloj: I would assume that a breaking change would have bumped the major version from 5 to 6, but we will have to test it out anyways. There is no rush on this now.

Surfrdan: According to https://docs.aws.amazon.com/amazon-mq/latest/developer-guide/activemq-version-management.html 5.16 is a major version, as it 5.18. They use defferent terminology to the sem,ver spec awkwardly. 

I'm happy to trial 5.18 but can we push this PR as is now and do that in a future release along with a bunch of other updates from Dependapot which have been missed. I'll open another ticket about that

tomcbe: Fine for me

whikloj (approved):
👍

PR #2067, opened 2023-12-01
Allow continuation of a reindex

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/jira/software/c/projects/FCREPO/issues/FCREPO-3907

Adds a new flag to allow the continuation of a failed re-indexing.

# How should this be tested?

In theory you need a lot of objects.
1. Start up the site either with empty indexes to with `fcrepo.rebuild.on.start=true` set.
2. Stop or wait for it to fail (assuming it does normally fail)
3. Set `fcrepo.rebuild.continue=true`, you can remove the `fcrepo.rebuild.on.start` flag but this flag takes precendence so you will still be okay.
4. Start Fedora.

The messages will now show a count of those skipped items as well as successful and errors. At the end you should have all your items available.

# Interested parties

### DIFF
--- a/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexService.java
+++ b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexService.java
@@ -35,6 +35,8 @@ import org.fcrepo.kernel.api.rdf.DefaultRdfStream;
 import org.fcrepo.kernel.api.services.MembershipService;
 import org.fcrepo.kernel.api.services.ReferenceService;
 import org.fcrepo.persistence.api.PersistentStorageSessionManager;
+import org.fcrepo.persistence.api.exceptions.ObjectExistsInOcflIndexException;
+import org.fcrepo.persistence.ocfl.api.FedoraOcflMappingNotFoundException;
 import org.fcrepo.persistence.ocfl.api.FedoraToOcflObjectIndex;
 import org.fcrepo.search.api.Condition;
 import org.fcrepo.search.api.InvalidQueryException;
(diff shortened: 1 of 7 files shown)


fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexService.java
```diff
@@ -113,6 +115,21 @@ public void indexOcflObject(final Transaction tx, final String ocflId) {
...
+                        throw new ObjectExistsInOcflIndexException(
+                                String.format("Skipping indexing of %s in transaction %s, because" +
+                                " it already exists in the index.", fedoraId, tx.getId())
+                        );
+                    } catch (FedoraOcflMappingNotFoundException e) {
+                        LOGGER.debug("Indexing object {} in transaction {}, because it does not yet exist in the " +
+                                "index.", fedoraId, tx.getId());
+                    }
+                }
+
```
mikejritter, line 132: Do you know if it's possible to be in a state where a resource has only been partially indexed? I was testing yesterday and I from what it looks like the transaction is tied to the entire object so it _should_ all be committed (or rolled back) at the same time, so I don't think this will be an issue but I haven't seen how larger indexing operations look after they've stopped. I might try to make a stress test later to see if I can find out at all.

mikejritter (commented):
I built and tested this to confirm that it works. I ended up removing some of the rows from the index and then used the flag to trigger the rebuild. 

Everything else looks good, just one question that I don't think will be an issue but figured I'd ask about.

fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/ReindexService.java
whikloj: My understanding is that the entire object is indexed or not indexed in a single transaction. So all of the various indexes should either be updated or not. But I would be happy to have you press on this to ensure I have not underestimated the problem and only partially solved the issue.

mikejritter: I haven't been able to produce any partially indexed objects, although my test dataset ended up being a little smaller than I'd liked. Given that this is working for NLM, and that the logic indexes each object in a single transaction I'm ok with this.

An unrelated note, and something I need to make a jira for, is that sufficiently large objects (e.g. archival group with many children/versions) can cause the reindex transaction to timeout. I'm not sure what type of limit for an atomic resource would cause this but I saw it with an archival group.

mikejritter (approved):

PR #2069, opened 2023-12-06
Enable configurable Pairtree pid minter

Surfrdan:
**Enable configurable pair tree pid minter.**
* * *

**JIRA Ticket**: https://fedora-repository.atlassian.net/jira/software/c/projects/FCREPO/issues/FCREPO-3916

At the request of the Samvera Fedora 6 Working group, I have enabled pair tree style PID minting in Fedora 6

# How should this be tested?

* set properties ```fcrepo.pid.minter.length```  (2) and ```firepower.pid.minter.count``` (4) to replicate the Fedora PairTree PIDs of old.  Easiest way is ``` mvn jetty:run -Dfcrepo.pid.minter.length=2 -Dfcrepo.pid.minter.count=4 ```
* Try creating new containers in the UI. PIDs should resemble PairTree IDs 
* If you don't set the properties, standard UUID PIDs will be generated

# Interested parties


fcrepo-kernel-api/src/test/java/org/fcrepo/kernel/api/services/functions/ConfigurableHierarchicalSupplierTest.java
```diff
@@ -18,14 +21,14 @@
...
-    public void setUp() {
-        fedoraPropsConfig = new FedoraPropsConfig();
-    }
+@SpringBootTest
+@SpringBootConfiguration
+@RunWith(SpringJUnit4ClassRunner.class)
+@TestPropertySource(properties = {
+    "fcrepo.pid.minter.length=2",
+    "fcrepo.pid.minter.count=4"
+})
```
whikloj, line 30: Assuming this sets the properties for all the tests, then the first two tests should be failing now. I'm having trouble pulling a ref to the merge of this PR. But reading these changes it appears that originally we set the length and count on one test, so setting it for all 3 should cause a failure.

fcrepo-kernel-api/src/test/java/org/fcrepo/kernel/api/services/functions/ConfigurableHierarchicalSupplierTest.java
```diff
@@ -18,14 +21,14 @@
...
+public abstract class ConfigurableHierarchicalSupplierTest {
```
whikloj, line 31: This probably doesn't need to be `abstract`

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -941,8 +941,8 @@ private FedoraId mintNewPid(final FedoraId fedoraId, final String slug) {
         }
         if (!isBlank(slug)) {
             pid = slug;
-        } else if (pidMinter != null) {
-            pid = pidMinter.get();
+        // else if (pidMinter != null) {
+        //    pid = pidMinter.get();
```
whikloj, line 945: This allowed users to develop their own class to mint pids, we shouldn't disable it. If it is not set, then the default pid minter is used.

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/AbstractResource.java
```diff
@@ -58,8 +58,7 @@ public class AbstractResource {
     @Optional
     protected Supplier<String> pidMinter;
 
-    // Mint non-hierarchical identifiers. To force pairtree creation as default, use
-    //  ConfigurableHierarchicalSupplier(int length, count) instead.
+    @Inject
     protected UniqueValueSupplier defaultPidMinter = new ConfigurableHierarchicalSupplier();
```
whikloj, line 62: Don't need `@Inject` here, you are instantiating a `new ConfigurableHierarchicalSupplier();` anyways.

fcrepo-kernel-api/src/test/java/org/fcrepo/kernel/api/services/functions/ConfigurableHierarchicalSupplierTest.java
```diff
@@ -39,11 +50,11 @@ public void testGetIdNoPairtree() {
...
+        final UniqueValueSupplier defaultPidMinter = new ConfigurableHierarchicalSupplier();
         final String id = defaultPidMinter.get();
         // With (desiredLength > 0 && desiredCount > 0) check to see that id contains (count + 1) parts
         final int parts = (id.split("/").length);
         assertEquals(5, parts);
     }
 
-}
\ No newline at end of file
+}
```
whikloj, line 60: So this test can be done using `setField` which allows you to "Inject" the class, but still change it in the test.
```java
@RunWith(MockitoJUnitRunner.Silent.class)
public class ConfigurableHierarchicalSupplierTest {

    private final FedoraPropsConfig propsConfig = new FedoraPropsConfig();

    private final UniqueValueSupplier defaultPidMinter = new ConfigurableHierarchicalSupplier();

    @Before
    public void setUp() {
        // Need to set the defaults 
        propsConfig.setFcrepoPidMinterLength(0);
        propsConfig.setFcrepoPidMinterCount(0);
        setField(defaultPidMinter, "fedoraPropsConfig", propsConfig);
    }

    @Test
    public void testGet() {
        final String id = defaultPidMinter.get();
        // No pairtrees is default; with no args check to see that id contains just the 1 pid part
        final int parts = (id.split("/").length);
        assertEquals(1, parts);
    }

    @Test
    public void testGetIdNoPairtree() {
        final String id = defaultPidMinter.get();
        // With (desiredLength,desiredCount=0), check to see that id contains 1 part and no slashes
        final int parts = (id.split("/").length);
        assertEquals(1, parts);
    }

    @Test
    public void testGetIdPairtreeParams() {
        // Alter the settings for this test.
        propsConfig.setFcrepoPidMinterLength(2);
        propsConfig.setFcrepoPidMinterCount(4);
        final String id = defaultPidMinter.get();
        // With (desiredLength > 0 && desiredCount > 0) check to see that id contains (count + 1) parts
        final int parts = (id.split("/").length);
        assertEquals(5, parts);
    }
}
```

whikloj (changes requested):

fcrepo-kernel-api/src/test/java/org/fcrepo/kernel/api/services/functions/ConfigurableHierarchicalSupplierTest.java
Surfrdan: I've made the recommended changes but I'm back to getting an NPE on the ``` final String id = defaultPidMinter.get();``` liens in each test

whikloj (approved):
👍

PR #2076, opened 2024-03-21
FCREPO-3897: Fix html ui error message when nesting AGs

mikejritter:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3897

This updates to use the mixin when creating the error message when creating a resource. The mixins are defined in common-node-actions.vsl and have the same value as the resource type you're trying to create, e.g. `archival group`.

# How should this be tested?

* Run fedora
```
mvn jetty:run -pl fcrepo-webapp -Dfcrepo.log=DEBUG
```
* Create an archival group in the html ui
* Create a nested archival group in the html ui
* The error message should now display `Error creating archival group`

# Interested parties
@fcrepo/committers


whikloj (approved):
Works as expected 👍

PR #2079, opened 2024-04-30
Add a SMT for OCFL path

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3935

Generates a Server Managed Triple `fedora:ocflPath` with the full path to the base of the resource.

# How should this be tested?

This is a bit hacky as the only way to know if something is in staging or the repo is to get that information from `fcrepo-storage-ocfl` and pass it back as part of the Resource Headers (I think). 

That seemed like a large change, which I think involves to change the resource headers object (in fcrepo-storage-ocfl) to store the path and change the conversion code from OCFL resource headers to Fedora resource headers to set this path. Then change the ResourceFactory to transfer the path from the resource headers to the FedoraResource. Then we could still use the ManagedPropertiesService to set the triple using the path.

This makes some assumptions, which if we ever make the path configurable would make it invalid. 🤷 

To test, build and run this code. Make an object in Fedora, view the metadata for that object and find a new `http://fedora.info/definitions/v4/repository#ocflPath` triple with the full path on disk.

For Archival Groups and their parts, they all point to the path of the Archival Group. 

# Interested parties

### DIFF
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/DisplayOcflPath.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/DisplayOcflPath.java
@@ -1,0 +1,40 @@
+/*
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree.
+ */
+package org.fcrepo.config;
+
+/**
+ * Display OCFL path mode enum
+ * @author whikloj
+ */
+public enum DisplayOcflPath {
+    NONE("none"),
+    RELATIVE("relative"),
+    ABSOLUTE("absolute");
+
+    private final String value;
+
+    DisplayOcflPath(final String value) {
+        this.value = value;
+    }
+
+    public String getValue() {
+        return value;
+    }
+
+    public static DisplayOcflPath fromString(final String value) {
+        for (final var mode : values()) {
+            if (mode.value.equalsIgnoreCase(value)) {
+                return mode;
+            }
+        }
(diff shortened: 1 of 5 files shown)


bbpennel:
One issue with this approach is that if you are using S3 as the storage backend, then the absolute paths returned don't match where the files are stored. For example, I tested with an S3 bucket running fcrepo locally with jetty:run, and i get back fedora:ocflPath properties with values like
`/git/fcrepo/fcrepo-webapp/target/fcrepo-home/data/ocfl-root/dca/c48/694/dcac48694357fce347ad3455ddf9fbc5d93127179a61dc2c0302305cdf32560f`

If I have to set to relative paths, it might be fine since I get back `dca/c48/694/dcac48694357fce347ad3455ddf9fbc5d93127179a61dc2c0302305cdf32560f` which exists within the S3 ocfl root.

We could probably still make it work by making the `ManagedPropertiesServiceImpl` aware of what the storage backend was. 

But honestly, including a physical path property from the storage layer via the headers seems like a more appropriate source for this info, even if a bit of a pain. We would have to be careful not to persist the property out into the header files too. 

I also found myself wondering if fedora:ocflPath or fedora:physicalPath would make more sense. Obviously ocfl is the only storage backend we have, and the the paths are relative to the ocfl root, so maybe there's no reason to abstract it. But on the other hand we've mostly kept the storage implementation contained to lower levels so far, aside from the existence of archival groups and configuration options. Mostly just thinking out loud though.

whikloj:
Closing this in favour of getting the real path from OCFL. #2090

PR #2080, opened 2024-05-02
URL encode user agent if there was a space in the user name

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3938

URL encodes the user name if we use it to build a user agent and it has a space, or URL encode the user agent if it has a space.

# How should this be tested?

Prior to this PR you need to get a user name with a space. I haven't tested this but I am guessing if you configure your Tomcat with a user "john smith" and create an object in Fedora using that login you should get a stack trace.

After you will not.

# Interested parties

### DIFF
--- a/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/ResourceOperationEventBuilder.java
+++ b/fcrepo-kernel-impl/src/main/java/org/fcrepo/kernel/impl/observer/ResourceOperationEventBuilder.java
@@ -6,17 +6,20 @@
 
 package org.fcrepo.kernel.impl.observer;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+import java.net.URI;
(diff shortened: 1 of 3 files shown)


Surfrdan:
Tested by creating a tomcat-users.xml user with a space character (not encoded) in the XML
``` 
<user name="test user3" password="testpass" roles="fedoraUser" />
``` 
Appears to work and authenticate using the browser and curl
```
curl -u "test user3:testpass" http://localhost:8080/fcrepo/rest
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix fedora: <http://fedora.info/definitions/v4/repository#> .
@prefix ldp: <http://www.w3.org/ns/ldp#> .

<http://localhost:8080/fcrepo/rest/>
...
```

Surfrdan (approved):

PR #2084, opened 2024-06-20
Remove unnecessary apostrophe from messages.

demiankatz:

This fixes a typo -- the contraction "it's" should be changed to the possessive "its."

# How should this be tested?

As long as the changed wording doesn't break any test assertions, I don't think this should be an impactful change.


whikloj (approved):
👍

PR #2087, opened 2024-07-19
FCREPO-3950 - Prevent containment of fcr:acl on reindex

bbpennel:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3950

Adds tests to replicate the bug, and then prevents adding containment relations (and by extension, membership relations) for fcr:acl resources. This applies to basic, direct and indirect containers.

# How should this be tested?

See the ticket for instructions on replicating the issue. Effectively, create a container, add an fcr:acl to it, and then rebuild the index. The new tests should replicate the issue too.

# Interested parties
@fcrepo/committers


bbpennel:
Even though childShouldNotBeContained is only checking one thing internally, I left it there anyways in case we end up encountering more special resource types that need to be skipped. I can remove it if that's preferable though.

Surfrdan (approved):
Tested with David Novak's data and ACLs and can confirm that neither the membership or containment is indexed on reindexing of the object now. The UI also retains the same view of the container between reindexes unlike before.

PR #2090, opened 2024-08-21
Return OCFL path in response

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3935

Alternative implementation to #2079 

Relies on: https://github.com/fcrepo/fcrepo-storage-ocfl/pull/58

The OCFL FileDetails object has a [getStorageRelativePath()](https://github.com/OCFL/ocfl-java/blob/main/ocfl-java-api/src/main/java/io/ocfl/api/model/FileDetails.java#L59-L66) function which returns the full path of the OCFL hidden Fedora header file when we read the headers. The related `fcrepo-storage-ocfl` PR passes this path into the storage layer ResourceHeaders, which are then transferred to the kernel level ResourceHeaders and up to the HTTP layer.

This does not show the storage specific prefix. At this level we only have an `objectSession` and (I think) we would need the entire repository (`io.ocfl.api.OcflRepository`) to get the storage (`io.ocfl.core.storage.OcflStorage`) to get the [`objectRootPath()`](https://github.com/OCFL/ocfl-java/blob/main/ocfl-java-core/src/main/java/io/ocfl/core/storage/OcflStorage.java#L182-L188).

~This method also does not have a way to stop this new information from being loaded. Because of the existing functionality, it requires a second touch of the file which (I imagine) will have some performance impacts.~
Edit: I have re-added the toggle property and have made changes in the `fcrepo-storage-ocfl` PR that hopefully reduces duplication and (by extension) performance impacts.

I have also stripped the path elements (the last 4 parts of the path) to make
`a3e/b5e/29e/a3eb5e29e71a3c6c8eec9ccb27622a454ec8b8f93393074d1e81223e18eca3f8/v1/content/.fcrepo/fcr-root.json`
to
`a3e/b5e/29e/a3eb5e29e71a3c6c8eec9ccb27622a454ec8b8f93393074d1e81223e18eca3f8`

# How should this be tested?

Build PRs, run Fedora, create an object, add the `fcrepo.ocfl.show_path=true` property and see the new `Fedora-Ocfl-Path` header.

For a container
```
> curl -I -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/test-container-1
HTTP/1.1 200 OK
Date: Tue, 10 Sep 2024 14:29:22 GMT
...
Last-Modified: Wed, 21 Aug 2024 17:11:11 GMT

Fedora-Ocfl-Path: 1ae/96b/c16/1ae96bc1695d8d30ebba20f010da0b9421afbd57cea05d2baa7b2d1f0717d35d

Link: <http://localhost:8080/rest/test-container-1>; rel="timegate"
...
Content-Type: text/turtle;charset=utf-8
Content-Length: 0
Server: Jetty(9.4.44.v20210927)
```

For a binary
```
> curl -I -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/test-container-1/a-binary
HTTP/1.1 200 OK
Date: Tue, 10 Sep 2024 14:33:50 GMT
...
Content-Type: image/jpeg
Accept-Ranges: bytes
Content-Disposition: attachment; size=2395723; creation-date="Tue, 10 Sep 2024 14:33:27 GMT"; modification-date="Tue, 10 Sep 2024 14:33:27 GMT"

Fedora-Ocfl-Path: a3e/b5e/29e/a3eb5e29e71a3c6c8eec9ccb27622a454ec8b8f93393074d1e81223e18eca3f8

Link: <http://localhost:8080/rest/test-container-1/a-binary/fcr:metadata>; rel="describedby"
...
Content-Length: 2395723
Server: Jetty(9.4.44.v20210927)
```

For a binary description
```
> curl -I -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/test-container-1/a-binary/fcr:metadata
HTTP/1.1 200 OK
Date: Tue, 10 Sep 2024 14:34:34 GMT
...
Last-Modified: Tue, 10 Sep 2024 14:33:27 GMT

Fedora-Ocfl-Path: a3e/b5e/29e/a3eb5e29e71a3c6c8eec9ccb27622a454ec8b8f93393074d1e81223e18eca3f8

Link: <http://localhost:8080/rest/test-container-1/a-binary>; rel="describes"
...
Content-Type: text/turtle;charset=utf-8
Content-Length: 0
Server: Jetty(9.4.44.v20210927)
```

For an Archival group
```
> curl -I -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/my-archival-group
HTTP/1.1 200 OK
Date: Tue, 10 Sep 2024 14:31:04 GMT
...
Last-Modified: Thu, 05 Sep 2024 16:29:17 GMT

Fedora-Ocfl-Path: 502/bb2/403/502bb2403dfb33624cf5b3cbf8e200fff81237fad89f0734b331ea35bd6a08c5

Link: <http://localhost:8080/rest/my-archival-group>; rel="timegate"
...
Content-Type: text/turtle;charset=utf-8
Content-Length: 0
Server: Jetty(9.4.44.v20210927)
```

and for an archival group member
```
> curl -I -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/my-archival-group/murphy-picture      
HTTP/1.1 200 OK
Date: Tue, 10 Sep 2024 14:35:51 GMT
...
Content-Type: image/jpeg
Accept-Ranges: bytes
Content-Disposition: attachment; size=2395723; creation-date="Thu, 5 Sep 2024 18:37:40 GMT"; modification-date="Thu, 5 Sep 2024 18:37:40 GMT"

Fedora-Ocfl-Path: 502/bb2/403/502bb2403dfb33624cf5b3cbf8e200fff81237fad89f0734b331ea35bd6a08c5

Link: <http://localhost:8080/rest/my-archival-group/murphy-picture/fcr:metadata>; rel="describedby"
...
Content-Length: 2395723
Server: Jetty(9.4.44.v20210927)
```

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -165,6 +165,8 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
 
     public static final String HTTP_HEADER_OVERWRITE_TOMBSTONE = "Overwrite-Tombstone";
 
+    private static final String HTTP_OCFL_PATH = "Fedora-Ocfl-Path";
+
     private static final String FCR_PREFIX = "fcr:";
     private static final Set<String> ALLOWED_FCR_PARTS = Set.of(FCR_METADATA, FCR_ACL);
 
@@ -679,7 +681,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
         if (ocflPropsConfig.isShowPath()) {
             final var contentPath = resource.getStorageRelativePath();
             if (contentPath != null) {
-                servletResponse.addHeader("Fedora-Ocfl-Path", contentPath.toString());
+                servletResponse.addHeader(HTTP_OCFL_PATH, contentPath.toString());
             }
         }
 
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -88,6 +88,7 @@ import javax.ws.rs.core.Request;
 import javax.ws.rs.core.Response;
 
 import org.fcrepo.config.DigestAlgorithm;
+import org.fcrepo.config.OcflPropsConfig;
 import org.fcrepo.http.api.services.EtagService;
 import org.fcrepo.http.api.services.HttpRdfService;
 import org.fcrepo.http.commons.api.rdf.HttpTripleUtil;
@@ -206,6 +207,9 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
 
     @Inject
     protected ResourceTripleService resourceTripleService;
+
+    @Inject
+    protected OcflPropsConfig ocflPropsConfig;
 
     protected abstract String externalPath();
 
@@ -670,6 +674,13 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
             }
             servletResponse.addHeader("Accept-Ranges", "bytes");
             servletResponse.addHeader(CONTENT_DISPOSITION, dispositionBuilder.build().toString());
+        }
+        // Add ocflPath header if desired
+        if (ocflPropsConfig.isShowPath()) {
+            final var contentPath = resource.getStorageRelativePath();
+            if (contentPath != null) {
+                servletResponse.addHeader("Fedora-Ocfl-Path", contentPath.toString());
+            }
         }
 
         addLinkAndOptionsHttpHeaders(resource);
--- a/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
+++ b/fcrepo-configs/src/main/java/org/fcrepo/config/OcflPropsConfig.java
@@ -135,6 +135,9 @@ public class OcflPropsConfig extends BasePropsConfig {
     @Value("${fcrepo.ocfl.s3.enable.checksum:true}")
     private boolean s3EnableChecksum;
 
+    @Value("${fcrepo.ocfl.show_path:false}")
+    private boolean showPath;
+
     private DigestAlgorithm FCREPO_DIGEST_ALGORITHM;
 
     /**
@@ -187,6 +190,7 @@ public class OcflPropsConfig extends BasePropsConfig {
                             .collect(Collectors.joining(", "))));
         }
         LOGGER.info("Fedora OCFL digest algorithm: {}", FCREPO_DIGEST_ALGORITHM.getAlgorithm());
+        LOGGER.info("Fedora OCFL show path: {}", showPath);
     }
 
     /**
@@ -545,5 +549,12 @@ public class OcflPropsConfig extends BasePropsConfig {
      */
     public int getS3MaxConcurrency() {
         return s3MaxConcurrency;
+    }
+
+    /**
+     * @return true if the path of the OCFL object should be shown
+     */
+    public boolean isShowPath() {
+        return showPath;
     }
 }
--- a/fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
+++ b/fcrepo-http-api/src/test/java/org/fcrepo/http/api/FedoraLdpTest.java
@@ -15,6 +15,7 @@ import org.apache.jena.rdf.model.RDFNode;
 
 import org.fcrepo.common.db.DbTransactionExecutor;
 import org.fcrepo.config.FedoraPropsConfig;
+import org.fcrepo.config.OcflPropsConfig;
 import org.fcrepo.http.api.services.EtagService;
 import org.fcrepo.http.api.services.HttpRdfService;
 import org.fcrepo.http.commons.api.rdf.HttpIdentifierConverter;
@@ -70,6 +71,7 @@ import javax.ws.rs.core.SecurityContext;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
+import java.nio.file.Path;
 import java.security.Principal;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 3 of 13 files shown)


bbpennel:
After looking at the fcrepo-storage-ocfl PR and this one, some thoughts:

1. it seems like the main advantage of this approach is that we don't have to duplicate the logic of producing a hash and chopping it up into a path. But on the other hand, that logic is unlikely to change since it would be costly to rearrange the OCFL root. So the chance of drift occurring between the two implementations is very low. So it's mainly encapsulating storage concerns at a more appropriate part of the stack.
2. It doesn't seem like there would be major performance implications with the implementation in this PR, just the extra step of `ocflFileDetails.getStorageRelativePath()` and the overhead of passing around the header in all cases, even when the user doesn't want to see it.
3. As you note, we would only get a relative path from this approach, so it's essentially on par with the other PR in terms of functionality.
4. I assume we'd still need to do a bit of extra processing in this PR to remove the `.fcrepo/fcr-root.json` part of the path, but putting that off until we decide if this approach is worth pursuing seems reasonable.
5. My understanding is that this PR would not result in the `storageRelativePath` header ever getting written out to the header file, is that correct?

So while part of me dislikes duplicating the path calculation so far up the stack, the fact that this approach doesn't get us the option to create correct absolute paths makes to seem less worthwhile to me. I think either approach would be fine, so if you have a preference at this point I'd support that. But perhaps we should go with the first PR and change it to only support relative paths.

whikloj:
Hey @bbpennel 
1. Yes, we get the actual path which is nice if we ever change it. But I agree that this seems unlikely right now.
2. I am less concerned once I pulled some of the pieces apart a little more. But I'm also not super confident there wouldn't be an impact.
3. Agreed
4. Or we don't remove the `.fcrepo/fcr-root.json` either. This gives you the version path of files too (I.e. .../content/v2/...) which might be a good thing 🤷 
5. Yeah, it is only added to the readHeaders and loadRootResource dynamically on its way up the chain. I also did not add it for staged headers,  so I believe (but didn't test) that you don't get the path for staged objects.

I have no preference and I agree with your concerns about the path calculations. We could also ask @pwinckles if there is a better way in ocfl-java.

The main benefit of the first PR is if you don't want the path it isn't even calculated. Where this PR passes it up to the resource for all objects. Not sure of that's a problem or not.

pwinckles:
@bbpennel @whikloj 

What if modify ocfl-java to expose [RepositoryConfig](https://github.com/OCFL/ocfl-java/blob/main/ocfl-java-core/src/main/java/io/ocfl/core/storage/RepositoryConfig.java) so Fedora can get a handle to the [OcflStorageLayoutExtension](https://github.com/OCFL/ocfl-java/blob/main/ocfl-java-core/src/main/java/io/ocfl/core/extension/storage/layout/OcflStorageLayoutExtension.java)? This would allow you to use `OcflStorageLayoutExtension.mapObjectId()` to get the path to the object. It would be relative the storage root, and not absolute.

This solution is more in line with the first PR, but avoids making assumptions about what the storage layout actually is (and re-implementing the algorithm).

Let me know if this is something you want to pursue. It's unfortunately a little tricky than I'd like to work into ocfl-java based on how I divided up the modules.

whikloj:
@pwinckles would your suggested plan this be any better than what I attempted with these PRs? I don't want you to have to perform a bunch of changes to ocfl-java for the same end result. 

Also, do you see a problem with how I hooked into ocfl-java to get the relative path?

pwinckles:
@whikloj I don't like the approach in the first PR because it makes assumptions about the layout that are not guaranteed to be true. I'm more okay with the version in this PR. However, I am a little scared by all of the changes in `DefaultOclfObjectSession`, which is a delicate class, and have not given them a thorough review.

> Or we don't remove the .fcrepo/fcr-root.json either. This gives you the version path of files too (I.e. .../content/v2/...) which might be a good thing

What is the goal here? To provide a binary path or a path to the containing ocfl object? It makes me uneasy when people start poking around in ocfl objects, because, while they are understandable by inspecting the filesystem, it has been my experience that people consistently have faulty mental models of how the object is laid out.

fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
```diff
@@ -671,6 +675,13 @@ protected void addResourceHttpHeaders(final FedoraResource resource,
             servletResponse.addHeader("Accept-Ranges", "bytes");
             servletResponse.addHeader(CONTENT_DISPOSITION, dispositionBuilder.build().toString());
         }
+        // Add ocflPath header if desired
+        if (ocflPropsConfig.isShowPath()) {
+            final var contentPath = resource.getStorageRelativePath();
+            if (contentPath != null) {
+                servletResponse.addHeader("Fedora-Ocfl-Path", contentPath.toString());
```
bbpennel, line 682: We're pretty inconsistent about it in this file, but it seems like we should have a constant for the header name.

bbpennel (approved):

bbpennel:
This seems to be working for me in basic testing with local storage and s3

pwinckles:
@whikloj @bbpennel Are you sure you want to return paths like:

```
        fedora:ocflPath          "97c/313/b71/97c313b71a3398c06831fbd81bd6d31aff914e40671fec95a7a78c1fcc17d0bf/v1/content/.fcrepo/fcr-root~fcr-desc.json" ;
```

Why not just to the object root?

```
        fedora:ocflPath          "97c/313/b71/97c313b71a3398c06831fbd81bd6d31aff914e40671fec95a7a78c1fcc17d0bf" ;
```

Pointing inside the object like that makes me uneasy, especially since the path you're giving to internal Fedora files and not to user content.

whikloj:
Hi @pwinckles, you are correct and so we are displaying the path to the object root only. I make the change in `FedoraResourceImpl` when we set the storage path. 
https://github.com/fcrepo/fcrepo/pull/2090/files#diff-efbd5f52986f25ae38372d0ed3e71eee6d63dfae02b1cb0d497b36aa80f5c15aR496-R502

pwinckles:
@whikloj Ah, I was going by what was in the PR description.

whikloj:
🤦 good point, I'll edit it.

PR #2091, opened 2024-09-06
Fix too long ranges and end start ranges

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/jira/software/c/projects/FCREPO/issues/FCREPO-3955

Fixes Fedora's handling of Range requests that are too long and if they don't specify a starting position.

# How should this be tested?

Create a binary and check it's content length.
Before this PR
Make a request for the binary with `Range: 0-<content length + 1>` get 416
Make a request for the binary with `Range: -5` get a response with `Content-Range: 0-5`
After this PR
Make a request for the binary with `Range: 0-<content length + 1>` get 206 with the full object 
Make a request for the binary with `Range: -5` get a response with `Content-Range: <content length - 5>-<content length - 1>`

# Interested parties

### DIFF
--- a/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/Range.java
+++ b/fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/Range.java
@@ -15,13 +15,11 @@ import java.util.regex.Pattern;
  * Range header parsing logic
  *
  * @author awoods
+ * @author whikloj
  */
 public class Range {
 
     private final long start;
-
-    // Did the Range actually specify a start?
-    private final boolean no_start;
 
     private final long end;
 
@@ -32,7 +30,7 @@ public class Range {
      * Unbounded Range
      */
     private Range() {
-        this(0, -1);
+        this(-1, -1);
     }
 
     /**
@@ -47,21 +45,10 @@ public class Range {
      * Left and right bounded range
      * @param start the start
      * @param end the end
-     * @param no_start whether there was a starting byte
-     */
-    private Range(final long start, final long end, final boolean no_start) {
-        this.start = start;
-        this.end = end;
-        this.no_start = no_start;
-    }
-
-    /**
-     * Left and right bounded range
-     * @param start the start
-     * @param end the end
      */
     private Range(final long start, final long end) {
-        this(start, end, false);
+        this.start = start;
+        this.end = end;
     }
 
     /**
@@ -69,7 +56,7 @@ public class Range {
      * @return true if the range imposes limits
      */
     public boolean hasRange() {
-        return !(start == 0 && end == -1);
+        return !(start == -1 && end == -1);
     }
 
     /**
@@ -79,7 +66,7 @@ public class Range {
     public long size() {
         if (end == -1) {
             return -1;
-        } else if (start == 0 && no_start) {
+        } else if (start == -1) {
             return end;
         }
         return end - start + 1;
@@ -90,7 +77,7 @@ public class Range {
      * @return start of the range, or -1 if no start was specified
      */
     public long start() {
-        return no_start ? -1 : start;
+        return start;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


bbpennel:
This is not really working for me in `mvn jetty:run`. If I request `Range: bytes=0-<contentsize+1>` I get back a 500 and a `Error 500 insufficient content written` message from jetty. Previously it gave me a 416, which was slightly nicer than a 500. I recognize that this is a jetty problem caused by the response size not matching the content length, but hopefully we can address it.

I think we need to make the content-length header match the functional size of the returned response. The content length gets set here using the `range` variable, but it's the range as was submitted by the user, not after the adjustments being made above:
https://github.com/fcrepo/fcrepo/blob/c2cb63fcc92e07a78f05fc1c86efca14aacbb33c/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java#L392
So my guess is the `range` needs to be readjusted before setting the content-length, and maybe also in the line before for rangeInputStream.

whikloj:
I was not getting the same errors you were seeing when I made requests but I was getting a warning from curl that the connection was closed with bytes still left to read. With these latest changes I don't get that error.

Before:
```
> curl -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/test-container-1/a-binary -H"Range: bytes=0-2395723" -i -o output.txt
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 2339k  100 2339k    0     0  14.3M      0 --:--:-- --:--:-- --:--:-- 14.3M
curl: (18) transfer closed with 1 bytes remaining to read
```

After
```
> curl -ufedoraAdmin:fedoraAdmin http://localhost:8080/rest/test-container-1/a-binary -H"Range: bytes=0-2395723" -i -o output.txt
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 2339k  100 2339k    0     0  10.5M      0 --:--:-- --:--:-- --:--:-- 10.5M
```

Also of note ocfl-java supports range requests as of v2.2.1 (https://github.com/OCFL/ocfl-java/issues/118)

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/Range.java
```diff
@@ -135,6 +119,111 @@ public static Range convert(final String source) {
...
+         * @param end the end of the range
+         * @param length the length of the content
+         */
+        protected RangeOfLength(final long start, final long end, final long length) {
+            this.start = start;
+            this.end = end;
+            // If a valid byte-range-set includes at least one byte-range-spec with a first-byte-pos that is less than
+            // the current length of the representation, or at least one suffix-byte-range-spec with a non-zero
+            // suffix-length, then the byte-range-set is satisfiable. Otherwise, the byte-range-set is unsatisfiable.
+            this.satisfiable = start < length && (end() == -1 || end > start) && (start != -1 && end != -1);
```
bbpennel, line 184: I think `end > start` should be `end >= start`. At the moment if I request `-H"Range: bytes=0-0"` I get a 416, but it should give me the first byte (even the comment down at line 223 agrees). We should probably add a case to RangeTest for this.

bbpennel (changes requested):

fcrepo-http-commons/src/main/java/org/fcrepo/http/commons/domain/Range.java
whikloj: I made that comment, so it could be wrong. I will make this change because I feel like that comment and you are correct.

bbpennel: I wasn't really sure from reviewing the spec what the expected outcome was, but I tried it against github (which supports range headers) and `0-0` does get me the first byte there. Plus there wouldn't be a way to get a single byte otherwise, since the ranges are inclusive so`0-1` returns 2 bytes.

bbpennel (approved):

PR #2092, opened 2024-09-18
FCREPO-3952 - Cleanup transactions around application restarts

bbpennel:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3952

When shutting down fedora, the application will attempt to rollback all active or expired-but-not-rolledback transactions, along with associated OCFL sessions. This should result in no leftover staged files or residual transaction database entries.

If the application shuts down in a non-graceful manner, the transaction tables also get truncated during startup and any leftover files in the staging directory are cleaned up.

ReferenceIndexService `reset` method now also clears the transaction table, which I'm guessing was overlooked before.

This PR also cleans up session information in the OCFL index when an OCFL session is rolled back (previously only committed OCFL sessions would roll back)

# How should this be tested?

This can be tested using `mvn jetty:run` and the h2 client which can be downloaded here:
http://h2database.com/html/download.html
You cannot connect to h2 at the same time that fedora is running. Use h2.sh found in the extracted h2 client directory, and the connection URL is `jdbc:h2:/path/to/fcrepo/fcrepo-webapp/target/fcrepo-home/data/fcrepo-h2;FILE_LOCK=SOCKET` with no username or password.

Or you can use a different database that is easier to connect to.

## Replicate issue
Prior to checking out this branch, to replicate the issue do the following steps:
```
curl -i -X POST "http://localhost:8080/rest/fcr:tx" -ufedoraAdmin:fedoraAdmin

# update the tx id based on the previous response
curl -H"Atomic-ID: http://localhost:8080/rest/fcr:tx/0a3ef984-4e8a-446e-9cdf-1a8f8b69d464" -X POST "http://localhost:8080/rest" -H"Slug: newone" -ufedoraAdmin:fedoraAdmin
```

Then shut down fedora and connect to h2. You should see the entries in the `CONTAINMENT_TRANSACTIONS` table. They will still be there even if you start up fedora and shut it down again. There will also be files in `/path/to/fcrepo/fcrepo-webapp/target/fcrepo-home/data/staging`.

## Verify fix with graceful shutdown
Then rebuild with this branch and start up fedora. If you checked h2 now `CONTAINMENT_TRANSACTIONS` would be empty. Now repeat the curl commands above and then shut down fedora. The `CONTAINMENT_TRANSACTIONS` table should be empty, along with the `staging` folder.

## Verify fix with hard shutdown
Now start fedora up again and perform the same curl calls. Then find the process id of fedora and kill it with `kill -9 <processid>`. Connect to h2 and you'll see entries in `CONTAINMENT_TRANSACTIONS`. Now start fedora up again, and if using h2 then shut down fedora again, connect to h2, and verify that `CONTAINMENT_TRANSACTIONS` is now empty, along with the staging folder.

## Testing with other types of content

You might also want to verify that references or membership get cleaned up from their respective transaction tables. I did test the membership table locally with an IndirectContainer based off [pcdm in action](https://github.com/awead/ldp-pcdm), which unfortunately needs some updating to work with fedora 6). 

It would also probably be prudent to verify that no committed database entries get cleaned up, by creating a few resources in fedora outside of a transaction prior to doing the testing steps above. (I also verified this)

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/ContainmentIndex.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/ContainmentIndex.java
@@ -106,6 +106,11 @@ public interface ContainmentIndex {
     void rollbackTransaction(final Transaction tx);
 
     /**
+     * Clear all transactions in the containment index.
+     */
+    void clearAllTransactions();
+
+    /**
      * Check if the resourceID exists in the containment index. Which should mean it exists.
      *
      * @param tx The transaction, or null if no transaction
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/MembershipService.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/services/MembershipService.java
@@ -80,6 +80,11 @@ public interface MembershipService {
     void rollbackTransaction(final Transaction transaction);
 
     /**
+     * Clear all transactions in the membership index.
+     */
+    void clearAllTransactions();
+
+    /**
      * Truncates the membership index. This should only be called when rebuilding the index.
      */
     void reset();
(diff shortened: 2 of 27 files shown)


whikloj (approved):
Works as expected 👍

PR #2099, opened 2024-10-23
FCREPO-3957 - Make range requests at storage/ocfl level

bbpennel:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3957

* Update to ocfl-java 2.2.1. Remove local aws dependencies and rely on versions on ocfl-aws instead to prevent unnecessary conflicts
* Use ocfl-java range methods when getting ranges from internally stored binaries. 
* Add methods for getting ranges to binaries. 
* Remove RangeRequestInputStream since it is no longer used. 
* Add more tests for ranges

# How should this be tested?

You will need to check out and build this fcrepo-storage-ocfl branch locally:
https://github.com/fcrepo/fcrepo-storage-ocfl/pull/61

And then build this branch locally and run `mvn jetty:run`.

All range requests should operate the same as they did before this PR. For example, if you have a binary named `http://localhost:8080/rest/e64153d9-02f8-4ff1-b482-dbf06d1b1fd1` then you can request a range from it with:

```
curl http://localhost:8080/rest/e64153d9-02f8-4ff1-b482-dbf06d1b1fd1 -H"Range: bytes=0-1081" -v -o /dev/null  -ufedoraAdmin:fedoraAdmin
```

The headers should include `Content-Range: bytes 0-10810/658424` or whatever range you requested, where the last part is the total file size.

# Interested parties
@fcrepo/committers


Surfrdan (approved):
Looks good to me. Initial GH Buildfailures were cured by purging cached build states and it now passes fine

PR #2101, opened 2024-11-14
Allow encoded forward slashes

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3933

Change configuration to allow encoded forward slashes (`%2F`) in the path since Shiro 1.13

# How should this be tested?

There is a test in fcrepo-webapp and this does seem to appear only with Tomcat.

Build the WAR before this PR and deploy it to Tomcat.
Perform a PUT with an encoded slash
```
curl -ufedoraAdmin:fedoraAdmin  -XPUT -d" @prefix dc: <http://purl.org/dc/elements/1.1/> . <> dc:title 'Some object' ." -H"Content-type: text/turtle" "http://localhost:8080/fcrepo/rest/admin_set%2Fdefault" -i
```

See a 400 Bad Request

Build with this PR, and try again. Get a 201 Created or 204 No Content (depending on if it exists already).

# Interested parties

### DIFF
--- a/fcrepo-webapp/src/main/java/org/fcrepo/webapp/AuthConfig.java
+++ b/fcrepo-webapp/src/main/java/org/fcrepo/webapp/AuthConfig.java
@@ -186,6 +186,7 @@ public class AuthConfig {
         final var filter = new InvalidRequestFilter();
         filter.setBlockNonAscii(false);
         filter.setBlockBackslash(false);
+        filter.setBlockEncodedForwardSlash(false);
         filter.setBlockSemicolon(false);
         return filter;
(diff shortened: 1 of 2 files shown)


whikloj:
It appears that the snapshot of fcrepo-storage-ocfl 6.4.0-SNAPSHOT has not been deployed lately. I can't seem to deploy the snapshot so I am not sure what we should do here.

whikloj:
Actually it does appear that fcrepo-storage-ocfl was deployed recently, maybe I have a mistake.

PR #2114, opened 2025-02-11
FCREPO-3985 - Update fcrepo-persistence-ocfl to junit 5

bbpennel:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-3985

Updates fcrepo-persistence-ocfl to use junit 5 for testing

# Interested parties
@fcrepo/committers

### DIFF
--- a/fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndexTest.java
+++ b/fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndexTest.java
@@ -5,18 +5,6 @@
  */
 package org.fcrepo.persistence.ocfl.impl;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotEquals;
-import static org.junit.Assert.assertThrows;
-import static org.junit.Assert.fail;
-import static org.mockito.Mockito.mock;
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.when;
-import static org.springframework.test.util.ReflectionTestUtils.setField;
-
-import java.util.UUID;
-
 import org.fcrepo.config.FlywayFactory;
 import org.fcrepo.config.OcflPropsConfig;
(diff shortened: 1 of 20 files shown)


fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndexTest.java
```diff
@@ -209,13 +189,11 @@ public void identifierTooLong() throws InvalidResourceIdentifierException {
         for (cnt = 0; cnt < 149; cnt++) {
             root = root + "/" + String.valueOf(cnt);
         }
-        try {
-            final FedoraId tempid = FedoraId.create(root);
+        final var finalRoot = root;
```
whikloj, line 192: Wondering if this was necessary?

whikloj (changes requested):
Just the one question, but we should also get a common set of import ordering because I think I keep putting statics on the top and you keep putting them on the bottom 🤣

fcrepo-persistence-ocfl/src/test/java/org/fcrepo/persistence/ocfl/impl/DbFedoraToOcflObjectIndexTest.java
bbpennel: it was needed because `root` needs to be modifiable in the block above, but the value has to be final in order for it to be accessible in the closure of the assertThrows

bbpennel:
I don't remember actually setting a particular ordering for my imports, just disabling "*" imports. I see a mix of the two orders elsewhere in the project, and the old eclipse config we provide doesn't really specify as far as I can tell https://github.com/fcrepo/fcrepo/blob/main/src/site/eclipse/fedora.importorder

I'm good with going with whatever the default is, which it seems like might be static first?

bbpennel:
@whikloj I've updated my settings and reordered static imports to the top in this PR

whikloj:
@bbpennel sorry I didn't mean you had to do that, we should just figure out a common ordering and then make it easy to implement in all the various IDEs... but thanks.

whikloj (approved):
👍

PR #2187, opened 2025-03-31
Reset session timeout between tests

whikloj:
**JIRA Ticket**: 

Some change in Github actions has a test that is setting the session timeout to 5 seconds causing errors in other tests. This might be due to change test ordering, concurrency or perhaps a failure in that particular test.

Regardless, this change is to set it to the default of 180000ms or 3 minutes between each test.

# How should this be tested?

Tests should pass on Github

# Interested parties

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
@@ -584,13 +584,7 @@ public class FedoraLdp extends ContentExposingResource {
             return createUpdateResponse(getFedoraResource(transaction, fedoraId), created.get());
         } finally {
             transaction.releaseResourceLocksIfShortLived();
-            if (requestBodyStream != null) {
-                try {
-                    requestBodyStream.close();
-                } catch (final IOException e) {
-                    LOGGER.error("Error closing request body stream", e);
-                }
-            }
+            IOUtils.closeQuietly(requestBodyStream);
         }
     }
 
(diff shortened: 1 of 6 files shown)


bbpennel (approved):

fcrepo-http-api/src/main/java/org/fcrepo/http/api/FedoraLdp.java
```diff
@@ -581,10 +581,16 @@ public Response createOrReplaceObjectRdf(
                 }
 
             LOGGER.debug("Finished creating resource with path: {}", externalPath());
-
             return createUpdateResponse(getFedoraResource(transaction, fedoraId), created.get());
         } finally {
             transaction.releaseResourceLocksIfShortLived();
+            if (requestBodyStream != null) {
```
bbpennel, line 587: We could use IOUtils.closeQuietly to close these streams since it'd do the null check and doesn't care if there's an error (I don't think we care), which would also avoid the codecov complaints about coverage.

fcrepo-http-api/src/test/java/org/fcrepo/integration/http/api/FedoraSearchIT.java
```diff
@@ -216,7 +216,7 @@ public void testModifiedGreaterAndLessThan() throws Exception {
         assertEquals(1, resources.size());
         final var externalFedoraId = resources.get(0);
         final var lessThanNow = MODIFIED + encode("<") + earlier;
-        final var greaterThanNow = MODIFIED + encode(">") + now;
+        final var greaterThanNow = MODIFIED + encode(">") + earlier;
```
bbpennel, line 219: I'm curious what this change was for? Also, is `now` not used anymore? It may be related to the failure `FedoraSearchIT.testModifiedGreaterAndLessThan:230 expected no results ==> expected: <0> but was: <1>`

bbpennel (approved):

PR #2191, opened 2025-04-03
Allow test to pass if timing causes memento to not list any membership

bbpennel:
Updates test to correctly pass if memento does not have any membership, which will happen if the proxy is modified the second after the memento is created.

# How should this be tested?
Tests should pass


whikloj (approved):
👍

PR #2192, opened 2025-04-04
Fcrepo kernel api tests

whikloj:
**JIRA Ticket**: 

Issue: #2171 

Deletes a bunch of no longer used Exceptions and their exception mappers and tests (if applicable).
Adds more tests for the some of the fcrepo-kernel-api classes

# How should this be tested?

Tests pass, code coverage goes ⬆️ 

# Interested parties


fcrepo-kernel-api/src/test/java/org/fcrepo/kernel/api/utils/RelaxedPropertiesHelperTest.java
```diff
@@ -0,0 +1,295 @@
...
+                ),
+                Triple.create(
+                        subject,
+                        HAS_MIME_TYPE.asNode(),
+                        NodeFactory.createLiteral("application/n-triples")
+                ),
+                Triple.create(
+                        subject,
+                        HAS_MIME_TYPE.asNode(),
+                        NodeFactory.createVariable("some-variable")
```
bbpennel, line 280: I'm honestly not totally sure what this means. Does it imply that if someone used a sparql variable to supply "not-a-mimetype" as the mimetype, that it would be accepted by fedora?

fcrepo-kernel-api/src/test/java/org/fcrepo/kernel/api/utils/WrappingStreamTest.java
```diff
@@ -0,0 +1,408 @@
...
+        assertEquals(objectB, any2.get().getObject());
+        assertFalse(any2.get().getObject().equals(objectA) || any2.get().getObject().equals(objectC));
+    }
+
+    @Test
+    public void testFindFirst() {
+        final var first = stream.findFirst();
+        assertTrue(first.isPresent());
+        assertEquals(objectA, first.get().getObject());
+        assertFalse(first.get().getObject().equals(objectB) || first.get().getObject().equals(objectB));
```
bbpennel, line 168: Is this actually supposed to be checking if its objectB twice?

fcrepo-kernel-api/src/test/java/org/fcrepo/kernel/api/utils/WrappingStreamTest.java
```diff
@@ -0,0 +1,408 @@
...
+        stream = generateTextStream();
+        final var first1 = stream.filter(s -> !s.getObject().equals(objectA)).findFirst();
```
bbpennel, line 170: I'd suggest making this a separate test method for modularity and readability

fcrepo-kernel-api/src/test/java/org/fcrepo/kernel/api/utils/RelaxedPropertiesHelperTest.java
whikloj: This is used to parse both incoming RDF as well as Sparql. So you can have a triple with a HAS_MIME_TYPE predicate and a variable. We can restrict it, I'm not sure if there is a valid use case I just didn't want to go overboard.

bbpennel: well either way we don't need to restrict it in this PR

bbpennel (approved):

PR #2195, opened 2025-04-09
More kernel api tests

whikloj:
**JIRA Ticket**: #2171 

Adds a couple more tests. I'm not sure if some of the no-op lines actually need tests or some of the enums that are defined in interfaces here but instantiated in other modules. Perhaps once those modules are complete the lines here will also show coverage. 

# How should this be tested?

Github actions runs tests, test coverage go up.

# Interested parties


bbpennel (approved):
That's a pretty intense exception test

PR #2196, opened 2025-04-09
2177 - Add tests for fcrepo-search-api

bbpennel:
**JIRA Ticket**: https://github.com/fcrepo/fcrepo/issues/2177

Add tests for fcrepo-search-api

# How should this be tested?

Running the tests

### DIFF
--- a/fcrepo-search-api/pom.xml
+++ b/fcrepo-search-api/pom.xml
@@ -47,9 +47,8 @@
 
     <!-- test gear -->
     <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <scope>test</scope>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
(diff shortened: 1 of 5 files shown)


fcrepo-search-api/pom.xml
```diff
@@ -47,9 +47,8 @@
...
     <!-- test gear -->
     <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-      <scope>test</scope>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
```
whikloj, line 54: Unrelated to your work here which is good. But it appears we didn't have any tests but we were importing junit and are still importing mockito. As it doesn't look like you are using it (and it is the old Junit 4 one) maybe we should remove it?

whikloj (approved):
👍

PR #2199, opened 2025-04-11
2164 - Test coverage for fcrepo-configs

bbpennel:
**JIRA Ticket**: https://github.com/fcrepo/fcrepo/issues/2164

Adds test coverage for fcrepo-configs

Minor adjustment to validation of s3 properties so it will disallow blank ocflS3Bucket when s3 is enabled. In tests, ocflS3Bucket was defaulting to "" rather than null.

# How should this be tested?

Run the tests.


whikloj (approved):
👍

PR #2202, opened 2025-04-29
Final test coverage or exclusion for fcrepo-kernel-api

whikloj:
**JIRA Ticket**: 

🤞 last one for fcrepo-kernel-api #2171 

Adds the new exclusion annotation, and adds a couple checks

# How should this be tested?

Tests pass

# Interested parties

### DIFF
--- a/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/identifiers/FedoraId.java
+++ b/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/identifiers/FedoraId.java
@@ -518,11 +518,7 @@ public class FedoraId {
         }
         if (processID.contains(FCR_VERSIONS)) {
             final String[] versionSplits = split(processID, FCR_VERSIONS);
-            if (versionSplits.length > 2) {
-                throw new InvalidResourceIdentifierException(String.format(
-                        "Path <%s> is invalid. May not contain multiple %s parts.",
-                        fullPath, FCR_VERSIONS));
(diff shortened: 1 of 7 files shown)


fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/identifiers/FedoraId.java
```diff
@@ -518,11 +518,7 @@ private String processIdentifier() {
         }
         if (processID.contains(FCR_VERSIONS)) {
             final String[] versionSplits = split(processID, FCR_VERSIONS);
-            if (versionSplits.length > 2) {
```
bbpennel, line 521: did you conclude this was unreachable?

bbpennel (approved):

fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/identifiers/FedoraId.java
whikloj: Yeah, in the private constructor we call this method [after we call](https://github.com/fcrepo/fcrepo/blob/main/fcrepo-kernel-api/src/main/java/org/fcrepo/kernel/api/identifiers/FedoraId.java#L113-L114) `checkForInvalidPath` which also checks for 2 instances of any of the special path elements (`fcr:acl`, `fcr:versions`, `fcr:tombstone` and `fcr:metadata`). So this doesn't get hit.

PR #2209, opened 2025-05-08
2157 - Update logback to 1.4 and slf4j to 2.0

bbpennel:
**JIRA Ticket**: https://github.com/fcrepo/fcrepo/issues/2157

Updates logback to most recent 1.4.x and slf4j to most recent 2.0.x

Note: 1.5.x is the most recent logback series, but it conflicts with dependencies in the older version of jetty we are currently using, causing some ITs to fail. So we can circle back to logback after updating jetty in


whikloj (approved):
👍

PR #2211, opened 2025-05-08
2152 - Upgrade to hamcrest 3

bbpennel:
**JIRA Ticket**: https://github.com/fcrepo/fcrepo/issues/2152

* Upgrade to hamcrest 3. 
* Remove system-rules which was no longer doing anything and pulling in old hamcrest. 
* Update awaitility

Note: there are still a few dependencies that pull in hamcrest 1.3, mainly jersey-test-framework-provider-grizzly2. We will want to upgrade that once we move over to the jakarta libraries. There is also ldp-testsuite which is the backbone of fcrepo-integration-ldp, but hasn't been updated since 2014. We'll need to decide what to do about that at some point.

# How should this be tested?

These are all test dependencies, so just running the tests.

### DIFF
--- a/fcrepo-stats-api/pom.xml
+++ b/fcrepo-stats-api/pom.xml
@@ -28,13 +28,13 @@
 
     <!-- test gear -->
     <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
+      <artifactId>mockito-junit-jupiter</artifactId>
       <scope>test</scope>
     </dependency>
   </dependencies>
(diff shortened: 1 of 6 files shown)


fcrepo-stats-api/pom.xml
```diff
@@ -28,13 +28,13 @@
...
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
+      <artifactId>mockito-junit-jupiter</artifactId>
```
whikloj, line 37: There are no tests in fcrepo-stats-api, should we just remove these dependencies?

whikloj: By "these" I mean `junit-jupiter` and `mockito-junit-jupiter`

bbpennel: A little depressing, but sure why not, they are removed. Looks like all the classes are simple beans, only one does [slightly more than return a value](https://github.com/fcrepo/fcrepo/blob/main/fcrepo-stats-api/src/main/java/org/fcrepo/stats/api/RepositoryStatsByMimeTypeResults.java#L34).

whikloj (approved):
👍

PR #2212, opened 2025-05-08
Remove felix and osgi config

bbpennel:
Remove felix and osgi config, everything is jars now, since we no longer work with OSGI/karaf


whikloj (approved):
👍

Issue #2144, opened 2025-02-27
HTTPClient 5


Surfrdan:
Upgrade from HTTPClient 4 to 5

Issue #2142, opened 2025-02-27
Commons Lang 3.17


Surfrdan:
Upgrade Commons Lang from 3.12 to 3.17

Issue #2126, opened 2025-02-27
Caffeine 3.2


Surfrdan:
Upgrade Caffeine from 3.0 to 3.2

PR #2216, opened 2025-05-15
Update dependencies and explicitly attach mockito as a javaagent

whikloj:
**JIRA Ticket**: 
- #2126
- #2135  
- #2141 
- #2142 
- #2143 
- #2144 
- #2145 

Updates a bunch of dependencies to the latest versions available. It also adds new arguments to the surefire plugin configuration based on https://javadoc.io/doc/org.mockito/mockito-core/latest/org.mockito/org/mockito/Mockito.html#0.3
This only works in modules that have a dependency on mockito.

# How should this be tested?

Should have no impact on function.

# Interested parties

### DIFF
--- a/fcrepo-auth-common/pom.xml
+++ b/fcrepo-auth-common/pom.xml
@@ -207,6 +207,13 @@
         <groupId>org.apache.felix</groupId>
         <artifactId>maven-bundle-plugin</artifactId>
       </plugin>
+      <plugin>
+        <artifactId>maven-surefire-plugin</artifactId>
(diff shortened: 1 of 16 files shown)


bbpennel (changes requested):

whikloj:
Yeah, it should work because it is supposed to pull the parent `argLine` in via the `${argLine}` or `@{argLine}` variable, but perhaps it isn't working. I have a different method working in my other branch, so I'll close this in favour of that work.

Issue #2128, opened 2025-02-27
Guava 33


Surfrdan:
Upgrade Guava from 32 to 33

bbpennel:
https://github.com/fcrepo/fcrepo/pull/2230

whikloj:
Closed with https://github.com/fcrepo/fcrepo/commit/7cc9778f26dbc4187b05f7fe217e4dc92e631acb

PR #2230, opened 2025-06-10
GH-2128 - Update to guava 33

bbpennel:
**JIRA Ticket**: https://github.com/fcrepo/fcrepo/issues/2128

* Update to guava 33.
* Replace javax.annotation with jakarta.annotation, since guava is no longer transitively pulling in javax.annotation


whikloj (approved):
👍 and thanks for catching those javax misses.

Issue #2227, opened 2025-06-05
net.jodah.failsafe


bbpennel:
This dependency has moved to https://mvnrepository.com/artifact/dev.failsafe/failsafe
It is only used for DbTransactionExecutor

PR #2234, opened 2025-06-10
GH-2227 - Update failsafe dependency, and remove unnecessary include

bbpennel:
**JIRA Ticket**: https://github.com/fcrepo/fcrepo/issues/2227

* Update failsafe dependency, and remove unnecessary include


whikloj (approved):
👍

Issue #2218, opened 2025-05-16
HTMLUnit


Surfrdan:
2.53 -> 4.11.0+

whikloj:
Resolved with https://github.com/fcrepo/fcrepo/commit/6cdcf351e96fe4f5dbae17fecbfe1a5d4a205144

PR #2235, opened 2025-06-12
GH-2218 - Upgrade to htmlunit 4.4

bbpennel:
**JIRA Ticket**: https://github.com/fcrepo/fcrepo/issues/2218

* Upgrades htmlunit from version 2 to 4.4. 
    * The most recent version is 4.13, but 4.5+ all fail because of an error parsing/executing jquery 1.12 which is used in the UI. We could probably upgrade further if we upgraded jquery, but I assume we are not going to prioritize that since we are planning to replace the UI.

### DIFF
--- a/fcrepo-webapp/src/test/java/org/fcrepo/integration/FedoraHtmlResponsesIT.java
+++ b/fcrepo-webapp/src/test/java/org/fcrepo/integration/FedoraHtmlResponsesIT.java
@@ -4,54 +4,52 @@
  * tree.
  */
 package org.fcrepo.integration;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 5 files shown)


fcrepo-webapp/src/test/java/org/fcrepo/integration/FedoraHtmlResponsesIT.java
```diff
@@ -216,6 +214,7 @@ public void testCreateNewObjectAndDeleteIt() throws IOException {
 
         final String pid = createNewObject();
         final HtmlPage page = webClient.getPage(serverAddress + pid);
+//        System.out.println("Page: " + page.asXml());
```
whikloj, line 217: Can be deleted

whikloj (changes requested):
One comment to remove, otherwise looks good.

whikloj (approved):
👍

Issue #2242, opened 2025-07-10
maven jetty plugin


Surfrdan:
Currently v11 which targets JDK 11.
v12 is out and supports JDK 21

PR #2249, opened 2025-07-25
GH-2242 - Update to jetty 12

bbpennel:
**JIRA Ticket**: https://github.com/fcrepo/fcrepo/issues/2242

* Updates jetty version to 12
* Updates dependencies and configuration accordingly
* Had to add bouncycastle in for checkstyle since it was erroring due to missing classes after updating jetty, the older version of jetty must have been somehow providing it

# How should this be tested?

`mvn jetty:run -pl fcrepo-webapp` should work, including with `-Dfcrepo.dynamic.test.port=<someportotherthan8080>`


bbpennel:
Noting that while I removed the IT for encoded slashes in resource names due to being unable to get it working with the jetty-maven-plugin, it does work in standalone jetty. In order to get jetty to permit it you have to make two changes. 

The first is to update jetty.xml to change the uriCompliance httpConfiguration option to `LEGACY`, or by setting the system property `jetty.httpConfig.uriCompliance` to `LEGACY`.

The second is to use a webAppContext file and set the `decodeAmbiguousURIs` to true on the servlet handler. If you don't do this second step, requests will mysteriously fail in shiro. Here is an example of what that context file could look like:
```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE Configure PUBLIC "-//Jetty//Configure//EN" "https://jetty.org/configure_10_0.dtd">

<Configure class="org.eclipse.jetty.ee10.webapp.WebAppContext"> 
  <Set name="contextPath">/fcrepo</Set> 
  <Set name="war">/path/to/fcrepo.war</Set>

  <Set name="securityHandler">
    <New class="org.eclipse.jetty.ee10.servlet.security.ConstraintSecurityHandler">
     <Set name="loginService">
       <New class="org.eclipse.jetty.security.jaas.JAASLoginService">
         <Set name="name">fcrepo</Set>
         <Set name="loginModuleName">fcrepo</Set>
       </New>
     </Set>
    </New>
  </Set>

  <Get id="servletHandler" name="servletHandler">
    <Set name="decodeAmbiguousURIs">true</Set>
  </Get>
</Configure>
```

PR #2252, opened 2025-08-14
Remove JQuery and bump htmlunit but removes some tests

Surfrdan:
**Exploratory attempt to remove outdated JQuery**

**JIRA Ticket**: https://github.com/fcrepo/fcrepo/issues/2245

Removes JQuery which is required by parts of Bootstrap 3 but I don't think we were using. 
Also bumps htmlunit but disables several UI tests which required a javascript based response.
htmlunit is unable to support modern javascript and is not really fit for purpose in a JS environment. 

# How should this be tested?

* Manually test the functions of the UI. In particular, those areas where automated tests have been removed. 

# Interested parties


Surfrdan:
Lessons learned so far:
 * Bumping htmlunit version breaks javascript functionality
 * Bumping bootstrap to 5 breaks CSS layout
 * Disabling javascript in the htmlunit web client breaks a number of tests
 * HTMLUnit probably unfit for purpose going forward.

Surfrdan:
CSS Layout would be mostly fixed now with Bootstrap upgraded from 3 to 5. No more JQuery requirement either so that issue has gone away.

fcrepo-http-api/src/main/resources/views/common-node-actions.vsl
```diff
@@ -3,122 +3,123 @@
...
+<div class="alert alert-warning d-flex align-items-center" role="alert">
+    <!-- Optional icon if you include Bootstrap Icons:
+    <i class="bi bi-exclamation-triangle-fill me-2"></i> -->
+    <div>
+        This is a <strong>historic version</strong> and cannot be modified.
+    </div>
+</div>
+    #if ($helpers.isRdfResource($rdf, $original, "http://fedora.info/definitions/v4/repository#", "Binary"))
+    <a href="$content" class="btn btn-success btn-lg">
+        <!-- <i class="bi bi-download me-1"></i> -->
```
bbpennel, line 16: Probably want to retain the icon. Bootstrap has it own icons, and while there's not an exact match, this one might be fine
https://icons.getbootstrap.com/icons/file-earmark-arrow-down/

bbpennel (changes requested):
While we don't need to perfectly reproduce the old appearance (bootstrap probably fixed a number of accessibility issues), there are a few parts that I think we should update:
* The biggest one is that predicate names are showing up in a light blue color instead of the previous near-black color. Screenshot 2025-08-21 at 1 34 20 PM
* The bread crumbs at the top were previously in a gray bar and indented a little bit more than the title. It'd be at least good to get it indented as much as the fields below: Screenshot 2025-08-21 at 1 34 26 PM
* Fields with multiple values are wrapping down to the beginning of the line rather than being offset by the field label. All values should line up along their left edge like a table Screenshot 2025-08-21 at 1 36 20 PM
* When at the root of the repo, the "Properties" header is blue because it is collapsible, but it's gray on all other pages. This one could probably be left as is, it likely only jumped out at me because I'm used to the interface
* On the search page, the X button is just a squished red square with nothing in it, and isn't in the right corner of the panel anymore 
Screenshot 2025-08-21 at 1 49 42 PM
* On the search page, "Add condition" does not look like a button anymore

Surfrdan:
I've addressed all of Ben's comments above in the latest commits

bbpennel:
For some reason this PR is now showing a ton of changes that are already on the main branch, which is a bit worrisome. Were commits [e1e0d42](https://github.com/fcrepo/fcrepo/pull/2252/commits/e1e0d42ea5d97bc1cd1d728d8b7184a3c9b1c064) and [53ef5c7](https://github.com/fcrepo/fcrepo/pull/2252/commits/53ef5c7c2c750160b306429cc4f69f4f9b2768c4) cherry-picked onto this branch? Most of the changes in this PR are unrelated at this point.

Otherwise, your updates look good for the most part. I'm still not seeing an icon on the download button, even though I see that you added `bi-file-earmark-arrow-down` to it.

It also looks like your IDE decided to reformat all the CSS rules into one liners in fcrepo-webapp/src/main/webapp/static/css/common.css, which is pretty hard to review and not really an improvement. Could you undo that if it's not too much work?

Surfrdan:
Thanks @bbpennel 

I have opened a ticket for the remedial work

https://fedora-repository.atlassian.net/browse/FCREPO-4034

PR #2258, opened 2025-09-26
Use variable for fcrepo-storage-ocfl version

bbpennel:
Minor update to use the existing variable for fcrepo-storage-ocfl's version, and to set it to the correct value

# How should this be tested?
build passing

# Interested parties
@fcrepo/committers

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
       (for scm, site-distribution, etc -->
     <project_name>fcrepo</project_name>
     <project_organization>fcrepo</project_organization>
-    <fcrepo-storage-ocfl.version>6.5.0-SNAPSHOT</fcrepo-storage-ocfl.version>
+    <fcrepo-storage-ocfl.version>6.4.0</fcrepo-storage-ocfl.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -22,7 +22,7 @@
       (for scm, site-distribution, etc -->
     <project_name>fcrepo</project_name>
     <project_organization>fcrepo</project_organization>
-    <fcrepo-storage-ocfl.version>6.5.0-SNAPSHOT</fcrepo-storage-ocfl.version>
```
whikloj, line 25: If we do this, then when you make a change to `fcrepo-storage-ocfl` locally and build it, you will need to remember to change this variable to `7.0.0-SNAPSHOT` to test it.

Also, if we are doing this perhaps we should be adding a github action to `fcrepo-storage-ocfl` that deploys `fcrepo/fcrepo:main`, changes the `fcrepo-storage-ocfl.version` to whatever the current version is and tries to run the Fedora tests?

whikloj (commented):
👍

PR #2302, opened 2026-04-08
Deduplicate all types

whikloj:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-4078

When using the rdf types pulled from the resource in the search index, we deduplicate them. But when we pull the rdf types from the resource and pass them to the search index, we don't deduplicate them. This deduplicates all types.

# How should this be tested?

See the same OCFL provided in the above ticket. Add `fcrepo.rebuild=true` to your properties and rebuild using that OCFL. Before this PR, it will fail with either duplicate key or foreign key failures. After it will succeed.

# Interested parties
@fcrepo/committers


bbpennel (approved):

bbpennel:
PR fixes the issue for me, and the test replicates the bug 👍

PR #2306, opened 2026-04-29
Fcrepo 4075 - Add Artemis support 

Surfrdan:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-4075

Adds Artemis support alongside existing ActiveMQ support 
ActiveMQ remains the default for 7.x but will be deprecated in Fedora 8

# How should this be tested?

* Test the fcrepo.jms.provider property with values `activemq` and `artemis`. ActiveMQ should remain the default. 
* make JMS or STOMP queries to 61616 or 61613 respectively. 
* Attempt to configure external Artemis and ActiveMQ brokers  in combination with this embedded broker as is popular with some users. 

# Interested parties
@rmassa initially enquired about this


fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
```diff
@@ -83,6 +90,13 @@ public class FedoraPropsConfig extends BasePropsConfig {
             ACTIVE_MQ_DIR_DEFAULT_VALUE + "').toAbsolutePath().toString()}}")
     private String activeMqDirectory;
 
+    @Value("${" + FCREPO_ARTEMIS_CONFIGURATION + ":classpath:/config/broker.xml}")
```
bbpennel, line 93: For clarity, it would make sense for the default artemis file to be named artemis.xml like how we have `activemq.xml`

fcrepo-configs/src/main/resources/config/activemq.xml
```diff
@@ -113,7 +113,7 @@
         <transportConnectors>
```
bbpennel, line 113: We should undo the changes to this file, its just the IDE going overboard on fixing indent, no need to add noise to the commit history

fcrepo-configs/src/test/java/org/fcrepo/config/FedoraPropsConfigTest.java
```diff
@@ -73,7 +73,7 @@ public void testDefaultValues() {
         assertEquals("localhost", config.getJmsHost());
         assertEquals("61616", config.getJmsPort());
         assertEquals("61613", config.getStompPort());
-        assertNotNull(config.getActiveMQConfiguration());
```
bbpennel, line 76: We should still be expecting amq config to have a default as well

fcrepo-jms/src/main/java/org/fcrepo/jms/DefaultMessageFactory.java
```diff
@@ -31,7 +31,8 @@
  */
 public class DefaultMessageFactory implements JMSEventMessageFactory {
 
-    private static final String JMS_NAMESPACE = "org.fcrepo.jms.";
```
bbpennel, line 34: Won't this break everyone's amq integration, including camel-tool-box? We should only use the underscore form when the user is using artemis.

My preference for how to handle this would be to change DefaultMessageFactory so that all the constants become static variables instead, and jmsNamespace gets automatically set based on the FedoraPropsConfig.FCREPO_JMS_PROVIDER. Instead of using the constants directly, we would get the property names with lazy getters, but they are only used in the class itself and the test for the class, so I don't think it would be a very invasive change.

fcrepo-jms/src/test/java/org/fcrepo/integration/jms/observer/JmsQueueActiveMQIT.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+/**
+ * Queue-based JMS integration test running against an embedded ActiveMQ Classic broker.
+ *
+ * @author acoburn
```
bbpennel: you should probably be the author rather than acoburn for all these new jms tests.

fcrepo-jms/src/test/java/org/fcrepo/jms/DefaultMessageFactoryTest.java
```diff
@@ -74,8 +75,32 @@ public void testBuildMessageNullUrl() throws JMSException {
...
     }
 
+    /**
+     * Artemis enforces the JMS spec rule that property names must be valid Java identifiers; ActiveMQ Classic does
+     * not. A regression that re-introduced dotted names would slip past the other unit tests (which use an
+     * ActiveMQ-Classic message) and only fail in the Artemis ITs. Asserting the constants here catches it at
+     * unit-test time.
+     */
+    @Test
+    public void testHeaderNamesAreValidJavaIdentifiers() {
```
bbpennel: This only really applies to artemis, we should have tests that very that when using amq we have the dot form and in artemis we have the underscore form

fcrepo-jms/src/test/resources/spring-test/broker-queue-artemis.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bbpennel, line 1: for all these config files (spring and brokers) I think we should specify in the filename whether they are for artemis or amq, this is inconsistently applied at the moment with one or the other being treated as the default

fcrepo-webapp/src/main/java/org/fcrepo/webapp/JmsConfig.java
```diff
@@ -43,6 +55,24 @@ static class JmsEnabled extends ConditionOnPropertyTrue {
...
+            return "activemq".equalsIgnoreCase(provider);
+        }
+    }
+
+    static class ArtemisConfigured implements Condition {
+        @Override
+        public boolean matches(final ConditionContext context, final AnnotatedTypeMetadata metadata) {
+            final var provider = context.getEnvironment().getProperty(FedoraPropsConfig.FCREPO_JMS_PROVIDER,
+                    "activemq");
+            return "artemis".equalsIgnoreCase(provider);
```
bbpennel, line 74: what happens if the property is set to something other than activemq or artemis? Does fedora fail to start?

bbpennel (changes requested):

fcrepo-configs/src/main/java/org/fcrepo/config/FedoraPropsConfig.java
Surfrdan: Thats the expected filename in Artemis just as activemq.xml is the expected filename in ActiveMQ. I think it would be counter intuitive to rename it https://artemis.apache.org/components/artemis/documentation/latest/configuration-index.html#broker-xml

fcrepo-jms/src/main/java/org/fcrepo/jms/DefaultMessageFactory.java
Surfrdan: Yeh thats a fair point. Technically it's breaking the JMS spec but the larger issue is that changing it would indeed break existing users AMQ setups. Lets leave the non-conformant namespace for AMQ for now and let it die with AMQ in 8.0 while Artemis gets the new conformant namespace from the outset

fcrepo-webapp/src/main/java/org/fcrepo/webapp/JmsConfig.java
```diff
@@ -60,8 +60,10 @@ static class ActiveMqConfigured implements Condition {
         public boolean matches(final ConditionContext context, final AnnotatedTypeMetadata metadata) {
             final var provider = context.getEnvironment().getProperty(FedoraPropsConfig.FCREPO_JMS_PROVIDER,
                     "activemq");
+            LOGGER.info("ActiveMQ support is deprecated and will likely be removed in version 8");
```
bbpennel: This should probably be at WARN level. I'd say its also fine to leave it ambiguous but definitive with something like "ActiveMQ support is deprecated and will be removed in a future version".

fcrepo-webapp/src/main/java/org/fcrepo/webapp/JmsConfig.java
Surfrdan: I've added a fallback to activemq now

bbpennel:
I think codecov is angry because there isn't any test coverage of JmsConfig but there were a number of changes to it

fcrepo-webapp/src/main/java/org/fcrepo/webapp/JmsConfig.java
```diff
@@ -66,40 +98,98 @@ public AbstractJMSPublisher jmsPublisher(final FedoraPropsConfig propsConfig) {
...
+        }
+
+        final Path tempFile = Files.createTempFile("fcrepo-artemis-", "-broker.xml");
+        tempFile.toFile().deleteOnExit();
+
+        try (InputStream input = configuration.getInputStream()) {
+            Files.copy(input, tempFile, StandardCopyOption.REPLACE_EXISTING);
+        }
+
+        return tempFile;
```
bbpennel, line 164: I forgot to ask about this in the first pass, but I was curious why the feature to pull in config was added for artemis but not amq, or any of the other config files? There doesn't seem to be anything particularly artemis specific about it.

bbpennel (approved):

bbpennel:
This PR looks good to me, pending testing. I left one other question on it

fcrepo-webapp/src/main/java/org/fcrepo/webapp/JmsConfig.java
Surfrdan: activemq reads an activemq.xml file off of the class path, whereas artemis needs it on the filesystem and passed as an argument

Issue #2315, opened 2026-06-10
Outdated version information in OCFL metadata


centromere:
I was looking through an OCFL object produced by Fedora 7 and discovered that the `message` field still references Fedora 6:

```json
{
  "versions": {
    "v1": {
      "created": "2026-03-15T16:17:54.62657246Z",
      "message": "Authored by Fedora 6",
      "user": {
        "name": "fedoraAdmin",
        "address": "info:fedora/fedoraAdmin"
      },
      "state": {
        "5ef0f9890f4631a2ac16d63918cdc636a0f564ee8ff3f25efdf1b2173431ba3837c2f35ad2053d91136505e5091b045bf9b679abec53a737af4286b2901f9426": [
          "fcr-container.nt"
        ],
        "e4272f8e504898235c6c843622b511647b929e2d8403e65590aee126949c79a07a625bdb6b0d26c993ed97164e35df16f4949f21901ccdf225d9451e06d3ddfd": [
          ".fcrepo/fcr-root.json"
        ]
      }
    }
  }
}
```
The string looks to be defined here:
https://github.com/fcrepo/fcrepo/blob/65cb7395ba43f6293ffac54a03c97b145f437c5b/fcrepo-persistence-ocfl/src/main/java/org/fcrepo/persistence/ocfl/impl/OcflPersistenceConfig.java#L107

Surfrdan:
PR https://github.com/fcrepo/fcrepo/pull/2316

whikloj:
Resolved with https://github.com/fcrepo/fcrepo/commit/7554ea9529d75d8ddb02350ba3db960b58f716fb

PR #2316, opened 2026-06-11
updating naming of fedora version from 6 to 7 in several locations

Surfrdan:
**Issue**: #2315 

updates naming of fedora version from 6 to 7 in several locations

# Interested parties


yurj:
Hi! Would it be useful to get the version automatically instead of using a fixed string? I don't know if it is an information we should store. If not, I would remove the version.

whikloj (approved):
Works as described

PR #2333, opened 2026-09-04
FCREPO-4093 Return post-update ETag from PATCH and PUT

Surfrdan:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-4093

* Regression introduced by FCREPO-4073 (#2298)
* Client-side note: https://github.com/fcrepo-exts/fcrepo-java-client/pull/75

Makes `PATCH` and `PUT` return the ETag of the resource as it is after the update, instead of as it was before.

FCREPO-4073 added a per-request `cachedRdfEtag` field to `ContentExposingResource` so that a HEAD/GET request computes the RDF etag once instead of twice. On a mutating request the cache is populated by `evaluateRequestPreconditions` before the update is applied, and then reused by `addCacheControlHeaders` when the response is built, so the response carried the pre-update value. `PATCH` already called `reloadResource()`, but the reloaded resource never reached the etag computation.

Changes:

* Add `invalidateCachedRdfEtag()` to `ContentExposingResource`, called from `reloadResource()` (used by `PATCH` and `FedoraVersioning`) and at the start of `createUpdateResponse()` (used by `PUT` and `POST`).
* `FedoraAcl.updateSparql` passed its pre-update `FedoraResource` object to `addCacheControlHeaders`. It now re-reads the ACL resource after the update.

# How should this be tested?

To reproduce on `main`:

1. `POST` a container, note the `ETag` from a `HEAD` on it.
2. `PATCH` the container with a SPARQL update and an `If-Match` for that ETag.
3. The `ETag` on the 204 response equals the ETag from step 1. A following `HEAD` returns a different, correct ETag.

Two integration tests in `FedoraLdpIT` cover this:

```bash
mvn -pl fcrepo-http-api verify -Dit.test='FedoraLdpIT#testPatchResponseReturnsUpdatedEtag+testPutResponseReturnsUpdatedEtag'
```

Each asserts the response ETag differs from the pre-request ETag and matches the ETag of a following `HEAD`. Both fail on unmodified source with "returned the ETag from before the update" and pass with this change.

Full `fcrepo-http-api` suite run: 583 tests, 2 failures. Both are pre-existing and unrelated:

* `LDPContainerIT.indirectContainerModifyProxyOnDemandVersioning` fails identically with this change reverted.
* `FedoraVersioningIT.testResponseForMultipleMementosInASecond` passes when run on its own (timing-sensitive under load).

Other modules' test suites were not run.

# Interested parties

@fcrepo/committers

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
+++ b/fcrepo-http-api/src/main/java/org/fcrepo/http/api/ContentExposingResource.java
@@ -421,7 +421,16 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
 
     protected FedoraResource reloadResource() {
         this.fedoraResource = null;
+        invalidateCachedRdfEtag();
         return resource();
+    }
+
+    /**
+     * Discard the etag computed earlier in this request. The cached value is only valid for the state of the
+     * resource at the time it was computed, so it must be dropped after the resource has been modified.
+     */
+    protected void invalidateCachedRdfEtag() {
+        this.cachedRdfEtag = null;
     }
 
     /**
@@ -872,6 +881,7 @@ public abstract class ContentExposingResource extends FedoraBaseResource {
      *         content depending on Prefer headers.
      */
     protected Response createUpdateResponse(final FedoraResource resource, final boolean created) {
+        invalidateCachedRdfEtag();
         addCacheControlHeaders(servletResponse, resource, transaction());
         addResourceLinkHeaders(resource, created);
         addExternalContentHeaders(resource);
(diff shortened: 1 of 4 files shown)


Surfrdan:
This PR was generated live during a workshop at Fedora Camp 2026 as an example of how much can be generated from a Claude prompt as simple as "Fix this jira issue https://fedora-repository.atlassian.net/browse/FCREPO-4093"

It has not deter undergone any filtering and self review before posting to GitHub as a PR but we went through the process to demonstrate end-to-end

bbpennel (approved):

PR #2334, opened 2026-09-11
FCREPO-4106 - Fix issue with delete and purge in the same TX

bbpennel:
**JIRA Ticket**: https://fedora-repository.atlassian.net/browse/FCREPO-4106

Fixes bug that occurred when deleting and purging a resource within the same long running transaction, which resulted in stale entries being left in the containment index, ultimately causing purges of any ancestors to fail.

This was actually primarily restoring a previous behavior that was removed in https://github.com/fcrepo/fcrepo/pull/2193/changes . That PR did correctly fix the issue of finding the parent of a deleted resource, but only if the delete was committed.

Co-Authored-By: chatgpt-5.6-terra

# How should this be tested?

New tests were added to replicate the bug. You can also manually replicate it with the steps in the ticket, and then verify they work with the fixed branch.

# Interested parties
@fcrepo/committers


bbpennel:
I have also confirmed this fixes the integration test in my repository

Surfrdan (approved):

